PR #152, opened 2023-01-04
Add new page/route for notebooks create and rename

derek-ho:
### Description

### Issues Resolved

### Check List
- [ ] UX Sign off of new page design and widgets 

references #9


.cypress/integration/2_notebooks.spec.js
```diff
@@ -49,9 +49,9 @@ describe('Adding sample data and visualization', () => {
     cy.get('.euiModalHeader__title[data-test-subj="confirmModalTitleText"]')
       .contains('Add samples')
       .should('exist');
-    cy.wait(100);
-    cy.get('.euiButton__text').contains('Yes').trigger('mouseover').click();
     cy.wait(100 * 5);
```
mengweieric, line 52: Minor: question what's difference between 100 * 5 vs 500? also could we avoid hard waiting here instead to have something maybe a timeout?

mengweieric (approved):
Thanks for the changes

pjfitzgibbons:
> Also need to address UX Review from [opensearch-project/observability#1308 (review)](https://github.com/opensearch-project/observability/pull/1308#pullrequestreview-1209192948)
> 
> "I left some comments. In general, we need some UX feedback here. In my opinion, it is a bad experience to create a whole page for 1 input box." - @ps48

@kgcreative Could you please add your comments or approval for this change?

derek-ho:
@pjfitzgibbons are we keeping the session storage or removing it? I remember you talked about removing it

kgcreative:
So, it does feel really awkward to have "create" start with just one field. And, I do think that the edit modal is much better than a full page edit for a single field as well.

I think what I would like to see instead, is for "Create" to launch a new notebook directly with either {Notebook N} or {New Notebook} as the title, and we add an inline [rename] link that pops the rename modal? (That way we don't hide it in the "actions" menu either. (In my ideal world, you could just click the title and edit it, but we're a ways off from that).

From there, we can trigger the bottom bar save/cancel

pjfitzgibbons:
@kgcreative ... so then if a user clicks "create notebook" and changes their mind... they need to "Delete" that newly created (and autosaved) notebook?
Is there some intuitive UX around this, since create in that scenario doesn't have a defacto "Cancel"?
Did that make ssense?

kgcreative:
@pjfitzgibbons -- would it make sense in that case to drop the user into a new notebook, with the previous "Create" modal open? -- Explicit action: "Create" -- saves the  notebook object with the given new name. | "Cancel" would return the user to their previous route? (Basically we skip autosave/creating the object until after the user explicitly saves the modal)

pjfitzgibbons:
@kgcreative Ok.  I thought I tried that (URL that lands on a modal open) and it didn't work.
We'll try a spike on that and see if we can make it go.

pjfitzgibbons:
@derek-ho ☝️    Can you do a spike and see if you can route a "create" url that lands on a page with the Create modal open?

derek-ho:
Updated behavior: @kgcreative @pjfitzgibbons @mengweieric @ps48 take another look and see if this is better
https://user-images.githubusercontent.com/26328171/216399241-1ec2bd0b-6ba8-47f4-806c-f8185c1d153f.mov

kgcreative:
That feels much better

ps48:
@derek-ho great to see this update. Looks much better than our previous approach.

public/components/notebooks/components/note_table.tsx
```diff
@@ -367,10 +380,9 @@ export function NoteTable(props: NoteTableProps) {
                 <EuiSpacer size="m" />
                 <EuiFlexGroup justifyContent="center">
                   <EuiFlexItem grow={false}>
-                    <EuiButton
+                    <EuiButton fill href="#/notebooks/create"
```
ps48, line 383: Let's revert this fill as it was previously. Usually, a page should only have 1 primary fill button to implicitly tell the user that this is the main button to click on.

public/components/notebooks/components/note_table.tsx
```diff
@@ -77,6 +80,13 @@ export function NoteTable(props: NoteTableProps) {
     props.fetchNotebooks();
   }, []);
 
+  useEffect(() => {
+    const url = window.location.hash.split('/')
+    if (url[url.length-1] === 'create') { 
+      createNote();
+    }
+  }, [location]);
+
```
ps48, line 89: Nice 😄

public/components/notebooks/components/note_table.tsx
```diff
@@ -114,7 +124,10 @@ export function NoteTable(props: NoteTableProps) {
     setModalLayout(
       getCustomModal(
         onCreate,
-        closeModal,
+        () => {
+          closeModal()
+          history.push('/notebooks')
```
ps48, line 129: Should this be the `/notebooks` or the create workflow originator page?

public/components/notebooks/components/note_table.tsx
derek-ho: I see... Fixed in [cc0763c](https://github.com/opensearch-project/dashboards-observability/pull/152/commits/cc0763cb6612d999418849812427d33a7ca931f8)

public/components/notebooks/components/note_table.tsx
derek-ho: Good catch - I missed this as I was mainly testing it from the notebooks page itself, not the connection from other pages. Fixed with h`history.goback()` in [cc0763c](https://github.com/opensearch-project/dashboards-observability/pull/152/commits/cc0763cb6612d999418849812427d33a7ca931f8)

ps48 (approved):

public/components/notebooks/components/note_table.tsx
pjfitzgibbons: @derek-ho FYI future code changes (don't address for this PR)

Array.at(-1) vs arr[arr.length - 1]
The usual practice is to access [length](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/length) and calculate the index from that — for example, array[array.length - 1]. The at() method allows relative indexing, so this can be shortened to array.at(-1).

pjfitzgibbons (approved):

pjfitzgibbons:
@kgcreative This has been pushed to 2.7.   As such, I would like to reconsider your idea ⬆️ that "create" should just drop the user into a note-book edit with the title "new" or whatever.

I have two questions around this :
1.  As a Notebook is "live" always,  does it make sense from the Dashboards / Dashboards list viewpoint that the Notebook-View URL can be == Notebook Edit URL ?

2.  Leaning-into the idea that Create Notebook should result in a New Notebook page - with edit-in-place Notebook Name ...  Is there anything to your knowledge preventing the use of edit-in-place for notebook name?

Thanks for your advice!

PR #219, opened 2023-01-23
fix trace analytics services cypress file

derek-ho:
Signed-off-by: Derek Ho <dxho@amazon.com>

### Description

### Issues Resolved

### Check List


.cypress/integration/5_trace_analytics_services.spec.js
```diff
@@ -181,16 +181,16 @@ describe('Testing traces Spans table verify table headers functionality', () =>
     cy.contains('analytics-service, frontend-client, recommendation').should('exist');
     cy.get('.euiLink.euiLink--primary').contains('authentication').should('exist').click();
     cy.get('.panel-title').contains('Spans').should('exist');
-    cy.get('.panel-title-count').contains('5').should('exist');
```
derek-ho, line 184: @joshuali925 I may need some sanity confirmation from you - I found some of these numbers of spans changed in the cypress test - was test data updated at any point after this test was written?

derek-ho: Only concern here being that it seems to me this test should always fail, it wouldn't be flaky - do we have a history of this test file failing all the time?

PR #220, opened 2023-01-24
Cypress test update and related linting fixes

mengweieric:
### Description
- Update event analytics cypress
- Fix linting errors

### Issues Resolved

### Check List

### DIFF
--- a/.cypress/integration/1_event_analytics.spec.js
+++ b/.cypress/integration/1_event_analytics.spec.js
@@ -281,7 +281,7 @@ describe('Saves a query on explorer page', () => {
       .focus()
       .type(SAVE_QUERY2, { force: true });
 
-    cy.get('[data-test-subj="eventExplorer__querySaveConfirm"]').click();
+    cy.get('[data-test-subj="eventExplorer__querySaveConfirm"]').click({ force: true });
     cy.wait(delay * 2);
     cy.get('.euiToastHeader__title').contains('successfully').should('exist');
 
@@ -778,7 +778,7 @@ describe.skip('Renders chart and verify Toast message if X-axis and Y-axis value
   });
 });
 
-describe.skip('Render Table View', () => {
+describe.only('Render Table View', () => {
   before(() => {
     landOnEventVisualizations();
     clearQuerySearchBoxText('searchAutocompleteTextArea');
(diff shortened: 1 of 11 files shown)


joshuali925 (approved):

common/types/explorer.ts
```diff
@@ -380,7 +380,7 @@ export interface VisMeta {
 }
 
 export interface VisualizationState {
-  query: Query;
+  queryState: Query;
```
derek-ho, line 383: any reason why this was changed?

mengweieric: linting fix, there's a global 'query' within the context where inner query will shadow the global one.

public/components/event_analytics/explorer/visualizations/config_panel/config_panes/config_controls/data_config_panel_fields.tsx
```diff
@@ -55,10 +55,10 @@ export const DataConfigPanelFields = ({
     if (visType === VIS_CHART_TYPES.HeatMap)
       return name === AGGREGATIONS ? list.length >= 1 : list.length >= 2;
     // condition for line and scatter for dimensions section.
-    return name === GROUPBY && (list.length >= 1 || !isEmpty(time_field));
+    return name === GROUPBY && (list.length >= 1 || !isEmpty(timeField));
   };
 
-  const { time_field, unit, interval } = dimensionSpan;
```
derek-ho, line 61: Again why was this field name changed?

mengweieric: Same linting fix, rules only accept camelCase, CamelCase, or CAMEL_CASE

.cypress/utils/event_analytics/constants.js
```diff
@@ -20,6 +20,7 @@ export const TEST_QUERIES = [
   },
   {
     query: 'source = opensearch_dashboards_sample_data_logs',
+    dateRangeDOM: YEAR_TO_DATE_DOM_ID,
```
derek-ho, line 23: Question: how is this index timestamp determined - is it when you add the data? Will this break next year?

.cypress/integration/1_event_analytics.spec.js
```diff
@@ -280,19 +229,14 @@ describe('Saves a query on explorer page', () => {
     cy.get('[data-test-subj="eventExplorer__querySaveName"]')
       .focus()
       .type(SAVE_QUERY2, { force: true });
-
-    cy.get('[data-test-subj="eventExplorer__querySaveConfirm"]').click();
-    cy.wait(delay * 2);
-    cy.get('.euiToastHeader__title').contains('successfully').should('exist');
-
+    cy.get('[data-test-subj="eventExplorer__querySaveConfirm"]').click({ force: true });
+    cy.get('.euiToastHeader__title').should('contain', 'successfully');
```
derek-ho, line 233: is it really ok to get rid of the delay here? Won't the request take some time?

derek-ho: Same concern in a few places

.cypress/integration/1_event_analytics.spec.js
```diff
@@ -18,79 +18,32 @@ import {
...
+  VIS_TYPE_HBAR,
+  VIS_TYPE_VBAR,
+  VIS_TYPE_HEATMAP,
+  FIELD_HOST,
+  FIELD_AGENT
 } from '../utils/event_analytics/constants';
 import { supressResizeObserverIssue, COMMAND_TIMEOUT_LONG } from '../utils/constants';
 import { clearQuerySearchBoxText } from '../utils/event_analytics/helpers';
 
-const renderHistogramChart = () => {
```
derek-ho, line 31: Why were these tests removed?

public/components/event_analytics/explorer/save_panel/__tests__/__snapshots__/save_panel.test.tsx.snap
```diff
@@ -738,7 +738,7 @@ exports[`Saved query table component Renders saved query table 1`] = `
             onMouseOver={[Function]}
           >
             <EuiSwitch
-              checked={true}
+              checked={false}
```
derek-ho, line 741: was this behavior not tested properly - this seems like a pretty big snapshot change?

public/components/event_analytics/explorer/save_panel/save_panel.tsx
```diff
@@ -50,59 +42,36 @@ export const SavePanel = ({
...
-  setMetricMeasure,
-  setMetricLabel,
-  metricChecked,
+  isSaveAsMetricEnabled,
 }: ISavedPanelProps) => {
   const [options, setOptions] = useState([]);
   const [checked, setChecked] = useState(false);
-  const [measure, setMeasure] = useState('');
-  const [label, setLabel] = useState([]);
+  const [svpnlError, setSvpnlError] = useState(null);
```
derek-ho, line 50: What are you doing this this value?

public/components/event_analytics/explorer/save_panel/save_panel.tsx
```diff
@@ -50,59 +42,36 @@ export const SavePanel = ({
...
-    setLabel(selectedLabels);
-    setMetricLabel(selectedLabels);
-  };
-
-  useEffect(() => {
-    if (metricChecked) {
-      setChecked(true);
-      setMeasure(metricMeasure);
-    }
-  }, [])
```
derek-ho, line 104: is this what was causing snapshot change? Is this deletion intended?

public/components/event_analytics/explorer/save_panel/save_panel.tsx
```diff
@@ -50,59 +42,36 @@ export const SavePanel = ({
...
 
-  const onMeasureChange = (selectedMeasures: string) => {
-    setMeasure(selectedMeasures);
-    setMetricMeasure(selectedMeasures);
-  };
-
-  const onLabelChange = (selectedLabels: React.SetStateAction<never[]>) => {
-    setLabel(selectedLabels);
-    setMetricLabel(selectedLabels);
-  };
```
derek-ho, line 97: these were just unused?

.cypress/integration/1_event_analytics.spec.js
mengweieric: We don't need these any more.

.cypress/utils/event_analytics/constants.js
mengweieric: timestamp is determined automatically in event explorer by looking for the first encountered date type in an index mappings. It will not break as long as the sample data is not changed.

public/components/common/search/search.tsx
```diff
@@ -221,15 +217,12 @@ export const Search = (props: any) => {
...
                   curVisId={curVisId}
-                  spanValue={spanValue}
                   setSubType={setSubType}
-                  metricMeasure={metricMeasure}
-                  setMetricMeasure={setMetricMeasure}
-                  setMetricLabel={setMetricLabel}
-                  metricChecked={metricChecked}
+                  isSaveAsMetricEnabled={
+                    isEqual(curVisId, 'line') && tempQuery.match(PPL_SPAN_REGEX) !== null
+                  }
```
derek-ho, line 225: Can you explain what is this doing/why this change was made here? Any issue linking the bug here?

public/components/event_analytics/explorer/explorer.tsx
```diff
@@ -1430,25 +1437,17 @@ export const Explorer = ({
...
   };
 
-  useEffect(() => {
-    if (isEqual(selectedContentTabId, TAB_CHART_ID)) {
-      const statsTokens = queryManager.queryParser().parse(tempQuery).getStats();
-      const updatedDataConfig = getDefaultVisConfig(statsTokens);
-      setSpanValue(!isEqual(typeof updatedDataConfig.span, 'undefined'));
-    }
-  }, [tempQuery, selectedContentTabId, curVisId]);
-
```
derek-ho, line 1451: why removed?

public/components/event_analytics/explorer/save_panel/save_panel.tsx
mengweieric: the linting rule doesn't allow using console.xxx statement, so here I've just done a quick change to make error as a state (so later we can pipe error to maybe toast UI) rather than using console as this approach may not help on knowing the error.

public/components/event_analytics/explorer/save_panel/save_panel.tsx
```diff
@@ -146,18 +115,17 @@ export const SavePanel = ({
...
           <EuiFormRow display="columnCompressedSwitch">
-          <EuiToolTip
-          content="Only Time Series visualization and a query including stats/span can be saved as Metric" >
-            <EuiSwitch
-              showLabel={true}
-              label="Save as Metric"
-              checked={checked}
-              onChange={onToggleChange}
-              compressed
-              disabled = {!((isEqual(curVisId, 'line')) && spanValue)}
```
derek-ho, line 157: Never mind I see that this is passed down now, but wondering if matching the regex is doing the same as this block of code: 
`useEffect(() => {
    if (isEqual(selectedContentTabId, TAB_CHART_ID)) {
      const statsTokens = queryManager.queryParser().parse(tempQuery).getStats();
      const updatedDataConfig = getDefaultVisConfig(statsTokens);
      setSpanValue(!isEqual(typeof updatedDataConfig.span, 'undefined'));
    }
  }, [tempQuery, selectedContentTabId, curVisId]);`

public/components/event_analytics/explorer/save_panel/save_panel.tsx
mengweieric: For now I don't see these functions are used at all, linting is complaining about unused stuff so I was just doing minimal removals here to pass linting checking

public/components/event_analytics/explorer/save_panel/save_panel.tsx
```diff
@@ -113,8 +82,8 @@ export const SavePanel = ({
           <EuiFormRow helpText="Search existing dashboards or applications by name">
             <EuiComboBox
               placeholder="Select dashboards/applications"
-              onChange={(options) => {
-                handleOptionChange(options);
+              onChange={(daOptions) => {
```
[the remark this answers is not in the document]
mengweieric: Similar reason, 'options' is a global defined variable, have to rename to avoid shadowing and pass the linting checking

public/components/event_analytics/explorer/save_panel/save_panel.tsx
mengweieric: yes intended, I recall this should also relate to unused variable issues.

public/components/common/search/search.tsx
mengweieric: This comes from linting autofix.

public/components/event_analytics/explorer/explorer.tsx
mengweieric: The way metric uses to determine disabled/enabled saving-as-metric, as far as I can see, only cares about if the query contains timespan on saving-panel open. This code piece does the work to record timespan but it triggers query manager parsing and setSpanValue on every query typing/changing which causes performance degradation.

public/components/event_analytics/explorer/save_panel/save_panel.tsx
mengweieric: yes the quick change i made is to match a span in query.

.cypress/integration/1_event_analytics.spec.js
mengweieric: The wait here is like a hard wait, event request has returned successfully, it would still wait there. we don't necessary need to wait here for a specific amount of time. I think wait is something we need to almost always universally avoid in our test code, we should work on gradually refactor code to use other approaches.

derek-ho (approved):
LGTM

ps48:
@mengweieric should we backport these changes to 2.x?

PR #256, opened 2023-02-07
Modify one legacy query in sample panels for timeseries that causes invalid data issue

mengweieric:
Signed-off-by: Eric Wei <menwe@amazon.com>

### Description
There is one sample data query that is not valid after we changed line to timeseries which restricts dimension to only be timespan. Changed the query to use timespan. 
Impact: users will still see invalid data in existing sample data, but any newly added sample data will not have this error. Need to decide if we should support BWC for this.

### Issues Resolved

### Check List


mengweieric:
Close this as we will remove data validation for timeseries for now to still support line chart @anirudha

PR #281, opened 2023-02-15
Panels create page

### LINKED ISSUE #108: Convert Operational Panels - New into full-page form
Currently Operational Panels  raises a UI Dialog to create a new item.

In order to integrate Operational Panels with Dashboards List, the New operation must navigate to an addressable URL -> a full-page form.

Will need to copy/re-write from previous PR https://github.com/opensearch-project/observability/pull/1368

Image

kavithacm:
### Description
Create page for custom panels

### Issues Resolved
#108 

### Check List

### DIFF
--- a/public/components/custom_panels/home.tsx
+++ b/public/components/custom_panels/home.tsx
@@ -93,7 +93,7 @@ export const Home = ({
   const createCustomPanel = (newCustomPanelName: string) => {
     if (!isNameValid(newCustomPanelName)) {
       setToast('Invalid Operational Panel name', 'danger');
-      history.goBack();
+      window.location.assign(`${_.last(parentBreadcrumbs)!.href}`);
       return;
     }
 
--- a/public/components/custom_panels/home.tsx
+++ b/public/components/custom_panels/home.tsx
@@ -9,7 +9,7 @@ import { Toast } from '@elastic/eui/src/components/toast/global_toast_list';
 import _ from 'lodash';
 import React, { ReactChild, useState } from 'react';
 // eslint-disable-next-line @osd/eslint/module_migration
-import { StaticContext } from 'react-router';
+import { StaticContext, Switch } from 'react-router';
 import { Route, RouteComponentProps } from 'react-router-dom';
 import PPLService from '../../services/requests/ppl';
 import DSLService from '../../services/requests/dsl';
@@ -95,6 +95,7 @@ export const Home = ({
       return;
     }
 
+    console.log("enters createCustomPanel ");
     return http
       .post(`${CUSTOM_PANELS_API_PREFIX}/panels`, {
         body: JSON.stringify({
@@ -287,6 +288,7 @@ export const Home = ({
     }
   };
 
+  console.log("renderProps.match.path: ", renderProps.match.path);
   return (
     <div>
       <EuiGlobalToastList
@@ -297,9 +299,10 @@ export const Home = ({
         side={toastRightSide ? 'right' : 'left'}
         toastLifeTimeMs={6000}
       />
+      <Switch>
       <Route
         exact
-        path={renderProps.match.path}
+        path={['/operational_panels/create', '/operational_panels']}
         render={(props) => {
           return (
             <ObservabilitySideBar>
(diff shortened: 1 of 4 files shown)


public/components/custom_panels/home.tsx
```diff
@@ -92,6 +93,7 @@ export const Home = ({
   const createCustomPanel = (newCustomPanelName: string) => {
     if (!isNameValid(newCustomPanelName)) {
       setToast('Invalid Operational Panel name', 'danger');
+      history.goBack();
```
derek-ho, line 96: I missed this edge case - I will add it in a following PR, but I am wondering what does UX look like here - if somebody tries creating a panel from dashboards but then it fails because its an invalid name, do we want to land them back to the page before? Would toast even show up? Or what exactly is the process here

kavithacm: If someone tries to create a panel but ends up with invalid name error, they should always be redirected to panels home page or else the create panel functionality breaks and the user won't be able to create a panel without manually refreshing or redirecting back to original page.  So I chnaged history.goback to window.location.assign so that if there's no history to go back to, it would land on panels home page and also not break the create panel functionality

derek-ho (approved):

public/components/custom_panels/home.tsx
```diff
@@ -92,6 +93,7 @@ export const Home = ({
   const createCustomPanel = (newCustomPanelName: string) => {
     if (!isNameValid(newCustomPanelName)) {
       setToast('Invalid Operational Panel name', 'danger');
+      window.location.assign(`${_.last(parentBreadcrumbs)!.href}`);
```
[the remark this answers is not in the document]
kavithacm: The parentBreadcrumbs are defined in app.tsx explicitly, so i don't think it'll ever be null. Adding the link of the file below for reference.
https://github.com/opensearch-project/dashboards-observability/blob/main/public/components/app.tsx#L121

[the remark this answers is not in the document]
mengweieric: +1 Could you share the reason why use this way?

public/components/custom_panels/home.tsx
```diff
@@ -9,8 +9,8 @@ import { Toast } from '@elastic/eui/src/components/toast/global_toast_list';
 import _ from 'lodash';
 import React, { ReactChild, useState } from 'react';
 // eslint-disable-next-line @osd/eslint/module_migration
```
mengweieric, line 11: This may not be related to this PR, but I wondering if we could remove these rules as long as you don't have issue with these rules

public/components/custom_panels/home.tsx
```diff
@@ -297,9 +299,10 @@ export const Home = ({
         side={toastRightSide ? 'right' : 'left'}
         toastLifeTimeMs={6000}
       />
+      <Switch>
       <Route
         exact
-        path={renderProps.match.path}
+        path={['/operational_panels/create', '/operational_panels']}
```
mengweieric, line 305: is 'Create panel' now the also the home?

public/components/custom_panels/home.tsx
```diff
@@ -297,9 +299,10 @@ export const Home = ({
         side={toastRightSide ? 'right' : 'left'}
         toastLifeTimeMs={6000}
       />
+      <Switch>
```
mengweieric, line 302: Seems like indent issue, did you bootstrap your app and see husky running when you committing your code?

kavithacm: Will check now

public/components/custom_panels/home.tsx
mengweieric: ideally i think we should validate the input at creating page, so whenever user click create, the name should be valid.

public/components/custom_panels/home.tsx
kavithacm: '/operational_panels' is still the home but when used '/operational_panels/create', just the url changes and in the background its still home page

public/components/custom_panels/home.tsx
kavithacm: I haven’t checked if its currently in use, will remove it if its not used

anirudha (approved):

Issue #9, opened 2022-11-23
Convert Notebooks - New into full-page form


pjfitzgibbons:
Currently Notebooks raises a UI Dialog to create a new item.

In order to integrate Notebooks with Dashboards List, the New operation must navigate to an addressable URL -> a full-page form.

Will need to copy/re-write from previous PR into this repo : https://github.com/opensearch-project/observability/pull/1308

![Image](https://user-images.githubusercontent.com/11318/203637864-93ccb523-9347-4f9d-8aca-2c93b1d1b7ad.gif)

PR #283, opened 2023-02-16
add edit as rename modal open support

derek-ho:
### Description
Creates an edit path that hooks into a specific notebook page with rename modal open - will be sued for dashboard list integration

### Issues Resolved

### Check List

### DIFF
--- a/public/components/notebooks/components/notebook.tsx
+++ b/public/components/notebooks/components/notebook.tsx
@@ -707,7 +707,7 @@ export class Notebook extends Component<NotebookProps, NotebookState> {
     // If the URL changes to /edit or the notebook name is loaded in with the URL as /edit.
     // This second condition is to update the rename modal with the loaded notebook name on page laod
     // if the page if first loaded in as /edit.
-    if (prevProps.location !== this.props.location || prevState.path !== this.state.path) {
+    if (prevProps.location.pathname !== this.props.location.pathname || prevState.path !== this.state.path) {
       if (this.props.location.pathname.split('/').at(-1) === 'edit') {
         this.showRenameModal();
       }
--- a/public/components/notebooks/components/notebook.tsx
+++ b/public/components/notebooks/components/notebook.tsx
@@ -705,7 +705,7 @@ export class Notebook extends Component<NotebookProps, NotebookState> {
 
   componentDidUpdate(prevProps: NotebookProps, prevState: NotebookState) {
     // If the URL changes to /edit or the notebook name is loaded in with the URL as /edit.
-    // This second condition is to update the rename modal with the loaded notebook name on page laod
+    // This second condition is to update the rename modal with the loaded notebook name on page load
     // if the page if first loaded in as /edit.
     if (prevProps.location.pathname !== this.props.location.pathname || prevState.path !== this.state.path) {
       if (this.props.location.pathname.split('/').at(-1) === 'edit') {
--- a/public/components/notebooks/components/notebook.tsx
+++ b/public/components/notebooks/components/notebook.tsx
@@ -679,6 +679,9 @@ export class Notebook extends Component<NotebookProps, NotebookState> {
   }
 
   componentDidMount() {
+    if (this.props.location.pathname.split('/').at(-2) === 'edit') {
+      this.showRenameModal();
+    }
     this.setBreadcrumbs('');
     this.loadNotebook();
     this.checkIfReportingPluginIsInstalled();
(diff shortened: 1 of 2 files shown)


derek-ho:
@kgcreative @kavck  we have a few options to support edit URL access:
- Use replace notebooks/${id} with notebooks/edit/${id} (this option makes sense if we agree that someone looking to edit a notebook or operational panel doesn't necessarily mean they want to rename it - maybe they want to add a paragraph or modify it in some other way) - this will also mean any existing links directing to notebooks/${id} should also be changed to go to notebooks/edit/${id} - (@pjfitzgibbons /others any edge case concerns where UUID === edit?)
- When user drops in on notebooks/edit/${id} they will drop in on their notebook page with the rename modal open - when they close it we will not manipulate the URL in any way (this option is the least work and we don't need to muck around with manipulating browser histories - less edge cases, but also has the drawback that if the user bookmarks the notebooks/operational panel with /edit/ in the url, they will forever be redirected to their notebook/panel with the rename modal opened)
- When user drops in on notebooks/edit/${id} they will drop in on their notebook page with the rename modal open - when they close it we will change their URL to notebooks/${id}, and also remove notebooks/edit/${id} from their browser history - this last part is to ensure they do not have to see the modal pop up again in case they click back in their browser. This has the most edge cases (still not bad), but does require manipulation of browser histories. 


Please comment on which one you prefer, if anything is unclear I can try to come up with videos of the options, thanks!

CC @pjfitzgibbons @kavithacm

derek-ho:
For more context, initial issue is here: https://github.com/opensearch-project/dashboards-observability/issues/9, we want to do the same thing but for edit as well to support @pjfitzgibbons's work for dashboard list

kgcreative:
@derek-ho - If I think about the breadcrumbs in an "edit" interface, i think we would show: notebook > notebook name > edit. If that's the case, i'm wondering if the path of notebooks/${id}/edit might make more sense / be more discoverable. Thoughts?

derek-ho:
@kgcreative I see... Yea notebooks/id/edit makes sense to me... The question still remains - how do we handle the other aspects of this new path, mainly: 
1. Does `notebooks/id/edit` just bring you to the notebooks page (same as 'viewing' it)? Or does "edit" coming from dashboard list integration have to connect with rename modal?
2. If the answer to 1) is yes, we are done. If the answer to 1) is no, and it has to be connected to the rename modal, do we want to modify the url such that the /edit path is only present if the modal is open (involves messing with browser history to prevent modal from popping up on browser back button), or does it just persist in the url (will have issues where if user bookmarks the page with /edit in the url, they will always land on their notebook with the rename modal open).

kgcreative:
Do we have a way to pass state through from the previous route so that it takes you to the "details" page, but then opens the modal? (not sure if there is a way to pass a param so we don't have to re-write the URL). Ideally, we would land the user on the "details" page with the "Rename" modal open.

derek-ho:
@kgcreative I believe since we will be coming from another part of the app, namely core dashboards itself we would have no easy way to pass in this info without url manipulation (@pjfitzgibbons maybe you can chime in here). So it seems like we would want to have the modal open and then change the url based on if the modal is closed/open?

kgcreative:
I think that makes sense @derek-ho.
I think the remaining open question for me is one of flow.

There are two different ways we could do this, and i'm not sure which one "feels" better.

1. Dashboards list -> "edit link" -> notebook/id/edit [modal open] -> ok/cancel -> notebook details page. [back button should take the user back to dashboards page].

2. Dashboards list -> "edit link" -> notebook/id/edit [modal open] -> ok/cancel -> dashboard list [success toast].

I think 1 feels more natural.

derek-ho:
https://user-images.githubusercontent.com/26328171/219794133-02392b6e-d32e-42ec-9660-22331626992c.mov

@kgcreative @kavck @pjfitzgibbons @kavithacm created this short video, I think this is the expected behavior, let me know if you folks see anything missed/have any other comments, thanks! Keep an eye on the URL 👀

public/components/notebooks/components/notebook.tsx
```diff
@@ -694,6 +703,17 @@ export class Notebook extends Component<NotebookProps, NotebookState> {
     }
   }
 
+  componentDidUpdate(prevProps: NotebookProps, prevState: NotebookState) {
+    // If the URL changes to /edit or the notebook name is loaded in with the URL as /edit.
+    // This second condition is to update the rename modal with the loaded notebook name on page laod
```
joshuali925, line 708: np
```suggestion
    // This second condition is to update the rename modal with the loaded notebook name on page load
```

joshuali925 (approved):

mengweieric (approved):

derek-ho:
@pjfitzgibbons can you make a final call here

pjfitzgibbons:
I"ll go w/ @kgcreative 's 1st option :
Dashboards list -> "edit link" -> notebook/id/edit [modal open] -> ok/cancel -> notebook details page. [back button should take the user back to dashboards page].

PR #346, opened 2023-04-05
initial PR

derek-ho:
### Description

### Issues Resolved

### Check List

### DIFF
--- a/public/components/application_analytics/components/app_table.tsx
+++ b/public/components/application_analytics/components/app_table.tsx
@@ -125,6 +125,10 @@ export function AppTable(props: AppTableProps) {
     showModal();
   };
 
+  const BadComponent = () => {
+    throw new Error('test');
+  };
+
   const deleteApp = () => {
     const applicationString = `application${selectedApplications.length > 1 ? 's' : ''}`;
     setModalLayout(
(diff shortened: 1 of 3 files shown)


derek-ho:
https://user-images.githubusercontent.com/26328171/230158385-5ca72655-d072-4305-95b6-bcb0f7d4c092.mov

ps48:
This is good start to not crash the whole app, which we have seen in the past when there are UI leaks and edge cases. We also, add a clear direction to the user on what to do next, something as simple as:
1. Link to our forum thread where users can post the screenshots of errors
2. Maybe a prompt to refresh the page

derek-ho:
> This is good start to not crash the whole app, which we have seen in the past when there are UI leaks and edge cases. We also, add a clear direction to the user on what to do next, something as simple as:
> 
> 1. Link to our forum thread where users can post the screenshots of errors
> 2. Maybe a prompt to refresh the page

Sounds good and like it will be useful 👍 . I have updated the component to show like the following:
Screen Shot 2023-04-05 at 2 33 31 PM

Maybe @kavck can chime in here on whether it is an appropriate message or if it can be tweaked a bit.

pjfitzgibbons:
Looks great so far! I agree wrap all the top-level components this way.

One thing maybe we can work with UX is any way to tell the user "something went wrong"... and not confuse them with the technical detail (backtrace)... while giving "us" (developers) the technical detail when it's reported.

I think other apps represent this in a "report this bug" kind of feature.

At minimum, you're on the right path to giving the user some next-step upon experiencing the failure.

mengweieric:
Looks good, but do we want to hide the error stack and only show error itself?

public/components/metrics/index.tsx
```diff
@@ -81,7 +82,7 @@ export const Home = ({
     setToasts([...toasts, { id: new Date().toISOString(), title, text, color } as Toast]);
   };
 
-  const onRefreshFilters = (startTime: ShortDate, endTime: ShortDate) => {
+  const onRefreshFilters = (st: ShortDate, et: ShortDate) => {
```
derek-ho, line 85: it doesn't seem like these are being used - should we remove?

public/components/metrics/index.tsx
```diff
@@ -116,7 +117,7 @@ export const Home = ({
 
   useEffect(() => {
     if (!editMode) {
-      selectedMetrics.length > 0 ? setIsTopPanelDisabled(false) : setIsTopPanelDisabled(true);
+      setIsTopPanelDisabled(!(selectedMetrics.length > 0));
```
derek-ho, line 120: eslint forced me to make this change - it is logically equivalent ?

derek-ho:
> Looks great so far! I agree wrap all the top-level components this way.
> 
> One thing maybe we can work with UX is any way to tell the user "something went wrong"... and not confuse them with the technical detail (backtrace)... while giving "us" (developers) the technical detail when it's reported.
> 
> I think other apps represent this in a "report this bug" kind of feature.
> 
> At minimum, you're on the right path to giving the user some next-step upon experiencing the failure.



> Looks good, but do we want to hide the error stack and only show error itself?

Regarding these two comments, this is the standard EUI/OUI component for error boundary - I'm not sure it is worth it to invest time in making a custom component for us that only shows the top level error, as the stack trace does provide valuable context. Open for discussions about this, though. In addition to reporting to github/forum - should we also ask them for screenshot or open console to show any previous context? Or is this too complex?

derek-ho:
Putting in draft since in its current form EUI/OUI does not allow to shim in a custom message/component to be rendered with the error boundary. Raised an issue with the oui repo to discuss whether we can add support on that end, or if we should implement a custom component that allows us to show the error message

public/components/metrics/index.tsx
ps48: looks good to me. The prior one is more readable though. An easy way to check this is to use the edit button can check manually.

public/components/metrics/index.tsx
ps48: good to remove

public/components/metrics/index.tsx
pjfitzgibbons: Uhh.. based on the terminology :
`setIsTopPanelDisabled(selectedMetrics.length == 0);`

@ps48 would that be readable enough?

PS: Tertiary conditions are almost always less readable.  This sentiment is > IMHO

derek-ho:
Closing this as stale - if maintainers want me to take another look happy to do so. Code changed a lot though

PR #356, opened 2023-04-14
bulk update separation

derek-ho:
### Description
Delete fixed (minus toast for delete and rename), separate out bulk update for event analytics add viz to operational panel flow. Still need to do bulk update for SO to add new and old viz

### Issues Resolved

### Check List

### DIFF
--- a/public/components/custom_panels/helpers/add_visualization_helper.ts
+++ b/public/components/custom_panels/helpers/add_visualization_helper.ts
@@ -58,7 +58,7 @@ export const addVisualizationPanel = (
   // client: ILegacyScopedClusterClient,
   // panelId: string,
   savedVisualizationId: string,
-  oldVisualizationId?: string,
+  oldVisualizationId: string | undefined,
   allPanelVisualizations: VisualizationType[]
 ) => {
   try {
--- a/public/components/custom_panels/home.tsx
+++ b/public/components/custom_panels/home.tsx
@@ -32,18 +32,12 @@ import {
   ObservabilityPanelAttrs,
   PanelType,
 } from '../../../common/types/custom_panels';
-import { ObservabilitySideBar } from '../common/side_nav';
 import { CustomPanelTable } from './custom_panel_table';
 import { CustomPanelView } from './custom_panel_view';
 import { isNameValid } from './helpers/utils';
 import { CustomPanelViewSO } from './custom_panel_view_so';
 import { coreRefs } from '../../framework/core_refs';
-<<<<<<< HEAD
-import { CustomPanelType } from '../../../common/types/custom_panels';
-import { deletePanel, fetchPanels } from './redux/panel_slice';
-=======
-import { fetchPanels } from './redux/panel_slice';
->>>>>>> 5f2b777a2c7963680f4bea3746e6badea65b9303
+import { deletePanel, fetchPanels, uuidRx } from './redux/panel_slice';
 
 // import { ObjectFetcher } from '../common/objectFetcher';
 
@@ -147,8 +141,6 @@ export const Home = ({
         console.error('create error', err);
       });
   };
-
-  const uuidRx = /^[0-9a-fA-F]{8}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{12}$/;
 
   const isUuid = (id) => !!id.match(uuidRx);
 
--- a/public/components/custom_panels/home.tsx
+++ b/public/components/custom_panels/home.tsx
@@ -12,6 +12,7 @@ import { StaticContext } from 'react-router';
 import { Route, RouteComponentProps, Switch } from 'react-router-dom';
 import { map, mergeMap, tap, toArray } from 'rxjs/operators';
 import { concat, from, Observable, of } from 'rxjs';
+import { useDispatch } from 'react-redux';
 import PPLService from '../../services/requests/ppl';
 import DSLService from '../../services/requests/dsl';
 import { CoreStart, SavedObjectsStart } from '../../../../../src/core/public';
@@ -39,8 +40,7 @@ import { SavedObject } from '../../../../../src/core/types';
 import { CustomPanelViewSO } from './custom_panel_view_so';
 import { coreRefs } from '../../framework/core_refs';
 import { CustomPanelType } from '../../../common/types/custom_panels';
-import { fetchPanels } from './redux/panel_slice';
-import { useDispatch } from 'react-redux';
+import { deletePanel, fetchPanels } from './redux/panel_slice';
 
 // import { ObjectFetcher } from '../common/objectFetcher';
 
@@ -80,7 +80,7 @@ export const Home = ({
   const [start, setStart] = useState<ShortDate>('');
   const [end, setEnd] = useState<ShortDate>('');
 
-  const dispatch = useDispatch()
+  const dispatch = useDispatch();
 
   const setToast = (title: string, color = 'success', text?: ReactChild, side?: string) => {
     if (!text) text = '';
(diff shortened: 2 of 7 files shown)


public/services/saved_objects/saved_object_savers/ppl/save_as_current_vis.ts
```diff
@@ -46,9 +46,12 @@ export class SaveAsCurrentVisualization extends SavedQuerySaver {
   }
 
   addToPanel({ selectedPanels, saveTitle, notifications, visId }) {
+    const uuidRx = /^[0-9a-fA-F]{8}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{12}$/;
+    const soPanels = selectedPanels.filter((id) => id.panel.id.match(uuidRx));
```
joshuali925, line 50: duplicate and unused variable?

public/services/saved_objects/saved_object_savers/ppl/save_as_current_vis.ts
```diff
@@ -46,9 +46,12 @@ export class SaveAsCurrentVisualization extends SavedQuerySaver {
...
+    const opsPanels = selectedPanels.filter((id) => !id.panel.id.match(uuidRx));
```
joshuali925, line 51: - use `uuidRx.test(id.panel.id)`
- `id` variable name is not accurate
- i'm not sure if the logic should be here, e.g. what happens to panels in .observability index? `save_as_current_vis` don't support them anymore?
- not sure if it helps but you can check .observability and .kibana visualizations handling here as reference, for example bulk delete https://github.com/opensearch-project/dashboards-observability/blob/775a77ebda4ab64abdd20bab01841ccf94165edd/public/services/saved_objects/saved_object_client/saved_objects_actions.ts#L74-L88

public/services/saved_objects/saved_object_savers/ppl/save_as_new_vis.ts
```diff
@@ -76,9 +76,12 @@ export class SaveAsNewVisualization extends SavedQuerySaver {
   }
 
   addToPanel({ selectedPanels, saveTitle, notifications, visId }) {
+    const uuidRx = /^[0-9a-fA-F]{8}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{12}$/;
```
joshuali925, line 79: similar comments and probably define regex globally

public/services/saved_objects/saved_object_savers/ppl/save_as_current_vis.ts
```diff
@@ -46,9 +47,8 @@ export class SaveAsCurrentVisualization extends SavedQuerySaver {
   }
 
   addToPanel({ selectedPanels, saveTitle, notifications, visId }) {
-    const uuidRx = /^[0-9a-fA-F]{8}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{12}$/;
-    const soPanels = selectedPanels.filter((id) => id.panel.id.match(uuidRx));
-    const opsPanels = selectedPanels.filter((id) => !id.panel.id.match(uuidRx));
+    const soPanels = selectedPanels.filter((panel) => panel.panel.id.test(uuidRx));
+    const opsPanels = selectedPanels.filter((panel) => !panel.panel.id.test(uuidRx));
```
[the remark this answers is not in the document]
derek-ho: Copy and paste issue, copied old file to new instead of new to old, let me push a fix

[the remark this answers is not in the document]
derek-ho: Fixed in [19b9c9f](https://github.com/opensearch-project/dashboards-observability/pull/356/commits/19b9c9f7dd45e1b1531e750b1c1cfaf74b614a5d)

joshuali925 (approved):

public/components/custom_panels/redux/panel_slice.ts
```diff
@@ -112,34 +113,47 @@ export const fetchPanel = (id) => async (dispatch, getState) => {
...
   }
 };
 
+export const addVizToPanels = (panels, vizId) => async (dispatch, getState) => {
+  forEach(panels, (oldPanel) => {
+    const panel = getState().customPanel.panelList.find((p) => p.id === oldPanel.panel.id);
+
+    const allVisualizations = panel!.visualizations;
+
+    const visualizationsWithNewPanel = addVisualizationPanel(vizId, undefined, allVisualizations);
```
derek-ho, line 146: @ps48 @joshuali925 can you verify this line makes sense? Since we want to add a viz to panel as a new one, pass in undefined into this function: https://github.com/opensearch-project/dashboards-observability/blob/5f2b777a2c7963680f4bea3746e6badea65b9303/public/components/custom_panels/helpers/add_visualization_helper.ts#L57?

public/services/saved_objects/saved_object_savers/ppl/save_as_new_vis.ts
```diff
@@ -76,9 +83,14 @@ export class SaveAsNewVisualization extends SavedQuerySaver {
   }
 
   addToPanel({ selectedPanels, saveTitle, notifications, visId }) {
+    const { dispatch } = this.dispatchers;
```
pjfitzgibbons, line 86: Please consider context here, as well as SRP-SingleResponsibilityPrinciple.

Wer are inside "saved_objects/saved_object_savers".... This, despite it's name, is the Observability index re-invention of saved objects.
We don't want to add any code here.  This whole module will (eventually) be deprecated/removed.

Also, wrt. SRP,  although "saving things" sounds like a single responsibliity, we can see from the code that saving to SavedObject-Core (via Redux/Dispatch) is _very_ different from using SavedObjectSavers.
These are different responsibilities.


Recommendation : 
Lift the filter/split of saved-object (UUID) and Observability (non-UUID) out of this method.  Pass the non-UUID into this method, and specifically call dispatch from the caller (which is a component, which is insdie React context, which has correct context for `useDispatch()` ).    This will also remove the prop-drilling of "dispatchers".

joshuali925: > we can see from the code that saving to SavedObject-Core (via Redux/Dispatch) is very different from using SavedObjectSavers

Why does this happen? I didn't look into impl details but essentially the difference between saving SO vs. observability object is the client. If the client can be abstracted, then the same operation should work for both cases

public/components/custom_panels/helpers/add_visualization_helper.ts
```diff
@@ -58,7 +58,7 @@ export const addVisualizationPanel = (
   // client: ILegacyScopedClusterClient,
   // panelId: string,
   savedVisualizationId: string,
-  oldVisualizationId?: string,
+  oldVisualizationId: string | undefined,
```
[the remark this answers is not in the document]
derek-ho: Maybe... I think what it is doing is makes it optional, which it cannot be since a required parameter comes after. This gets rid of it but maybe the more right thing is to put it at the end...

public/components/custom_panels/home.tsx
```diff
@@ -358,7 +356,7 @@ export const Home = ({
                 chrome={chrome}
                 parentBreadcrumbs={customPanelBreadCrumbs}
                 cloneCustomPanel={cloneCustomPanel}
-                deleteCustomPanel={deleteCustomPanel}
+                deleteCustomPanel={deleteCustomPanelSO}
```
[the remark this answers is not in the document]
derek-ho: I can modify this in a follow up PR

public/services/saved_objects/saved_object_savers/ppl/save_as_new_vis.ts
derek-ho: @joshuali925 @pjfitzgibbons can we merge this in as is? I understand it may not be right but theres a few other p0 items today, once we make the final go-no go call I can refactor to be correct

public/components/custom_panels/redux/panel_slice.ts
ps48: Yes, this is correct for adding panels to a visualization. For replace workflow, you can use the same function and pass OldVisualizationId.

PR #350, opened 2023-04-10
Left-Nav app registrations for Observability - WIP

pjfitzgibbons:
### Description
Modify Observability plugin registration to make Observability a "top level" grouping on the OSD Left Navigation fly-out.

The P3/P4 result will look like :

image (3)

Our P1 goal is only the Observability Grouping :
Applications
Logs
Metrics
Traces
Notebooks

Note that Traces does not have an indpendent module at this time.
Also note that "Logs" is our current Event Explorer

The "Observability" link under the "OpenSearch Plugins" group will remain, with it's default page-link as Event Explorer.

### Issues Resolved
resolves #34 

### Check List
- [ ] New functionality includes testing.
  - [ ] All tests pass, including unit test, integration test and doctest
- [ ] New functionality has been documented.
  - [ ] New functionality has javadoc added
  - [ ] New functionality has user manual doc added
- [ ] Commits are signed per the DCO using --signoff

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #360, opened 2023-04-15
Fix trace analytics navigaiton

ps48:
### Description
Fix trace Analytics and add side nav for child components

### Issues Resolved
#350 

### Check List

### DIFF
--- a/public/components/trace_analytics/trace_side_nav.tsx
+++ b/public/components/trace_analytics/trace_side_nav.tsx
@@ -1,0 +1,74 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import {
+  EuiButton,
+  EuiFlexGroup,
+  EuiFlexItem,
+  EuiPage,
+  EuiPageBody,
+  EuiPageSideBar,
+  EuiSideNav,
+  EuiSideNavItemType,
+  EuiSwitch,
+} from '@elastic/eui';
+import React from 'react';
+
+export function TraceSideBar(props: { children: React.ReactNode }) {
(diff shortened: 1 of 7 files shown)


public/components/trace_analytics/trace_side_nav.tsx
```diff
@@ -0,0 +1,74 @@
...
+  EuiPage,
+  EuiPageBody,
+  EuiPageSideBar,
+  EuiSideNav,
+  EuiSideNavItemType,
+  EuiSwitch,
+} from '@elastic/eui';
+import React from 'react';
+
+export function TraceSideBar(props: { children: React.ReactNode }) {
```
joshuali925, line 19: `ObservabilitySideBar` doesn't work?

joshuali925 (approved):

pjfitzgibbons (commented):
When opening the Traces module, the Dashboard is displayed.
What would be displayed on Dashboard?
How is that different than the Jaeger or Services views ?

ps48:
> When opening the Traces module, the Dashboard is displayed. What would be displayed on Dashboard? How is that different than the Jaeger or Services views ?

Trace Analytics by default shows a Dashboards with has an accumulated view of all traces and services. `Jaeger` and `Data Prepper` are sources for traces. The Traces view and Services show trace group and services drill down respectively. 

The overall functionality doesn't change here. It is the same as it was earlier. More info [here](https://opensearch.org/docs/1.2/observability-plugin/trace/ta-dashboards/)

public/components/trace_analytics/trace_side_nav.tsx
ps48: @joshuali925 It had some issues with home page so I just added this part to keep things simple. Will remove observability side bar later.

PR #369, opened 2023-04-17
Fix crash at loading app analytics tabs

joshuali925:
### Description
For future refactoring: it would be better to use core chrome service to get current breadcrumbs, then append new ones to it, rather than passing down parentBreadcrumbs as props.

Passing parent down can be inconsistent (this case breadcrumb vs array<breadcrumbs>) and redundant

### Issues Resolved

### Check List


public/components/trace_analytics/components/dashboard/__tests__/__snapshots__/dashboard.test.tsx.snap
```diff
@@ -305,7 +327,29 @@ exports[`Dashboard component renders dashboard 1`] = `
...
         "setBadge": [MockFunction],
         "setBrand": [MockFunction],
-        "setBreadcrumbs": [MockFunction],
+        "setBreadcrumbs": [MockFunction] {
+          "calls": Array [
+            Array [
+              Array [
+                undefined,
+                Object {
+                  "href": "#/trace_analytics/home",
```
rupal-bq, line 336: is `trace_analytics` in url works? I thought it was removed in earlier pr

anirudha (approved):
thanks

rupal-bq (approved):
LGTM

public/components/trace_analytics/components/dashboard/__tests__/__snapshots__/dashboard.test.tsx.snap
joshuali925: that is probably from test files https://github.com/opensearch-project/dashboards-observability/blob/544c2f50521501c87deacf3136b0596686467821/public/components/trace_analytics/components/dashboard/__tests__/dashboard.test.tsx#L39

The unit test doesn't care what it should be in reality, as long as dummy arguments are all valid it serves its purpose

PR #427, opened 2023-04-25
change samples to saved object

derek-ho:
### Description
Saves sample panels as saved object instead of observability ops panel

### Issues Resolved
[List any issues this PR will resolve]

### Check List
- [ ] New functionality includes testing.
  - [ ] All tests pass, including unit test, integration test and doctest
- [ ] New functionality has been documented.
  - [ ] New functionality has javadoc added
  - [ ] New functionality has user manual doc added
- [ ] Commits are signed per the DCO using --signoff

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #436, opened 2023-04-27
change samples to saved object (#427)

derek-ho:
* change samples to saved object

* add sample to so

* fix saved object sample

* fix up

* add back toast

* revert file

* refactoring

* use constant

---------

(cherry picked from commit ce06ccb9fc4acd75d3ccbfbacab9db8a521de252)

### Description

### Issues Resolved

### Check List


ps48 (approved):

common/constants/custom_panels.ts
```diff
@@ -11,3 +13,79 @@ export const CREATE_PANEL_MESSAGE = 'Enter a name to describe the purpose of thi
 export const CUSTOM_PANELS_SAVED_OBJECT_TYPE = 'observability-panel';
 
 export const CUSTOM_PANEL_SLICE = 'customPanel';
+
+export const samplePanelName = '[Logs] Web traffic Panel';
+
+export const createDemoPanel = (savedVisualizationIds: string[]) => {
+  return {
```
pjfitzgibbons, line 20: I think if we're explicitly indexing the savedVisualizationIds then we should at least assert the list-length on the way in.

An alternative would be to put the list of panel-locations into its own array and zip() the savedVisualizationIds and panelLocations... then take whatever has [not-null, not-null].

Might be a lot easier to read both the locations list (could be listed horizontally), and the structure of the visualization (listed once in a map() from the zip() result).

pjfitzgibbons (approved):

Issue #530, opened 2023-06-12
[BUG] Log patterns UI regressions


joshuali925:
**What is the bug?**
Bug 1. Patterns UI opens when a query is sent, but does not show actual pattern unless user hide and show patterns again.
Bug 2. clicking on a pattern line should filter logs by the pattern.

**How can one reproduce the bug?**
1. Go to event analytics
2. Run `source = opensearch_dashboards_sample_data_logs`
3. See `No patterns found.` on page (bug 1)
4. Click "Hide patterns", then "Show patterns"
5. Patterns show up correctly
6. Click on a pattern line, it gets highlighted but does not filter events (bug 2)

**What is the expected behavior?**
Pattern should show up on initial page load, and filter should work.

**What is your host/environment?**
 - Plugins

**Do you have any screenshots?**

**Do you have any additional context?**
Bug 1 is introduced by #444, bug 2 probably appeared after event analytics refactoring

PR #513, opened 2023-06-05
BugFix Application Analytics timestamp query error

TackAdam:
### Description
[Describe what this change achieves]
BugFix Application Analytics timestamp query error.

### Issues Resolved
[List any issues this PR will resolve]

### Check List
- [ ] New functionality includes testing.
  - [ ] All tests pass, including unit test, integration test and doctest
- [ ] New functionality has been documented.
  - [ ] New functionality has javadoc added
  - [ ] New functionality has user manual doc added
- [ ] Commits are signed per the DCO using --signoff

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #538, opened 2023-06-13
Bugfix: Event Analytics Filters

TackAdam:
### Description
Fixed filters not being refreshed/ needing to hide then show.
Fixed click on filters not being applied visually
Patterns working fully in Application Analytics and Logs. On refresh the patterns field is cleared to allow a new search or filter to be applied

### Issues Resolved
https://github.com/opensearch-project/dashboards-observability/issues/530

### Check List

### DIFF
--- a/public/components/event_analytics/explorer/explorer.tsx
+++ b/public/components/event_analytics/explorer/explorer.tsx
@@ -727,7 +727,7 @@ export const Explorer = ({
       if (availability !== true) {
         await updateQueryInStore(tempQuery);
       }
-      await fetchData();
+      await fetchData(startTime, endTime);
     },
     [tempQuery, query]
   );
--- a/public/components/event_analytics/explorer/explorer.tsx
+++ b/public/components/event_analytics/explorer/explorer.tsx
@@ -556,7 +556,6 @@ export const Explorer = ({
                     <EuiHorizontalRule margin="xs" />
                     <LogPatterns
                       selectedIntervalUnit={selectedIntervalRef.current}
-                      setTempQuery={setTempQuery}
                       handleTimeRangePickerRefresh={handleTimeRangePickerRefresh}
                     />
                   </>
--- a/public/components/event_analytics/explorer/log_patterns/log_patterns.tsx
+++ b/public/components/event_analytics/explorer/log_patterns/log_patterns.tsx
@@ -69,6 +69,15 @@ const EventPatterns = ({
       })
     );
     handleTimeRangePickerRefresh(true);
+    // after rendering the patterns visual, we want the pattern to be reset for future searches
+    await dispatch(
+      changeQuery({
+        tabId,
+        query: {
+          [FILTERED_PATTERN]: '',
+        },
+      })
+    );
   };
 
   const showToastError = (errorMsg: string) => {
--- a/public/components/event_analytics/explorer/log_patterns/log_patterns.tsx
+++ b/public/components/event_analytics/explorer/log_patterns/log_patterns.tsx
@@ -25,7 +25,6 @@ export interface LogPatternProps {
     text: string;
     value: string;
   };
-  setTempQuery: () => string;
   handleTimeRangePickerRefresh: (flag: boolean) => {};
   patterns: PatternTableData[];
   query: IQuery;
@@ -33,7 +32,6 @@ export interface LogPatternProps {
 
 const EventPatterns = ({
   selectedIntervalUnit,
-  setTempQuery,
   handleTimeRangePickerRefresh,
   patterns,
   query,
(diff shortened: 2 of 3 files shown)


public/components/event_analytics/explorer/log_patterns/log_patterns.tsx
```diff
@@ -49,11 +49,16 @@ const EventPatterns = ({
     requestParams: { tabId },
   });
 
+  // refresh filters on opening page
```
derek-ho, line 52:
```suggestion
  // refresh patterns on opening page
```

public/components/event_analytics/explorer/log_patterns/log_patterns.tsx
```diff
@@ -49,11 +49,16 @@ const EventPatterns = ({
...
+  useEffect(() => {
+    onPatternSelection('');
+  }, []);
+
   const onPatternSelection = async (pattern: string) => {
     if (query[FILTERED_PATTERN] === pattern) {
       return;
     }
-    dispatch(
+    await dispatch(
```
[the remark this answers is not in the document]
TackAdam: Adding comment above, but some reference on why it is needed:
https://stackoverflow.com/questions/63686900/is-it-possible-to-await-for-dispatch-is-finish
https://www.reddit.com/r/reactjs/comments/w5roly/whyawait_dispatch_works_redux/

public/services/data_fetchers/ppl/ppl_data_fetcher.ts
TackAdam, line 1: https://github.com/opensearch-project/dashboards-observability/pull/449
Fixed a timestamp over-riding bug by removing the use of rawQueryStr
https://github.com/opensearch-project/dashboards-observability/pull/513
With patterns originally not working this caused errors in Logs-Explorer that required this rawQueryStr to be built. After the changes introduced to get patterns working correctly the UpdateQueryState no longer required this change and can be reverted.

public/components/event_analytics/explorer/explorer.tsx
```diff
@@ -727,7 +727,7 @@ export const Explorer = ({
       if (availability !== true) {
         await updateQueryInStore(tempQuery);
       }
-      await fetchData();
+      await fetchData(startTime, endTime);
```
derek-ho, line 730: Can you create an issue/discussion point regarding this change? I think with this change all instances of fetchdata will be called with startTime/endTime, so we can remove the dependency on startTime/endTime from the query in redux - or vice versa. In my opinion it is bad to have two references/sources of truth to anything (timerange/patterns, etc.), so either we should remove it from the redux query if it is not being used, or only rely on time from query in redux - not both @mengweieric thoughts?

public/components/event_analytics/explorer/log_patterns/log_patterns.tsx
```diff
@@ -49,21 +49,35 @@ const EventPatterns = ({
...
       changeQuery({
         tabId,
         query: {
           [FILTERED_PATTERN]: pattern,
         },
       })
     );
-    // workaround to refresh callback and trigger fetch data
-    await setTempQuery(query[RAW_QUERY]);
```
derek-ho, line 65: same comment here about the tempquery - can we remove it all together? If it is necessary can we consolidate it into the query in redux instead of having a separate useState variable?

derek-ho (approved):

public/components/event_analytics/explorer/explorer.tsx
mengweieric: Agreed with an issue to track and document this and to have a consistent way of set time range in both app and explorer sides. We may want to remove explicit start/end time dependency for explorer and instead, dispatch timerange to app query state. This only does the trick to make it work but would potentially cause more issues and bugs

public/services/data_fetchers/ppl/ppl_data_fetcher.ts
ps48: Should we also remove, `buildRawQuery` function from Utils?

TackAdam: Still needed in one other location in ppl_data_fetcher line 70
this.queryIndex = this.getIndex(buildRawQuery(query, appBaseQuery));

public/components/event_analytics/explorer/log_patterns/log_patterns.tsx
TackAdam: Removed tempQuery from log patterns, It is still used in explorer to handle query searches and changes.

mengweieric (approved):
Also tested on event explorer side?

TackAdam:
> Also tested on event explorer side?

Yes, seems to be all working.

PR #713, opened 2023-07-18
Integrations base path fix

### LINKED ISSUE #714: [BUG] Integrations plugin doesn't work if base path is configured
**What is the bug?**
When run with a base path, the integrations plugin fails in a few places. Images aren't loaded, and the "Try it" button fails as well.

1. Start OpenSearch Dashboards with `yarn start`. Normally we've developed with `--no-base-path`, and the test environments reflect this as well.

Regardless of if `--no-base-path` is passed or not, the plugin should function the same way.

 - Mac
 - Commit 5c96cc62

N/A

N/A

Swiddis:
### Description
The assumption that dashboards would always be run with `--no-base-path` turns out to be false, so we need to update a few paths to apply the base path. This PR does those updates, switching from fetch to the http client for all API calls.

### Issues Resolved
#714 

### Check List


public/components/integrations/components/add_integration_flyout.tsx
```diff
@@ -131,10 +131,16 @@ export function AddIntegrationFlyout(props: IntegrationFlyoutProps) {
   const fetchDataSourceMappings = async (
     targetDataSource: string
   ): Promise<{ [key: string]: { properties: any } } | null> => {
-    return fetch(`/api/console/proxy?path=${targetDataSource}/_mapping&method=GET`, {
-      method: 'POST',
-      headers: [['osd-xsrf', 'true']],
-    })
+    return fetch(
+      http.basePath.prepend(`/api/console/proxy?path=${targetDataSource}/_mapping&method=GET`),
```
joshuali925, line 135: why is this going through console (devtools) 's api?

also i think if you use `http.get('/api/console')` or `http.get('./api/console')` it works without manually adding base path

public/components/integrations/components/add_integration_flyout.tsx
```diff
@@ -131,10 +131,16 @@ export function AddIntegrationFlyout(props: IntegrationFlyoutProps) {
   const fetchDataSourceMappings = async (
     targetDataSource: string
   ): Promise<{ [key: string]: { properties: any } } | null> => {
-    return fetch(`/api/console/proxy?path=${targetDataSource}/_mapping&method=GET`, {
-      method: 'POST',
-      headers: [['osd-xsrf', 'true']],
-    })
+    return fetch(
```
joshuali925, line 134: todo item fetch should be replaced by osd's http client. compatibility is better

public/components/integrations/components/add_integration_flyout.tsx
```diff
@@ -91,7 +91,7 @@ export const doPropertyValidation = (
 };
 
 export function AddIntegrationFlyout(props: IntegrationFlyoutProps) {
-  const { onClose, onCreate, integrationName, integrationType } = props;
+  const { onClose, onCreate, integrationName, integrationType, http } = props;
```
joshuali925, line 94: note only. clients/services like http might be better to be stored as context value or somewhere global. passing through props can get messy since many things need it

YANG-DB (approved):

public/components/integrations/components/add_integration_flyout.tsx
ps48: You can use this: https://github.com/opensearch-project/dashboards-observability/blob/main/public/framework/core_refs.ts

Swiddis: Just switched to using coreRefs

derek-ho (approved):

Issue #757, opened 2023-07-26
[BUG]Integration loading of a dashboard fails for ndjson empty lines


YANG-DB:
**What is the bug?**

Integration loading of a dashboard fails when empty spaces appear in the ndjson file

This works well in the savedObject import api 

**What is the expected behavior?**
Success loading the ndjson dashboard even if it has empty lines - fix the integration parser to trim the ndjson

Swiddis:
Duplicate of #743

Issue #743, opened 2023-07-24
[BUG] Integration asset parsing behavior doesn't match Saved Object API


Swiddis:
**What is the bug?**
While integration assets and the saved object API both use NDjson for sharing objects, the two have different behavior. In particular, it seems like the saved object API can handle empty lines, while the integration assets raises an exception. [These are both in line with the spec](https://github.com/ndjson/ndjson-spec#32-parsing), but they should not disagree with each other.

**How can one reproduce the bug?**
1. Take any valid NDJson Saved Object export.
2. Add a newline to the end.
3. Import that same export with the newline and observe that the import is successful.
4. Try to load the same export as an integration asset and the integration fails to validate.

**What is the expected behavior?**
The saved object API and the integration asset parser should have the same behavior.

**What is your host/environment?**
- Mac, commit dd278ade

**Do you have any screenshots?**
N/A

**Do you have any additional context?**
Documentation on how the saved object API parses ndjson would be helpful, in addition to the newline issue there could be other implementation quirks that we might want to copy. It makes a better user experience to have the same parser behavior everywhere.

Swiddis:
See: [Source code for current saved object collection](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/6bab7eff724913d4657a015b268fd6d00d7bfaf2/src/core/server/saved_objects/import/collect_saved_objects.ts#L51-L102)

PR #758, opened 2023-07-26
fix integration json loading parser so that it will trim unneeded emp…

YANG-DB:
trim empty spaces from the ndjson file and match the existing behavior of the savedObject import dashboard that works even when empty spaces appear in the ndjson file

### Description
Integration loading of a dashboard fails when empty spaces appear in the ndjson file

This works well in the savedObject import api

### What is the expected behavior?
Success loading the ndjson dashboard even if it has empty lines - fix the integration parser to trim the ndjson

### Issues Resolved
https://github.com/opensearch-project/dashboards-observability/issues/757

### Check List

### DIFF
--- a/server/adaptors/integrations/repository/integration.ts
+++ b/server/adaptors/integrations/repository/integration.ts
@@ -211,7 +211,7 @@ export class Integration {
       );
       try {
         const ndjson = await fs.readFile(sobjPath, { encoding: 'utf-8' });
-        const asJson = '[' + ndjson.replace(/\n/g, ',') + ']';
+        const asJson = '[' + ndjson.trim().replace(/\n/g, ',') + ']';
         const parsed = JSON.parse(asJson);
         result.savedObjects = parsed;
       } catch (err: any) {


server/adaptors/integrations/repository/integration.ts
```diff
@@ -211,7 +211,7 @@ export class Integration {
       );
       try {
         const ndjson = await fs.readFile(sobjPath, { encoding: 'utf-8' });
-        const asJson = '[' + ndjson.replace(/\n/g, ',') + ']';
+        const asJson = '[' + ndjson.trim().replace(/\n/g, ',') + ']';
```
Swiddis, line 214: This breaks NDJson compliance. The [spec](https://github.com/ndjson/ndjson-spec#32-parsing) says that the two permissible behaviors are to ignore all blank lines, or raise an exception on any blank line. If we want to switch to the former mode of operation, we need to handle all blank lines, not just ones at the start or end of the file.

Swiddis, line 214: Should probably include a test case that has an empty line to avoid regressions if we refactor this later.

Swiddis (changes requested):

server/adaptors/integrations/repository/integration.ts
YANG-DB: as stated - regardless of the spec it should work identically to how the savedObject import API work - if that works with the empty lines so does this code

Swiddis: Where is the saved object import format specified? It's mentioned in #743 that other implementation quirks make this nontrivial. We shouldn't assume they share the same behavior based on one example.

YANG-DB: why should we test this case ? its not transitioning the data but only removed none necessary empty lines - if none exist it does nothing...
how would we possibly break this in the future ?

Swiddis (approved):

server/adaptors/integrations/repository/integration.ts
pjfitzgibbons: If it's simple to test, "how would we possibly break this in the future" is exactly the kind of unknown to test-proof.  since this is a data-modifying function, I'd lean toward producing some sort of test.

pjfitzgibbons, line 214: As commented in resovled : If it's simple to test, "how would we possibly break this in the future" is exactly the kind of unknown to test-proof. since this is a data-modifying function, I'd lean toward producing some sort of test.

pjfitzgibbons (approved):

PR #782, opened 2023-07-31
Make integration category and template lists dynamic

### LINKED ISSUE #780: [BUG]Integration categories not visible
**What is the bug?**
I cant see all the categories of the available integratiobs

As shown here the `aws` integrations are not present nor selectable for filtering
Screenshot 2023-07-31 at 2 06 15 PM

ALL categories should be available for filtering within the integration list

Swiddis:
### Description
Removes instance of hardcoded integration categories list. Also removes an unrelated hardcoded template list that results in a very similar bug in the added integration list.

### Issues Resolved
Resolves #780 

### Check List

### DIFF
--- a/common/constants/integrations.ts
+++ b/common/constants/integrations.ts
@@ -7,5 +7,3 @@ export const OPENSEARCH_DOCUMENTATION_URL = 'https://opensearch.org/docs/latest/
 export const ASSET_FILTER_OPTIONS = ['index-pattern', 'search', 'visualization', 'dashboard'];
 // TODO get this list dynamically from the API
 export const INTEGRATION_TEMPLATE_OPTIONS = ['nginx'];
-// TODO get this list dynamically from the API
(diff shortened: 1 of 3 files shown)


YANG-DB (approved):

common/constants/integrations.ts
```diff
@@ -5,7 +5,3 @@
 
 export const OPENSEARCH_DOCUMENTATION_URL = 'https://opensearch.org/docs/latest/integrations/index';
 export const ASSET_FILTER_OPTIONS = ['index-pattern', 'search', 'visualization', 'dashboard'];
```
derek-ho, line 7: if it is easy can we also include the asset filter not being hard coded?

Swiddis: Should be identical to how we removed the template list, though since we're repeating the deduplicate+sort logic a bunch we should try to centralize it to a general method somewhere. Can look into it later

Swiddis: Update: Since due dates are shifted let's just merge the current version and deal with this in a separate PR @derek-ho

PR #818, opened 2023-08-03
Fix index name collisions on integration sample creation.

### LINKED ISSUE #812: [BUG]Conflicting logs index templates
**What is the bug?**
when attempting to add multiple integrations that contain log templates - the patterns collide and prevent correct creation of the specific integration's log index template

Setup AWS-ELB integration
Setup Nginx integration
The logs index templates are conflicting since they both share the next index pattern:
```Json5
    "index_patterns": [
        "ss4o_logs-*-*"
    ],
    "data_stream": {},
    "template": {
   ....
```

Since the logs index templates are not created in advance for all types of possible log combinations, 
We need a clear distinction between each log associated for each integration:

1. Each integration will include the log mapping with a distinct name:
- ELB will contain `logs-elb` where the index pattern inside the index template will be

```Json5
    "index_patterns": [
        "ss4o_logs-elb-*"
    ],
    "data_stream": {},
    "template": {
   ....
```

- nginx will contain `logs-nginx` where the index pattern inside the index template will be
```Json5
    "index_patterns": [
        "ss4o_logs-nginx-*"
    ],
    "data_stream": {},
    "template": {
   ....
```

And so on for the rest of the common log index templates.
----

This will prevent the collision both in the index-pattern and both in the index mapping template name that will be very explicit for which integration it addresses - this also makes the template priority redundant ...

## Simplified query access
In addition to simplify the queries access to the index - each mapping will contain a shortened version of the index name - AKA alias:

For  `ss4o_logs-elb-*` the template will contain the next index alias:
```json5
    "aliases": {
      "logs-elb": {}
    },
```
This will simplify the queries and access patterns to the generated log indices and define the standard access convention in a meaningful format .

[Avoid index pattern collisions
](https://www.elastic.co/guide/en/elasticsearch/reference/current/index-templates.html)

Swiddis:
### Description
When updating an index template with the same name, OpenSearch rejects the request if it invalidates any old indices. This affects the integration plugin's sample data creation, since it was previously only using type information for creating indices, and this led to collisions with index patterns. This PR also adds name information to the created indices, in order to avoid these collisions.

### Issues Resolved
Resolves #812 

### Check List

### DIFF
--- a/public/components/integrations/components/integration.tsx
+++ b/public/components/integrations/components/integration.tsx
@@ -67,7 +67,7 @@ export function Integration(props: AvailableIntegrationProps) {
     } else {
       payload.index_patterns = [dataSourceName];
       return fetch(
-        `/api/console/proxy?path=_index_template/${integration.name}_${componentName}_${version}&method=POST`,
+        `/api/console/proxy?path=_index_template/ss4o_${componentName}_${version}_${integration.name}&method=POST`,
         {
           method: 'POST',
           headers: [
--- a/public/components/integrations/components/integration.tsx
+++ b/public/components/integrations/components/integration.tsx
@@ -43,8 +43,7 @@ export function Integration(props: AvailableIntegrationProps) {
       template: { mappings: { _meta: { version: string } } };
       composed_of: string[];
       index_patterns: string[];
-    },
-    dataSourceName: string
+    }
   ): Promise<{ [key: string]: { properties: any } } | null> => {
     const version = payload.template.mappings._meta.version;
     return http
@@ -88,7 +87,7 @@ export function Integration(props: AvailableIntegrationProps) {
 
   const createDataSourceMappings = async (targetDataSource: string): Promise<any> => {
     const data = await http.get(`${INTEGRATIONS_BASE}/repository/${integrationTemplateId}/schema`);
-    const error = null;
+    let error = null;
     const mappings = data.data.mappings;
     mappings[integration.type].composed_of = mappings[integration.type].composed_of.map(
       (templateName: string) => {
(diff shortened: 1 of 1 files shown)


public/components/integrations/components/integration.tsx
```diff
@@ -67,7 +67,7 @@ export function Integration(props: AvailableIntegrationProps) {
     } else {
       payload.index_patterns = [dataSourceName];
       return fetch(
-        `/api/console/proxy?path=_index_template/${componentName}_${version}&method=POST`,
+        `/api/console/proxy?path=_index_template/ss4o_${componentName}_${version}_${integration.name}&method=POST`,
```
YANG-DB, line 70: why are we creating a unique component for each integration ?
we should not duplicate similar schema components (http, cloud, communication )
the code should first check is the component exists and verify it has the same version - if does nothing to update 
Otherwise - add the new template with (including the version as part of the template name)

YANG-DB (changes requested):

public/components/integrations/components/integration.tsx
Swiddis: The part that creates non-duplicated components exists a few lines before this part, this is after component creation.

The issue only is with template creation, which *is* unique per-integration for most integrations, due to differing dependencies.

derek-ho (approved):

YANG-DB:
please add a concrete example on how the index pattern that is explicitly Written in the index template is. being changed by the code - is this a good solution to change without allowing the user to be aware of this change ???

Swiddis:
> please add a concrete example on how the index pattern that is explicitly Written in the index template is. being changed by the code - is this a good solution to change without allowing the user to be aware of this change ???

It only really affects which index the sample data is created in, the user is still expected to configure their own indices that follow the right conventions if they want to set up their own system. For running on prod, they'll want to remove all the sample data anyways, so it shouldn't really affect them?

YANG-DB (approved):

PR #822, opened 2023-08-03
Changed log mappings to be apache-specific

### LINKED ISSUE #812: [BUG]Conflicting logs index templates

danieldong51:
### Description
Changes log mapping names to be apache-specific, to avoid conflict with other log index patterns.

### Issues Resolved
#812 
[Backport comments](https://github.com/opensearch-project/dashboards-observability/pull/800#discussion_r1283513166)

### Check List

### DIFF
--- a/server/adaptors/integrations/__data__/repository/apache/apache-1.0.0.json
+++ b/server/adaptors/integrations/__data__/repository/apache/apache-1.0.0.json
@@ -29,7 +29,7 @@
             "version": "1.0.0"
         },
         {
-            "name": "logs",
+            "name": "logs-apache",
             "version": "1.0.0"
         }
     ],
--- a/server/adaptors/integrations/__data__/repository/apache/schemas/logs-apache-1.0.0.mapping.json
+++ b/server/adaptors/integrations/__data__/repository/apache/schemas/logs-apache-1.0.0.mapping.json
@@ -1,9 +1,12 @@
 {
     "index_patterns": [
-        "ss4o_logs-*-*"
+        "ss4o_logs-apache-*"
     ],
     "data_stream": {},
     "template": {
+        "aliases": {
+            "logs-apache": {}
+        },
         "mappings": {
             "_meta": {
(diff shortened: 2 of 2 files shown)


Swiddis:
Same as for #821: if #818 goes through this change shouldn't be necessary

YANG-DB (approved):

Swiddis (approved):

PR #901, opened 2023-08-10
Make OSD Bootstrap more sturdy in CI

Swiddis:
### Description
As the Bootstrap command in CI is inconsistent, this has been causing a lot of delays in getting CI feedback in PRs. This PR seeks to address this with a few safeguards:

1. Advice from the [OSD Developer Guide](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/DEVELOPER_GUIDE.md) was applied to update the network timeout. In their own CI they use this network setting and also have a backup call in case it fails. They also use a `clean` which may address some inconsistency, though it should be a no-op in a CI setting.
2. `fail-fast` is disabled in the CI, so even if the measures from (1) are unsuccessful, we will at least receive some feedback from the other two runs without needing a full re-run.
3. The results are cached to both reduce the effects of the issue and speed up CI overall.

### Issues Resolved
N/A

### Check List

### DIFF
--- a/.github/workflows/dashboards-observability-test-and-build-workflow.yml
+++ b/.github/workflows/dashboards-observability-test-and-build-workflow.yml
@@ -14,6 +14,8 @@ jobs:
     strategy: 
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
+      # Since Windows is inconsistent, we want to let other OSes run on a fail
+      fail-fast: false
     runs-on: ${{ matrix.os }}
 
     steps:
@@ -51,11 +53,12 @@ jobs:
         with:
           path: OpenSearch-Dashboards/plugins/dashboards-observability
 
-
       - name: Plugin Bootstrap
         run: |
           cd OpenSearch-Dashboards/plugins/dashboards-observability
-          yarn osd bootstrap
+          yarn cache clean
+          yarn config set network-timeout 1000000 -g
+          yarn osd bootstrap || yarn osd bootstrap
 
       - name: Test all dashboards-observability modules
         run: |
--- a/.github/workflows/dashboards-observability-test-and-build-workflow.yml
+++ b/.github/workflows/dashboards-observability-test-and-build-workflow.yml
@@ -53,13 +53,13 @@ jobs:
         with:
           path: OpenSearch-Dashboards/plugins/dashboards-observability
 
-      - name: Cache OpenSearch Dashboards Bootstrap Results
+      - name: Load Plugin Bootstrap cache
         uses: actions/cache@v3
         with:
           path: |
             **/node_modules
             **/target
-          key: ${{ runner.os }}-bootstrap-${{ hashFiles('**/yarn.lock', '**/package.json') }}
+          key: ${{ runner.os }}-bootstrap-${{ hashFiles('OpenSearch-Dashboards/yarn.lock', 'OpenSearch-Dashboards/plugins/dashboards-observability/yarn.lock') }}
 
       - name: Plugin Bootstrap
         run: |
--- a/.github/workflows/dashboards-observability-test-and-build-workflow.yml
+++ b/.github/workflows/dashboards-observability-test-and-build-workflow.yml
@@ -48,30 +48,24 @@ jobs:
       - run: node -v
       - run: yarn -v
 
+      - name: Checkout Dashboards Observability
+        uses: actions/checkout@v2
+        with:
+          path: OpenSearch-Dashboards/plugins/dashboards-observability
+
       - name: Cache OpenSearch Dashboards Bootstrap Results
         uses: actions/cache@v3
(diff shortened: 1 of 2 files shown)


YANG-DB (approved):

pjfitzgibbons:
@Swiddis Love this!
Do you have any way to test this in local dev?  (docker instructions somewhere?)
If so, can you post a terminal log of the test.  I just want to get a feel for what this actually does at CI runtime?

Swiddis:
> Do you have any way to test this in local dev? (docker instructions somewhere?)

Not at the moment, I'm mostly just relying on an actual run to hang. It looks like [this one](https://github.com/opensearch-project/dashboards-observability/actions/runs/5827274994/job/15802981263) is pretty close to expected behavior, since it hung for multiple minutes while waiting on network calls. I'm still experimenting a bit with caching.

Swiddis:
Results are in

| OS | No Cache | Cache |
|---|---|---|
| Ubuntu | 4:28 | 0:20 |
| Windows | 11:51 | 2:04 |
| Mac | 8:54 | 2:12 |

Should be good to merge now

.github/workflows/dashboards-observability-test-and-build-workflow.yml
```diff
@@ -51,11 +53,20 @@ jobs:
...
+            **/target
+          key: ${{ runner.os }}-bootstrap-${{ hashFiles('OpenSearch-Dashboards/yarn.lock', 'OpenSearch-Dashboards/plugins/dashboards-observability/yarn.lock') }}
+          restore-keys: ${{ runner.os }}-bootstrap-
 
       - name: Plugin Bootstrap
         run: |
-          cd OpenSearch-Dashboards/plugins/dashboards-observability
-          yarn osd bootstrap
+          cd OpenSearch-Dashboards
+          yarn config set network-timeout 1000000 -g
```
pjfitzgibbons, line 68: Is this seconds?

pjfitzgibbons (approved):

.github/workflows/dashboards-observability-test-and-build-workflow.yml
Swiddis: Milliseconds, this PR uses the same number as is used in the [OpenSearch Dashboards CI](https://github.com/opensearch-project/OpenSearch-Dashboards/commit/038968d154c8bcc9d9c75e7abbf9b9435b9a0e42), which in turn is inspired by a [related issue on yarn](https://github.com/yarnpkg/yarn/issues/8242#issuecomment-776561223) that suggests a similar approach with a number that's also in that ballpark.

.github/workflows/dashboards-observability-test-and-build-workflow.yml
```diff
@@ -51,11 +53,20 @@ jobs:
         with:
           path: OpenSearch-Dashboards/plugins/dashboards-observability
 
+      - name: Load Plugin Bootstrap cache
+        uses: actions/cache@v3
+        with:
+          path: |
+            **/node_modules
+            **/target
```
joshuali925, line 61: why is it caching target?

.github/workflows/dashboards-observability-test-and-build-workflow.yml
```diff
@@ -51,11 +53,20 @@ jobs:
...
+          key: ${{ runner.os }}-bootstrap-${{ hashFiles('OpenSearch-Dashboards/yarn.lock', 'OpenSearch-Dashboards/plugins/dashboards-observability/yarn.lock') }}
+          restore-keys: ${{ runner.os }}-bootstrap-
 
       - name: Plugin Bootstrap
         run: |
-          cd OpenSearch-Dashboards/plugins/dashboards-observability
-          yarn osd bootstrap
+          cd OpenSearch-Dashboards
+          yarn config set network-timeout 1000000 -g
+          yarn osd bootstrap || yarn osd bootstrap
```
joshuali925, line 69: this is ok but there's also a retry workflow https://github.com/marketplace/actions/retry-step

.github/workflows/dashboards-observability-test-and-build-workflow.yml
Swiddis: It showed up when I scanned for what files bootstrap was making, I assumed it may be important. I don't know the technical details of what bootstrap does, so I might have overselected.

joshuali925: i'm not sure why bootstrap would touch target. target is the transpiled javascript from source code (public ts/tsx) to be executed in browser, `yarn start` (optimizing phase) and `yarn build` would update target.

caching it probably won't cause an issue even if your key does not consider source code changes (because `yarn build` will update target anyways), but i don't think it should be cached. it's not really related to dependencies

PR #947, opened 2023-08-22
Refactor Integrations class to rely on CatalogReader

Swiddis:
### Description
As part of continuing work on enabling remote catalogs, more refactoring for existing code is being done. In particular, we're introducing the concept of a `CatalogReader` which can be freely swapped out to enable new input sources, such as a URL or a zip file.

The main conversion for the `Integrations` class is done, but before this PR is ready, there is still some rippling changes to apply, and more tests need updating. In particular, with the introduction of a `Result` type in #943, the error handling code is being refactored considerably. This PR is just opened early to get an initial look.

### Issues Resolved
N/A

### Check List


server/adaptors/integrations/repository/catalog_reader.ts
```diff
@@ -0,0 +1,52 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+type IntegrationPart = 'assets' | 'data' | 'schemas' | 'static';
+
+interface CatalogReader {
```
YANG-DB, line 8: the idea of an interface for catalog is great - but the method names for this interface are too explicit for file based - can u change the methods to be more generic so that both the `local_catalog_reader` and `network_catalog_reader` will have same common API
This interface should focus on the domain entities such as: 
 - getCatalogMetadata
 - getIntegrationMetadata
 - getIntegration

server/adaptors/integrations/repository/integration.ts
```diff
@@ -33,77 +33,47 @@ function compareVersions(a: string, b: string): number {
...
-  }
-}
-
 /**
  * The Integration class represents the data for Integration Templates.
  * It is backed by the repository file system.
  * It includes accessor methods for integration configs, as well as helpers for nested components.
  */
 export class Integration {
+  reader: CatalogReader;
```
YANG-DB, line 42: why is the catalog reader part of the integration ?
I think that Integration should be agnostic of where its coming from (file / network)

YANG-DB (changes requested):

server/adaptors/integrations/repository/catalog_reader.ts
Swiddis: I'm not sure this is what the catalog reader is for, because that's already what the `Repository` and `Integration` classes are doing. They do all of that abstraction of fetching and validating the parts of a catalog, the reader specifically is about the types of low-level scans (file or network) that they are able to do. `getCatalogMetadata` is already `Repository.getIntegrationList`, and `getIntegrationMetadata` is already `Integration.getConfig`.

server/adaptors/integrations/repository/integration.ts
Swiddis: For the same reason that the integration currently depends on the file system: `Integration` is a high-level wrapper around a part of a catalog containing integration data. This PR is meant to abstract away the interface for low-level catalog access.

server/adaptors/integrations/__test__/kibana_backend.test.ts
pjfitzgibbons, line 1: I think we're supposed to avoid ES-centric names if possible.  Maybe `integrations_fetch_service` ?

server/adaptors/integrations/__test__/kibana_backend.test.ts
```diff
@@ -298,6 +300,130 @@ describe('IntegrationsKibanaBackend', () => {
         404
       );
     });
+
+    it('should reject with a 404 if static data is not found', async () => {
+      const templateName = 'template1';
+      const staticPath = 'path/to/static';
+      mockRepository.getIntegration.mockResolvedValue({
+        getStatic: jest.fn().mockResolvedValue({
```
pjfitzgibbons, line 308: An aside comment : I'm happy to see the { ok, error } pattern is working out so well in these files!

server/adaptors/integrations/__test__/validators.test.ts
```diff
@@ -29,7 +29,7 @@ const validInstance: IntegrationInstance = {
 
 describe('validateTemplate', () => {
   it('Returns a success value for a valid Integration Template', () => {
-    const result: ValidationResult<IntegrationTemplate> = validateTemplate(validTemplate);
+    const result: Result<IntegrationTemplate> = validateTemplate(validTemplate);
```
pjfitzgibbons, line 32: Whoa.  Learn something new every day -- sometimes twice!   Result is great!

server/adaptors/integrations/integrations_kibana_backend.ts
pjfitzgibbons, line 1: As above, avoid ES-centric naming - maybe `integrations_fetch_service` instead.

server/adaptors/integrations/__test__/builder.test.ts
```diff
@@ -121,22 +131,24 @@ describe('IntegrationInstanceBuilder', () => {
         dataSource: 'instance-datasource',
         name: 'instance-name',
       };
-      sampleIntegration.deepCheck = jest.fn().mockResolvedValue(false);
+      sampleIntegration.deepCheck = jest
+        .fn()
+        .mockResolvedValue({ ok: false, error: new Error('Mock error') });
```
pjfitzgibbons, line 136: I see the introduction of Result further down in the changes.
Should these uses also be converted to Result?

server/adaptors/integrations/repository/catalog_reader.ts
pjfitzgibbons: How do Repository and Integration handle network vs local ?
What is the network oriented corollary class to this one ?

server/adaptors/integrations/repository/integration.ts
pjfitzgibbons: How large-scale would be that redesign, really ?
At least in this PR you could set the naming to point the story in the right direction, which could make the future refactor simpler to understand.

pjfitzgibbons (changes requested):
I would like to discuss design/architecture of these changes.

Swiddis:
After talking with Peter a bit about potential directions for building on the architecture of this, I've worked out some next steps to maybe work out both comments, will see about getting this PR ready in the next couple days with some corresponding issues for further architecture work.

server/adaptors/integrations/__test__/builder.test.ts
Swiddis: Only have converted part of it to result to avoid making the PR too large, intend to do more later

server/adaptors/integrations/repository/integration.ts
Swiddis: Renamed the class to make its purpose more clear, now the integration reader and integration data are separate entities

server/adaptors/integrations/repository/catalog_reader.ts
Swiddis: Have changed the names to make it better, and gave more responsibility to the class to remove some further file-specific handling from the calling code.

YANG-DB (approved):

PR #969, opened 2023-09-06
Upgrade cypress/requests from 2 to 3

Swiddis:
### Description
The outdated version doesn't affect us directly, but it does cause issues in automatic auditing, so we may as well just bump it. Similar PR: https://github.com/opensearch-project/dashboards-reporting/pull/174

### Issues Resolved
N/A

### Check List


YANG-DB (approved):

package.json
```diff
@@ -53,7 +53,7 @@
     "@types/react-test-renderer": "^16.9.1",
     "@types/sanitize-filename": "^1.6.3",
     "antlr4ts-cli": "^0.5.0-alpha.4",
-    "cypress": "^12.8.1",
+    "cypress": "^13.0.0",
```
derek-ho, line 56: You can enforce cypress request 3.0 without upgrading - https://github.com/opensearch-project/dashboards-reporting/pull/174 I would recommend that unless you have tested all files already

derek-ho: Also FTR will be 12 soon - I would recommend staying on same major version as them

package.json
```diff
@@ -53,7 +53,7 @@
     "@types/react-test-renderer": "^16.9.1",
     "@types/sanitize-filename": "^1.6.3",
     "antlr4ts-cli": "^0.5.0-alpha.4",
-    "cypress": "^12.8.1",
+    "cypress": "12",
```
ps48, line 56: Any reason for dropping this down to 12?

ps48 (approved):

package.json
Swiddis: Good catch, happened while upgrading and re-downgrading.

PR #983, opened 2023-09-08
Move query_utils to /public

pjfitzgibbons:
### Description
Move query_utils module to /public.
This module contains only functions used in browser front-end code.  Having this module in /common causes a server restart on every change.  The query_utils module will be updated extensively in Metrics Edit feature.

### Check List

### DIFF
--- a/public/components/common/query_utils.ts
+++ b/public/components/common/query_utils.ts
@@ -8,12 +8,12 @@ import { isEmpty } from 'lodash';
 import {
   DATE_PICKER_FORMAT,
   PPL_DEFAULT_PATTERN_REGEX_FILETER,
-} from '../../common/constants/explorer';
+} from '../../../common/constants/explorer';
 import {
   PPL_INDEX_INSERT_POINT_REGEX,
   PPL_INDEX_REGEX,
   PPL_NEWLINE_REGEX,
-} from '../../common/constants/shared';
+} from '../../../common/constants/shared';
 
 /**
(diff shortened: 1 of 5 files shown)


public/components/common/query_utils.ts
```diff
@@ -55,6 +55,7 @@ export const preprocessQuery = ({
 }) => {
   let finalQuery = '';
 
+  console.log('preprocessQuery', { rawQuery, finalQuery });
```
derek-ho, line 58: remove?

derek-ho (approved):
Other than extra console.log LGTM

.nvmrc
```diff
@@ -0,0 +1 @@
+16
```
ps48, line 1: Not sure if this was an intended addition. Ideally, we should use nvmrc from OSD.

.prettierrc
```diff
@@ -0,0 +1,5 @@
+{
```
ps48, line 1: Are we sticking to this prettier config for the whole repo?

report.txt
```diff
@@ -0,0 +1,191 @@
+Targets
```
ps48, line 1: I think we should remive this report and https://github.com/opensearch-project/dashboards-observability/pull/983/files#diff-df3e51bb2577d27a12a41e5efd1630341367497497b26a0763d5749e45dc8e42 file.

ps48 (approved):

Issue #930, opened 2023-08-15
[FEATURE] Show all buckets in selected time range for count distribution


joshuali925:
**Is your feature request related to a problem?**
In event analytics, the count histogram is only visible for when there is data in the aggregated bucket. For example, the selected time range is this year, the data exists somewhere in January, and although the selected time range is 1 year (data should be aggregated per each month), the time range in the histogram is 1 month.
<img width="1223" alt="image" src="https://github.com/opensearch-project/dashboards-observability/assets/28062824/2d9e3984-4ed9-4b2e-8b69-2e9911490d7c">

It doesn't look nice with an extra wide bar in the chart. Also it is difficult to tell how spread the data is (e.g. is it 379 hits in a year, or are they all in the same month)

**What solution would you like?**
The data points shown in histogram should be determined by aggregated time period, not by when is data available.

This happens because backend only returned buckets for available data. In the example, since all data is in January, the aggregation query buckets by month (because 1 year is selected) and will only return 1 bucket (1 month). While it is expected behavior for `aggs`, conventionally frontend should create the empty buckets for the rest of 11 months so that the histogram shows 12 months of data.

**What alternatives have you considered?**
N/A

**Do you have any additional context?**
N/A

PR #990, opened 2023-09-11
Explorer empty bucket

paulstn:
### Description
Includes empty buckets in log explorer for data points without any hits, in order to display the chart accurately.

### Issues Resolved
https://github.com/opensearch-project/dashboards-observability/issues/930

### Check List


public/components/event_analytics/explorer/visualizations/count_distribution/count_distribution.tsx
```diff
@@ -31,9 +40,62 @@ export const CountDistribution = ({ countDistribution }: any) => {
     },
   ];
 
+  // fill the final data with the exact right amount of empty buckets
+  function fillWithEmpty(processedData: any) {
```
joshuali925, line 44: might be better to move this function outside and add tests. also could `processedData` have a type? `any` makes it difficult to reuse the function

paulstn: Is public/components/event_analytics/utils/utils.tsx a good file to have this function in or do you have somewhere else in mind?
I'll also make sure processData can have a type, but what is a good place to put that type definition?

joshuali925: you can put it there. for the type maybe make the function more generic? for example it takes in `x` and `y` values instead of `processedData` which is a very specific single element tuple, and processed `intervalPeriod` instead of `selectedInterval`.

if the type of `processedData` is used across the project then you can probably leave it and add a type in some types.ts

paulstn: Made this change

public/components/event_analytics/utils/utils.tsx
```diff
@@ -459,3 +462,57 @@ export const getContentTabTitle = (tabID: string, tabTitle: string) => {
...
+ * @param selectedInterval Moment unitOfTime used to dictate how long each interval is
+ * @param startTime starting time of x values
+ * @param endTime ending time of x values
+ * @returns an object with buckets and values where the buckets are all of the new x values and
+ * values are the corresponding values which include y values that are 0 for empty data
+ */
+export const fillTimeDataWithEmpty = (
+  xVals: string[],
+  yVals: number[],
+  selectedInterval: any,
```
joshuali925, line 480: is this string? or better would be something like `'m' | 'd' | 'w' | .....`

joshuali925 (approved):

public/components/event_analytics/utils/utils.tsx
mengweieric, line 480: Similar thing here, please add corresponding type

paulstn: An example string in xVals could be "2023-01-01 00:00:00", but any datetime string can work, because it will be parsed by moment.js. So theoretically, as long as the string looks like a date, this method is useful for any time related data.

public/components/event_analytics/utils/utils.tsx
```diff
@@ -459,3 +462,57 @@ export const getContentTabTitle = (tabID: string, tabTitle: string) => {
...
+  startTime: string,
+  endTime: string
+) => {
```
mengweieric, line 483: Do we also want to add return type for this function?

mengweieric (approved):

public/components/event_analytics/utils/utils.tsx
paulstn: I've added the return type

public/components/event_analytics/utils/utils.tsx
paulstn: I created a new type and put it into common/types/explorer.ts to include all possible moment.js friendly interval strings.

paulstn: I created a new type and put it into common/types/explorer.ts to include all possible moment.js friendly interval strings.

PR #992, opened 2023-09-12
Add access control tab content

derek-ho:
### Description
Add access control tab content to datasource page

### Issues Resolved

### Check List


public/components/data_connections/components/access_control_tab.tsx
```diff
@@ -0,0 +1,205 @@
...
+  ] = useState<Array<{ label: string }>>([]);
+
+  useEffect(() => {
+    coreRefs.http!.get('/api/v1/configuration/roles').then((data) =>
+      setRoles(
+        Object.keys(data.data).map((key) => {
+          return { label: key };
+        })
+      )
+    );
```
derek-ho, line 58: Assumes security plugin is installed and user has permission to access it (will be addressed with the permission PR)

public/components/data_connections/components/access_control_tab.tsx
```diff
@@ -0,0 +1,205 @@
...
+    `acceleration-1`
+  );
+
+  const [finalQueryPermissionsPlaceHolder, setFinalQueryPermissionsPlaceHolder] = useState<
+    Array<{ label: string }>
+  >([]);
+  const [
+    finalAccelerationPermissionsPlaceHolder,
+    setFinalAccelerationPermissionsPlaceHolder,
+  ] = useState<Array<{ label: string }>>([]);
```
derek-ho, line 49: Placeholder to update the state (this will be replaced with API call to POST, and then showing the list will be the result of the API call to GET)

YANG-DB (approved):

public/components/data_connections/components/access_control_tab.tsx
```diff
@@ -0,0 +1,205 @@
...
+            <EuiButton
+              data-test-subj="createButton"
+              onClick={
+                mode === 'view'
+                  ? () => {
+                      setMode('edit');
+                    }
+                  : () => {
+                      setMode('view');
+                    }
```
ps48, line 168: I think we can make it more readable or change it to a function. 
```
const changeMode = () => { 
     mode === 'view' ? setMode('edit') : setMode('view');
}
onClick = {changeMode}

OR 

onClick = { () => setMode(mode === 'view' ? 'edit' : 'view')}
```

public/components/data_connections/components/access_control_tab.tsx
ps48: What happens if security plugin is not installed. Are we not showing the whole data connections plugin? or Do we have a fallback view?

common/types/data_connections.ts
```diff
@@ -0,0 +1,15 @@
...
+ */
+
+import { EuiRadioGroupOption } from '@elastic/eui';
+
+export interface PermissionsFlexItem {
+  roles: Array<{ label: string }>;
+  selectedRoles: Array<{ label: string }>;
+  selectedRadio: string;
+  setSelectedRoles: (selectedRoles: Array<{ label: string }>) => void;
+  setSelectedRadio: (selectedRadio: string) => void;
```
ps48, line 13: Minor, but Ideally types here should be something like: `React.Dispatch<React.SetStateAction<Array<{ label: string }>>`

public/components/data_connections/components/access_control_tab.tsx
```diff
@@ -0,0 +1,205 @@
...
+              </EuiButtonEmpty>
+            </EuiFlexItem>
+            <EuiFlexItem grow={false}>
+              <EuiButton onClick={saveChanges} size="s" iconType="check" fill>
+                Save
+              </EuiButton>
+            </EuiFlexItem>
+          </EuiFlexGroup>
+        </EuiBottomBar>
+      ) : null}
```
ps48, line 202: Minor can be changed to `mode === 'edit' && <view/>` no need to use null.

ps48 (approved):

public/components/data_connections/components/access_control_tab.tsx
```diff
@@ -0,0 +1,205 @@
...
+                  : '-'}
+              </EuiText>
+            </EuiFlexItem>
+          </EuiFlexGroup>
+        </EuiFlexItem>
+      </EuiFlexGroup>
+    );
+  };
+
+  const renderEditAccessControlDetails = () => {
```
pjfitzgibbons, line 115: Please turn this into a formal React FC.
This is a quality issue for us throughout our plugin.

public/components/data_connections/components/access_control_tab.tsx
```diff
@@ -0,0 +1,205 @@
...
+    return (
+      <EuiFlexGroup direction="column">
+        <QueryPermissionsFlexItem
```
pjfitzgibbons, line 118: Please do not call these components "...FlexItem".  The component's purpose is not a consequence of it's inclusion in a flex-group.

public/components/data_connections/components/access_control_tab.tsx
```diff
@@ -0,0 +1,205 @@
...
+                    }
+              }
+              fill={mode === 'view' ? true : false}
+            >
+              {mode === 'view' ? 'Edit' : 'Cancel'}
+            </EuiButton>
+          </EuiFlexItem>
+        </EuiFlexGroup>
+        <EuiHorizontalRule />
+        {mode === 'view' ? renderViewAccessControlDetails() : renderEditAccessControlDetails()}
```
pjfitzgibbons, line 177: Consider setting this value to a component-constant (`AccessControlDetails`).  You have several of these conditions - they can all be collected into one place... thereby reducing cyclomatic complexity.

Containing these conditions into one location will also raise the visibility of refactoring / simplification opportunities.

public/components/data_connections/components/query_permissions_flex_item.tsx
```diff
@@ -0,0 +1,69 @@
...
+          <EuiRadioGroup
+            options={radios}
+            idSelected={selectedRadio}
+            onChange={(id) => setSelectedRadio(id)}
+            name="query-radio-group"
+            legend={{
+              children: <span>Access level</span>,
+            }}
+          />
+          {selectedRadio === QUERY_RESTRICT ? (
```
pjfitzgibbons, line 46: Consider moving this condition to a component-constant.  Name it something that tells the story about what it is / why it's there.

We are left to imagine the meaning of  `selectedRadio === QUERY_RESTRICT`... often the wrong meaning.

pjfitzgibbons (changes requested):

public/components/data_connections/components/access_control_tab.tsx
derek-ho: I think current plan is to show as much as possible:
- if users have access to datasources APIs, we show them the page, and if they do not then we will show Fallback UI
- if users have access to datasource APIs, but do not have access to security plugin or security plugin is not installed, we will not be able to show the dropdown with pre-defined roles, but we will still allow them to manually type in the roles. Created a follow up issue to avoid scope-creep: https://github.com/opensearch-project/dashboards-observability/issues/1022. Will not implement manually type to create roles in this PR, but in follow up PR.

common/types/data_connections.ts
derek-ho: Done in: [fc96304](https://github.com/opensearch-project/dashboards-observability/pull/992/commits/fc96304362b085696c2e5df8e7c638b69c142180), thanks! I didn't know this

public/components/data_connections/components/access_control_tab.tsx
derek-ho: Done in: [fc96304](https://github.com/opensearch-project/dashboards-observability/pull/992/commits/fc96304362b085696c2e5df8e7c638b69c142180), thanks!

public/components/data_connections/components/access_control_tab.tsx
derek-ho: I tried to address this in: [fc96304](https://github.com/opensearch-project/dashboards-observability/pull/992/commits/fc96304362b085696c2e5df8e7c638b69c142180), please let me know if I understood what you meant correctly.

public/components/data_connections/components/access_control_tab.tsx
derek-ho: Agree. Done in: [fc96304](https://github.com/opensearch-project/dashboards-observability/pull/992/commits/fc96304362b085696c2e5df8e7c638b69c142180), thanks for the suggestion.

public/components/data_connections/components/query_permissions_flex_item.tsx
derek-ho: Not quite sure I am following what you are suggesting here. The issue is that right now, the radio groups and modifying the radio groups need to be passed down from the parent to this specific component, but this component also needs to know some specifics about the radio groups in order to render correctly. This situation sounds like maybe the parent should not be doing that, and should just allow this component to render everything and control as needed. However this means that the parent needs to pass down the whole data connection details so that this child component can make the appropriate API call to update it. Does this tradeoff make sense?

public/components/data_connections/components/access_control_tab.tsx
derek-ho: Not quite sure I understand what you are suggesting/how it would be cleaner than having a state variable. Can you point me to some examples I can look at?

public/components/data_connections/components/data_connection.tsx
```diff
@@ -37,64 +40,68 @@ export const DataConnection = (props: any) => {
...
         setDatasourceDetails({
           allowedRoles: data.allowedRoles,
           name: data.name,
           cluster: data.properties['emr.cluster'],
+          connector: data.connector,
+          properties: data.properties,
         })
       )
       .catch((err) => {
         if (err.body.statusCode === 403) {
```
pjfitzgibbons, line 72: Without re-throwing, you're effectively swallowing all other errors except 403.  Is that really what we want?
If any/ever error should be met with setHasAccess(false), then just set it without condition.
This might be a case to wrap this in a function so you can tell a story `checkStatusForbiddenOrThrow(err)`.

public/components/data_connections/components/query_permissions.tsx
```diff
@@ -0,0 +1,87 @@
...
+          <EuiText size="s" className="overview-content">
+            Control which OpenSearch roles have query permissions on this data source.{' '}
+            <EuiLink external={true} href={OPENSEARCH_DOCUMENTATION_URL} target="_blank">
+              Learn more
+            </EuiLink>
+          </EuiText>
+        </EuiFlexItem>
+        <EuiFlexItem>
+          <EuiRadioGroup
+            options={radios}
```
pjfitzgibbons, line 74: I still very much recommend renaming to `availableRoles`, `selectedRoles`, `setSelectedRoles`
Future implemention might be checkbox list.

public/components/data_connections/components/query_permissions.tsx
```diff
@@ -0,0 +1,87 @@
...
+            idSelected={selectedRadio}
+            onChange={(id) => setSelectedRadio(id)}
+            name="query-radio-group"
+            legend={{
+              children: <span>Access level</span>,
+            }}
+          />
+          {selectedRadio === QUERY_RESTRICTED && <ConfigureRoles />}
```
pjfitzgibbons, line 82: `selectedAccessLevel`

PR #997, opened 2023-09-12
WIP Attempt fix Available Metrics Spec

pjfitzgibbons:
### Description
Attempt to fix Metrics Available Metric spec file.

Test should assert "Available Metrics 1 of 1" appears on page after mock loaded request of Observability Metric data from SavedObjects.

Currently, test fails - expected "Available Metrics 1 of 1" not found on string "Selected Metrics 0 of 0 Available Metrics 0 of 0"

I have not been able to determine any change that will update mocked SavedObject getBulk in a reactive way and show that the <MetricAccordion> widget recieves updated state data and re-renders.

This PR is in contrast to PR #996

### Check List

### DIFF
--- a/public/components/metrics/sidebar/__tests__/sidebar.test.tsx
+++ b/public/components/metrics/sidebar/__tests__/sidebar.test.tsx
@@ -3,7 +3,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import { configure, mount } from 'enzyme';
+import { ReactWrapper, configure, mount } from 'enzyme';
 import Adapter from 'enzyme-adapter-react-16';
 import React from 'react';
 import { waitFor } from '@testing-library/react';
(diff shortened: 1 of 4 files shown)


public/components/metrics/sidebar/__tests__/sidebar.test.tsx
```diff
@@ -37,16 +41,59 @@ describe('Side Bar Component', () => {
...
+
+    const el = document.createElement('div');
+    act(() => {
+      ReactDOM.render(<Sidebar />, el);
+    });
+    expect(el.innerHTML).toContain('Available Metrics 0 of 0');
+    await act(async () => {
+      objectListResolver({ observabilityObjectList: [sampleSavedMetric] });
     });
+    expect(el.innerHTML).toContain('Available Metrics 1 of 1');
```
Swiddis, line 96: In a lot of the flaky tests I've been working with lately, I've seen a lot of these types of string assertions. Fine, but I wonder if there's a better/more stable way to check this.

pjfitzgibbons: This test is asserting that the output values (the literal rendered string "Available Metrics 1 of 1") are rendered after a side-effect update from redux store.

More stable than... ?
1. display it (the blank-ish state)
2. change something
3. assert the display changed

public/components/metrics/sidebar/__tests__/sidebar.test.tsx
```diff
@@ -37,16 +41,59 @@ describe('Side Bar Component', () => {
...
+    coreRefs.pplService = new PPLService(httpClientMock);
+    coreRefs.pplService.fetch = jest.fn(() => {
+      console.log('mock pplService fetch');
+      return new Promise((resolve) => {
+        resolve({
+          data: { DATA_SOURCES: ['datasource1', 'datasource2'] },
+          then: () => Promise.resolve(),
+        });
+      });
+    });
```
derek-ho, line 86: Can use jest.fn().mockResolvedValue() instead

public/components/metrics/sidebar/__tests__/sidebar.test.tsx
```diff
@@ -37,16 +41,59 @@ describe('Side Bar Component', () => {
...
+
+    const el = document.createElement('div');
+    act(() => {
```
derek-ho, line 89: try await act here?

pjfitzgibbons: Can only await an `act(() =>` that is specifically async.   Otherwise, act is speicfically a sync call and acts that way in all respects.
The first act() is setting up render, which is not an async call.
Evidence : The initial page render is "all blank" b/c dependency data is entirely missing.  The first visible page render is "Available Metrics 0 of 0", which is non-blank, so ensures first render cycle has executed.

PR #1041, opened 2023-09-21
modified explorer data grid to follow discover look and feel

paulstn:
### Description
Adds data grid to log explorer that has the look and feel of the new Discover data grid
image

### Issues Resolved
Look and feel changes to Explorer Data Grid.
Known needed changes (p1-2 items): sidebar field order implementation, surrounding events implementation

### Check List


public/components/event_analytics/explorer/sidebar/sidebar.tsx
```diff
@@ -89,22 +93,34 @@ export const Sidebar = (props: ISidebarProps) => {
...
-      );
+      let nextFields;
+      if (
+        explorerFields.selectedFields.length === 0 &&
+        storedExplorerFields.selectedFields.length !== 0
+      ) {
+        nextFields = toggleFields(storedExplorerFields, field, SELECTED_FIELDS, AVAILABLE_FIELDS);
+      } else {
+        nextFields = toggleFields(explorerFields, field, SELECTED_FIELDS, AVAILABLE_FIELDS);
+      }
```
ps48, line 121: Minor: Can this be simplified? Suggesting something below (can be better)
```
    const selectedFields =
      explorerFields.selectedFields.length === 0 && storedExplorerFields.selectedFields.length > 0
        ? storedExplorerFields
        : explorerFields;

    const nextFields = toggleFields(selectedFields, field, SELECTED_FIELDS, AVAILABLE_FIELDS);

Comment (ps48) [reply public/components/event_analytics/explorer/sidebar/sidebar.tsx:121]: Also I see something similar in `handleAddField` do you want to create a function and reuse code?

File: public/components/event_analytics/explorer/events_views/doc_flyout.tsx [hunk 2]
```diff
@@ -124,13 +124,13 @@ export const DocFlyout = ({
...
       <div className="obsExplorer">
-        {populateDataGrid(
+        {/* {populateDataGrid(
           explorerFields,
           getHeaders(explorerFields.queriedFields, DEFAULT_COLUMNS.slice(1), true),
           <tr className="osdDocTable__row">{memorizedTds}</tr>,
           getHeaders(explorerFields.selectedFields, DEFAULT_COLUMNS.slice(1), true),
           <tr className="osdDocTable__row">{memorizedTds}</tr>
-        )}
+        )} */}
```
ps48, line 133: Do you want to remove this commented piece?

public/components/event_analytics/explorer/events_views/data_grid.tsx
```diff
@@ -3,148 +3,271 @@
...
+        }
+        return `${rows[trueIndex][columnId]}`;
+      }
+      return null;
+    },
+    [rows, pageFields, explorerFields]
+  );
+
+  // ** Pagination config
+  const [pagination, setPagination] = useState({ pageIndex: 0, pageSize: 25 });
```
ps48, line 219: Is this default value same as discover?

paulstn: Yes

public/components/event_analytics/explorer/events_views/doc_flyout.tsx
paulstn: Will do

public/components/event_analytics/explorer/sidebar/sidebar.tsx
paulstn: Yep, will make the change now

public/components/event_analytics/explorer/__tests__/data_grid.test.tsx
```diff
@@ -28,21 +37,44 @@ describe('Datagrid component', () => {
...
+      <Provider store={store}>
+        <DataGrid
+          http={http}
+          pplService={pplService}
+          rows={DATA_GRID_ROWS}
+          rowsAll={[]}
+          explorerFields={explorerFields}
+          timeStampField={'timestamp'}
+          rawQuery={
+            "source = opensearch_dashboards_sample_data_logs | where match(request,'filebeat')"
```
ps48, line 61: Minor: Might want to move this to test constants: https://github.com/opensearch-project/dashboards-observability/blob/main/test/event_analytics_constants.ts

paulstn: Moved

derek-ho:
Can you add a screenshot of the new feel?

public/components/event_analytics/explorer/events_views/data_grid.tsx
derek-ho: Seems like page size default for me is 100 - can we double check this/pull this value if it is set globall?

paulstn: You're right - changed the default page size to 100

public/components/event_analytics/explorer/events_views/data_grid.tsx
```diff
@@ -3,148 +3,271 @@
...
+    storedSelectedColumns,
+  } = props;
+  const [rowRefs, setRowRefs] = useState<
+    Array<RefObject<{ closeAllFlyouts(openDocId: string): void }>>
+  >([]);
+  const { getEvents } = useFetchEvents({
+    pplService,
+    requestParams,
+  });
+  const sortingFields: MutableRefObject<GridSortingColumn[]> = useRef([]);
```
joshuali925, line 71: why is this a ref? the component should rerender if sort changes

also type would be `MutableRefObject<EuiDataGridSorting['columns']>` if you use the upstream type

common/types/explorer.ts
```diff
@@ -401,3 +401,8 @@ export type MOMENT_UNIT_OF_TIME =
   | 's'
   | 'milliseconds'
   | 'ms';
+
+export interface GridSortingColumn {
+  id: string;
+  direction: 'asc' | 'desc';
+}
```
joshuali925, line 408: this is good but prefer to reuse existing type declaration `EuiDataGridSorting['columns']`, so we don't need to change type if upstream changes

public/components/event_analytics/explorer/events_views/data_grid.tsx
```diff
@@ -3,148 +3,271 @@
...
+    redoQuery();
+  };
+
+  const setPage = (page: number[]) => {
+    pageFields.current = page;
+    redoQuery();
+  };
+
+  // creates the header for each column listing what that column is
+  const dataGridColumns = useMemo(() => {
```
joshuali925, line 117: any perf issue needs useMemo? i don't think datagrid itself is memorized, not sure if useMemo/useCallback helps here

public/components/event_analytics/explorer/events_views/data_grid.tsx
```diff
@@ -3,148 +3,271 @@
...
+      {
+        id: 'inspectCollapseColumn',
+        headerCellRender: () => null,
+        rowCellRender: ({ rowIndex }: { rowIndex: number }) => {
+          return (
+            <FlyoutButton
+              ref={null}
+              http={http}
+              key={null}
+              docId={'undefined'}
```
joshuali925, line 171: why defining ref and key to be null? and why is docId `'undefined'`?

public/components/event_analytics/explorer/events_views/data_grid.tsx
```diff
@@ -3,148 +3,271 @@
...
+        if (columnId === '_source') {
+          return (
+            <EuiDescriptionList type="inline" compressed>
+              {Object.keys(rows[trueIndex]).map((key) => (
+                <Fragment key={key}>
+                  <EuiDescriptionListTitle className="osdDescriptionListFieldTitle">
+                    {key}
+                  </EuiDescriptionListTitle>
+                  <EuiDescriptionListDescription
+                    dangerouslySetInnerHTML={{ __html: dompurify.sanitize(rows[trueIndex][key]) }}
```
joshuali925, line 201: why is this needed rather than passing as children?

public/components/event_analytics/explorer/events_views/docViewRow.tsx
```diff
@@ -36,7 +36,7 @@ interface IDocViewRowProps {
   onFlyoutOpen: (docId: string) => void;
 }
 
-export const DocViewRow = forwardRef((props: IDocViewRowProps, ref) => {
+export const FlyoutButton = forwardRef((props: FlyoutButtonProps, ref) => {
```
joshuali925, line 39: rename the file as well?

test/event_analytics_constants.ts
```diff
@@ -571,3 +571,6 @@ export const HORIZONTAL_BAR_TEST_VISUALIZATIONS_DATA = {
 export const EXPLORER_START_TIME = 'Aug 28, 2023 @ 20:00:00.406';
 
 export const EXPLORER_END_TIME = 'Aug 28, 2023 @ 20:00:00.408';
+
+export const EXPLORER_DATA_GRID_QUERY =
+  "source = opensearch_dashboards_sample_data_logs | where match(request,'filebeat')";
```
joshuali925, line 576: unused?

joshuali925 (commented):
minor comments

common/types/explorer.ts
paulstn: modified

public/components/event_analytics/explorer/events_views/data_grid.tsx
paulstn: I was modeling this after Discover's data grid, and they have memoized table attributes. They also have their entire data grid memoized, would it be better if I did that as well? I assumed this was done to stop the rerendering of constant (barring user change) attributes such as columns but would this have no effect without the entire table memoization?

public/components/event_analytics/explorer/events_views/docViewRow.tsx
paulstn: I would, but more in depth flyout changes are slated for p1 so I was going to get to that (along with many other changes) later.

test/event_analytics_constants.ts
paulstn: This is used in `public/components/event_analytics/explorer/__tests__/data_grid.test.tsx` as a specific raw query being passed into the data grid for testing.

public/components/event_analytics/explorer/events_views/data_grid.tsx
paulstn: I assume the thinking was to sanitize the returned data before displaying to prevent XSS but I don't think it's needed here. I changed it back to passing the row value as a child

test/event_analytics_constants.ts
joshuali925: move to test files if it's only used by tests?

public/components/event_analytics/explorer/events_views/data_grid.tsx
joshuali925: sanitize is needed for `dangerouslySetInnerHTML`, i'm asking is it necessary to `dangerouslySetInnerHTML` in the first place? if not probably just pass as child

public/components/event_analytics/explorer/events_views/data_grid.tsx
joshuali925: it can help if we run into performance issues, but right now useMemo doesn't really help since data grid is not memorized

paulstn: Should I memoize data grid, remove useMemo in my code, or just leave it alone?

public/components/event_analytics/explorer/events_views/data_grid.tsx
paulstn: done, i'm passing it in as a child.

test/event_analytics_constants.ts
paulstn: Isn't this the test constant file? If not, where would it be?

public/components/event_analytics/explorer/events_views/data_grid.tsx
joshuali925: i think you can remove or leave it if you don't see performance issues right now. we can add optimizations later when needed

test/event_analytics_constants.ts
joshuali925: my bad, i misread the filename, you can ignore this comment..

public/components/event_analytics/explorer/events_views/data_grid.tsx
paulstn: I'll leave it in, I believe we might need to include some performance optimizations in later

public/components/event_analytics/explorer/events_views/data_grid.tsx
paulstn: Changed the type accordingly.
This needs to be a ref because the state won't persist across rerenders otherwise. I tested it out by removing the useRef() (and all associated) portions and an error occurs where the portion on the toolbar the says "Sort fields" doesn't keep track of which fields were sorted (and their directions).

public/components/event_analytics/explorer/events_views/data_grid.tsx
paulstn: ref is a vestigial prop that will need to be removed later. 
key and docId are important, but for 'surrounding events', which as a whole isn't implemented yet but will be put in for p1. it won't have an effect on any other part of the flyout currently.

public/components/event_analytics/explorer/events_views/data_grid.tsx
paulstn: There is also an issue with having useState, it manifests itself as sort having a delay. Could be caused by getEvents, data cell rendering, or a few other things. Github Issue will be created after merge.

public/components/event_analytics/utils/__tests__/utils.test.tsx
```diff
@@ -114,4 +116,27 @@ describe('Utils event analytics helper functions', () => {
...
+              direction: 'asc',
+            },
+          ],
+        },
+        {
+          current: [0, 100],
+        },
+        () => {}
+      )
+    ).toBe(undefined);
```
joshuali925, line 140: i'm not sure how is expecting undefined validating the function?

you can probably at least do this
```suggestion
      const getEvents = jest.fn();
      redoQuery(
        'now/y',
        'now',
        "source = opensearch_dashboards_sample_data_logs | where match(request,'filebeat')",
        'timestamp',
        {
          current: [
            {
              id: 'timestamp',
              direction: 'asc',
            },
          ],
        },
        {
          current: [0, 100],
        },
        getEvents
      )
      expect(getEvents).toBeCalledWith(xxx)
```

public/components/event_analytics/explorer/events_views/data_grid.tsx
```diff
@@ -3,148 +3,251 @@
...
+        rowCellRender: ({ rowIndex }: { rowIndex: number }) => {
+          return (
+            <FlyoutButton
+              ref={null}
+              http={http}
+              key={null}
+              docId={'undefined'}
+              doc={rows[rowIndex % pageFields.current[1]]}
+              selectedCols={explorerFields.queriedFields}
+              timeStampField={timeStampField}
```
mengweieric, line 141: Should timeStampField be part of the useMemo dependency?

public/components/event_analytics/explorer/sidebar/sidebar.tsx
```diff
@@ -87,24 +91,41 @@ export const Sidebar = (props: ISidebarProps) => {
     });
   };
 
+  const checkWithStoredFields = () => {
+    if (
+      explorerFields.selectedFields.length === 0 &&
+      storedExplorerFields.selectedFields.length !== 0
+    ) {
+      return storedExplorerFields;
+    } else {
```
mengweieric, line 100: else can be removed? only with return statement?

mengweieric (approved):
minor comments

public/components/event_analytics/utils/__tests__/utils.test.tsx
paulstn: I didn't think to have jest test for a dummy getEvents, this seems like it would test the function well. Thanks for suggesting this, I've implemented it.

public/components/event_analytics/explorer/events_views/data_grid.tsx
paulstn: timeStampField is just a constant that is passed in as a prop from explorer.tsx. the intention for it being a prop is that if the user wants to define a time field instead of the default one, they can sort by that. but this is not implemented yet. I can still add it to the dependency array, though, for the future

public/components/event_analytics/explorer/sidebar/sidebar.tsx
paulstn: removed else

joshuali925 (approved):

PR #1054, opened 2023-09-27
Create prometheus datasource flow

derek-ho:
### Description
- Adds create prometheus datasource from UI flow
- Uses Application navigateToApp in order to navigate between different plugins
- Shows different linking cards in the data connections page based on the datasource type

### Issues Resolved

### Check List


public/components/datasources/components/new/review_s3_datasource_configuration.tsx
```diff
@@ -13,6 +13,7 @@ import {
   EuiFlexItem,
   EuiButton,
 } from '@elastic/eui';
+import { Role } from 'common/types/data_connections';
```
derek-ho, line 16: Missed this in the previous PR

opensearch_dashboards.json
```diff
@@ -16,7 +16,8 @@
     "savedObjects",
     "uiActions",
     "urlForwarding",
-    "visualizations"
+    "visualizations",
+    "management"
```
derek-ho, line 20: This is to add redirection to Management datasources on the opensearch card - not sure if we should add this here or in separate PR since it seems like that may or may not happen

public/components/datasources/components/new/configure_prometheus_datasource.tsx
```diff
@@ -0,0 +1,252 @@
...
+    switch (selectedAuthOption) {
+      case 'basic':
+        return (
+          <>
+            <EuiFormRow label="Username">
+              <EuiFieldText
+                placeholder={'Username placeholder'}
+                value={username}
+                onChange={(e) => setUsername(e.target.value)}
+                onBlur={(e) => setUsernameForRequest(e.target.value)}
```
derek-ho, line 93: @joshuali925 @ps48 was trying to figure a way around all this boiler plate code - issue is that when form is in a child component, re-rendering is causing it to lose focus when input changes. Am working around it with onBlur and temp var. in child component, which defaults to parent variable on load - is there a better way to do this?

public/components/datasources/components/new/new_datasource_card_view.tsx
```diff
@@ -16,33 +16,51 @@ import {
...
+    {
+      name: 'OPENSEARCH',
+      displayName: 'OpenSearch',
+      description: 'Connect to self managed OpenSearch clusters',
+      displayIcon: <EuiIcon type="logoOpenSearch" size="xl" />,
+      onClick: () =>
+        application!.navigateToApp(MANAGEMENT_APP_ID, {
+          path: 'opensearch-dashboards/dataSources',
+        }),
+    },
```
derek-ho, line 47: May want to just remove this card due to scope creep - wdyt?

public/components/datasources/components/manage/data_connection.tsx
```diff
@@ -103,6 +107,48 @@ export const DataConnection = (props: any) => {
...
+    switch (datasourceDetails.connector) {
+      case 'S3GLUE':
+        return (
+          <EuiFlexGroup>
+            <EuiFlexItem>
+              <EuiCard
+                icon={<EuiIcon size="xxl" type="discoverApp" />}
+                title={'Query data'}
+                description="Query your data in Data Explorer or Observability Logs."
+                onClick={() => application!.navigateToApp(observabilityLogsID)}
```
derek-ho, line 121: Need another set of eyes here that this is the right way to navigate between apps

public/components/datasources/components/manage/data_connection.tsx
```diff
@@ -103,6 +108,99 @@ export const DataConnection = (props: any) => {
...
+              />
+            </EuiFlexItem>
+            <EuiFlexItem>
+              <EuiCard
+                icon={<EuiIcon size="xxl" type="bolt" />}
+                title={'Accelerate performance'}
+                description="Accelerate performance through OpenSearch indexing."
+                onClick={() => {}}
```
derek-ho, line 130: @sumukhswamy @ps48 still not sure what this card is supposed to do, since acceleration is no longer another page?

public/components/datasources/components/manage/data_connection.tsx
```diff
@@ -103,6 +108,99 @@ export const DataConnection = (props: any) => {
...
+              />
+            </EuiFlexItem>
+            <EuiFlexItem>
+              <EuiCard
+                icon={<EuiIcon size="xxl" type="database" />}
+                title={'Tables'}
+                description="Manually Define Tables"
+                onClick={() => application!.navigateToApp('opensearch-query-workbench')}
```
derek-ho, line 138: Will keep it like this for now, but I think ideally there should be additional path needed for workbench to correctly display the right datasource?

public/components/datasources/components/new/new_datasource_card_view.tsx
```diff
@@ -16,33 +16,37 @@ import {
...
+  onClick: () => void;
 }
 
-const Datasources: DatasourceCard[] = [
-  {
-    name: 'OPENSEARCH',
-    displayName: 'OpenSearch',
-    description: 'Connect to self managed OpenSearch clusters',
-    displayIcon: <EuiIcon type="logoOpenSearch" size="xl" />,
-  },
```
derek-ho, line 34: Removed opensearch for now - if we can add it back later in the right way will do so - this specific card needs to not be shown if datasources is not there

public/framework/core_refs.ts
```diff
@@ -3,7 +3,13 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import { ChromeStart, HttpStart, IToasts } from '../../../../src/core/public';
+import {
+  ApplicationStart,
+  ChromeStart,
+  HttpStart,
+  IToasts,
+  IUiSettingsClient,
```
ps48, line 11: are we using `IUiSettingsClient` anywhere?

public/components/datasources/components/new/configure_prometheus_datasource.tsx
ps48: What is the cause of re-rendering? Suspecting: Should the switch statement be in a useEffect()?

public/components/datasources/components/manage/connection_configuration.tsx
```diff
@@ -127,7 +126,7 @@ export const ConnectionConfiguration = (props: ConnectionConfigurationProps) =>
           </EuiFormRow>
           <EuiFormRow label="Password">
             <EuiFieldPassword
-              type={dual ? 'dual' : undefined}
+              type={'dual'}
```
ps48, line 129: Minor: We can just keep this`type="dual"`

public/components/datasources/components/manage/data_connection.tsx
ps48: IMO This is good!

public/components/datasources/components/manage/data_connection.tsx
ps48: This will redirect to query workbench with accelerate table flyout open. Can keep it blank for now.

public/components/datasources/components/manage/data_connection.tsx
ps48: @sumukhswamy  Can help you out here. We need to pass in `current data source` here in URL.

public/components/datasources/components/new/configure_prometheus_datasource.tsx
derek-ho: What I am thinking - I am passing down the value for that field (for example to persist across navigating to different part of the form and coming back). However, when the child component's onChange is triggered, this updates the parent's value of the variable. This then causes all the children to re-render, and the field loses focus, since it isn't focused by default. The way I am working around this for now/in this PR is to only trigger the parent's update `setValue` hook `onBlur` (i.e. when the user stops typing into the field). However, this is causing a lot of boiler plate with needing another temporary value in the child component. Is there a cleaner way around this?

public/components/datasources/components/new/configure_datasource.tsx
```diff
@@ -73,13 +94,37 @@ export function Configure(props: ConfigureDatasourceProps) {
...
+            currentPassword={password}
+            setPasswordForRequest={setPassword}
+            selectedQueryPermissionRoles={selectedQueryPermissionRoles}
+            setSelectedQueryPermissionRoles={setSelectedQueryPermissionRoles}
+            currentAccessKey={accessKey}
+            currentSecretKey={secretKey}
+            setAccessKeyForRequest={setAccessKey}
+            setSecretKeyForRequest={setSecretKey}
+            currentRegion={region}
+            setRegionForRequest={setRegion}
```
ps48, line 125: Question: Is there a way to reduce the number of props here?

derek-ho: This is side effect of this issue: https://github.com/opensearch-project/dashboards-observability/pull/1054#discussion_r1340408705, if there is a cleaner way to do it I am all for it, but I couldn't find one with quick search

ps48 (approved):
Added minor comments

public/components/datasources/components/new/configure_datasource.tsx
derek-ho: This wouldn't be an issue if we render the form inline, but rendering the form as a child is causing issues

opensearch_dashboards.json
ps48: removing out should be good here. As the UI is also removed for this as of now.

public/components/datasources/components/new/configure_datasource.tsx
derek-ho: Maybe mis-understood your question here - I don't think there is a way to do so, since I need to know the values in the parent in order to send the request and have a way of updating them in the child. Thus I need to pass in value and setter for each of the possible configuration properties. Consolidating it into an object might be cleaner, but I think this  way makes it clear what is going in and as what. Leaning towards keeping it like this for now

public/components/datasources/components/manage/data_connection.tsx
derek-ho: Will keep it blank for now, when endpoint is ready will update it separately

public/components/datasources/components/manage/data_connection.tsx
derek-ho: Will keep it general for now, and update it later when it is ready

public/framework/core_refs.ts
derek-ho: Missed this - was testing out to see about reading feature flag values, but that is not priority right now, thanks for the catch!

public/components/datasources/components/manage/connection_configuration.tsx
derek-ho: Done

public/components/datasources/components/manage/data_connection.tsx
```diff
@@ -103,6 +108,99 @@ export const DataConnection = (props: any) => {
...
```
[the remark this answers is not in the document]
derek-ho: Good suggestion - made it into a `Default` component, and use it for default and S3 case

Swiddis (approved):

PR #1052, opened 2023-09-27
Update UI for integrations setup

Swiddis:
### Description
Per UX team feedback, this PR updates the UI for integrations setup, and also restores the original functionality for creating integrations. Integration tests still need to be rewritten and the validation functionality needs to be finished.

image

### Issues Resolved
N/A

### Check List
- [X] New functionality includes testing.
  - [X] All tests pass, including unit test, integration test and doctest
- [ ] New functionality has been documented.
  - [ ] New functionality has javadoc added
  - [ ] New functionality has user manual doc added
- [X] Commits are signed per the DCO using --signoff

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #1057, opened 2023-09-29
Setup S3 connection with integrations

Swiddis:
### Description

Followup to #1052, this PR adds actual S3 integration functionality.

Work in progress -- Main querying pipeline is good, but need to do some work for the UI.

### Issues Resolved
N/A

### Check List


public/components/integrations/components/__tests__/__snapshots__/setup_integration.test.tsx.snap
```diff
@@ -50,27 +50,30 @@ exports[`Integration Setup Page Renders integration setup page as expected 1`] =
                         Object {
                           "connectionDataSource": "",
                           "connectionType": "index",
-                          "displayName": "sample Integration",
+                          "displayName": "[object Object] Integration",
```
derek-ho, line 53: This looks wrong, can you take another look @Swiddis

Swiddis: Woops, forgot to update a test -- fixed

YANG-DB (approved):

public/components/integrations/components/setup_integration.tsx
```diff
@@ -118,10 +119,52 @@ const suggestDataSources = async (type: string): Promise<Array<{ label: string }
...
+  query: string,
+  trackProgress: (step: number) => void
+): Promise<Result<object>> => {
+  // Used for polling
+  const sleep = (ms: number) => {
+    return new Promise((resolve) => setTimeout(resolve, ms));
+  };
+
+  try {
+    console.log('- Posting query');
```
joshuali925, line 132: nit remove logs

kavithacm, line 132: Are these console comments necessary ? also in line no 151

public/components/integrations/components/setup_integration.tsx
```diff
@@ -118,10 +119,52 @@ const suggestDataSources = async (type: string): Promise<Array<{ label: string }
...
+        trackProgress(3);
+        return { ok: true, value: poll };
+      } else if (poll.status === 'FAILURE') {
+        return { ok: false, error: new Error('FAILURE status', { cause: poll }) };
+      }
+      await sleep(3000);
+    }
+  } catch (err: any) {
+    console.error(err);
+    return { ok: false, error: err };
```
joshuali925, line 166: i think this works but usually it's `error, loading, data`. If loading is true then display loading, else if error is not null then display error, otherwise data is useable. for example like https://swr.vercel.app/#overview

public/components/integrations/components/setup_integration.tsx
```diff
@@ -231,16 +296,64 @@ export function SetupBottomBar({
...
+                console.log('Beginning queries');
+                // Queries must exist because we disable s3 if they're not present
+                for (const query of assets.data.queries!) {
+                  console.log('Query:', query);
+                  const queryStr = query.query.replace('${TABLE}', config.connectionDataSource);
+                  const result = await runQuery(queryStr, (step) => setProgress(progress + step));
+                  if (!result.ok) {
+                    console.error('Query failed', result.error);
+                    setLoading(false);
+                    setToast('Something went wrong.', 'danger');
```
kavithacm, line 335: can we change this toast to more appropriate wording?

public/components/integrations/components/setup_integration.tsx
```diff
@@ -169,7 +212,22 @@ export function SetupIntegrationForm({
       <EuiSpacer />
       <EuiFormRow label="Data Source" helpText="Select a data source to connect to.">
         <EuiSelect
-          options={integrationConnectionSelectorItems}
+          options={integrationConnectionSelectorItems.map((item) => {
+            const copy: { value: string; text: string; disabled?: boolean } = Object.assign(
+              {},
+              item
```
joshuali925, line 218: nit use `{...item}` to copy object

public/components/integrations/components/setup_integration.tsx
derek-ho: +1

public/components/integrations/components/__tests__/__snapshots__/setup_integration.test.tsx.snap
```diff
@@ -55,22 +36,8 @@ exports[`Integration Setup Page Renders integration setup page as expected 1`] =
...
-                            Object {
-                              "name": "logs",
-                              "version": "1.0.0",
-                            },
-                          ],
-                          "license": "Apache-2.0",
                           "name": "sample",
-                          "type": "logs",
-                          "version": "2.0.0",
+                          "type": "",
```
derek-ho, line 40: Why are you removing type here

public/components/integrations/components/setup_integration.tsx
```diff
@@ -231,16 +296,64 @@ export function SetupBottomBar({
...
+                const http = coreRefs.http!;
+
+                console.log('Fetching assets');
+                const assets = await http.get(
+                  `${INTEGRATIONS_BASE}/repository/${integration.name}/assets`
+                );
+                progress += 1;
+                setProgress(progress);
+
+                console.log('Beginning queries');
```
kavithacm, line 326: can we remove unwanted console log comments?

derek-ho, line 326: Please remove console logs

public/components/integrations/components/setup_integration.tsx
```diff
@@ -118,10 +119,52 @@ const suggestDataSources = async (type: string): Promise<Array<{ label: string }
...
+    const queryId = (
+      await http.post(CONSOLE_PROXY, {
+        body: JSON.stringify({ query, lang: 'sql' }),
+        query: {
+          path: '_plugins/_async_query',
+          method: 'POST',
+        },
+      })
+    ).queryId;
+    while (true) {
```
derek-ho, line 143: Is there better way than while true?

derek-ho (approved):
Approving to unblock, left a few comments

public/components/integrations/components/setup_integration.tsx
```diff
@@ -231,16 +296,64 @@ export function SetupBottomBar({
...
-                false,
-                integration.name,
-                config.displayName,
-                template,
-                setToast,
-                config.displayName,
-                config.connectionDataSource
-              );
+              // Not sure why if I make an incrementer it doesn't change the prop,
+              // But since it doesn't, we track our progress manually.
```
joshuali925, line 300: could you explain a bit what's an incrementer?

derek-ho:
@Swiddis can you fix DCO

joshuali925 (approved):

public/components/integrations/components/setup_integration.tsx
Swiddis: What's more appropriate wording? We don't have any error information in the response that we can use

public/components/integrations/components/setup_integration.tsx
Swiddis: If I make a helper `const incrementProgress = () => setProgress(progress + 1)`, it doesn't correctly identify when progress updates, so it always computes `0 + 1`. Similar situation if I use `setProgress(progress + 1)` directly. The workaround is to manually track progress locally and use that instead of the actual progress variable.

public/components/integrations/components/setup_integration.tsx
Swiddis: Not sure what kind of polling timeout to put on it, I've heard some queries can take up to 10 mins and don't have any reference for when I should just tell the user that it isn't working

public/components/integrations/components/__tests__/__snapshots__/setup_integration.test.tsx.snap
Swiddis: Stopped filling a default value when the integration isn't loaded yet -- We only use the type when we go to integration adding so it's safe to keep empty until it's loaded.

public/components/integrations/components/setup_integration.tsx
joshuali925: use `setProgress((progress) => progress + 1)`. it's a bit of counterintuitive issue in react where the value of progress is saved in the function

Swiddis:
Dealt with small-scale PR feedback -- More large-scale changes will need to happen later. Just waiting on tests before merging.

PR #1221, opened 2023-11-01
implements explorer data grid column reordering

paulstn:
### Description
Implements data grid column reordering within the column toolbar drag and drop. Also implements it within the sidebar fields drag and drop, where the order of the selected fields can be manipulated but the available fields will still be sorted alphabetically.

### Issues Resolved
This PR also resolves an issue within app analytics where the sidebar isn't populated with any fields.

### Check List

### DIFF
--- a/public/components/event_analytics/explorer/events_views/data_grid.tsx
+++ b/public/components/event_analytics/explorer/events_views/data_grid.tsx
@@ -138,6 +138,7 @@ export function DataGrid(props: DataGridProps) {
         visibleColumns: columns,
         setVisibleColumns: (visibleColumns: string[]) => {
           const fields: IField[] = [];
+          // fields within visibleColumns are re-created to be of type IField
           visibleColumns.map((col) => {
             fields.push({ name: col, type: getFieldTypes(col, explorerFields) });
           });
@@ -280,7 +281,7 @@ export function DataGrid(props: DataGridProps) {
           toolbarVisibility={{
             showColumnSelector: {
               allowHide: false,
-              allowReorder: true,
+              allowReorder: explorerFields.selectedFields.length > 0,
             },
             showFullScreenSelector: false,
             showStyleSelector: false,
--- a/public/components/event_analytics/explorer/events_views/data_grid.tsx
+++ b/public/components/event_analytics/explorer/events_views/data_grid.tsx
@@ -15,18 +15,21 @@ import {
 } from '@elastic/eui';
 import moment from 'moment';
 import { MutableRefObject } from 'react';
+import { useDispatch } from 'react-redux';
 import { IExplorerFields, IField } from '../../../../../common/types/explorer';
 import {
   DATE_DISPLAY_FORMAT,
   DEFAULT_EMPTY_EXPLORER_FIELDS,
   DEFAULT_SOURCE_COLUMN,
   DEFAULT_TIMESTAMP_COLUMN,
+  SELECTED_FIELDS,
 } from '../../../../../common/constants/explorer';
 import { HttpSetup } from '../../../../../../../src/core/public';
 import PPLService from '../../../../services/requests/ppl';
 import { FlyoutButton } from './docViewRow';
 import { useFetchEvents } from '../../hooks';
 import { redoQuery } from '../../utils/utils';
+import { updateFields } from '../../redux/slices/field_slice';
 
 interface DataGridProps {
   http: HttpSetup;
@@ -41,6 +44,7 @@ interface DataGridProps {
   startTime: string;
   endTime: string;
   storedSelectedColumns: IField[];
+  tabId: string;
 }
 
 export function DataGrid(props: DataGridProps) {
@@ -56,11 +60,13 @@ export function DataGrid(props: DataGridProps) {
     requestParams,
     startTime,
     endTime,
+    tabId,
   } = props;
   const { fetchEvents } = useFetchEvents({
     pplService,
     requestParams,
   });
+  const dispatch = useDispatch();
   const selectedColumns =
     explorerFields.selectedFields.length > 0
       ? explorerFields.selectedFields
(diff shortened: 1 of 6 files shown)


public/components/event_analytics/explorer/events_views/data_grid.tsx
```diff
@@ -131,7 +137,22 @@ export function DataGrid(props: DataGridProps) {
       return {
         visibleColumns: columns,
         setVisibleColumns: (visibleColumns: string[]) => {
-          // TODO: implement with sidebar field order (dragability) changes
+          if (explorerFields.selectedFields.length > 0) {
+            const fields: IField[] = [];
+            // fields within visibleColumns are re-created to be of type IField
+            visibleColumns.map((col) => {
```
mengweieric, line 143: Should use forEach as you are not using the newly generated array at all

public/components/event_analytics/explorer/events_views/data_grid.tsx
```diff
@@ -262,7 +283,7 @@ export function DataGrid(props: DataGridProps) {
           toolbarVisibility={{
             showColumnSelector: {
               allowHide: false,
-              allowReorder: true,
+              allowReorder: explorerFields.selectedFields.length > 0,
```
mengweieric, line 286: If by default only have timestamp and _source for initial state, then there's no selected fields explorerFields.selectedFields, but datagrid should still allow user to reorder timnestamp and _source is it?

public/components/event_analytics/explorer/sidebar/sidebar.tsx
```diff
@@ -139,14 +135,32 @@ export const Sidebar = (props: ISidebarProps) => {
     source: any;
     draggableId: string;
   }) => {
+    console.log(destination);
```
mengweieric, line 138: remove console.log?

public/components/event_analytics/utils/utils.tsx
```diff
@@ -442,3 +442,15 @@ export const redoQuery = (
     setData(res.jsonData);
   });
 };
+
+// method to return the type of a field from its name
+export const getFieldTypes = (newFieldName: string, explorerFields: IExplorerFields) => {
+  let fieldType: string = '';
```
mengweieric, line 448: we don't need this line, we can actually turn it into something 

```
export const getFieldTypes = (newFieldName: string, explorerFields: IExplorerFields): string => {
  for (const field of explorerFields.availableFields) {
    if (field.name === newFieldName) return field.type;
  }
  for (const field of explorerFields.selectedFields) {
    if (field.name === newFieldName) return field.type;
  }
  return '';
}
```

public/components/event_analytics/explorer/events_views/data_grid.tsx
paulstn: Here i'm actually disabling reordering when no fields are selected, so that no reordering is done when the default columns appear in data grid. The reason I do this is because the default columns are 'artificial fields' and I don't want them to appear as objects in the sidebar.

public/components/event_analytics/explorer/sidebar/sidebar.tsx
paulstn: removed

public/components/event_analytics/explorer/sidebar/sidebar.tsx
```diff
@@ -55,79 +56,74 @@ export const Sidebar = (props: ISidebarProps) => {
...
-        })
-      );
-      dispatch(
-        sortFields({
-          tabId: tabID,
-          data: [modifiedField],
-        })
-      );
-    });
+  const updateStoreFields = (fieldsData: ExplorerFields) => {
```
mengweieric, line 90: We need useCalkback here is it? As I'm seeing you add this as part of dependency of another useCallback

public/components/event_analytics/explorer/events_views/data_grid.tsx
paulstn: changed to forEach

public/components/event_analytics/explorer/sidebar/sidebar.tsx
```diff
@@ -55,79 +56,74 @@ export const Sidebar = (props: ISidebarProps) => {
...
+      updateFields({
+        tabId,
+        data: {
+          ...fieldsData,
+        },
+      })
+    );
+  };
+
+  const sortStoreFields = (fieldName: string) => {
```
mengweieric, line 101: useCallback?

public/components/event_analytics/explorer/events_views/data_grid.tsx
mengweieric: then it doesn't match with existing behavior in discover

public/components/event_analytics/utils/utils.tsx
paulstn: modified, and now using forEach instead of map

public/components/event_analytics/explorer/events_views/data_grid.tsx
paulstn: existing behavior is discover is bugged a little bit, the default columns don't get reordered when done through the data grid even though the names move and the sidebar also has some inconsistencies. I extrapolated that since the default columns don't actually move themselves, then I should disable the ability to attempt at doing so.

paulstn: existing behavior is discover is bugged a little bit, the default columns don't get reordered when done through the data grid even though the names move and the sidebar also has some inconsistencies. I extrapolated that since the default columns don't actually move themselves, then I should disable the ability to attempt at doing so.

public/components/event_analytics/explorer/sidebar/sidebar.tsx
paulstn: added

public/components/event_analytics/explorer/explorer.tsx
```diff
@@ -899,12 +900,12 @@ export const Explorer = ({
           <EuiSplitPanel.Outer className="eui-yScroll" hasBorder={true} borderRadius="none">
             {!appLogEvents && (
               <EuiSplitPanel.Inner paddingSize="s" color="subdued" grow={false}>
-                <DataSourceSelection tabId={initialTabId} />
+                <DataSourceSelection tabId={tabId} />
```
mengweieric, line 903: please remove the import from top 

```
import { initialTabId } from '../../../framework/redux/store/shared_state';
```

mengweieric (approved):

public/components/event_analytics/explorer/explorer.tsx
paulstn: removed

mengweieric:
@paulstn please resolve the conflicts

PR #1265, opened 2023-11-21
Added HAProxy Integration

Pothula-Kalyani-ML:
### Description
Added new observability integration for HAProxy
### Issues Resolved
Ability to monitor HAProxy Access Logs
### Check List
- [X] New functionality has been documented.
  - [X] New functionality has user manual doc added
### Results
image
image


server/adaptors/integrations/__data__/repository/haproxy/info/test/.gitignore
```diff
@@ -0,0 +1 @@
+otel-collector/log.json
```
Swiddis, line 1: Where is otel-collector involved? It's not in the docker compose

server/adaptors/integrations/__data__/repository/haproxy/info/test/docker-compose.yaml
```diff
@@ -0,0 +1,118 @@
...
+      - "9200"
+    healthcheck:
+      test: ["CMD", "curl", "-f", "http://localhost:9200/_cluster/health?wait_for_status=yellow"]
+      interval: 5s
+      timeout: 25s
+      retries: 4
+    networks:
+      - opensearch-net # All of the containers will join the same Docker bridge network
+  opensearch-dashboards:
+    image: pothulakalyanireddy/dashboard-app:dev
```
Swiddis, line 38: In general I'd like to avoid relying on external accounts for test images, would it be possible to use e.g. the Dockerfile in the repository root to set up an instance? Alternatively we can remove this section entirely and assume that the OSD instance is run on the outside and connecting to the OS instance here, that's how I usually do development with integrations.

For version compatibility with something based on this branch,  the opensearch image would also need to be `opensearchstaging/opensearch:3.0.0`.

server/adaptors/integrations/__data__/repository/haproxy/info/test/docker-compose.yaml
```diff
@@ -0,0 +1,118 @@
...
+            [opensearch]
+    networks:
+      - opensearch-net
+  redis:
+    image: redis
+    ports:
+      - 6357:6357
+    networks:
+      - opensearch-net
+  flask-app:
```
Swiddis, line 103: I'm not sure I follow why this app and redis are included? They don't seem to be referenced by the haproxy setup at all.

server/adaptors/integrations/__data__/repository/haproxy/info/test/fluent-bit/fluent-bit.conf
```diff
@@ -0,0 +1,30 @@
...
...c.code.." "..c.size.." \""..c.error.."\"" else d["body"]=c.remote.." "..c.frontend.." "..c.Backend..""..c.logasap.." "..c.fc_hex_c_ca_res..""..c.sq_bq..""..c.reqcookie..""..c.rescookie..""..c.terminationstate.."["..os.date("%d/%b/%Y:%H:%M:%S %z").."] \""..c.method.." "..c.path.." HTTP/1.1\" "..c.code.." "..c.size.." \""..c.agent.."\"" end d["trace_id"]="102981ABCD2901"d["span_id"]="abcdef1010"d["attributes"]={}d["attributes"]["data_stream"]={}d["attributes"]["data_stream"]["dataset"]="haproxy.access"d["attributes"]["data_stream"]["namespace"]="production"d["attributes"]["data_stream"]["type"]="logs"d["event"]={}d["event"]["category"]={"web"}d["event"]["name"]="access"d["event"]["domain"]="haproxy.access"d["event"]["kind"]="event"d["event"]["result"]="success"d["event"]["type"]={"access"}d["http"]={}d["http"]["request"]={}d["http"]["request"]["method"]=c.method;d["http"]["response"]={}d["http"]["response"]["bytes"]=tonumber(c.size)d["http"]["response"]["status_code"]=c.code;d["http"]["flavor"]="1.1"d["http"]["url"]=c.path;d["communication"]={}d["communication"]["source"]={}d["communication"]["source"]["address"]="127.0.0.1"d["communication"]["source"]["ip"]=c.remote;return 1,b,d end
```
Swiddis, line 17: This filter looks to be generating data that isn't compatible with the provided schemas, if you set up the indices before importing data then fluentbit reports ingestion issues and no data shows up. For working with these filters better I've taken to [putting the filter in a separate file](https://gist.github.com/Swiddis/90a3767e01d4d0698436b3580adb9385), it makes debugging these issues a lot easier.

It actually looks like the issues are because this is an older version of the same filter, so the version in the Gist should work if you add the below haproxy parser and an appropriate formatter.

server/adaptors/integrations/__data__/repository/haproxy/info/test/docker-compose.yaml
```diff
@@ -0,0 +1,118 @@
...
+    networks:
+      - opensearch-net
+  haproxy:
+    container_name: haproxy
+    image: haproxytech/haproxy-alpine:2.4
+    volumes:
+      # - ./haproxy-otel/opentelemetry_module.conf:/etc/nginx/conf.d/opentelemetry_module.conf
+      - ./haproxy.cfg:/usr/local/etc/haproxy/haproxy.cfg:ro
+    ports:
+      - 8080:8080
```
Swiddis, line 70: If I understand correct, this should be 8405, since that's the port haproxy is listening on. 8080 isn't set up in the config.

Swiddis (changes requested):
Thanks for the contribution, and especially for including a docker-compose file. It helps a ton with reviewing faster. There are still a few questions I have about getting it set up, I think the bulk of the issue is just that this is built off of an old test compose that needs a couple updates to accommodate the new setup.

server/adaptors/integrations/__data__/repository/haproxy/info/test/docker-compose.yaml
Pothula-Kalyani-ML: Hi, with this version of opensearchstaging/opensearch:3.0.0 the dashboard version (2.11.0)  opensearchproject/opensearch-dashboards:latest is incompatible facing issue of (OpenSearch dashboard server is not yet ready) 
. I tried using opensearchstaging/opensearch-dashboards:3.0.0 in this I could not find integrations. 
image

server/adaptors/integrations/__data__/repository/haproxy/info/test/.gitignore
Pothula-Kalyani-ML: yes , it is not being used, so removed it.

server/adaptors/integrations/__data__/repository/haproxy/info/test/fluent-bit/fluent-bit.conf
Pothula-Kalyani-ML: Made a commit with your suggestions.

server/adaptors/integrations/__data__/repository/haproxy/info/test/docker-compose.yaml
Pothula-Kalyani-ML: Yes, your right that should be 8405, updated this in my latest commit.

server/adaptors/integrations/__data__/repository/haproxy/info/test/docker-compose.yaml
Swiddis: It's because the staging image doesn't include plugins, I'm not sure if we have a widely-used method to add a dashboard image that has some plugins installed. I think a good route is to just remove the dashboards image from this compose file and let the user set it up externally.

server/adaptors/integrations/__data__/repository/haproxy/info/test/fluent-bit/otel-converter.lua
```diff
@@ -0,0 +1,82 @@
...
+                bytes=tonumber(record.size),
+                status_code=tonumber(record.code)
+            },
+            flavor="1.1",
+            url=record.path
+        },
+        communication={
+            source={
+                address="127.0.0.1",
+                ip=record.remote
```
Swiddis, line 77:
```suggestion
                address=record.remote,
                ip=string.match(record.remote, "%d+\\.%d+\\.%d+\\.%d+") -- If it has an IP, add it
```
Looks like I had a bug here with how I was handling IP, it doesn't work when `record.remote` has a port in it.

server/adaptors/integrations/__data__/repository/haproxy/info/test/docker-compose.yaml
Pothula-Kalyani-ML: Thank you so much for the support.

server/adaptors/integrations/__data__/repository/haproxy/info/test/fluent-bit/otel-converter.lua
Pothula-Kalyani-ML: Updated the file as requested.

Swiddis (approved):
LGTM!

YANG-DB (approved):

Issue #1039, opened 2023-09-21
[BUG] Outdated Cypress tests in `9_integrations`


Swiddis:
**What is the bug?**
In the [latest test run](https://github.com/opensearch-project/dashboards-observability/actions/runs/6255381133/job/16984612121), 4/10 Cypress tests for integrations are failing due to the new set up UI.

**How can one reproduce the bug?**
CI action, also cypress tests locally.

**What is the expected behavior?**
The tests should pass.

**What is your host/environment?**
 - OS: Mac
 - Commit: d67ec95

**Do you have any screenshots?**

**Do you have any additional context?**
Peter has talked a lot about implementing feature flags, that would be a good move here, so we can continue development without needing to comment out tests for a broken UI. It would allow future work to be better about making small PRs, too.

Swiddis:
Tests are broken again -- Community contribution added 2 more assets for 6, and the tests still expect 4.

Issue #1040, opened 2023-09-21
[BUG] Tracking issue for failing Cypress tests.


Swiddis:
**What is the bug?**
Now that we have a running E2E action, we need to fix our Cypress tests in CI conditions. I suspect many of the failures are due to the CI environment itself not being set up correctly, lots of them may be outdated, and maybe a handful could indicate bugs that have been untested for a while. This issue is a tracking issue for fixing our tests.

- #1039 
- #1278 

**How can one reproduce the bug?**
Run the Cypress E2E action. Because the actions support manual running, you can try it on your fork directly, but they're also run nightly.

**What is the expected behavior?**
The tests should pass.

**What is your host/environment?**
 - Defined in [the workflow file](https://github.com/opensearch-project/dashboards-observability/blob/main/.github/workflows/integration-tests-workflow.yml).

**Do you have any screenshots?**
N/A

**Do you have any additional context?**
Once we fix them at home, we can look at fixing the tests in [FTR](https://github.com/opensearch-project/opensearch-dashboards-functional-test/).

Swiddis:
Superseded by #1338

Issue #957, opened 2023-08-24
[Cypress12] Ensure cypress functional tests run with Cypress12


cwperks:
The [opensearch-dashboards-functional-test repo](https://github.com/opensearch-project/opensearch-dashboards-functional-test/issues/801) is working on upgrading the version of Cypress used in the repo from v9 to v12.

There is a feature branch in the repository called [`feature/cypress12`](https://github.com/opensearch-project/opensearch-dashboards-functional-test/tree/feature/cypress12) to work on the upgrade and ensure that all repositories that have functional tests in the repo can verify that their tests run successfully with the upgraded version of cypress. If any updates to tests are required please submit a PR against the function test repo and base the PR on the `feature/cypress12` branch to show the tests running with the upgraded version of Cypress. 

Please refer to the [migration guide](https://docs.cypress.io/guides/references/migration-guide) from Cypress to reference when updating tests. 

If no work is required to update the tests then please close this issue and indicate that no update is required.

Swiddis:
We've already migrated to Cypress 12, no further work needed.

cwperks:
@Swiddis This issue is about `dashboards-observability` tests in the functional test repo. Have you checked to see if those run with Cypress 12 or if any updates are required?

I saw that the tests failed in this run: https://github.com/opensearch-project/opensearch-dashboards-functional-test/actions/runs/5933730596/job/16089570575

derek-ho:
@Swiddis this is for FTR repo: https://github.com/opensearch-project/opensearch-dashboards-functional-test. We have not yet verified this runs with cypress 12

PR #1290, opened 2023-12-07
[CI/CD] Refactor Cypress Workflow

RyanL1997:
### Description
- Fix the integration test case in the Cypress integration suite
- Add a dashboard status check in the integration CI workflow
- Separate cypress test suites into its own GH runner
- Setup cypress tests directories for different testing categories

The current state is that the workflow of Dashboards observability plugin E2E test is able to be run and the the cypress framework can work properly (see log here: https://github.com/RyanL1997/dashboards-observability/actions/runs/7145888298/job/19462486749). However, the test cases is still failing, and it should be resolved in the follow up PR for #1040.

### Issues Resolved
* Resolve https://github.com/opensearch-project/dashboards-observability/issues/1039
* Relate https://github.com/opensearch-project/dashboards-observability/issues/1040

### Check List


Swiddis (approved):
Thanks!

package.json
```diff
@@ -53,7 +53,7 @@
     "@types/react-test-renderer": "^16.9.1",
     "@types/sanitize-filename": "^1.6.3",
     "antlr4ts-cli": "^0.5.0-alpha.4",
-    "cypress": "^12.8.1",
+    "cypress": "^13.6.1",
```
mengweieric, line 56: Should we include major version bump in this PR without testing?

Swiddis: Some time back I looked through the Cypress 12 -> 13 migration guide and none of the breaking changes were affecting us. Integration tests are still failing with both versions, though.

ps48: Should we be on the same version as FTR, cause we have some tests here and some basic tests in FTR. Wouldn't it be difficult for developers to maintain syntax of two different versions? 

I'm fine with either, just putting my perspective for long term easiness.

.github/workflows/integration-tests-workflow.yml
```diff
@@ -114,13 +114,29 @@ jobs:
...
+              exit 1
+            fi
+            sleep 10
+          done
 
       - name: Install Cypress
         run: |
           cd ./OpenSearch-Dashboards/plugins/dashboards-observability
-          npx cypress install
+          npm install cypress --save-dev
```
joshuali925, line 139: cypress is already defined in package.json, why is the installation a separate step with npm?

joshuali925 (approved):

package.json
Swiddis: There's already an open issue to migrate the FTR tests to 12: #957. The compatible route is to stay on 12, but in the 12->13 case specifically, it should bring extra performance gains under the hood without any code difference overhead.

mengweieric: Same thoughts here, both are fine, but seems like there's no direct and necessary benefit from this version bump for this PR, we could do it with separate PR or later in a step by step manner.

.github/workflows/integration-tests-workflow.yml
RyanL1997: Good catch. I was using this one only to get the latest version of cypress in my local. I have reverted this line back.

RyanL1997:
Transferring some of the conversation between me and @ps48 here: this PR will also refactor the current workflow structure and separate the test cases into parallel runs

package.json
RyanL1997: > There's already an open issue to migrate the FTR tests to 12: https://github.com/opensearch-project/dashboards-observability/issues/957. The compatible route is to stay on 12, but in the 12->13 case specifically, it should bring extra performance gains under the hood without any code difference overhead.

Thanks for the reference @Swiddis. For some additional context, that issue was originally kicked off by the security team, security frontend needs the cypress version >11 for testing external auth. However, since the FTR has its own overwrite commands with basic auth only (such as `cy.visit()`, `cy.request()`) and tons of tests were relying on these commands, it was hard to do the refactor work. But later on the security team introduced the Cypress 13 in their own repo. In addition, compare to Cypress 9, newer version does have a huge performance improvement.

At the same time, I agree with @mengweieric and @ps48. I can remove the version bump in this one then and maybe we can do the version bump as a follow-up. However, from my perspective, I think it is fine to stay in Cypress 12 then, and we don't need to downgrade our version to match with FTR. Cuz we are using this dependency only for our own workflow, and we don't have any syntax conflict at this point.

RyanL1997:
The refactor has been done. The test cases have been separated into its own GH runner. We have also setup the directory for future splitting of the test classes. The sample run is here: https://github.com/RyanL1997/dashboards-observability/actions/runs/7145888298/job/19462486749

 (cc: @ps48, @Swiddis , @mengweieric, @joshuali925 )

package.json
RyanL1997: The version bump of this PR has been removed.

Swiddis:
Still LGTM -- One nit: `notebook` is the only one that doesn't have a `_test` suffix, would like to keep them consistent.

.github/workflows/integration-tests-workflow.yml
```diff
@@ -23,10 +23,20 @@ jobs:
       matrix:
         os: [ubuntu-latest]
         java: [11]
+        testgroups: [
+          app_analytics_test,
+          notebook,
```
Swiddis, line 28: This group doesn't exist

.github/workflows/integration-tests-workflow.yml
```diff
@@ -129,26 +155,15 @@ jobs:
...
 
-      - name: Cypress tests
-        uses: cypress-io/github-action@v2
-        with:
-          working-directory: OpenSearch-Dashboards/plugins/dashboards-observability
-          command: yarn cypress:parallel --browser chrome
-          wait-on: 'http://localhost:5601'
-        env:
-          CYPRESS_CACHE_FOLDER: ${{ matrix.cypress_cache_folder }}
+      - name: Run Cypress tests for test gourp of ${{ matrix.testgroups }}
```
Swiddis, line 163: Typo

mengweieric (approved):
Thanks for the fix.

derek-ho (approved):

.github/workflows/integration-tests-workflow.yml
RyanL1997: good catch! fixed

PR #1301, opened 2023-12-15
Add linter CI

mengweieric:
### Description
Add linter CI to lint only changed files including
- *.ts, 
- *.tsx
- *.js

The linting will be ran only on changes matching above patterns in that PR.

Sample linting run result
https://github.com/mengweieric/dashboards-observability/actions/runs/7215956806/job/19661217812#step:8:1

Sample annotation when linting errors found
https://github.com/mengweieric/dashboards-observability/commit/305287356b68ccd4a1a14873acf1ccd130a8eb9a#diff-2790e5916efcabbaa1ad67efe53928cc72a31ecdf3094cfd0c9582471f91530eL85

### Issues Resolved

### Check List

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -14,7 +14,37 @@ module.exports = {
     '@elastic/eslint-config-kibana',
     'plugin:@elastic/eui/recommended',
     'plugin:react-hooks/recommended',
+    "eslint:recommended",
+    "plugin:cypress/recommended",
+    "plugin:import/recommended",
+    "prettier"
   ],
+  env: {
+    'cypress/globals': true,
+  },
+  plugins: [
+    'cypress',
+  ],
+  rules: {
+    '@osd/eslint/no-restricted-paths': [
+      'error',
+      {
+        basePath: __dirname,
+        zones: [
+          {
+            target: ['(public|server)/**/*'],
+            from: ['../../packages/**/*','packages/**/*'],
+          },
+        ],
+      },
+    ],
+    // Add cypress specific rules here
+    'cypress/no-assigning-return-values': 'error',
+    'cypress/no-unnecessary-waiting': 'error',
+    'cypress/assertion-before-screenshot': 'warn',
+    'cypress/no-force': 'warn',
+    'cypress/no-async-tests': 'error',
+  },
   overrides: [
     {
       files: ['**/*.{js,ts,tsx}'],
@@ -29,4 +59,5 @@ module.exports = {
       },
     },
   ],
+  "ignorePatterns": ["**/*.d.ts"]
 };
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,61 @@
+name: Lint
+
+on:  [push, pull_request]
+
+env:
+  PLUGIN_NAME: dashboards-observability
+  OPENSEARCH_DASHBOARDS_VERSION: 'main'
+
+jobs:
+  build:
+    name: Lint
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout OpenSearch Dashboards
+        uses: actions/checkout@v2
+        with:
+          repository: opensearch-project/Opensearch-Dashboards
+          ref: ${{ env.OPENSEARCH_DASHBOARDS_VERSION }}
+          path: OpenSearch-Dashboards
+
+      - name: Checkout dashboards observability
+        uses: actions/checkout@v2
+        with:
+          path: OpenSearch-Dashboards/plugins/${{ env.PLUGIN_NAME }}         
+
+      - name: Get node and yarn versions
+        working-directory: ${{ env.WORKING_DIR }}
+        id: versions_step
+        run: |
+          echo "::set-output name=node_version::$(cat ./OpenSearch-Dashboards/.nvmrc | cut -d"." -f1)"
+          echo "::set-output name=yarn_version::$(node -p "(require('./OpenSearch-Dashboards/package.json').engines.yarn).match(/[.0-9]+/)[0]")"
+
+      - name: Setup node
+        uses: actions/setup-node@v1
+        with:
+          node-version: ${{ steps.versions_step.outputs.node_version }}
+          registry-url: 'https://registry.npmjs.org'
+
+      - name: Install correct yarn version for OpenSearch Dashboards
+        run: |
+          npm uninstall -g yarn
+          echo "Installing yarn ${{ steps.versions_step.outputs.yarn_version }}"
+          npm i -g yarn@${{ steps.versions_step.outputs.yarn_version }}
+
+      - name: Bootstrap the plugin
+        working-directory: OpenSearch-Dashboards/plugins/${{ env.PLUGIN_NAME }}  
+        run:
+          yarn osd bootstrap
+
+      - name: lint code base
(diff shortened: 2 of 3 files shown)


.github/workflows/lint.yml
```diff
@@ -0,0 +1,62 @@
...
+
+      - name: Install correct yarn version for OpenSearch Dashboards
+        run: |
+          npm uninstall -g yarn
+          echo "Installing yarn ${{ steps.versions_step.outputs.yarn_version }}"
+          npm i -g yarn@${{ steps.versions_step.outputs.yarn_version }}
+
+      - name: Bootstrap the plugin
+        run: |
+          cd OpenSearch-Dashboards/plugins/dashboards-observability
```
RyanL1997, line 49: nit: similar to the L29, you can use `working-directory` here to replace the `cd`

.github/workflows/lint.yml
```diff
@@ -0,0 +1,62 @@
...
+      contents: read
+      packages: read
+      statuses: write
+
+    steps:
+      - name: Checkout OpenSearch Dashboards
+        uses: actions/checkout@v2
+        with:
+          repository: opensearch-project/Opensearch-Dashboards
+          ref: ${{ env.OPENSEARCH_DASHBOARDS_VERSION }}
```
RyanL1997, line 20: The `env` section is missing, and this is not defined.

.github/workflows/lint.yml
```diff
@@ -0,0 +1,62 @@
...
+          npm i -g yarn@${{ steps.versions_step.outputs.yarn_version }}
+
+      - name: Bootstrap the plugin
+        run: |
+          cd OpenSearch-Dashboards/plugins/dashboards-observability
+          yarn osd bootstrap
+
+      - name: lint code base
+        run: |
+          cd OpenSearch-Dashboards/plugins/dashboards-observability
```
RyanL1997, line 54: nit: same as the above.

.github/workflows/lint.yml
```diff
@@ -0,0 +1,62 @@
+name: Lint
+
+on:  [push, pull_request]
+
+jobs:
+  build:
+    name: Lint
+    runs-on: ubuntu-latest
+
+    permissions:
```
RyanL1997, line 10: Just for my knowledge, what are these setups?

RyanL1997 (commented):
Hi @mengweieric, thanks for putting this together. I just left some comments

.github/workflows/lint.yml
RyanL1997: `working-directory: OpenSearch-Dashboards/plugins/${{ env.PLUGIN_NAME }}`

.github/workflows/lint.yml
mengweieric: just granting minimal permissions for workflow but I removed as we don't actually need this much at this stage.

RyanL1997 (approved):
LGTM!

ps48 (approved):

YANG-DB (approved):

.eslintrc.js
```diff
@@ -14,7 +14,37 @@ module.exports = {
...
+    "prettier"
   ],
+  env: {
+    'cypress/globals': true,
+  },
+  plugins: [
+    'cypress',
+  ],
+  rules: {
+    '@osd/eslint/no-restricted-paths': [
```
joshuali925, line 29: what does this do?

.github/workflows/lint.yml
```diff
@@ -0,0 +1,61 @@
...
+      - name: Bootstrap the plugin
+        working-directory: OpenSearch-Dashboards/plugins/${{ env.PLUGIN_NAME }}  
+        run:
+          yarn osd bootstrap
+
+      - name: lint code base
+        working-directory: OpenSearch-Dashboards/plugins/${{ env.PLUGIN_NAME }}
+        run: |
+          git fetch origin main
+          CHANGED_FILES=$(git diff --name-only --diff-filter=ACMRTUXB origin/main | grep -E "\.(js|ts|tsx)$")
```
joshuali925, line 55: why `--diff-filter`? seems unnecessary

.github/workflows/lint.yml
```diff
@@ -0,0 +1,61 @@
...
+          if [ -n "$CHANGED_FILES" ]; then
+            echo "Linting changed files..."
+            yarn lint $CHANGED_FILES
```
joshuali925, line 58: why not combine with the unit test CI?

also probably not an issue but i believe files with spaces won't work

joshuali925 (approved):

.github/workflows/lint.yml
mengweieric: This is for not including files that are deleted, and also with this it explicitly shows what are included and what are not. I actually tried before to run CI without these flags but it was getting errors, and it runs successfully after adding it back. without flags: https://github.com/mengweieric/dashboards-observability/actions/runs/7254224280/job/19762669894. Do you any suggestions it can be written differently without those flags but do the work?

.github/workflows/lint.yml
mengweieric: I don't think it cannot be combined with unit test CI, and saw other plugin has this integrated with their unit tests CI, so it should be fine. The initial considerations are

- In our case since we have tech debt for these linting errors, and no enforced rule so far to ensure anyone who make changes to any file, they are responsible for fixing all existing linting errors in the file(s) they modify. Then we may want to skip failing lintings in sometime, but if it's merged with unit test CI, the following Unit test run will be blocked in those cases then we cannot verify what are results of unit test running.
- other reasons like they are doing different checks, I'm just thinking conceptually we should separate them maybe?
- And if we want to do granularly controls like not runing linting on push but only on pull_request, may take additional effort to do them differently in one CI maybe

.eslintrc.js
mengweieric: this restricts certain import paths based on specified patterns: https://github.com/opensearch-project/OpenSearch-Dashboards/blob/d8cbc170d1b8f07b0c2938ce32fe3ff125911030/packages/osd-eslint-plugin-eslint/rules/no_restricted_paths.js

I borrowed this block from [core](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/.eslintrc.js#L269) and [security plugin](https://github.com/opensearch-project/security-dashboards-plugin/blob/main/.eslintrc.js#L29) just thinking we could directly benefit from it if we have existing issues or having issues in the future

.github/workflows/lint.yml
mengweieric: Yes, spaces should not be an issue I think. Will we have (or allow) file(s) with space in the names?

.github/workflows/lint.yml
joshuali925: if you just want to exclude deleted, it should be `--diff-filter=d`?

.github/workflows/lint.yml
joshuali925: i see.

> spaces should not be an issue I think

i raised up since it might cause a bug and can be fixed using a bash array, but if we don't have space in filenames then i'm ok with it

.github/workflows/lint.yml
mengweieric: The --diff-filter we specify here are actually specifying what we select from git diff if I'm not mistaken: https://git-scm.com/docs/git-diff#Documentation/git-diff.txt---diff-filterACDMRTUXB82308203, therefore here without 'D' we skipped deleted files but select all others.

PR #1358, opened 2024-01-12
Revert "Merge feature/explorer-query-assistant to main"

paulstn:
Reverts opensearch-project/dashboards-observability#1325


ps48 (approved):
Lint is failing due to lint issues in live-tail. This is not an intentional update in the original PR so we should be good here.

ps48:
Only 1 unit test fails on main, the rest of the cypress tests are working/failing as expected. Only major question I have is on: https://github.com/opensearch-project/dashboards-observability/commit/4539a8b92434f24380fba0e828b909f372e84916, as it changes the time range for logs explorer in general. We'll have to fix this as well. 
We can close the revert if the above two things are handled in main and a manual backport PR to 2.x is raised for the previously merged commits.

PR #1360, opened 2024-01-12
Fix test failures from query assistant

paulstn:
Note: There is an unresolved issue where while query assistant is enabled, and a query is generated and run using query assistant, it'll use the 15 minute time constraint instead of the 40 years. However, this issue goes away when clicking on the run button again. Fix pending.

### Description

### Issues Resolved

### Check List


common/constants/shared.ts
```diff
@@ -257,5 +257,6 @@ export const ASYNC_QUERY_SESSION_ID = 'async-query-session-id';
 
 export const DIRECT_DUMMY_QUERY = 'select 1';
 
+export const NON_QUERY_ASSISTANT_START_TIME = 'now-15m';
 export const QUERY_ASSISTANT_FIXED_START_TIME = 'now-40y';
```
joshuali925, line 261: it's hard to tell what the difference is based on the naming. maybe 
```suggestion
export const DEFAULT_START_TIME = 'now-15m';
export const QUERY_ASSIST_START_TIME = 'now-40y';
```

public/components/common/search/search.tsx
```diff
@@ -271,16 +272,16 @@ export const Search = (props: any) => {
 
   useEffect(() => {
     // set index and olly query assistant question if changed elsewhere
-    if (!queryRedux.ollyQueryAssistant) return;
-    if (queryRedux.index.length > 0) {
+    if (!queryRedux?.ollyQueryAssistant) return;
+    if (queryRedux?.index.length > 0) {
```
joshuali925, line 276: i thought there are default values for these?

public/components/event_analytics/redux/slices/query_slice.ts
```diff
@@ -28,7 +30,7 @@ const initialQueryState = {
   [PATTERN_REGEX]: PPL_DEFAULT_PATTERN_REGEX_FILETER,
   [FILTERED_PATTERN]: '',
   [SELECTED_TIMESTAMP]: '',
-  [SELECTED_DATE_RANGE]: ['now-40y', 'now'],
+  [SELECTED_DATE_RANGE]: [coreRefs.queryAssistEnabled ? QUERY_ASSISTANT_FIXED_START_TIME : NON_QUERY_ASSISTANT_START_TIME, 'now'],
```
joshuali925, line 33: `coreRefs` are set in plugin start stage, could this const be initialized before that?

common/constants/shared.ts
paulstn: Sounds good to me, will make the change

public/components/common/search/search.tsx
paulstn: There are default values, I'm changing this back and also modifying the test I aimed to pass originally.

public/components/event_analytics/redux/slices/query_slice.ts
joshuali925: i tried it and got `now-15m` with query assist enabled because coreRefs were initialized after query_slice import

paulstn: Changed back to the explicit now-15m timestamp, leaving the timestamp issues for another pr

Swiddis (approved):

joshuali925 (approved):

PR #1369, opened 2024-01-18
[2.x] Backport query assist

joshuali925:
### Description
backports #1325, #1359, #1360, #1376, #1379

cc @paulstn i picked the 3 commits for query assist to 2.x, if you have other fixes for it please let me know or update this PR

### Issues Resolved
[List any issues this PR will resolve]

### Check List
- [x] New functionality includes testing.
  - [x] All tests pass, including unit test, integration test and doctest
- [ ] New functionality has been documented.
  - [ ] New functionality has javadoc added
  - [ ] New functionality has user manual doc added
- [x] Commits are signed per the DCO using --signoff

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #1378, opened 2024-01-22
improve lint workflow to avoid fast fail

joshuali925:
### Description
make it fail after linting all files

### Issues Resolved
https://github.com/opensearch-project/dashboards-observability/pull/1369#issuecomment-1904937003

### Check List

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -62,18 +62,12 @@ jobs:
 
       - name: Lint Changed Files
         run: |
-          jq -r '.[]' ${HOME}/files_modified.json ${HOME}/files_added.json | sort | uniq > /tmp/changed_files.txt
-          CHANGED_FILES=$(cat /tmp/changed_files.txt)
-          echo "These are the changed files: $CHANGED_FILES"
+          CHANGED_FILES=($(jq -r '.[]' ${HOME}/files_modified.json ${HOME}/files_added.json | grep '.\+.\(js\|ts\|tsx\)$' | sort -u))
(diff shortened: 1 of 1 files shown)


.github/workflows/lint.yml
```diff
@@ -62,18 +62,12 @@ jobs:
 
       - name: Lint Changed Files
         run: |
-          jq -r '.[]' ${HOME}/files_modified.json ${HOME}/files_added.json | sort | uniq > /tmp/changed_files.txt
-          CHANGED_FILES=$(cat /tmp/changed_files.txt)
-          echo "These are the changed files: $CHANGED_FILES"
+          CHANGED_FILES=($(jq -r '.[]' ${HOME}/files_modified.json ${HOME}/files_added.json | grep '.\+.\(js\|ts\|tsx\)$' | sort -u))
```
mengweieric, line 65: I remember one of the issues we were trying to tackle was when a file is renamed, both 'modified' and 'added' shas will show the same renamed file which lead to later lint that file more than once, is it? @RyanL1997 , is this considered in the new change as I saw uniq was removed?

mengweieric (approved):

.github/workflows/lint.yml
joshuali925: prefer `sort -u` over `sort | uniq`. they have the same output

RyanL1997: @mengweieric Yes, that shouldn't be a problem. Since the above step `Check Changes of Files` has filtered the output

RyanL1997 (approved):

Swiddis (approved):

Issue #1363, opened 2024-01-16
[MAINTENANCE] Fix all Integrations lints.


Swiddis:
**Is your feature request related to a problem?**
Since the new Lint action was introduced [by Eric](https://github.com/opensearch-project/dashboards-observability/pull/1301), there are lots of lint issues lying dormant in Integrations. This is a catchall issue for fixing these "zombie lints", as indicated in e.g. [this PR feedback](https://github.com/opensearch-project/dashboards-observability/pull/1288#pullrequestreview-1824613261).

**What solution would you like?**
There are four parts to the integration codebase that can be fixed together or separately. The majority of them should be autofixable, but some will require some manual tweaking:

- [ ] Adaptor Backend `server/adaptors/integrations`
- [ ] Routing `server/routes/integrations`
- [ ] React Frontend `public/components/integrations`
- [ ] Cypress Tests `.cypress/integration/integrations_test`

**What alternatives have you considered?**
N/A

**Do you have any additional context?**
N/A

Swiddis:
Routing lints covered as part of #1381

PR #1381, opened 2024-01-24
Add Index-based adaptor for integrations

### LINKED ISSUE #1380: [FEATURE] Integrations: Index Reader
Integrations currently can only work on file-system-based or in-memory integrations. To have long-term live persisted updates, we need to support reading integrations from an OS index.

The OS index reader can fallback to a json reader if the serialization is the same, though a smarter implementation may be needed to scale further.

N/A

#1253 #1244

Swiddis:
### Description
Index Adaptor is implemented as wrapper around the JSON reader. Solution seems to work fine for the current integrations, but may need to be refactored later for larger counts. The PR also includes refactoring to start using the adaptor as a source for integrations in the TemplateManager, and several smaller lint fixes as part of gradual work on #1363. When merged, users can add integrations by POSTing their serialized forms to `.kibana`. A Zip reader would still be helpful since integration serialization is specific to this code, but once the serialization is ported to another tool it should be straightforward to upload bundles like this. I expect this to be the method used by the Catalog repository to perform automatic uploading, instead of a zip file which is easier for manual uploads.

I ran into issues while testing the functionality with document length, many existing integrations are too large to be serialized directly. It turns out OS has a document length limit of 1 MB. As a short-term workaround, optimizing images and removing sample data is sufficient to get all current integrations under 0.5 MB (image optimization in general would help in a future PR). In the long term we plan to move images and sample data out of the integration data entirely, they'll be handled by the catalog.

This PR also makes integrations objects manageable in the Saved Objects table:
image

### Issues Resolved
Resolves #1380 

### Check List


YANG-DB (approved):

joshuali925 (approved):

server/adaptors/integrations/repository/repository.ts
```diff
@@ -3,44 +3,66 @@
...
     );
     return integrations.filter((x) => x !== null) as IntegrationReader[];
   }
 
-  async getIntegration(integPath: string): Promise<IntegrationReader | null> {
-    if ((await this.reader.getDirectoryType(integPath)) !== 'integration') {
-      console.error(`Requested integration '${integPath}' does not exist`);
+  async getIntegration(integrationName: string): Promise<IntegrationReader | null> {
+    const maybeIntegrations = await Promise.all(
+      this.readers.map((reader) => this.getReaderIntegration(reader, integrationName))
```
mengweieric, line 46: I'm not quite familiar with how the entire index adaptor works, but just wondering if there could have any performance issues as I saw there are recursive file traversals as well as file system operations. How did we perform performance tests and would this be a potential issue to the overall cluster health if the cluster is undergoing high traffic with less computing resource available.

Swiddis: Short answer is that it's not very performant: #1389. Among other things I want to try adding a cache in a future PR, which should fix the bulk of the issues, but I still need to figure out a framework for more thorough profiling to make sure I'm optimizing the right things.

Altogether there shouldn't be more than a couple dozen megabytes of integrations content at any point, so I don't see any issues with adding an in-memory cache.

mengweieric (approved):

Issue #1442, opened 2024-02-15
[FEATURE]Integration to allow generic Core API call


YANG-DB:
**Is your feature request related to a problem?**
In order to take full advantage of all the existing API and functionalities OpenSearch and its plugins provide, we need a generic way for the integration framework to be able to call any Core / Plugin API with the request body given as an integration asset.

**What solution would you like?**
Allowing a generic REST call strategy that will allow adding any type of API  - Core or Plugin to be installed and maintained as part of the integration's assets.

**What alternatives have you considered?**

**Do you have any additional context?**
 - https://github.com/opensearch-project/flow-framework/issues/522
 - https://github.com/opensearch-project/observability/issues/1805
 - https://github.com/opensearch-project/opensearch-catalog/issues/123

PR #1470, opened 2024-02-26
[Feature] Data sources associated objects tab

RyanL1997:
### Description
Add Data sources associated objects tab

### Issues Resolved

### Demo
Without data:
Screenshot 2024-02-26 at 4 18 47 PM

With data:
Screenshot 2024-02-26 at 4 19 40 PM

### Check List

### DIFF
--- a/common/types/data_connections.ts
+++ b/common/types/data_connections.ts
@@ -13,6 +13,15 @@ export interface PermissionsConfigurationProps {
   hasSecurityAccess: boolean;
 }
 
+export interface AssociatedObject {
+  id: string;
+  name: string;
+  database: string;
+  type: string;
+  createdByIntegration: string;
+  accelerations: string;
+}
+
 export type Role = EuiComboBoxOptionOption;
 
 export type DatasourceType = 'S3GLUE' | 'PROMETHEUS';
--- a/opensearch_dashboards.json
+++ b/opensearch_dashboards.json
@@ -1,7 +1,7 @@
 {
   "id": "observabilityDashboards",
-  "version": "3.0.0.0",
-  "opensearchDashboardsVersion": "3.0.0",
+  "version": "2.11.1.0",
+  "opensearchDashboardsVersion": "2.11.1",
   "server": true,
   "ui": true,
   "requiredPlugins": [
--- a/opensearch_dashboards.json
+++ b/opensearch_dashboards.json
@@ -1,7 +1,7 @@
 {
   "id": "observabilityDashboards",
-  "version": "2.11.1.0",
-  "opensearchDashboardsVersion": "2.11.1",
+  "version": "3.0.0.0",
+  "opensearchDashboardsVersion": "3.0.0",
   "server": true,
   "ui": true,
   "requiredPlugins": [
--- a/public/components/datasources/components/manage/associated_objects_tab.tsx
+++ b/public/components/datasources/components/manage/associated_objects_tab.tsx
@@ -14,6 +14,7 @@ import {
   EuiHorizontalRule,
   EuiButton,
   EuiSpacer,
+  EuiEmptyPrompt,
 } from '@elastic/eui';
 import { AssociatedObject } from 'common/types/data_connections';
 import { AccelerationsRecommendationCallout } from './accelerations_recommendation_callout';
@@ -72,6 +73,24 @@ export const AssociatedObjectsTab: React.FC<AssociatedObjectsTabProps> = ({
     );
   };
 
+  const noDataMessage = (
+    <EuiEmptyPrompt
+      title={<h2>You have no associated objects</h2>}
+      body={<p>Add or config tables from your data source or use Query Workbench.</p>}
+      actions={
+        <EuiButton
+          color="primary"
+          fill
+          onClick={() => window.open('https://example.com', '_blank')}
+          iconType="popout"
+          iconSide="left"
+        >
+          Query Workbench
+        </EuiButton>
+      }
+    />
+  );
+
   const columns = [
     {
       field: 'name',
@@ -79,7 +98,7 @@ export const AssociatedObjectsTab: React.FC<AssociatedObjectsTabProps> = ({
       sortable: true,
       'data-test-subj': 'nameCell',
       render: (name: string) => (
-        <EuiLink href="https://oui.opensearch.org/latest/" target="_blank">
+        <EuiLink href="https://example.com" target="_blank">
           {name}
         </EuiLink>
       ),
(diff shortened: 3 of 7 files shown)


opensearch_dashboards.json
```diff
@@ -1,7 +1,7 @@
 {
   "id": "observabilityDashboards",
-  "version": "3.0.0.0",
```
RyanL1997, line 3: this will be removed

public/components/datasources/components/manage/data_connection.tsx
```diff
@@ -165,6 +166,35 @@ export const DataConnection = (props: any) => {
...
     },
+    {
+      id: 'associated_objects',
+      name: 'Associated Objects',
+      disabled: false,
+      content: (
+        <AssociatedObjectsTab
+          associatedObjects={[
+            {
+              id: '1',
```
RyanL1997, line 177: @ps48 do you want me to remove this sample data? Or how are we gonna test this?

public/components/datasources/components/manage/associated_objects_tab.tsx
```diff
@@ -0,0 +1,176 @@
...
+      incremental: true,
+      schema: {
+        fields: { name: { type: 'string' }, database: { type: 'string' } }, // Adjust according to your data's fields
+      },
+    },
+  };
+
+  const pagination = {
+    initialPageSize: 10,
+    pageSizeOptions: [5, 10, 20],
```
sejli, line 147: Nitpick: We can probably increase this to `10, 25, 50` in some other [OSD components](https://github.com/search?q=org%3Aopensearch-project%20pageSizeOptions&type=code)

public/components/datasources/components/manage/associated_objects_tab.tsx
```diff
@@ -0,0 +1,176 @@
...
+      <EuiSpacer />
+      <EuiPanel>
+        <AssociatedObjectsHeader />
+        <EuiHorizontalRule />
+        <AccelerationsRecommendationCallout />
+        <EuiSpacer />
+        <EuiInMemoryTable
+          items={associatedObjects}
+          columns={columns}
+          search={search}
```
sejli, line 168: May need to move this out for an external search bar component since the `EuiInMemoryTable` doesn't have support for search bar **and** filters, but that can be done in a future PR.

public/components/datasources/components/manage/associated_objects_tab.tsx
```diff
@@ -0,0 +1,176 @@
...
+import { AccelerationsRecommendationCallout } from './accelerations_recommendation_callout';
+
+interface AssociatedObject {
+  id: string;
+  name: string;
+  database: string;
+  type: string;
+  createdByIntegration: string;
+  accelerations: string;
+}
```
sejli, line 27: Could be exported out of [`dashboards-observability/common/types/data_connections.ts`](https://github.com/opensearch-project/dashboards-observability/blob/main/common/types/data_connections.ts), since other files could use the same object type.

public/components/datasources/components/manage/data_connection.tsx
ps48: Let's keep the sample data here.

sejli:
In order to fix test and build workflows, `yarn test -u` should be run to update the snapshots. That should fix those workflows.

public/components/datasources/components/manage/associated_objects_tab.tsx
RyanL1997: done

ps48 (commented):
@RyanL1997 For the below lint error let's add a `console.error` to log the error whenever it occurs. 
```
136:15  error    'err' is defined but never used. Allowed unused args must match /^_/u   
```
Thanks! for quick the updates on the PR

ps48 (approved):

public/components/datasources/components/manage/associated_objects_tab.tsx
RyanL1997: lets do this as a follow up

RyanL1997:
@ps48 @sejli , I have addressed the comments above and also fixed that lint ( that was irrelevant from this change btw. ). And I'm updating the screenshots to the description.

common/types/data_connections.ts
```diff
@@ -13,6 +13,15 @@ export interface PermissionsConfigurationProps {
   hasSecurityAccess: boolean;
 }
 
+export interface AssociatedObject {
+  id: string;
+  name: string;
+  database: string;
+  type: string;
+  createdByIntegration: string;
+  accelerations: string;
```
joshuali925, line 22: would any of these fields be a smaller type than `string`, or all of them can be arbitrary strings?

public/components/datasources/components/manage/associated_objects_tab.tsx
```diff
@@ -0,0 +1,197 @@
...
+  return (
+    <>
+      <EuiSpacer />
+      <EuiPanel>
+        <AssociatedObjectsHeader />
+        <EuiHorizontalRule />
+        <AccelerationsRecommendationCallout />
+        <EuiSpacer />
+        {associatedObjects.length > 0 ? (
+          <EuiInMemoryTable
```
joshuali925, line 182: would there be a case that `associatedObjects` gets large and doesn't fit in memory?

joshuali925 (approved):

Swiddis (approved):

common/types/data_connections.ts
```diff
@@ -13,6 +13,15 @@ export interface PermissionsConfigurationProps {
   hasSecurityAccess: boolean;
 }
 
+export interface AssociatedObject {
+  id: string;
+  name: string;
+  database: string;
+  type: string;
+  createdByIntegration: string;
```
Swiddis, line 21: I don't know if `createdByIntegration` is a field that `AssociatedObject` would contain, I'd expect this to be queried from the integrations that store all the contained object information. Do you see us modifying Integrations to add fields like this to arbitrary objects? If so, it might be worth adding this requirement in #1442

RyanL1997: Transferring some of the discussion over here: for now, since we should able to have this field with the source of integration, lets have this field for now for this skeleton PR. We can sync up on this again when we actually do the connection of actual data content.

public/components/datasources/components/manage/associated_objects_tab.tsx
RyanL1997: Good point, I think for the memory fitting issue, we may need to do something like virtualization with react-window to limit the size of the list. However, since this this is a skeleton for now, and lets just leave like this. I will test it when we connect the actual data content, so that we can decide which approaches we need for addressing the memory issue. I will not resolve this comment btw until we have a concrete solution on this.

Issue #1464, opened 2024-02-22
[BUG] Issue with Log Explorer for S3 datasources query failure handling 


ps48:
**What is the bug?**
1. Whenever users run a query on an S3 datasource with an invalid syntax the query status goes in failed state but never stops loading the async loader. Users have to manually click the cancel button to close async loader UI element. 
2. Subsequently, when a valid query is run the loader starts with and "failed" status and then goes into the running state. This gives users the impression that even the valid query has failed. 

**What is the expected behavior?**
1. If query fails then we should show a toast message with the error details and close the async loader UI element.
2. Subsequently, a valid query run shouldn't start with a failed status. 

**What is your host/environment?**
 - OS: Linux
 - Version 2.11
 - Plugins Dashboards observability

**Do you have any screenshots?**
<img width="1792" alt="Screenshot 2024-02-21 at 2 33 10 PM" src="https://github.com/opensearch-project/dashboards-observability/assets/4348487/c9ba5206-1e8a-428e-88c8-e910f99c1b6f">

**Do you have any additional context?**
Add any other context about the problem.

PR #1476, opened 2024-03-01
Flint bug fix explorer failure

paulstn:
### Description
Within Explorer:
- Stopped continuous polling after getting failure.
- Displayed failure in toast message.
- Cleared out previous results and previous polling status upon starting a new search.
- Standardized 'case' for the status corresponding with direct queries.
- Renamed 'sql_query' to 'direct_query' due to the file supporting both sql and ppl

Failure Toast:
image
Clicking on `See full error` to bring up modal:
image

### Issues Resolved
https://github.com/opensearch-project/dashboards-observability/issues/1464

### Check List


common/types/explorer.ts
```diff
@@ -424,11 +424,11 @@ export interface GridSortingColumn {
 }
 
 export enum DirectQueryLoadingStatus {
-  SUCCESS = 'SUCCESS',
-  FAILED = 'FAILED',
-  RUNNING = 'RUNNING',
-  SCHEDULED = 'SCHEDULED',
-  CANCELED = 'CANCELED',
+  SUCCESS = 'success',
```
ps48, line 427: Should we add `waiting` here. Is that a valid status?

public/components/event_analytics/explorer/explorer.tsx
```diff
@@ -977,6 +977,7 @@ export const Explorer = ({
                   endTime={endTime}
                   setStartTime={setStartTime}
                   setEndTime={setEndTime}
+                  notifications={notifications}
```
ps48, line 980: Let's use `coreRefs.notifications`directly in direct_search.tsx instead of passing it through props.

public/components/common/search/direct_search.tsx
```diff
@@ -220,31 +243,32 @@ export const DirectSearch = (props: any) => {
...
+    const status = anyCaseStatus?.toLowerCase();
 
     if (status === DirectQueryLoadingStatus.SUCCESS || datarows) {
-      // stop polling
-      stopPolling();
-      setIsQueryRunning(false);
+      stopPollingWithStatus(status);
+      // update page with data
+      dispatchOnGettingHis(pollingResult, '');
+    } else if (status === DirectQueryLoadingStatus.FAILED) {
```
ps48, line 253: Should we also handle the status canceled here?

common/types/explorer.ts
paulstn: We could add it, but it currently wouldn't be used for anything since the raw status is displayed to the UI.

public/components/common/search/direct_search.tsx
paulstn: Canceled is handled separately (in direct_query_running.tsx), as that's where cancel is triggered, and so the polling is stopped from there directly.

public/components/event_analytics/explorer/explorer.tsx
paulstn: made this change

sumukhswamy (approved):

ps48 (approved):

public/components/common/search/direct_search.tsx
ps48: Ideally we should keep it in a single place if possible. Not a hard requirement in the PR 😄

PR #1479, opened 2024-03-04
Changed Explorer Data Grid useage of timestamp

paulstn:
### Description
In data grid, uses the set default timestamp for all operations such as displaying a time column, properly defining a time range, and sorting by timestamp. 

### Check List


ps48 (approved):

mengweieric:
Do you have an issue link or could you share more context and details of this change?

paulstn:
> Do you have an issue link or could you share more context and details of this change?

@mengweieric There is no issue link, but I can share more details. Previously, there would be issues where the default timestamp would not be used as the time field in the grid (the first column), and the current time (when the data is retrieved) would instead be used. I've modified the code to use the correct default time field. This also solved an issue where sorting might not return any data, this was because the id of the column wasn't set to the same timestamp being used in the query for the time range, I've made sure these values are now the same. Hope this helps, let me know if you need more detail!

public/components/event_analytics/explorer/events_views/data_grid.tsx
```diff
@@ -65,14 +70,18 @@ export function DataGrid(props: DataGridProps) {
...
   });
+
+  const query = useSelector(selectQueries)[tabId];
+  const defaultTimestamp = query[SELECTED_TIMESTAMP];
+
   const selectedColumns =
     explorerFields.selectedFields.length > 0
       ? explorerFields.selectedFields
-      : DEFAULT_EMPTY_EXPLORER_FIELDS;
+      : [{ name: defaultTimestamp, type: 'timestamp' }, ...DEFAULT_EMPTY_EXPLORER_FIELDS];
```
mengweieric, line 80: DEFAULT_EMPTY_EXPLORER_FIELDS has already contained ```{ name: 'timestamp', type: 'timestamp' }``` will this lead to selectedColumns to have:
```
selectedColumns = [
{ name: defaultTimestamp, type: 'timestamp' }
  { name: 'timestamp', type: 'timestamp' },
  { name: '_source', type: 'string' },
]
```
?

public/components/event_analytics/explorer/events_views/data_grid.tsx
```diff
@@ -114,8 +123,12 @@ export function DataGrid(props: DataGridProps) {
   const dataGridColumns = () => {
     const columns: EuiDataGridColumn[] = [];
     selectedColumns.map(({ name }) => {
-      if (name === 'timestamp') {
-        columns.push(DEFAULT_TIMESTAMP_COLUMN);
+      if (name === defaultTimestamp) {
```
mengweieric, line 126: This seems to be a little bit confusing to me. The map is supposed to help you create a new mapped columns just like you defined in the first place, but later you are doing a push in the map.

Please consider either use map or forEach. Seems like map here should be good for your case.

public/components/event_analytics/explorer/events_views/data_grid.tsx
```diff
@@ -65,14 +70,18 @@ export function DataGrid(props: DataGridProps) {
     pplService,
     requestParams,
   });
+
+  const query = useSelector(selectQueries)[tabId];
```
mengweieric, line 74: Seems like all you need related to tabId is for getting redux query state, you can use context. the tabId has been already exposed. tabId is not required as props for data grid I believe.

public/components/event_analytics/explorer/events_views/data_grid.tsx
```diff
@@ -114,8 +123,12 @@ export function DataGrid(props: DataGridProps) {
   const dataGridColumns = () => {
     const columns: EuiDataGridColumn[] = [];
     selectedColumns.map(({ name }) => {
-      if (name === 'timestamp') {
-        columns.push(DEFAULT_TIMESTAMP_COLUMN);
+      if (name === defaultTimestamp) {
+        columns.push({
+          ...DEFAULT_TIMESTAMP_COLUMN,
+          display: `Time (${defaultTimestamp})`,
```
mengweieric, line 129: we need i18n for all these user facing text

public/components/event_analytics/explorer/events_views/data_grid.tsx
```diff
@@ -114,8 +123,12 @@ export function DataGrid(props: DataGridProps) {
   const dataGridColumns = () => {
     const columns: EuiDataGridColumn[] = [];
     selectedColumns.map(({ name }) => {
-      if (name === 'timestamp') {
-        columns.push(DEFAULT_TIMESTAMP_COLUMN);
+      if (name === defaultTimestamp) {
+        columns.push({
+          ...DEFAULT_TIMESTAMP_COLUMN,
```
mengweieric, line 128: I see that you are only overriding display and id, but there's also a schema in DEFAULT_TIMESTAMP_COLUMN as 'datetime', will this be a problem if the defaultTimestamp is not a schema of 'datetime'?

public/components/event_analytics/explorer/__tests__/data_grid.test.tsx
```diff
@@ -45,13 +45,23 @@ describe('Datagrid component', () => {
       Promise.resolve((sampleEmptyPanel as unknown) as HttpResponse)
     );
 
+    const tabId = 'explorer-tab-_fbef9141-48eb-11ee-a60a-af33302cfb3c';
+
     const http = httpClientMock;
     const pplService = new PPLService(httpClientMock);
-    const store = createStore(rootReducer, applyMiddleware(thunk));
+    const preloadedState = {
```
mengweieric, line 52: Could you add tests for default time and user selected timestamp cases?

mengweieric (commented):
Please add tests to cover code changes

public/components/event_analytics/explorer/events_views/data_grid.tsx
paulstn: Actually, I modified DEFAULT_EMPTY_EXPLORER_FIELDS to remove this line: `{ name: 'timestamp', type: 'timestamp' }`, because fundamentally there cannot a default timestamp name, as that is used to map the data to the columns.

public/components/event_analytics/explorer/events_views/data_grid.tsx
paulstn: I am not fully understanding your confusion here, in this method I have selectedColumns which I essentially reformat into an array of `EuiDataGridColumn`, for which I'm pushing into. Could you clarify what's wrong here, is it with the timestamp case specifically?

public/components/event_analytics/explorer/events_views/data_grid.tsx
paulstn: the schema 'datetime' is a value specific to DataGrid, it's the only time related schema there is and assigning it to something that's not time related wouldn't cause anything to break.

public/components/event_analytics/explorer/events_views/data_grid.tsx
paulstn: implemented

public/components/event_analytics/explorer/__tests__/data_grid.test.tsx
paulstn: added that extra test

public/components/event_analytics/explorer/events_views/data_grid.tsx
paulstn: changed

public/components/common/search/direct_search.tsx
```diff
@@ -258,6 +296,16 @@ export const DirectSearch = (props: any) => {
 
     if (status === DirectQueryLoadingStatus.SUCCESS || datarows) {
```
mengweieric, line 296: will this `|| datarows` always be true?

mengweieric (approved):

public/components/common/search/direct_search.tsx
```diff
@@ -250,6 +262,32 @@ export const DirectSearch = (props: any) => {
...
+  const getDirectQueryTimestamp = (schema: Array<{ name: string; type: string }>) => {
+    const timestamp: IDefaultTimestampState = {
+      hasSchemaConflict: false, // schema conflict bool used for OS index w/ different mappings, not needed here
+      default_timestamp: '',
+      message: 'Index does not contain a valid time field.',
+    };
+
+    for (let i = 0; i < schema.length; i++) {
+      const fieldMapping = schema[i];
+      console.log('fieldmapping:', fieldMapping);
```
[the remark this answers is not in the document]
paulstn: removed

public/components/common/search/direct_search.tsx
```diff
@@ -250,6 +262,32 @@ export const DirectSearch = (props: any) => {
       });
   };
 
+  const getDirectQueryTimestamp = (schema: Array<{ name: string; type: string }>) => {
+    const timestamp: IDefaultTimestampState = {
+      hasSchemaConflict: false, // schema conflict bool used for OS index w/ different mappings, not needed here
+      default_timestamp: '',
+      message: 'Index does not contain a valid time field.',
```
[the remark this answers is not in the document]
paulstn: the message field isn't used anywhere currently, I have it this way in case it'll be needed in the future. I'll add it in now though, just in case

public/components/common/search/direct_search.tsx
paulstn: I didn't write this code originally, but I believe it won't be, as the only way that the datarows field will be inside of the polling result is if results are returned successfully.

Issue #1462, opened 2024-02-21
[FEATURE] Multiple installation flows for Integrations.


Swiddis:
**Is your feature request related to a problem?**
As integration setups become more complex, there's a growing need to support multiple installation flows tailored to different use cases. Presently, the setup process is streamlined for a singular installation path, which poses limitations as demands for varied setups arise. This is briefly discussed in #1444, which talks about making the installation process more flexible.

**What solution would you like?**
To address this, I propose enhancing the integration setup process to accommodate multiple installation flows, with a `flows` config option. This can be added to the config as a top-level field showing each flow:

```json5
{
  "name": "example"
  // ...
  "flows": [
    {
      "name": "Queries",
      "description": "Ingestion pipeline for logging use cases with a low querying frequency. Includes sample saved queries.",
      "required": true
    },
    {
      "name": "Dashboards",
      "description": "Optimized for high querying frequency, includes visual dashboards for additional insights."
    }
  ],
  // ...
}
```

These flows will be rendered in a checkbox-list on the frontend, where a user can select any number of flows to install with. When we install assets, each step can have a set of flows it's a part of (default: all flows).

```json5
{
  "name": "example"
  // ...
  "assets": {
    "savedObjects": {
      "name": "example_dashboards",
      "version": "1.0.0",
      "flows": ["Dashboards"] // Dashboards flow: installs only if the dashboards flow is selected
    },
    "queries": [
      {
        "name": "create_table", // No flows: always installs
        "version": "1.0.0",
        "language": "sql"
      },
      {
        "name": "saved_queries",
        "version": "1.0.0",
        "language": "sql",
        "flows": ["Queries"] // Queries flow: installs only if queries selected (equivalent to no flows if queries is required)
      },
      {
        "name": "create_mv",
        "version": "1.0.0",
        "language": "sql",
        "flows": ["Dashboards"]
      }
    ]
  },
  // ...
}
```

This refactoring is equally applicable if #1444 is implemented.

**What alternatives have you considered?**

Separate integrations could be made for each flow. This avoids the need for the feature, but complicates the selection process when a user is onboarding.

**Do you have any additional context?**

N/A

PR #1498, opened 2024-03-07
Update integration format for better handling of multiple asset types

### LINKED ISSUE #1444: [FEATURE] Refactor integration setup process

A critical part of the integration building process is setting up external assets, which define the ways that integrations can read and process data. Currently, assets are stored as [references in the config](https://github.com/opensearch-project/dashboards-observability/blob/2409cf7ddfaa255653a5156d0fc9f3a7d6a9457d/server/adaptors/integrations/__data__/repository/nginx/nginx-1.0.0.json#L41-L58), split by the type of asset (today: saved objects and queries).

As part of the setup process, we go through all of these assets and follow hardcoded steps to install each category of assets. Queries are sent to the asynchronous Flint API, saved objects are sent through a high-level client in-code. This has worked reasonably well so far, as it makes adding more asset types as easy as adding a definition to the `assets` field, and integrating it with the existing setup process.

But with growing feature demand, this approach is proving insufficient. There are issues like #1442 demanding being able to set up assets associated with arbitrary OpenSearch APIs, and in-progress features for supporting multiple installation flows[^1] within one integration that would require configuring which assets are installed. Furthermore, splitting the assets into distinct entities like this complicates [serialization](https://github.com/opensearch-project/dashboards-observability/blob/2409cf7ddfaa255653a5156d0fc9f3a7d6a9457d/server/adaptors/integrations/repository/integration_reader.ts#L453-L489) and [bundling](https://github.com/Swiddis/opensearch-catalog/blob/a50f6990f3deb210949cd08d20dcd49791db9839/cli/src/bundle/bundle.py). A different approach is needed.

Currently, the assets field of integrations looks like this:

```json
"assets": {
  "savedObjects": {
    "name": "dashboards",
    "version": "1.0.0"
  },
  "queries": [
    {
      "name": "create_table",
      "version": "1.0.0",
      "language": "sql"
    },
    {
      "name": "create_mv",
      "version": "1.0.0",
      "language": "sql"
    }
  ]
},
```

I propose updating the format to look like this:

```json
"assets": [
  {
    "name": "create_table",
    "version": "1.0.0",
    "extension": "sql",
    "type": "query"
  },
  {
    "name": "create_mv",
    "version": "1.0.0",
    "extension": "sql",
    "type": "query"
  },
  {
    "name": "dashboards",
    "version": "1.0.0",
    "extension": "ndjson",
    "type": "savedObjectBundle"
  }
],
```

By unifying the categories, serialization switches from handling each case to writing one main handler. This also allows more flexible installation ordering: if there are queries to run after installing a saved object bundle, that's trivially addable. The install logic will remain case-by-case, but it's somewhat unavoidable if different objects require fundamentally different interfaces to install, and it lets us rework the type-by-type logic to something like `for (const asset of assets) { install(asset); }`[^2].

One concern is backwards compatibility: since dynamic catalogs was released with the old serialization, at least the OS integrations plugin should be robust to both versions. This can be added on-read by having a validation check which automatically converts the old format to the new format, but there are still some details with how to make migrations work with e.g. the `.kibana` type mapping.

We can continue adding case-by-case steps, which has the drawbacks mentioned above, and is better for backwards compatibility. To implement ordering we could add a field to each type like `order`, where steps run sorted by `order` (default: 0, and support negative values). I worry this will continue to get worse as we add more asset types, but is minimal work upfront.

We could also provide some sort of helper method which takes the current format and converts it to this new format. That method can be relatively robust, and simplify a lot of operations on the asset list without requiring a change to the serialization format. 

Also related: workflow integrations with https://github.com/opensearch-project/observability/issues/1805. Additionally, the already-cited request for arbitrary API calls #1442.

[^1]: To support distinct installation paths, a further `flows?: string[]` field can be introduced (default: all flows), which specifies when a step should be run. If dashboards in the above assets are only installed for a full installation, a user can add the entry `"flows": ["full"]`. There may need to be more work to determine how to fully support installation flows, I'll leave that for another issue.
[^2]: A sophisticated implementation might include dependency logic and run steps in parallel, but this is unlikely to be necessary for the amount of effort it would add.

Swiddis:
### Description
To support multiple installation paths for integrations better, part of that is refactoring the current install flow to make the different types of assets more uniform. This PR:
- Updates the integration type
- Updates frontend/backend usage of integration assets for new type
- Fixes several lints around the frontend in preparation for #1462 
- Adds a migration path for any installed dynamic integrations

Next steps are refactoring the installation flow to support multiple install paths, using these extensability updates.

Notice: I noticed while writing this that we don't have any unit/integration tests that ensure that all types of assets are being correctly installed. This will be included as part of the installation workflows PR, the current PR is already fairly big as-is. Followed the boy scout rule for some other misc. changes like removing dead code and fixing misused types.

### Issues Resolved
Closes #1444 

### Check List


server/adaptors/integrations/types.ts
```diff
@@ -50,38 +52,19 @@ interface SerializedIntegrationStatics {
...
-  queries?: Array<{
-    name: string;
-    version: string;
-    language: string;
-  }>;
+interface IntegrationAsset {
+  name: string;
+  version: string;
+  extension: string;
+  type: SupportedAssetType;
```
Swiddis, line 59: This is the core change, most of the refactoring is following the consequences of this type update

server/plugin.ts
```diff
@@ -190,6 +191,9 @@ export class ObservabilityPlugin
           },
         },
       },
+      migrations: {
+        '3.0.0': migrateV1IntegrationToV2Integration,
```
Swiddis, line 195: Not 100% sure I've set up migration correctly, messed with it a bit to get the type checking to work out but I'm not sure when this migration triggers.

YANG-DB: Hi
How would migration be handled an by whom ? 
For example if I have a migration script - where can I specify it ?

Swiddis: Migrations run on OSD startup, it's handled in OSD and will apply to all of the installed templates that are on a lower version. But I'm not sure how OSD detects the version.

For content, I migrated all of the prebuilt content, but I'll need to update the bundle script for the new version.

YANG-DB (approved):

ps48 (approved):

Issue #1251, opened 2023-11-15
[BUG] UI sessions should be coupled with flint datasources


ps48:
**What is the bug?**
Currently when flint sessions are enabled on a domain. UI uses the same session from backend for different datasources. This is a requirement from the SQL plugin to handle multiple sessions on UI.

More context: https://github.com/opensearch-project/dashboards-query-workbench/issues/216

PR #1500, opened 2024-03-08
Catalog cache and Session update for async queries

### LINKED ISSUE #1484: Glue DataCatalog Cache (GDC) Design Document
## Overview

OpenSearch S3 Zero ETL (Flint) feature provides a unique solution to allow users run direct queries with/w/o acceleration with their data stored in S3. This possesses a unique problem where UI needs to display the catalog including datasources, databases and tables to enable the user to run any query. The fetch the catalog UI needs to run a sequence of show databases, tables and accelerations queries which take ~10-20s to respond each. This adds up a lot of  UI wait time for the user, eventually losing interest in the solution. Glue DataCatalog Cache (GDC) aims to solve this issue by running the show databases, tables and accelerations in the background and storing them in cache so UI retrieval is faster and users can focus on working with the overall solution.

## Motivation

With Glue DataCatalog Cache we aim to solve the below items:

1. Reduce user wait time on UI to see catalog data
2. Reduce no. of queries sent to the backend to fetch catalog information
3. Give users choice to hard reload the cached data

### OpenSearch Dashboards elements using GDC:

The Query Workbench and Datasources dashboards plugins will be the first UI elements to start using the Glue DataCatalog cache in OS version 2.13. Later, In OS 2.14 we’ll have Data Explorer/Discover Cohesion project will utilize the same cache object to load tables and accelerations. 

#### Workbench 2.13
Image (6)
Image (7)

#### DataSources 2.13
Image (8)

## Requirements

1. Save a cached view of the Glue data catalog
2. Persist cache across browser reloads
3. Restrict cache access to the users who have access to the datasource
4. Allow users to invoke a hard reload on all cached data
5. Allow users to invoke a hard reload on partial cached data (only accelerations, only tables and databases)

## Architecture Design

Today, with every page load of workbench plugin for a datasource we auto-run the queries to load databases. Thereafter, users need to click on loaded databases to fetch tables for the particular database and further click on the tables to loaded the acceleration indices. The would have worked well for the UI if the catalog load was instantaneous instead of ~10-20s per load. 

In the Glue DataCatalog Cache, we propose to load top 4000 elements including databases, tables and accelerations (10k elements is the OSD api limit, but UI can become sluggish well before 10k). Not limiting to a certain number of elements may lead to infinite loading time. 

#### Approach 1:

```
+---------------------------+
| 1. Show Databases         |
+---------------------------+
       |
       V
+---------------------------+
| 2. For each Database      |
|    Show Tables            |
|    Show Materialized Views|
+---------------------------+
       |
       V
+---------------------------+
| 3. For each Table         |
|    Show Indexes           |
|    (Skip Index & Covering |
|     Index Information)    |
+---------------------------+
```

1. Show Databases: Initially, we query the Glue Data Catalog with `SHOW DATABASES` query to list all available databases.
2. For Each Database, Show Tables and Materialized Views: For each database obtained in the previous step, we retrieve a list of tables and materialized views present within that database using `SHOW TABLES` and `SHOW MATERIALIZED VIEWS`.
3. For Each Table, run show Indexes: we run the queries `SHOW SKIPPING INDEX` and `SHOW COVERING INDEX` to retrieve information about its skipping and covering indexes for the particular table

#### Approach 2 (Preferred): 

```
+---------------------------+
| 1. Show Databases         |
+---------------------------+
       |
       V
+---------------------------+
| 2. For each Database      |
|    Show Tables            |
+---------------------------+
       |
       V
+---------------------------+
| 3. Show Accelerations     |
|    for All Tables         |
+---------------------------+
```

1. Show Databases: Again, we start by querying the Glue Data Catalog to list all available databases.
2. For Each Database: Show Tables: For each database obtained in the previous step, we retrieve a list of tables present within that database.
3. Show Accelerations for All Tables: After obtaining the list of tables from all databases, we directly run `SHOW ACCELERATIONS`  query to retrieve information about all accelerations associated with all tables and databases. This step consolidates all tables from all databases to fetch acceleration details in a single operation, rather than querying each table individually.

### Cache Data Structure

```
{
  "version": "1.0",
  "dataSources": [
    {
      "name": "DataSource1",
      "lastUpdated": "2024-02-20T12:00:00Z", // Saved in UTC
      "status": "Updated", // Failed, Empty, Loading
      "databases": [
        {
          "name": "Database1",
          "materializedViews": [
            {
              "name": "MaterializedView1"
            },
            {
              "name": "MaterializedView2"
            }
          ],
          "tables": [
            {
              "name": "Table1",
              "columns": [
                {"name": "column1", "dataType": "datatype1"},
                {"name": "column2", "dataType": "datatype2"},
                {"name": "column3", "dataType": "datatype3"}
              ],
              "skippingIndex": {
                "indexName": "SkippingIndex1"
              },
              "coveringIndices": [
                {
                  "indexName": "CoveringIndex1"
                },
                {
                  "indexName": "CoveringIndex2"
                }
              ]
            },
            {
              "name": "Table2",
              "columns": [
                {"name": "column4", "dataType": "datatype4"},
                {"name": "column5", "dataType": "datatype5"}
              ],
              "skippingIndex": {
                "indexName": "SkippingIndex2"
              },
              "coveringIndices": []
            }
          ]
        }
      ]
    }
  ]
}
```

This JSON structure provides a cached representation of the data catalog, including databases, tables, materialized views, and associated index information.

* **LastUpdated:** Indicates the timestamp when the cache of the data catalog was last updated. In this case, it was last updated on February 20, 2024, at 12:00:00 (UTC).
* **Version:** Specifies the version of the cache. This helps in tracking changes and updates to the cache format. The current version is 1.0.
* **Status:** Indicates the status of the cache. In this case, it's marked as "Updated", suggesting that the cache is currently in use and up-to-date.
* **DataSources:** Represents different data sources available in the data catalog.
    * Name: The name of the data source.
    *  **Databases:** Represents databases within each data source.
        * Name: The name of the database.
        * **MaterializedViews:** Represents materialized views within the database. These are precomputed views stored on disk.
        * Name: The name of the materialized view.
        * **Tables:** Represents tables within the database.
        * Name: The name of the table.
            * **SkippingIndex:** Represents the skipping index associated with the table. A skipping index is an index that may not cover all the columns but can still be useful for direct queries.
                *  IndexName: The name of the skipping index.
            * **CoveringIndices:** Represents covering indices associated with the table. A covering index is an index that includes all the columns required for a query, allowing the query to be answered directly from the index without accessing the table.
                * IndexName: The name of the covering index.

#### Approach 1: Kibana Index 

In this approach, we opt to store the cache in the `.kibana` index, leveraging Kibana's saved objects feature. Since the Glue DataCatalog exclusively contains metadata without user-specific data, it is suitable for storage within the Kibana index of the respective user. This approach ensures that the cache is stored per user, aligning with the lifecycle of the user's interactions. To facilitate storing and retrieving the cached information, plugins utilize the OSD SavedObjects Client.

#### Approach 2: Index Managed by the SQL plugin

**Option 2.1:** Re-use the `.ql-datasources` index. This index is currently managed for storing Glue datasources and associated metadata, such as datasource type, connector type, and authentication type. However, adopting this option would necessitate significant refactoring since adding any information to this index is presently restricted to super admins of OpenSearch. Additionally, utilizing this index for storing cache information related to the catalog may not align well with the Separation of Responsibility principle, as it could introduce complexities and potential conflicts with the primary purpose of the index.

**Option 2.2:** Create a new flint index `.cache-datasources` managed by the sql plugin. This requires the SQL plugin to manage one index per datasource. This approach provides more flexibility and autonomy in managing cache-related data independently from other indices and metadata.

#### Approach 3: Browser storage

**Option 3.1 Local Storage: (preferred)** In this approach, the cache is stored directly within the browser's local storage. This option offers simplicity and speed since data retrieval occurs locally without the need for server communication. However, browser storage may have limitations in terms of storage capacity and data persistence, particularly across different devices or sessions.

**Option 3.2 Session storage**: Similar to browser storage, session storage involves storing the cache within the browser's session storage. This option provides a temporary storage solution tied to the current browser session, ensuring data persistence only for the duration of the session. Session storage offers enhanced security and privacy since data is cleared upon session termination but may not be suitable for long-term storage requirements.

### Other Design Considerations

SQL plugin handles the management of cache by running the Show databases, tables and accelerations queries internally and providing APIs to the OSD plugin for consuming the cached information. This approach may pose an issue in the longer term when we head towards creating a Managed Flint Service. 

### Workflows

Image (9)

* **Initialization and Cache Refresh:**
    * The User loads the UI for the first time or requests a cache refresh.
    * The OpenSearchDashboards component activates and checks or updates the cache status.
    * OpenSearchDashboards sends a query to OpenSearchSQL to show databases.
    * OpenSearchSQL forwards the query request to Spark, which accesses the metadata store (AWSGlue).
    * OpenSearchSQL receives the response from Spark and sends it back to OpenSearchDashboards.
    * OpenSearchDashboards then proceeds to show tables in multiple databases, similar to the initial steps of showing databases.
    * Lastly, OpenSearchDashboards queries for accelerations using OpenSearchSQL, and the process repeats with Spark accessing the metadata store.
* **Subsequent UI Loads:**
    * After the initial load or cache refresh, subsequent UI loads involve a simpler process.
    * The User loads the UI again, and OpenSearchDashboards checks the cache status.
    * OpenSearchDashboards then proceeds to show results to the User.

## Security

Today, the access controls of datasources if by default restricted to the admin of the OpenSearch domain. To add more users we need to edit access permissions of the datasource and add security roles to it. Users with the security role will get access to the datasource. While, we keep GDC in one of the storage options we should adhere to the access controls maintained over the datasources.  

## Appendix

1. SQL Async queries Interface: https://github.com/opensearch-project/sql/blob/main/docs/user/interfaces/asyncqueryinterface.rst
1. OpenSearch Spark Accelerations: https://github.com/opensearch-project/opensearch-spark/blob/main/docs/index.md
1. OpenSearch Spark Design: https://github.com/opensearch-project/opensearch-spark/issues/4
1. AWS OpenSearch Direct Query Setup: https://docs.aws.amazon.com/opensearch-service/latest/developerguide/direct-query-s3.html

ps48:
### Description
Catalog cache and Session update for async queries

### Issues Resolved
#1484 
#1251 

### Check List

### DIFF
--- a/public/framework/catalog_cache/cache_loader.test.tsx
+++ b/public/framework/catalog_cache/cache_loader.test.tsx
@@ -5,14 +5,16 @@
 
 import { CachedDataSourceStatus } from '../../../common/types/data_connections';
 import {
+  mockShowDatabasesPollingResult,
+  mockShowIndexesPollingResult,
+  mockShowTablesPollingResult,
+} from '../../../test/datasources';
+import {
   updateAccelerationsToCache,
   updateDatabasesToCache,
   updateTablesToCache,
 } from './cache_loader';
 import { CatalogCacheManager } from './cache_manager';
-
-// Mock CatalogCacheManager
-// jest.mock('./cache_manager');
 
 interface LooseObject {
   [key: string]: any;
@@ -38,23 +40,6 @@ const localStorageMock = (() => {
 })();
 
 Object.defineProperty(window, 'localStorage', { value: localStorageMock });
-
-// // Mock the behavior of CatalogCacheManager
-// const mockAddOrUpdateDataSource = jest.fn();
-// const mockGetOrCreateDataSource = jest.fn().mockImplementation((dataSourceName: string) => ({
-//   name: dataSourceName,
-//   databases: [],
-//   lastUpdated: '', // or use an actual date if needed
-//   status: CachedDataSourceStatus.Empty,
-// }));
-
-// // Mock the methods used by updateDatabasesToCache
-// jest.mock('./cache_manager', () => ({
-//   CatalogCacheManager: {
-//     addOrUpdateDataSource: mockAddOrUpdateDataSource,
-//     getOrCreateDataSource: mockGetOrCreateDataSource,
-//   },
-// }));
 
 describe('loadCacheTests', () => {
   beforeEach(() => {
@@ -88,12 +73,7 @@ describe('loadCacheTests', () => {
 
     it('should update cache with new databases when polling result is not null', () => {
       const dataSourceName = 'TestDataSource';
-      const pollingResult = {
-        schema: [{ name: 'namespace', type: 'string' }],
-        datarows: [['Database1'], ['Database2']],
-      };
-
-      updateDatabasesToCache(dataSourceName, pollingResult);
+      updateDatabasesToCache(dataSourceName, mockShowDatabasesPollingResult);
 
       // Verify that addOrUpdateDataSource is called with the correct parameters
       expect(CatalogCacheManager.addOrUpdateDataSource).toHaveBeenCalledWith({
@@ -144,17 +124,6 @@ describe('loadCacheTests', () => {
     it('should update cache with new tables when polling result is not null', () => {
       const dataSourceName = 'TestDataSource';
       const databaseName = 'TestDatabase';
-      const pollingResult = {
-        schema: [
-          { name: 'namespace', type: 'string' },
-          { name: 'tableName', type: 'string' },
-          { name: 'isTemporary', type: 'boolean' },
-        ],
-        datarows: [
-          ['TestDatabase', 'Table1', false],
-          ['TestDatabase', 'Table2', false],
-        ],
-      };
 
       CatalogCacheManager.addOrUpdateDataSource({
         databases: [
@@ -169,7 +138,7 @@ describe('loadCacheTests', () => {
         lastUpdated: new Date().toUTCString(),
         status: CachedDataSourceStatus.Updated,
       });
-      updateTablesToCache(dataSourceName, databaseName, pollingResult);
+      updateTablesToCache(dataSourceName, databaseName, mockShowTablesPollingResult);
 
       // Verify that updateDatabase is called with the correct parameters
       expect(CatalogCacheManager.updateDatabase).toHaveBeenCalledWith(
@@ -208,51 +177,36 @@ describe('loadCacheTests', () => {
     });
 
     it('should save new accelerations cache when polling result is not null', () => {
-      const pollingResult = {
-        schema: [
-          {
-            flint_index_name: 'Index1',
-            kind: 'mv',
-            database: 'DB1',
-            table: 'Table1',
-            index_name: 'Index1',
-            auto_refresh: false,
-            status: 'Active',
-          },
-          {
-            flint_index_name: 'Index2',
-            kind: 'skipping',
-            database: 'DB2',
-            table: 'Table2',
-            index_name: 'Index2',
-            auto_refresh: true,
-            status: 'Active',
-          },
-        ],
-        datarows: [],
-      };
-
-      updateAccelerationsToCache(pollingResult);
+      updateAccelerationsToCache(mockShowIndexesPollingResult);
 
       // Verify that saveAccelerationsCache is called with the correct parameters
       expect(CatalogCacheManager.saveAccelerationsCache).toHaveBeenCalledWith({
         version: '1.0',
         accelerations: [
           {
-            flintIndexName: 'Index1',
-            type: 'materialized',
-            database: 'DB1',
-            table: 'Table1',
-            indexName: 'Index1',
+            flintIndexName: 'flint_mys3_default_http_logs_skipping_index',
+            type: 'skipping',
+            database: 'default',
+            table: 'http_logs',
+            indexName: 'skipping_index',
             autoRefresh: false,
             status: 'Active',
           },
           {
-            flintIndexName: 'Index2',
-            type: 'skipping',
-            database: 'DB2',
-            table: 'Table2',
-            indexName: 'Index2',
+            flintIndexName: 'flint_mys3_default_http_logs_status_clientip_and_day_index',
+            type: 'covering',
+            database: 'default',
+            table: 'http_logs',
+            indexName: 'status_clientip_and_day',
+            autoRefresh: true,
+            status: 'Active',
+          },
+          {
+            flintIndexName: 'flint_mys3_default_http_count_view',
+            type: 'materialized',
+            database: 'default',
+            table: '',
+            indexName: 'http_count_view',
             autoRefresh: true,
             status: 'Active',
           },
--- a/public/framework/catalog_cache/cache_loader.test.tsx
+++ b/public/framework/catalog_cache/cache_loader.test.tsx
@@ -3,6 +3,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+import { CATALOG_CACHE_VERSION } from '../../../common/constants/data_sources';
 import { CachedDataSourceStatus } from '../../../common/types/data_connections';
 import {
   mockShowDatabasesPollingResult,
@@ -169,7 +170,7 @@ describe('loadCacheTests', () => {
 
       // Verify that saveAccelerationsCache is called with the correct parameters
       expect(CatalogCacheManager.saveAccelerationsCache).toHaveBeenCalledWith({
-        version: '1.0',
+        version: CATALOG_CACHE_VERSION,
         accelerations: [],
         lastUpdated: expect.any(String),
         status: CachedDataSourceStatus.Failed,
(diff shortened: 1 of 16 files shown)


public/framework/catalog_cache/cache_manager.ts
```diff
@@ -0,0 +1,204 @@
...
+    const cachedDataSource = this.getDataSourceCache().dataSources.find(
+      (ds) => ds.name === dataSourceName
+    );
+    if (!cachedDataSource) {
+      throw new Error('DataSource not found exception: ' + dataSourceName);
+    }
+
+    const cachedDatabase = cachedDataSource.databases.find((db) => db.name === databaseName);
+    if (!cachedDatabase) {
+      throw new Error('Database not found exception: ' + databaseName);
```
Swiddis, line 144: From an API usability standpoint, I wonder if it's good design to throw an error on a cache miss. Since it's a pretty regular and probably expected occurrence, it might make more sense to return a `CachedDatabase | null` (or reuse the `ok: boolean` pattern from `AsyncApiResponse`) so TypeScript can do more heavy lifting for soundness checking.

test/datasources.ts
```diff
@@ -806,3 +806,53 @@ export const mockRoleData = {
     },
   },
 };
+
+export const mockShowDatabasesPollingResult = {
```
Swiddis, line 810: Might be worthwhile to type hint these with the intended types in case the type changes. Coming from [my recent PR](https://github.com/opensearch-project/dashboards-observability/pull/1498) where I paid the cost of not doing that sooner, and had to manually find/fix all the broken mocks.

public/framework/catalog_cache/cache_loader.tsx
```diff
@@ -0,0 +1,268 @@
...
+    DirectQueryLoadingStatus.SCHEDULED
+  );
+
+  const {
+    data: pollingResult,
+    loading: _pollingLoading,
+    error: pollingError,
+    startPolling,
+    stopPolling: stopLoading,
+  } = usePolling<any, any>((params) => {
```
Swiddis, line 177: I know linting complains a lot about `any`: it is ok here? Better than `unknown`?

public/framework/catalog_cache/cache_loader.test.tsx
```diff
@@ -0,0 +1,219 @@
...
+        status: CachedDataSourceStatus.Failed,
+      });
+    });
+
+    it('should save new accelerations cache when polling result is not null', () => {
+      updateAccelerationsToCache(mockShowIndexesPollingResult);
+
+      // Verify that saveAccelerationsCache is called with the correct parameters
+      expect(CatalogCacheManager.saveAccelerationsCache).toHaveBeenCalledWith({
+        version: '1.0',
```
Swiddis, line 184: Should this be a constant as well?

public/framework/catalog_cache/cache_manager.ts
ps48: This is by design right now. We want the cache users to know where the get request is failing. I'll check with @RyanL1997 @sejli what do they expect `thrown error or an error response`. I remember UI having special cases where they see something missing with toast errors or callouts.

public/framework/catalog_cache/cache_loader.test.tsx
ps48: Good catch 👍🏽

public/framework/catalog_cache/cache_loader.tsx
ps48: I tried `unknown` earlier but that gives a TS error. `any` works right now cause the usePolling hook uses `any`  internally.

test/datasources.ts
ps48: This is good point, updated the mocks!

public/framework/catalog_cache/cache_loader.tsx
```diff
@@ -0,0 +1,278 @@
...
+      });
+  };
+
+  useEffect(() => {
+    // cancel direct query
+    if (!pollingResult) return;
+    const { status: anyCaseStatus, datarows, error } = pollingResult;
+    const status = anyCaseStatus?.toLowerCase();
+
+    if (status === DirectQueryLoadingStatus.SUCCESS || datarows) {
```
paulstn, line 241: I see you're reusing some of the polling logic here, I'm wondering if its significantly better to abstract the actual polling+handling of responses such that both the cache loader and the direct query search don't have to deal with polling-specific details. If need be, we can open an issue and I can do this refactoring.

Swiddis (approved):

public/framework/catalog_cache/cache_loader.tsx
Swiddis: Maybe there should be a more general wrapper for polling? I also have written some manual polling logic elsewhere.

ps48: I totally agree, I'm planning to create hook for this in my next PR for create acceleration flyout (will surely re-use it here and refactor).

ps48: Created an issue here: https://github.com/opensearch-project/dashboards-observability/issues/1503

paulstn (approved):

PR #1501, opened 2024-03-08
Update integrations to allow custom checkpoint locations

### LINKED ISSUE #1468: [BUG] Integrations MV checkpoint is hardcoded to improper location
**What is the bug?**
When the integrations add a checkpoint directory, they do it under the S3 bucket location given. This leads to issues when Spark tries to read the checkpoint data that it falsely assumes is unindexed data to be read.

1. Create an S3 integration
2. The checkpoint is under the S3 bucket location

The user should be able to specify the checkpoint location, and it shouldn't be under the data bucket.

 - N/A

N/A

[Example checkpoint initialization](https://github.com/opensearch-project/dashboards-observability/blob/403eeceeecab62bdea08aed69d2fccec029a75f8/server/adaptors/integrations/__data__/repository/aws_elb/assets/create_mv-1.0.0.sql#L44).

Swiddis:
### Description
Per #1468, integrations has an improperly hardcoded checkpoint location. This PR adds the ability for users to input their own location.

Before:
image

After:
image

### Issues Resolved
Closes #1468 

### Check List


public/components/integrations/components/setup_integration.tsx
```diff
@@ -301,16 +307,35 @@ export function SetupIntegrationForm({
...
-            isInvalid={isBlurred && !config.connectionLocation.startsWith('s3://')}
+            isInvalid={isBucketBlurred && !config.connectionLocation.startsWith('s3://')}
             error={["Must be a URL starting with 's3://'."]}
           >
             <EuiFieldText
               value={config.connectionLocation}
               onChange={(event) => updateConfig({ connectionLocation: event.target.value })}
               placeholder="s3://"
-              isInvalid={isBlurred && !config.connectionLocation.startsWith('s3://')}
+              isInvalid={isBucketBlurred && !config.connectionLocation.startsWith('s3://')}
```
ps48, line 317: do we need to do a Regex check for location and bucket checkpoint instead of a prefix check? 
https://github.com/opensearch-project/dashboards-observability/blob/cea2e854a090792f01d81301d59920a13a569650/common/constants/data_sources.ts#L39

ps48 (approved):

public/components/integrations/components/setup_integration.tsx
Swiddis: Might be overkill? I think UI-wise we already communicate that we need a URL and the check is enough that a user would have to be trying to break it. Not sure if we should try and play the "Write a URL validator" game ourselves unless we have a library specifically for it.

seankao-az (approved):

Issue #1539, opened 2024-03-15
[FEATURE] Add i18n support for create acceleration flyout


ps48:
**Is your feature request related to a problem?**
Add i18n support for acceleration flyout 

**What solution would you like?**
Currently the user facing text is standardized with i18n in create acceleration flyout. We should add i18n to our standards for easy translation to other languages.

ps48:
This has been moved to OSD core

Issue #1534, opened 2024-03-14
[Enhancement] Add the types for datasources acceleration related classes


RyanL1997:
### Enhancement for the datasource acceleration related classes
**1. Add the types for datasources acceleration related classes:**
- settings
- maapings
- indexInfo
- acceleration
The above are the constant that requires the type. 

2. Handle the edge case that the incoming data from the above endpoint contains invalid value (`null`, `undefine`):
- do the check
- have a default template for UI rendering

### Reference:
https://github.com/opensearch-project/dashboards-observability/pull/1521

PR #1540, opened 2024-03-15
[Feature] Acceleration Actions Implementation

RyanL1997:
### Description
Implement the actions for acceleration:
- `Delete`: drop the index
- `Vacuum`: delete the index
- `Sync`: refresh a single acceleration's content

Enhancement:
- Localize the time for refresh timestamp
- Remove the sql definition tab and class
- Fix the refresh interval content when there is `null`
- Fix types issues
- Implement the reset of the acceleration flyout page
- Added the data source Eui link for navigating to data source

### Issues Resolved
* Relate #1534 
* Follow-up #1539 

### Check List


RyanL1997:
### DEMO VIDEO V1 OF MAR/15

https://github.com/opensearch-project/dashboards-observability/assets/109499885/ee7f0438-a820-499e-98e2-631bcd5437d0

RyanL1997:
### DEMO VIDEO V1 OF MAR/15 11:49PM PST

https://github.com/opensearch-project/dashboards-observability/assets/109499885/1e825c72-f9bb-4f65-b70f-5caab2cac39a

RyanL1997:
### DEMO VIDEO V3 OF MAR/16 04:52PM PST

https://github.com/opensearch-project/dashboards-observability/assets/109499885/fd7ddba5-bf1c-42b9-a088-bbbf2442b164

public/components/datasources/components/manage/accelerations/utils/acceleration_utils.tsx
```diff
@@ -80,12 +123,12 @@ export const getRefreshButtonIcon = () => {
   return 'inputOutput';
 };
 
-export const onRefreshButtonClick = (acceleration: any) => {
+export const onRefreshIconClick = (acceleration: CachedAcceleration) => {
   // TODO: send request to refresh
-  console.log('refreshing', acceleration.name);
+  console.log('refreshing', acceleration.indexName);
 };
```
sejli, line 129: Is there still implementation work to be done on this?

public/components/datasources/components/manage/accelerations/acceleration_table.tsx
```diff
@@ -95,13 +134,13 @@ export const AccelerationTable = ({
...
 
-  const handleRefresh = () => {
+  const handleRefresh = useCallback(() => {
     console.log('Initiating refresh...');
     if (!isCatalogCacheFetching(accelerationsLoadStatus)) {
       setIsRefreshing(true);
       startLoadingAccelerations(dataSourceName);
     }
-  };
+  }, [startLoadingAccelerations, isCatalogCacheFetching, setIsRefreshing]);
```
sejli, line 143: Do we need all of these dependencies here? Would `accelerationsLoadStatus` be the only dependency, since it's the only thing that will change?

public/components/datasources/components/manage/accelerations/acceleration_operation.tsx
```diff
@@ -0,0 +1,102 @@
...
+      setOperationSuccess(true);
+      setToast(operationSuccessMessage, 'success');
+    } else if (loadStatus === DirectQueryLoadingStatus.FAILED) {
+      setIsOperating(false);
+      setOperationSuccess(false);
+      setToast(operationFailureMessage, 'danger');
+    }
+
+    setCurrentStatus(loadStatus);
+  }, [loadStatus, setToast, accelerationToOperate, dataSource, operationType, currentStatus]);
```
sejli, line 74: Do we need this state? Feel like we could just have this `useEffect()` depend on the `loadStatus` from the `useDirectQuery()` hook. Our logic for these types of loading operations are usually something like:
1. `startLoading()`, set some state that indicates that the component is loading
2. Check if the component is done loading, i.e., `DirectQueryLoadingStatus.SUCCESS`. If so, perform success operations.
3. Check if the component failed, i.e., `DirectQueryLoadingStatus.FAILED`. If so, perform failure operations.

A very simple example of this can be found [here](https://github.com/opensearch-project/dashboards-observability/blob/main/public/components/datasources/components/manage/accelerations/create_accelerations_flyout/selectors/selector_helpers/load_databases.tsx#L32-L43). I think that could reduce the complexity, since checking we could perform the operations for `loadStatus === DirectQueryLoadingStatus.SCHEDULED` once we start loading, since it's guaranteed to happen.

Also, we probably don't need this many dependencies in the array.

public/components/datasources/components/manage/accelerations/utils/acceleration_utils.tsx
RyanL1997: I will remove this. Good catch

RyanL1997: removed.

public/components/datasources/components/manage/accelerations/acceleration_operation.tsx
RyanL1997: I was following the the example you mentioned. However, since our case required us to handle the toast pop up by using all these dependencies, so it became more complicated. For example, first, we need the `dataSource` and and `accelerationToOperate` to setup the displayed index name on the toast msg correctly. After that we also need the `operationType` for toast msgs. In addition, since we are bonding our toast msg with these status, and once the `use_ polling` class start fetching status by `startLoading`, it will keep fetching until it is `SUCCESS` or `FAILED` which means the logic here:
``` javascript
    if (loadStatus === DirectQueryLoadingStatus.SCHEDULED) {
      setIsOperating(true);
      setToast(operationInProgressMessage, 'success');
```
The above toast will keep being showed since the `DirectQueryLoadingStatus.SCHEDULED` will be there for a while. So the check of the `currentStatus` is guaranteed that the toast msg will only be poped once if there is no status change.

sejli: > For example, first, we need the dataSource and and accelerationToOperate to setup the displayed index name on the toast msg correctly. 
Got it. `accelerationToOperate` is consistently changing, and the `useEffect()` should run once it changes since the acceleration will be set. `datasource` is initialized when the hook is initialized, so it won't be changing and thus, I think it's not necessary in the dependency array. You can test this on your side.
> So the check of the currentStatus is guaranteed that the toast msg will only be poped once if there is no status change.
Makes sense. I think my main comment was to try and avoid unnecessary runs of `useEffect()`. If this is what works, looks good to me.

public/components/datasources/components/manage/accelerations/acceleration_details_flyout.tsx
```diff
@@ -101,18 +123,28 @@ export const AccelerationDetailsFlyout = (props: AccelerationDetailsFlyoutProps)
       });
   };
 
+  useEffect(() => {
+    if (operationSuccess !== undefined) {
+      if (operationSuccess) {
```
paulstn, line 128: Won't both of these if statements result the exact same way? (operationSuccess) is the same as resulting to 'if operationSuccess is defined'

RyanL1997: good catch. just clear it up.

public/components/datasources/components/manage/accelerations/acceleration_operation.tsx
RyanL1997: > datasource is initialized when the hook is initialized, so it won't be changing and thus, I think it's not necessary in the dependency array. 

Good catch! Fixed.

public/components/datasources/components/__tests__/testing_constants.ts
```diff
@@ -3,6 +3,8 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+import { CachedAcceleration } from '../../../../../common/types/data_connections';
```
ps48, line 6: All the. testing constants should be moved to `https://github.com/opensearch-project/dashboards-observability/blob/main/test/datasources.ts`

paulstn (approved):

public/components/datasources/components/manage/accelerations/acceleration_details_flyout.tsx
```diff
@@ -18,22 +18,19 @@ import {
...
-} from './utils/acceleration_utils';
+import { onDiscoverIconClick, AccelerationActionType } from './utils/acceleration_utils';
 import { coreRefs } from '../../../../../framework/core_refs';
 import { OpenSearchDashboardsResponse } from '../../../../../../../../src/core/server/http/router';
 import { CachedAcceleration } from '../../../../../../common/types/data_connections';
+import { useAccelerationOperation } from './acceleration_operation';
+import { AccelerationActionOverlay } from './acceleration_action_overlay';
 
 export interface AccelerationDetailsFlyoutProps {
   index: string;
```
ps48, line 29: Should we remove the `index` prop as discussed earlier? We get this already from `acceleration` prop

public/plugin.tsx
```diff
@@ -101,7 +101,12 @@ export const [
   getRenderAccelerationDetailsFlyout,
   setRenderAccelerationDetailsFlyout,
 ] = createGetterSetter<
-  (index: string, acceleration: CachedAcceleration, dataSourceName: string) => void
+  (
```
ps48, line 104: we should also update flyout params in types: https://github.com/opensearch-project/dashboards-observability/blob/main/public/types.ts

public/components/datasources/components/manage/accelerations/acceleration_details_flyout.tsx
RyanL1997: No. I found out that this "index" is actually the displayed name we processed. For example if it is a skiiping index, it doesn't have the "indexName". So this `index` is actually referencing to the index we are showing in the rows and detail flyout. We cannot remove this.

public/plugin.tsx
RyanL1997: fixed.

public/components/datasources/components/manage/accelerations/acceleration_details_flyout.tsx
ps48: I think we need to have that as a function call inside the flyout too. Otherwise, we cannot expect workbench to reimplement the logic.

RyanL1997: Transferring the discussion with @dai-chen here: lets keep the skipping index name just as `skipping_index` and we add the full path of this skipping index in the pop up toast and title of overlay: ${datasource}.${database}.${table}.

public/components/datasources/components/__tests__/testing_constants.ts
RyanL1997: done.

mengweieric (approved):

PR #1530, opened 2024-03-13
Data Sources: Installed Integrations Table

Swiddis:
### Description
First PR in a coming series adding installed integrations to the data sources viewer. Some WIP upcoming parts:
- [x] Snapshot Tests
- [x] Search Bar
- [x] Loading integrations to the table dynamically (need to add linked data source(s) to the integration instance fields).
- [x] Adding action buttons on the right
- [x] Empty state
image

This PR is mostly here to help get some initial work in for preliminary looks at the 2.13 release, but isn't release-ready yet so I'll hold off on the backport label for now.

### Issues Resolved
N/A (need to write an issue for the feature).

### Check List
- [ ] New functionality includes testing.
  - [ ] All tests pass, including unit test, integration test and doctest
- [ ] New functionality has been documented.
  - [ ] New functionality has javadoc added
  - [ ] New functionality has user manual doc added
- [X] Commits are signed per the DCO using --signoff

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #1544, opened 2024-03-15
Add actual integration queries to table

Swiddis:
### Description
Adds the ability to dynamically load integrations into the installed integrations table, and reactivates the tab. Also adds some terminal spacing to the S3 setup page since it was getting clipped by the bottom bar.

TODO: Cypress tests

Empty Table:
image

When an empty test integration is added that matches the data source:
image

### Issues Resolved
N/A

### Check List


Swiddis:
Depends on #1530

server/adaptors/integrations/types.ts
```diff
@@ -124,6 +124,7 @@ interface AssetReference {
   assetId: string;
   isDefaultAsset: boolean;
   description: string;
+  status?: string;
```
ps48, line 127: Can the status be from an enum?

public/components/integrations/components/setup_integration.tsx
```diff
@@ -415,6 +415,9 @@ export function SetupIntegrationForm({
               </EuiFormRow>
             </>
           ) : null}
+          {/* Bottom bar will overlap content if there isn't some space at the end */}
+          <EuiSpacer />
```
ps48, line 419: nit. Euispacer has props for size="l"/"xl". May be use that.

public/components/integrations/components/added_integration.tsx
```diff
@@ -27,11 +27,24 @@ import React, { useEffect, useState } from 'react';
...
+export const IntegrationHealthBadge = ({ status }: { status?: string }) => {
+  switch (status) {
+    case undefined:
+      return <EuiHealth color="warning">Unknown</EuiHealth>;
+    case 'available':
+      return <EuiHealth color="success">Active</EuiHealth>;
+    case 'partially-available':
+      return <EuiHealth color="warning">Partially Available</EuiHealth>;
+    default:
+      return <EuiHealth color="danger">Critical</EuiHealth>;
```
ps48, line 44: not a blocker: Is there any documentation or comment where i can learn about how we determine the status? 
As a user I would like to learn more about what statuses represent

ps48 (approved):

paulstn (approved):

public/components/integrations/components/setup_integration.tsx
Swiddis: Tried xxl but it was still too small, 2 default spacers did the trick

public/components/integrations/components/added_integration.tsx
Swiddis: We documented it in the design somewhere but I don't think we've added it to OSS docs, I'll see about finding a home for the info

server/adaptors/integrations/types.ts
Swiddis: I have a few things to fix with status so I made #1545, will work on it soon

PR #1560, opened 2024-03-18
Allow browsing integrations in Flyout from Data Sources page

Swiddis:
### Description
As part of in-progress work for allowing integration installation in a Flyout, this PR does the first step of copying the integration browsing component to the Flyout, with some additional refactoring for the changes (mostly around how category filters are rendered).

image

### Issues Resolved
N/A

### Check List
- [x] New functionality includes testing.
  - [x] All tests pass, including unit test, integration test and doctest
- [ ] New functionality has been documented.
  - [ ] New functionality has javadoc added
  - [ ] New functionality has user manual doc added
- [X] Commits are signed per the DCO using --signoff

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #1561, opened 2024-03-18
Add integration installation to data sources flyout

Swiddis:
### Description
Continuing off #1560, this PR adds the ability to install integrations from the Flyout.

https://github.com/opensearch-project/dashboards-observability/assets/31739405/85ca4171-5cbc-4866-92f3-88f77c3449ba

### Issues Resolved
N/A

### Check List


public/components/datasources/components/manage/integrations/installed_integrations_table.tsx
```diff
@@ -148,9 +149,25 @@ export const InstallIntegrationFlyout = ({
     ),
   };
 
+  const [installingIntegration, setInstallingIntegration] = useState<string | null>(null);
```
Swiddis, line 152: To allow installing in a flyout: We add the optional parameter `setInstallingIntegration` that is used to conditionally override the href to the setup page with an onClick that sets `installingIntegration` to the selected integration. (See diff in `available_integration_table.tsx`.)

public/components/integrations/components/setup_integration.tsx
```diff
@@ -635,29 +655,81 @@ export function SetupIntegrationPage({ integration }: { integration: string }) {
...
+            integration={template}
+            loading={showLoading}
+            setLoading={setShowLoading}
+            setSetupCallout={setSetupCallout}
+            unsetIntegration={unsetIntegration}
+          />
+        </EuiBottomBar>
+      </>
+    );
+  } else if (renderType === 'flyout') {
```
Swiddis, line 688: Bottom bar doesn't render in a flyout if we keep the existing EuiPage component, so I added multiple render types that swap the top-level components between those for an independent page or those for a flyout body. Slightly annoying but not sure what a better way is.

ps48 (approved):

public/components/datasources/components/manage/integrations/installed_integrations_table.tsx
RyanL1997: nit: if we switch to use `navigateToApp()`, do we still need this?

Reference:
Implementation of the EuiLink for datasource: https://github.com/opensearch-project/dashboards-observability/blob/31d506a1672803060d6abb7699ae099fbeb0b628/public/components/datasources/components/manage/accelerations/flyout_modules/acceleration_details_tab.tsx#L107

PR #1586, opened 2024-03-20
Lifting useLoadTableColumnsToCache out of Flyouts

sejli:
### Description
* Flyouts had an issue where `startLoading` would poll indefinitely when the flyout gets closed. Lifting the `useLoadTableColumnsToCache` hook out of flyouts stops this behavior. 
* Updated UI on empty associated objects table

### Issues Resolved
N/A

### Check List
- [ ] New functionality includes testing.
  - [ ] All tests pass, including unit test, integration test and doctest
- [ ] New functionality has been documented.
  - [ ] New functionality has javadoc added
  - [ ] New functionality has user manual doc added
- [x] Commits are signed per the DCO using --signoff

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #1598, opened 2024-03-20
Updating usePolling to cleanup after unmount

sejli:
### Description
* A simpler fix that achieves the same thing as #1586 
* When the component that initializes `usePolling()` unmounts before `stopPolling()` is called, it causes the `setInterval` to become a zombie process in the background. Updated the hook such that we check if unmounted before exiting the `startPolling()` function and clear if true.
* Updated UI for empty state
* Allow failed cache state to start reloading

### Issues Resolved
N/A

### Check List

### DIFF
--- a/public/components/datasources/components/manage/associated_objects/utils/associated_objects_tab_empty.tsx
+++ b/public/components/datasources/components/manage/associated_objects/utils/associated_objects_tab_empty.tsx
@@ -3,10 +3,9 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import { EuiButton, EuiEmptyPrompt, EuiLink, EuiText } from '@elastic/eui';
+import { EuiEmptyPrompt, EuiText } from '@elastic/eui';
 import React from 'react';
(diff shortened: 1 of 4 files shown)


public/components/datasources/components/manage/associated_objects/utils/associated_objects_tab_empty.tsx
```diff
@@ -3,28 +3,16 @@
...
-
-  const QueryWorkbenchButton = (
-    <EuiButton
-      iconSide="right"
-      onClick={() => application!.navigateToApp('opensearch-query-workbench')}
-      iconType="popout"
-    >
-      Query Workbench
-    </EuiButton>
-  );
```
ps48, line 27: Why are we removing this? Shouldn't be present when the object list is empty?

sejli: Nice catch, added it back.

ps48 (approved):

Swiddis (approved):

PR #1609, opened 2024-03-21
Fixed bugs in explorer redirection

paulstn:
### Description
- allow for redirection in associated objects skipping indices
- redirection in datasources table implemented
- extended associated object type
- fixed url redirection to stop causing crashes and unexpected issues, used mainly in the discover to explorer flow
- extended the query text area character limit now to 2^14
- change the direct query visualization error to open acc flyout in Q.W.

- s3 no results page changes with sample queries and glue data cache support to populate a functioning query, shown in the following flows:|

s3 PPL empty page (no query run yet):
image
s3 SQL empty page (no query run yet):
image
s3 no results (post query run w/ no data returned):
image

### Issues Resolved

### Check List


common/types/data_connections.ts
```diff
@@ -35,11 +35,12 @@ export interface Acceleration {
 }
 
 export interface AssociatedObject {
+  tableName: string;
```
paulstn, line 38: @RyanL1997 @sejli are you fine with this additional field in associated object? using it for defining the table name in skipping index redirection

sejli: Sounds good to me, as long as it's updated wherever this type is used.

ps48 (approved):

common/constants/autocomplete.ts
```diff
@@ -6,6 +6,8 @@
 
 import { BaseItem } from '@algolia/autocomplete-core';
 
+export const textAreaMaxLength = 16384;
```
Swiddis, line 9: Just curious: where is this limit from? I don't see 2^14 often.

paulstn: arbitrary amount where I looked at how long certain sql queries could be, i made a value judgement. we can always change it later, though

public/components/event_analytics/explorer/explorer.tsx
```diff
@@ -276,28 +276,32 @@ export const Explorer = ({
...
-      );
-      dispatch(
-        changeQuery({
-          tabId,
-          query: { [RAW_QUERY]: queryToRun },
-        })
-      );
-      setTempQuery(queryToRun);
+      if (datasourceName && datasourceType) {
+        dispatch(
```
Swiddis, line 280: Another question out of curiosity: what's the benefit of `dispatch` over `useEffect`? I'm not sure what'd motivate me to use one over the other, I haven't looked deeply into redux.

Swiddis (approved):

public/components/event_analytics/explorer/explorer.tsx
paulstn: the extent of my knowledge is that dispatch would be used for updating redux state, i'm not sure about if useEffect could be used for the same thing

PR #1655, opened 2024-03-28
prevent logged out datasources call

paulstn:
### Description
prevent logged out datasources call

### Issues Resolved

### Check List

### DIFF
--- a/public/plugin.tsx
+++ b/public/plugin.tsx
@@ -389,21 +389,27 @@ export class ObservabilityPlugin
     const { dataSourceService, dataSourceFactory } = startDeps.data.dataSources;
 
     // register all s3 datasources
-    if (startDeps.securityDashboards) {
-      core.http.get(SECURITY_PLUGIN_ACCOUNT_API).then(() => {
-        dataSourceFactory.registerDataSourceType(S3_DATASOURCE_TYPE, S3DataSource);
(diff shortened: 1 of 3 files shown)


public/plugin.tsx
```diff
@@ -388,16 +389,18 @@ export class ObservabilityPlugin
...
-    core.http.get(`${DATACONNECTIONS_BASE}`).then((s3DataSources) => {
-      s3DataSources.map((s3ds) => {
-        dataSourceService.registerDataSource(
-          dataSourceFactory.getDataSourceInstance(S3_DATASOURCE_TYPE, {
-            name: s3ds.name,
-            type: s3ds.connector.toLowerCase(),
-            metadata: s3ds,
-          })
-        );
+    core.http.get(SECURITY_PLUGIN_ACCOUNT_API).then(() => {
```
ps48, line 392: Would it make sense to check if the response of the accounts API call is `401` and only then not trigger the datasource registration? Otherwise if the security plugin doesn't exist it might return a `404` and in this scenario we still go ahead with datasource registration.

paulstn: I changed the code to originally to perform this, but I changed it again to rely on whether the security plugin would be installed so that we can avoid any extra error codes.

YANG-DB (approved):

Swiddis (approved):

PR #1736, opened 2024-04-23
Enhance Explorer to use describe command

paulstn:
### Description
This change allows for users to query in Log Explorer with the describe command, instead of requiring the `source = <index>` pattern.
To enact this change, the following adjustments were made:
- When processing the query, while the code is looking for the index in `source = <index>`, there will also be an additional check to see if the query follows the pattern `describe <index>`, and if it does, supplying the logic with the index as if it were found in the first pattern. This allows for the describe query to be processed at all.
- The timestamp may or may not be in the index, but the describe command is not time-based. And so in order to display that properly within Log Explorer and make sure the user workflows are guided in a non-time-based manner, the timestamp is overridden in a few key places to be empty. 

The following workflows were manually tested:
- A regular workflow, with a regular query and a timestamp
- A basic describe query workflow, where the query was just `describe <index>`, including the following:
  - Selecting fields within the UI
  - A properly displayed data grid
  - Modifications to timestamp (should have no effect)
- A more complex describe query workflow, in which everything above was kept constant except for the query which included some selected fields
- Saving and loading the basic and complex describe queries
- Creating and viewing an application with a describe query

### Issues Resolved
- Also fixed an issue where app analytics was crashing in addition to the above

### Check List


common/constants/shared.ts
```diff
@@ -93,6 +93,7 @@ export const PPL_INDEX_INSERT_POINT_REGEX = /(search source|source|index)\s*=\s*
 export const PPL_INDEX_REGEX = /(search source|source|index)\s*=\s*([^|\s]+)/i;
 export const PPL_WHERE_CLAUSE_REGEX = /\s*where\s+/i;
 export const PPL_NEWLINE_REGEX = /[\n\r]+/g;
+export const PPL_DESCRIBE_INDEX_REGEX = /(describe)\s*([^|\s]+)/i;
```
ykmr1224, line 96: Should we change `\s*` to `\s+`? (current regex would match with `describeXXX`)

public/components/common/query_utils/index.ts
```diff
@@ -219,6 +220,14 @@ export const getIndexPatternFromRawQuery = (query: string): string => {
   return getPromQLIndex(query) || getPPLIndex(query);
 };
 
+export const getDescribeQueryIndexFromRawQuery = (query: string): string => {
+  const matches = query.match(PPL_DESCRIBE_INDEX_REGEX);
+  if (matches) {
+    return matches[2];
+  }
+  return '';
```
ykmr1224, line 228: I'd recommend returning `undefined` explicitly instead of returning empty string. (It would enforce the check of `undefined`)

ykmr1224 (changes requested):
A few comments.

common/constants/shared.ts
paulstn: That's right, I had `\s*` because `source = <index>` allowed no whitespace but upon checking the ppl API it seems like no whitespace wouldn't work for `describe<index>`, thanks for the catch!

public/components/common/query_utils/index.ts
paulstn: To try and make this change with the least disruption possible, I followed the current method that an index is retrieved from the query here: https://github.com/opensearch-project/dashboards-observability/blob/411f33a154264b391df8dff751735dd51ebde67a/public/components/common/query_utils/index.ts#L210, which returns an empty string. Later on in the code the index as an empty string is used for some other purposes, and because of the legacy nature of this code, I decided to stick with whatever is done.
However, after looking through the code again, it seems like the final method I settled on for the describe query doesn't directly pass the parsed index anywhere else, meaning I can write it in as undefined. 
TLDR: Making this change now.

anirudha (approved):

mengweieric:
have we tested visualization tab? I understand we will not support visualization for describe query but is it affected?

public/components/event_analytics/explorer/explorer.tsx
```diff
@@ -643,11 +650,14 @@ export const Explorer = ({
...
+                        timeStampField={
+                          !query[FINAL_QUERY].match(PPL_DESCRIBE_INDEX_REGEX)
+                            ? queryRef.current![SELECTED_TIMESTAMP]
+                            : ''
+                        }
                         rawQuery={appBasedRef.current || queryRef.current![RAW_QUERY]}
                         totalHits={
-                          (isDefaultDataSourceType || appLogEvents) &&
-                          query[SELECTED_TIMESTAMP] !== ''
+                          showTimeBasedComponents
```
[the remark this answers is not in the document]
paulstn: I've leveraged some of the no-timestamp index workflows for this new case, we could get UX signoff if needed but since it uses existing views with no changes to those, I don't think it's entirely necessary.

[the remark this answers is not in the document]
mengweieric: Sure. This is not a blocker, good to reach out to UX for a quick check and sign-off, more of a flow wise.

mengweieric (approved):

PR #1825, opened 2024-05-07
Refactor away integrations adaptor class

Swiddis:
### Description
When integrations was originally written, the integration management happened in a backend plugin. This management was since moved to the frontend. Part of this migration involved making an adaptor interface that would support both management types during the migration, but now that the backend management has been absent for a while this interface is a redundant thin wrapper around the underlying IntegrationManager class. The wrapper doesn't do much except make the code slightly trickier to navigate and less obvious.

This PR refactors the adaptor away, and invokes the manager directly.

### Issues Resolved
N/A

### Check List


YANG-DB (approved):

joshuali925 (approved):
i don't think it hurts to use an interface even if there is only one implementation, but either way works

Issue #1731, opened 2024-04-19
[BUG]Integration changes dashboard ID


YANG-DB:
**What is the bug?**
When installing an Integration, it changes the dashboard's ID 

**How can one reproduce the bug?**
1) install an integration
2) review its name differs from the original given id

**What is the expected behavior?**
Integration should not change dashboard's id - or at least allow a parameter to keep the original ID

**Do you have any additional context?**
The installing of an integration should keep the same save concepts that already exist in the import save objects API and allow different importing options:
 - always create new ID
 - always override with imported ID
 - let user resolve conflicts

Swiddis:
Working as intended, it's meant to change IDs while preserving all references in order to allow creating multiple copies of integration instances without ID collisions.

YANG-DB:
Hi
This capability is still needed since some cases required to save the original ID - this should operate in a similar way to the existing dashboards-management import API...

PR #1833, opened 2024-05-17
Integration override enabled

YANG-DB:
### Description
Add the ability for an integration's dashboards to use its original ID's to install .
It has the practical effect of `override` in the case where an older integration was already installed

This is to be accompanied with a UX checkbox to allow user's selection whether to generate new ID or use existing to override - exactly similar to the savedObject import dialog

### Issues Resolved
https://github.com/opensearch-project/dashboards-observability/issues/1731
https://github.com/opensearch-project/dashboards-observability/issues/1835
https://github.com/opensearch-project/opensearch-catalog/issues/170

### Check List


Swiddis:
What happens if you create two integrations with overriding, and delete one of the integrations?

YANG-DB:
I will add a checkbox to indicate whether to override the dashboard id or not (similar to the savedObjects import dialog) so that the user will be aware

> What happens if you create two integrations with overriding, and delete one of the integrations?
This is the same as with dashboards -  the responsibility if on the user who is installing the integration template -

PR #1871, opened 2024-06-04
Lake formation support updates

amsiglan:
### Description
This PR 
- Updates the query used for fetching table for security lake since `EXTENDED` keyword is not supported
- Add a new connection `Security Lake` type for integrations that have at least one workflow for security lake data sources.
- Filters integrations based on the data source

### Check List

### DIFF
--- a/common/types/integrations.ts
+++ b/common/types/integrations.ts
@@ -1,0 +1,6 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+export type IntegrationConnectionType = 's3' | 'index' | 'securityLake';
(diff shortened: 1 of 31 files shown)


RyanL1997 (approved):

public/components/event_analytics/explorer/no_results.tsx
```diff
@@ -127,7 +132,9 @@ export const NoResults = ({ tabId }: any) => {
                     <p>Show a list of tables within a database</p>
                     <EuiSpacer size="s" />
                     <CreatedCodeBlock
-                      code={`SHOW TABLE EXTENDED IN ${datasourceName}.<database> LIKE '*'`}
+                      code={`SHOW ${
+                        dataSourceConnectionType === 'SECURITYLAKE' ? 'TABLES' : 'TABLE EXTENDED'
```
ps48, line 136: if this is just `show tables` query, do we want to filter them based on hive or non-hive provider?

amsiglan: Security lake only supports Iceberg tables so we don't need any filtering

ps48 (approved):

common/types/integrations.ts
```diff
@@ -0,0 +1,6 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+export type IntegrationConnectionType = 's3' | 'index' | 'securityLake';
```
ps48, line 6: Nit. (not blocking) Some places I see `securityLake` in lower case and some place in upper case. Could be better to keep it consistent.

ps48 (approved):

Issue #1916, opened 2024-06-18
[BUG] Trace analytics UI: Clicking Span resets scroll bar


TackAdam:
**What is the bug?**
In Trace analytics - Trace Analytics page, clicking a span is reseting the scroll bar to the top.

**How can one reproduce the bug?**

1. Go to https://playground.opensearch.org/app/observability-traces#/traces
2. Filter the date to include the last 5 years and refresh.
<img width="1197" alt="Screenshot 2024-06-18 at 2 00 20 PM" src="https://github.com/opensearch-project/dashboards-observability/assets/105462877/915233f8-b917-44c3-a507-653009774de2"><br />
3. Click on the Trace ID to the left of "client_pay_order" <br />

<img width="474" alt="Screenshot 2024-06-18 at 2 00 56 PM" src="https://github.com/opensearch-project/dashboards-observability/assets/105462877/fe3c4ddc-df67-48fc-b8f2-a9804ebb7195">

4. Scroll down the list of spans and click "analyics-service Logging.Controller.save"
<img width="1410" alt="Screenshot 2024-06-18 at 2 01 13 PM" src="https://github.com/opensearch-project/dashboards-observability/assets/105462877/87b75d84-c086-4f5a-acd0-29751103324f">
Click here: <br />
<img width="571" alt="Screenshot 2024-06-18 at 2 01 26 PM" src="https://github.com/opensearch-project/dashboards-observability/assets/105462877/fd2a3faf-1bc5-4294-bee0-0df5a173cc94">

5. See error: As the flyout opens the scrollbar is reset to the top.
<img width="1430" alt="Screenshot 2024-06-18 at 2 01 39 PM" src="https://github.com/opensearch-project/dashboards-observability/assets/105462877/c4594a4d-e3f4-4df8-98b5-585b46d626cb">

**What is the expected behavior?**
The scrollbar should maintain its position prior to the click occurring.

**What is your host/environment?**
 - Plugins

**Do you have any screenshots?**

**Do you have any additional context?**
Add any other context about the problem.

TackAdam:
Fixed with https://github.com/opensearch-project/dashboards-observability/pull/1917

PR #1917, opened 2024-06-19
[Bug fix] Trace analytics scroll bar reset

TackAdam:
### Description
This change resets the scroll bar to the position it was at prior to opening the fly-out.

### Issues Resolved
https://github.com/opensearch-project/dashboards-observability/issues/1916

### Check List

### DIFF
--- a/public/components/trace_analytics/components/traces/span_detail_panel.tsx
+++ b/public/components/trace_analytics/components/traces/span_detail_panel.tsx
@@ -15,7 +15,7 @@ import {
 } from '@elastic/eui';
 import debounce from 'lodash/debounce';
 import isEmpty from 'lodash/isEmpty';
-import React, { useEffect, useMemo, useState } from 'react';
+import React, { useEffect, useMemo, useRef, useState } from 'react';
 import { HttpSetup } from '../../../../../../../src/core/public';
 import { Plt } from '../../../visualizations/plotly/plot';
 import { TraceAnalyticsMode } from '../../home';
@@ -187,7 +187,19 @@ export function SpanDetailPanel(props: {
 
   const [currentSpan, setCurrentSpan] = useState('');
 
+  const spanContainerRef = useRef<HTMLDivElement | null>(null);
+  const prevScrollPositionRef = useRef<number | null>(null);
+
+  useEffect(() => {
+    if (spanContainerRef.current) {
+      setTimeout(() => {
+        spanContainerRef.current.scrollTop = prevScrollPositionRef.current || 0;
+      }, 0);
+    }
+  }, [currentSpan]);
+
   const onClick = (event: any) => {
+    const currentScrollPosition = spanContainerRef.current?.scrollTop || 0;
     if (!event?.points) return;
     const point = event.points[0];
     if (fromApp) {
@@ -195,6 +207,7 @@ export function SpanDetailPanel(props: {
     } else {
       setCurrentSpan(point.data.spanId);
     }
+    prevScrollPositionRef.current = currentScrollPosition;
   };
 
   const renderFilters = useMemo(() => {
@@ -279,7 +292,7 @@ export function SpanDetailPanel(props: {
           </>
         )}
         <EuiHorizontalRule margin="m" />
-        <div style={{ overflowY: 'auto', maxHeight: 500 }}>
+        <div style={{ overflowY: 'auto', maxHeight: 500 }} ref={spanContainerRef}>
           {toggleIdSelected === 'timeline' ? (
             <Plt
               data={data.gantt}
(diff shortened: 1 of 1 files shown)


joshuali925 (approved):
lgtm

public/components/trace_analytics/components/traces/span_detail_panel.tsx
```diff
@@ -187,14 +187,27 @@ export function SpanDetailPanel(props: {
...
   const [currentSpan, setCurrentSpan] = useState('');
 
+  const spanContainerRef = useRef<HTMLDivElement | null>(null);
+  const prevScrollPositionRef = useRef<number | null>(null);
+
+  useEffect(() => {
+    if (spanContainerRef.current) {
+      setTimeout(() => {
+        spanContainerRef.current.scrollTop = prevScrollPositionRef.current || 0;
+      }, 0);
```
ps48, line 197: Question: why do we need `setTimeout` with delay 0 here?

TackAdam: Without it, closing the fly-out will reset the scroll bar. I believe this happens because the fly-out table closing behavior is handled in span_detail_table and it needs a delay for the use-effect.

TackAdam:
This fix results in a slight flicker as the scroll bar adjust after being rendered to the correct position. I have not been able to find a fix that prevents this. Have tried using useMemo as it is handled in the SpanList menu but the click event for Timeline is happening in a different area.

joshuali925:
scrollbar resetting on rerender is not normal behavior, it might be caused by the plot rerendering. if recovering scroll position flickers, you can try to memorize the plot component instead

Swiddis:
I see that we [rely on a memo](https://github.com/TackAdam/dashboards-observability/blob/4fa37057e35d4e4cc40708ecc22a4b34544e19bf/public/components/trace_analytics/components/traces/span_detail_panel.tsx#L250-L268) for generating the span list and [inject that value directly](https://github.com/TackAdam/dashboards-observability/blob/4fa37057e35d4e4cc40708ecc22a4b34544e19bf/public/components/trace_analytics/components/traces/span_detail_panel.tsx#L305), but I wonder why we need a memo at all? Can we use the spans as JSX and let React do the rendering? I suspect that memo is where the reset is sneaking in to begin with.

joshuali925:
i don't think the span list is relevant for this bug, since triggering the bug doesn't need user to go to the span list view, that component was never mounted

Swiddis:
Ah, got confused between this being the span view as a whole and there being a timeline option under the span view. But the same thing applies to the timeline, both of them are being memoized.

joshuali925:
the gantt chart is not memorized, memorizing it could workaround this bug. you can try adding this above `return`,
```tsx
  const ganttChart = useMemo(
    () => (
      <Plt
        data={data.gantt}
        layout={layout}
        onClickHandler={onClick}
        onHoverHandler={onHover}
        onUnhoverHandler={onUnhover}
      />
    ),
    [data.gantt, layout]
  );
```

and change 
https://github.com/opensearch-project/dashboards-observability/blob/4fa37057e35d4e4cc40708ecc22a4b34544e19bf/public/components/trace_analytics/components/traces/span_detail_panel.tsx#L296-L306

to
```ts
{toggleIdSelected === 'timeline' ? ganttChart : spanDetailTable}
```

TackAdam:
Implemented Josh's suggestion and tested. All working correctly.

ps48 (approved):

joshuali925:
for completeness and good practice, the dependency array should include all dependencies. so the correct way would be using `useCallback` to memorize the event handlers
```tsx
  const onClick = useCallback(
    (event: any) => {
      if (!event?.points) return;
      const point = event.points[0];
      if (fromApp) {
        props.openSpanFlyout(point.data.spanId);
      } else {
        setCurrentSpan(point.data.spanId);
      }
    },
    [props.openSpanFlyout, setCurrentSpan, fromApp]
  );
```

```tsx
  const onHover = useCallback(() => {
    const dragLayer = document.getElementsByClassName('nsewdrag')?.[0];
    dragLayer.style.cursor = 'pointer';
  }, []);

  const onUnhover = useCallback(() => {
    const dragLayer = document.getElementsByClassName('nsewdrag')?.[0];
    dragLayer.style.cursor = '';
  }, []);
```
and put them in the dependency array
```tsx
  const ganttChart = useMemo(
    () => (
      <Plt
        data={data.gantt}
        layout={layout}
        onClickHandler={onClick}
        onHoverHandler={onHover}
        onUnhoverHandler={onUnhover}
      />
    ),
    [data.gantt, layout, onClick, onHover, onUnhover]
  );
```

some functions like `setCurrentSpan` would never change. but if something like `fromApp` changes, `onClick` will be re-created with the correct behavior. And since `onClick` changed and `ganttChart` depends on it, `ganttChart` component will also be re-created with the correct click behavior.

it likely won't make a real difference, since `fromApp` shouldn't change

PR #1920, opened 2024-06-21
Refactor of Trace analytics to remove duplicated checks.

TackAdam:
Refactor of Trace analytics to improve readability by reducing duplicated checks.

### Description
Files/calls using <mode === 'data_prepper'> and <mode === 'jaeger'> multiple times replaced with isDataPrepper and isJaeger to improve readability.

Also removed currPage argument from onTableChange in /trace_analytics/components/traces/traces_table.tsx call  as it was not used in the function and was throwing an error.

### Check List

### DIFF
--- a/public/components/trace_analytics/components/traces/traces_content.tsx
+++ b/public/components/trace_analytics/components/traces/traces_content.tsx
@@ -43,6 +43,8 @@ export function TracesContent(props: TracesProps) {
   const [redirect, setRedirect] = useState(true);
   const [loading, setLoading] = useState(false);
   const [trigger, setTrigger] = useState<'open' | 'closed'>('closed');
+  const isDataPrepper = mode === 'data_prepper';
+  const isJaeger = mode === 'jaeger';
 
   const DataSourceMenu = dataSourceManagement?.ui?.getDataSourceMenu<DataSourceViewConfig>();
   useEffect(() => {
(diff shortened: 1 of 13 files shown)


public/components/trace_analytics/components/traces/traces_content.tsx
```diff
@@ -43,6 +43,8 @@ export function TracesContent(props: TracesProps) {
   const [redirect, setRedirect] = useState(true);
   const [loading, setLoading] = useState(false);
   const [trigger, setTrigger] = useState<'open' | 'closed'>('closed');
+  const isDataPrepper = mode === 'data_prepper';
```
RyanL1997, line 46: I was wondering if we can define this into an util file, so that we don't need to re define it in all the related classes. Maybe we can setup a util file under here: https://github.com/opensearch-project/dashboards-observability/tree/main/public/components/trace_analytics

public/components/trace_analytics/components/traces/traces_content.tsx
```diff
@@ -43,6 +43,8 @@ export function TracesContent(props: TracesProps) {
...
+  const isJaeger = mode === 'jaeger';
```
RyanL1997, line 47: same here

RyanL1997 (commented):
Hi @TackAdam , thanks for taking this on! In generally, this is LGTM. I just left a comment for setting up a util file for trace related constants, since we are reusing them in multiple places.

Issue #1440, opened 2024-02-14
[FEATURE] Support Multiple Data Source in Trace Analytics Dashboards Plugin


BionIT:
**Is your feature request related to a problem?**
With multiple data source enabled, users of OpenSearch Dashboards are able to create data source connections and query local and remote OpenSearch clusters. Enabling multiple data source at dashboards plugin would ensure users that have been using the multiple data source feature to have a consistent experience in the dashboard, and allows users to use the same dashboard for different clusters.

**What solution would you like?**
Support multiple data source in the Trace analytics dashboards plugin and allow user to use one dashboard for different backend clusters

**What alternatives have you considered?**

**Do you have any additional context?**

PR #2051, opened 2024-08-09
[FEATURE] MDS support in Integrations for observability plugin

sumukhswamy:
### Description
Multi data source support in Integrations for observability plugin

https://github.com/user-attachments/assets/6b1aab11-5af1-4d88-a68c-6dbb0a93b33d

### Issues Resolved
https://github.com/opensearch-project/dashboards-observability/issues/1440

### Check List


public/components/integrations/components/setup_integration_inputs.tsx
```diff
@@ -95,7 +98,9 @@ const suggestDataSources = async (
         }) ?? []
       );
     } else if (type === 's3' || type === 'securityLake') {
-      const result = (await http.get(DATACONNECTIONS_BASE)) as Array<{
+      const result = (await http.get(
+        `${DATACONNECTIONS_BASE}/dataSourceMDSId=${dataSourceMDSId ?? ''}`
```
Swiddis, line 102: Is this usage correct? This is a URL with an `=` in it instead of a request param? The `http.get` method already has a way of using request params with automatic undefined-handling.

public/components/integrations/components/setup_integration_inputs.tsx
```diff
@@ -205,9 +211,20 @@ export function IntegrationConnectionInputs({
   dataSourceEnabled: boolean;
   dataSourceManagement: DataSourceManagementPluginSetup;
   savedObjectsMDSClient: SavedObjectsStart;
+  handleSelectedDataSourceChange: (
+    dataSourceMDSId: string | undefined,
+    dataSourceMDSLabel: string | undefined
+  ) => void;
   lockConnectionType?: boolean;
 }) {
+  const [dataSourceMDSId, setDataSourceMDSId] = useState<string | undefined>('');
```
Swiddis, line 220: What's the difference between `undefined` and empty string here?

server/adaptors/integrations/integrations_manager.ts
```diff
@@ -157,6 +157,8 @@ export class IntegrationsManager {
     templateName: string,
     name: string,
     indexPattern: string,
+    dataSourceMDSId: string | undefined,
+    dataSourceMDSLabel: string | undefined,
```
Swiddis, line 161: Too small to worry about now, but I notice there's a lot of `var: string | undefined` instead of `var?: string`. I'll probably end up rewriting these next time I'm messing with this code since the inconsistent syntax bugs me.

Swiddis (commented):
Generally LGTM, have a few questions

public/components/integrations/components/setup_integration_inputs.tsx
sumukhswamy: wanted to keep it consistent to how other plugins were using and by adding the mds-id here the path is getting differentiated for the other locations where this api is being used

public/components/integrations/components/setup_integration_inputs.tsx
sumukhswamy: I can make this consistent to be an empty string in all locations

server/adaptors/integrations/integrations_manager.ts
sumukhswamy: I will change it in the cases I have written

Swiddis (approved):
Tentative approval based on above

public/components/integrations/components/added_integration.tsx
```diff
@@ -119,6 +130,22 @@ export function AddedIntegration(props: AddedIntegrationProps) {
...
     return (
       <EuiPageHeader style={{ justifyContent: 'spaceBetween' }}>
+        {dataSourceEnabled && (
+          <DataSourceMenu
+            setMenuMountPoint={setActionMenu}
+            componentType={'DataSourceView'}
+            componentConfig={{
+              activeOption: [
+                {
+                  id: data?.references?.[0]?.dataSourceMDSId,
```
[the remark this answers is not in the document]
RyanL1997: is this line coming from the data source picker? From the call back function for getting the selected cluster id?

public/components/integrations/components/setup_integration_inputs.tsx
RyanL1997: > What's the difference between undefined and empty string here?
`' '` represents the local cluster. I guess the undefined is trying to handle the case that when the MDS is not enabled and the ID is undefined, so at that time we just need to handle the request towards local cluster.

> I can make this consistent to be an empty string in all locations
Ideally, we should still handle the `undefined` and `' '` separately. But just wanna make sure we check the MDS flag correctly so that it will not go thru the MDS endpoint when MDS is disabled.

public/components/integrations/components/__tests__/testing_constants.ts
```diff
@@ -169,6 +169,12 @@ export const addedIntegrationData: AddedIntegrationsTableProps = {
           },
         ],
         id: 'ad7e6e30-0b99-11ee-b27c-c9863222e9bf',
+        references: [
+          {
+            dataSourceMDSId: 'f4cf2c60-2f4d-11ef-bf2d-65348e7e70ea',
+            dataSourceMDSLabel: 'dataSource',
+          },
+        ],
```
ps48, line 177: Can we have two test constants here, one with MDS one without MDS?

public/components/integrations/components/added_integration.tsx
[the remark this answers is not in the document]
sumukhswamy: this is coming in from the saved objects created during the set up process of the integrations.

public/components/integrations/components/__tests__/testing_constants.ts
sumukhswamy: sure

public/components/integrations/components/added_integration_overview_page.tsx
```diff
@@ -33,6 +33,7 @@ export interface AddedIntegrationType {
   assets: any[];
   addedBy: string;
   id: string;
+  references: [];
```
ps48, line 36: Should this be optional?

public/components/integrations/components/create_integration_helpers.ts
```diff
@@ -204,7 +206,8 @@ export const doExistingDataSourceValidation = async (
 
 const createComponentMapping = async (
   componentName: string,
-  payload: ComponentMappingPayload
+  payload: ComponentMappingPayload,
```
ps48, line 209: Can you please create an issue in `https://github.com/opensearch-project/opensearch-js/issues` to add missing components in the js client.

RyanL1997 (approved):

public/components/integrations/components/__tests__/testing_constants.ts
```diff
@@ -169,6 +169,104 @@ export const addedIntegrationData: AddedIntegrationsTableProps = {
...
+        id: 'ad7e6e30-0b99-11ee-b27c-c9863222e9bf',
+        references: [
+          {
+            id: '',
+            name: '',
+            tye: 'data-source',
+          },
+        ],
```
ps48, line 269: `WithoutMDS` shouldn't have references here.

server/adaptors/integrations/integrations_builder.ts
```diff
@@ -4,14 +4,16 @@
...
 import { SavedObjectsClientContract } from '../../../../../src/core/server';
 import { IntegrationReader } from './repository/integration_reader';
-import { SavedObjectsBulkCreateObject } from '../../../../../src/core/public';
 import { deepCheck } from './repository/utils';
 
 interface BuilderOptions {
   name: string;
   indexPattern: string;
+  dataSourceMDSId: string | undefined;
+  dataSourceMDSLabel: string | undefined;
```
ps48, line 16: can be changed to:
```
dataSourceMDSId?: string;
dataSourceMDSLabel?: string;

File: server/adaptors/integrations/__test__/builder.test.ts [hunk 10]
```diff
@@ -307,6 +314,13 @@ describe('IntegrationInstanceBuilder', () => {
         tags: undefined,
         creationDate: expect.any(String),
         assets: refs,
+        references: [
+          {
+            id: undefined,
+            name: undefined,
+            type: 'data-source',
+          },
+        ],
```
ps48, line 323: Ideally, we shouldn't add a reference if Instance was build without MDS. Should remove this altogether without MDS.

public/components/integrations/components/setup_integration.tsx
```diff
@@ -138,16 +142,20 @@ const addIntegration = async ({
...
+  dataSourceMDSId,
+  dataSourceMDSLabel,
   setIsInstalling,
 }: {
   config: IntegrationSetupInputs;
   integration: IntegrationConfig;
   setLoading: (loading: boolean) => void;
   setCalloutLikeToast: (title: string, color?: Color, text?: string) => void;
+  dataSourceMDSId: string | undefined;
+  dataSourceMDSLabel: string | undefined;
```
ps48, line 154: same here undefined can be removed and the filed can be made optional.

ps48: just search for all references of `string | undefined` in the commit

PR #2097, opened 2024-08-26
[Enhancement] Deregister dashboards, applications, logs in MDS

sumukhswamy:
### Description
De-register plugins when in NEO

### Issues Resolved

### Check List

### DIFF
--- a/public/plugin.tsx
+++ b/public/plugin.tsx
@@ -32,9 +32,6 @@ import {
   observabilityApplicationsID,
   observabilityApplicationsPluginOrder,
   observabilityApplicationsTitle,
-  observabilityGettingStartedID,
-  observabilityGettingStartedPluginOrder,
-  observabilityGettingStartedTitle,
   observabilityIntegrationsID,
   observabilityIntegrationsPluginOrder,
   observabilityIntegrationsTitle,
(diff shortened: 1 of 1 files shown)


public/plugin.tsx
```diff
@@ -307,13 +307,15 @@ export class ObservabilityPlugin
...
-    });
+    if (!setupDeps.dataSource) {
+      core.application.register({
+        id: observabilityApplicationsID,
+        title: observabilityApplicationsTitle,
+        category: OBSERVABILITY_APP_CATEGORIES.observability,
+        order: observabilityApplicationsPluginOrder,
+        mount: appMountWithStartPage('applications'),
+      });
+    }
```
ps48, line 318: What happens when we have enabled the new navigation and MDS together? Do have to modify `plugin_nav` file to not register applications there where MDS is enabled?

TackAdam:
LGTM.
Checked functionality and seem correct.
For old navigation mds being enabled no longer registers Applications, Logs, or Dashboards.
Old Navigation MDS off:
OldNav_MDS_Off
Old Navigation MDS on: (application, logs, and dashboards present)
OldNav_MDS_On


For new navigation mds on now disables applications:
NewNav_MDS_On
And it remains when disabled:(mds off)
NewNac_MDS_Off

public/plugin.tsx
TackAdam: It does not break anything as the core navigation group just fails to find observabilityApplicationsID and doesn't register anything to the left side. It would be cleaner to handle this condition though.

TackAdam (approved):

public/plugin.tsx
ps48: > just fails to find observabilityApplicationsID

@TackAdam @sumukhswamy  Is this leading to some error on browser or Dashboards app? If so then let's handle that in this PR

ps48 (approved):

public/plugin.tsx
sumukhswamy: no error seen

PR #2103, opened 2024-08-28
Page Header] New page header for integrations and UI updates

TackAdam:
### Description
New page header for integration and ui updates

Old Navigation:
| Before | After |
| ------ | ----- |
| Before Image 1 | After Image 1 |
| Before Image 2 | After Image 2 |
| Before Image 3 | After Image 3 |
| Before Image 4 | After Image 4 |
| Before Image 5 | After Image 5 |

New Navigation:
| Before | After |
| ------ | ----- |
| Before Image 1 | After Image 1 |
| Before Image 2 | After Image 2 |
| Before Image 3 | After Image 3 |
| Before Image 4 | After Image 4 |
| Before Image 5 | After Image 5 |

### Check List


public/components/integrations/components/integration_overview_panel.tsx
```diff
@@ -4,65 +4,76 @@
...
-                }}
-                fill
-                disabled={props.loading}
-                data-test-subj="add-integration-button"
-                data-click-metric-element="integrations.set_up"
-              >
-                Set Up
-              </EuiButton>
+              <EuiTitle data-test-subj="eventHomePageTitle" size="l">
+                <h3>{config.displayName || config.name}</h3>
```
Swiddis, line 73: I see we're meant to be prioritizing the display name here (this is correct), but in the screenshot it's falling back to the non-display name. Can you check why/fix it?

TackAdam: While under the new navigation the headers are generated by dashboards directly. They pull from the breadcrumbs and display it as the title while removing the breadcrumb.
This part is rendered by the following code:
```
  return newNavigation ? (
    <HeaderControlledComponentsWrapper
      components={[
        <EuiFlexGroup gutterSize="s" alignItems="flexEnd" responsive={false}>
          {buttons}
        </EuiFlexGroup>,
      ]}
    />
```
The only way to change this would be to update the breadcrumb to be the display name. Not sure if we want that change though, or how it wold effect the old navigation.

Swiddis: Ah, got it, so it's displaying the URL path instead of the title -- I'm not sure if I like that (since it's what the display name is for and the integ IDs are not necessarily always readable) but if it's inconvenient to change then I guess we can keep it?

Swiddis: I see in the [breadcrumbs docs](https://oui.opensearch.org/1.12/#/navigation/breadcrumbs) that it's possible to have separate `href` and `text`, is that applicable here at all?

Swiddis (approved):

public/plugin.tsx
```diff
@@ -32,6 +32,9 @@ import {
   observabilityApplicationsID,
   observabilityApplicationsPluginOrder,
   observabilityApplicationsTitle,
+  observabilityGettingStartedID,
+  observabilityGettingStartedTitle,
+  observabilityGettingStartedPluginOrder,
```
ps48, line 37: I see this in the notebooks PR too. Can we remove them from here?

sumukhswamy (approved):

public/plugin.tsx
TackAdam: Done with update.

PR #2140, opened 2024-09-06
[MDS] Add support for register data sources during the absence of local cluster

RyanL1997:
### Description
Add support for register plugin during the absence of local cluster

### Issues Resolved
- Resolve the 500 errors when there is no backend plugins to consume the requests

### Check List

### DIFF
--- a/public/plugin.tsx
+++ b/public/plugin.tsx
@@ -186,7 +186,6 @@ export class ObservabilityPlugin
 
     setupOverviewPage(setupDeps.contentManagement!);
     this.featureFlagStatus = !!setupDeps.dataSource;
-    console.log('Feature Flag Status:', this.featureFlagStatus);
 
     // redirect legacy notebooks URL to current URL under observability
     if (window.location.pathname.includes('notebooks-dashboards')) {
--- a/public/plugin.tsx
+++ b/public/plugin.tsx
@@ -169,6 +169,7 @@ export class ObservabilityPlugin
   constructor(initializerContext: PluginInitializerContext) {
     this.config = initializerContext.config.get<PublicConfig>();
   }
+  private featureFlagStatus: boolean = false;
 
   public setup(
     core: CoreSetup<AppPluginStartDependencies>,
@@ -184,6 +185,8 @@ export class ObservabilityPlugin
     });
 
     setupOverviewPage(setupDeps.contentManagement!);
+    this.featureFlagStatus = !!setupDeps.dataSource;
+    console.log('Feature Flag Status:', this.featureFlagStatus);
 
     // redirect legacy notebooks URL to current URL under observability
     if (window.location.pathname.includes('notebooks-dashboards')) {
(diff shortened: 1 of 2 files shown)


public/plugin.tsx
```diff
@@ -184,6 +185,8 @@ export class ObservabilityPlugin
     });
 
     setupOverviewPage(setupDeps.contentManagement!);
+    this.featureFlagStatus = !!setupDeps.dataSource;
+    console.log('Feature Flag Status:', this.featureFlagStatus);
```
sumukhswamy, line 189: console log necessary?

sumukhswamy (approved):

public/plugin.tsx
RyanL1997: good catch. Removed

public/plugin.tsx
```diff
@@ -459,55 +461,57 @@ export class ObservabilityPlugin
...
-                metadata: {
-                  ...s3ds.properties,
-                  ui: {
-                    label: s3ds.name,
-                    typeLabel: getDataSourceTypeLabel(s3ds.connector.toLowerCase()),
-                    groupType: s3ds.connector.toLowerCase(),
-                    selector: {
-                      displayDatasetsAsSource: false,
+    // register all s3 datasources only if mds feature flag is disabled
+    if (!this.featureFlagStatus) {
```
ps48, line 465: can we directly use `setupDeps.dataSource` here?

RyanL1997: we can't access the `setupDeps` from the `start` function, we can only access to the `startDeps`, However, the `startDeps.dataSource.dataSourceEnabled` is also not accessible, when the dataSource plugin is not registered.

ps48: got it. this makes sense for now. Ideally with should stop the registration at setup itself. Adding if condition here: https://github.com/opensearch-project/dashboards-observability/blob/3ad2bf0448f6a705f449f42e62be8398f840ea37/public/plugin.tsx#L264C11-L264C31

ps48 (approved):

public/plugin.tsx
```diff
@@ -169,6 +169,7 @@ export class ObservabilityPlugin
   constructor(initializerContext: PluginInitializerContext) {
     this.config = initializerContext.config.get<PublicConfig>();
   }
+  private featureFlagStatus: boolean = false;
```
joshuali925, line 172: can this variable naming be more explicit? looking at `featureFlagStatus` it's hard to tell what the features is..

RyanL1997: Yes I will change that.

RyanL1997: Since the backport is failing, I will manually change it in 2.x first and backport. I will push another commit for change it on main.

RyanL1997: I have updated here https://github.com/opensearch-project/dashboards-observability/pull/2144.

PR #2154, opened 2024-09-10
fix(log explorer): show PPL errors in toasts correctly

joshuali925:
### Description
This PR has these fixes
- fix(log explorer): set loading state to false if query failed 
   - this is a regression caused by https://github.com/opensearch-project/dashboards-observability/pull/1902
- update fetch call to pass in data source id 
   - data source id is always undefined because log explorer doesn't support multi-datasource. passing undefined fixes the parameters so UI shows PPL errors in toasts
- revert icon to use prepend
   - using `prepend` property the icon will be smaller than its original size, but since the text field is now `compact` and original size doesn't fit anymore, we can revert back to use `prepend`.

### Issues Resolved

### Check List

### DIFF
--- a/public/components/event_analytics/hooks/use_fetch_events.ts
+++ b/public/components/event_analytics/hooks/use_fetch_events.ts
@@ -52,7 +52,7 @@ export const useFetchEvents = ({ pplService, requestParams }: IFetchEventsParams
   ) => {
     setIsEventsLoading(true);
     return pplService
-      .fetch({ query, format }, errorHandler)
+      .fetch({ query, format }, undefined, errorHandler)
       .then((res: any) => handler(res))
       .catch((err: any) => {
         console.error(err);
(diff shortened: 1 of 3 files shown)


public/components/event_analytics/hooks/use_fetch_events.ts
```diff
@@ -52,7 +52,7 @@ export const useFetchEvents = ({ pplService, requestParams }: IFetchEventsParams
   ) => {
     setIsEventsLoading(true);
     return pplService
-      .fetch({ query, format }, errorHandler)
+      .fetch({ query, format }, undefined, errorHandler)
```
Swiddis, line 55: ?

Swiddis (approved):

public/components/event_analytics/hooks/use_fetch_events.ts
joshuali925: it's regression from MDS PR, see description and 7b5cc9f commit message:
> - update fetch call to pass in data source id 
>   - data source id is always undefined because log explorer doesn't support multi-datasource. passing undefined fixes the parameters so UI shows PPL errors in toasts

sejli (approved):

PR #2177, opened 2024-09-20
[Bug] Services Data Picker, UI Fixes

TackAdam:
### Description
1. Bugfix for services data picker where declaring it twice was causing it to not render after closing fly-out and reset the selected data source to default. Also fix data source changing when navigation to traces through fly-out.
2. UI fix to Metrics switched refresh button to icon for super date picker
3. UI fix to overview page changing title of dashboard and adding ui link pop out icon
4. Add URL redirection paths to account for changed functionality. (Comment Below)
5. Add testing for the url re-direction.(Comment Below)
Bugfix:

https://github.com/user-attachments/assets/7909bee3-b474-4845-ba17-609a7ed43071

Metrics page after:
Screenshot 2024-09-19 at 8 58 43 PM

Before:
Screenshot 2024-09-19 at 9 06 50 PM

Overview page after:
Screenshot 2024-09-19 at 8 57 44 PM
Screenshot 2024-09-19 at 9 05 58 PM

Before:
Screenshot 2024-09-19 at 9 06 35 PM

### Check List

### DIFF
--- a/common/constants/shared.ts
+++ b/common/constants/shared.ts
@@ -60,13 +60,13 @@ export const observabilityMetricsTitle = 'Metrics';
 export const observabilityMetricsPluginOrder = 5092;
 
 export const observabilityTracesNewNavID = 'observability-traces-nav';
-export const observabilityTracesNewNavURL = observabilityTracesNewNavID + '#/traces';
+export const observabilityTracesNewNavURL = observabilityTracesNewNavID;
 export const observabilityTracesID = 'observability-traces';
 export const observabilityTracesTitle = 'Traces';
 export const observabilityTracesPluginOrder = 5093;
 
 export const observabilityServicesNewNavID = 'observability-services-nav';
-export const observabilityServicesNewNavURL = observabilityServicesNewNavID + '#/services';
+export const observabilityServicesNewNavURL = observabilityServicesNewNavID;
 export const observabilityServicesID = 'observability-services';
 export const observabilityServicesTitle = 'Services';
 export const observabilityServicesPluginOrder = 5092;
--- a/public/components/overview/components/dashboard_controls.tsx
+++ b/public/components/overview/components/dashboard_controls.tsx
@@ -16,6 +16,7 @@ import {
 import { OnTimeChangeProps } from '@opensearch-project/oui/src/eui_components/date_picker/super_date_picker/super_date_picker';
 import React from 'react';
 import { useObservable } from 'react-use';
+import { FormattedMessage } from '@osd/i18n/react';
 import { coreRefs } from '../../../framework/core_refs';
 import { HOME_CONTENT_AREAS } from '../../../plugin_helpers/plugin_overview';
 import { redirectToDashboards } from '../../getting_started/components/utils';
@@ -58,12 +59,18 @@ export function DashboardControls() {
             </EuiText>
           </EuiFlexItem>
           <EuiFlexItem grow={false}>
-            <EuiLink
-              onClick={() => redirectToDashboards('/view/' + dashboardState?.dashboardId)}
-              external={true}
+            <EuiToolTip
+              content={
+                <FormattedMessage id="dashboard.popout.tooltip" defaultMessage="Go to dashboard" />
+              }
             >
-              <EuiIcon type="popout" />
-            </EuiLink>
+              <EuiLink
+                onClick={() => redirectToDashboards('/view/' + dashboardState?.dashboardId)}
+                external={true}
+              >
+                <EuiIcon type="popout" />
+              </EuiLink>
+            </EuiToolTip>
           </EuiFlexItem>
         </EuiFlexGroup>
       </EuiFlexItem>
--- a/public/components/overview/components/dashboard_controls.tsx
+++ b/public/components/overview/components/dashboard_controls.tsx
@@ -61,7 +61,10 @@ export function DashboardControls() {
           <EuiFlexItem grow={false}>
             <EuiToolTip
               content={
-                <FormattedMessage id="dashboard.popout.tooltip" defaultMessage="Go to dashboard" />
+                <FormattedMessage
+                  id="observability.dashboard.popout.tooltip"
+                  defaultMessage="Go to dashboard"
+                />
               }
             >
               <EuiLink
--- a/public/components/overview/components/dashboard_controls.tsx
+++ b/public/components/overview/components/dashboard_controls.tsx
@@ -7,6 +7,7 @@ import {
   EuiButtonIcon,
   EuiFlexGroup,
   EuiFlexItem,
+  EuiIcon,
   EuiLink,
   EuiSuperDatePicker,
   EuiText,
@@ -48,15 +49,23 @@ export function DashboardControls() {
   };
 
   return isDashboardSelected ? (
-    <EuiFlexGroup gutterSize="s" justifyContent="spaceBetween">
-      <EuiFlexItem grow={true}>
-        <EuiText size="s" className="obsOverviewDashboardHeader">
-          <p>
-            <EuiLink onClick={() => redirectToDashboards('/view/' + dashboardState?.dashboardId)}>
-              {dashboardState?.dashboardTitle}
+    <EuiFlexGroup gutterSize="s" alignItems="center" justifyContent="spaceBetween">
+      <EuiFlexItem grow={false}>
+        <EuiFlexGroup alignItems="center" gutterSize="s">
+          <EuiFlexItem grow={false}>
+            <EuiText size="m" className="obsOverviewDashboardHeader">
+              <h4>{dashboardState?.dashboardTitle}</h4>
+            </EuiText>
+          </EuiFlexItem>
+          <EuiFlexItem grow={false}>
+            <EuiLink
+              onClick={() => redirectToDashboards('/view/' + dashboardState?.dashboardId)}
+              external={true}
+            >
+              <EuiIcon type="popout" />
             </EuiLink>
-          </p>
-        </EuiText>
+          </EuiFlexItem>
+        </EuiFlexGroup>
       </EuiFlexItem>
       <EuiFlexItem grow={false}>
         <EuiFlexGroup justifyContent="flexEnd" alignItems="center" gutterSize="s">
--- a/public/components/trace_analytics/components/common/__tests__/legacy_route_helpers.test.tsx
+++ b/public/components/trace_analytics/components/common/__tests__/legacy_route_helpers.test.tsx
@@ -15,6 +15,24 @@ import {
 import { coreRefs } from '../../../../../framework/core_refs';
 
 describe('ConvertLegacyTraceAnalyticsUrl', () => {
+  let originalLocation: Location;
+
+  beforeEach(() => {
+    originalLocation = window.location;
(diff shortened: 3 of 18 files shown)


public/components/trace_analytics/home.tsx
```diff
@@ -141,20 +150,40 @@ export const Home = (props: HomeProps) => {
...
+            notifications: props.notifications,
+            fullWidth: true,
+            onSelectedDataSources: onSelectedDataSource,
+            dataSourceFilter: dataSourceFilterFn,
+            activeOption: dataSourceMDSId,
+          }}
+        />
+      );
+    }
+  }, [
```
RyanL1997, line 181: Maybe we do something like this:

```suggestion
const sharedProps = {
  setMenuMountPoint: props.setActionMenu,
  componentConfig: {
    activeOption: dataSourceMDSId,
    fullWidth: true,
    dataSourceFilter: dataSourceFilterFn,
  },
};

return dataSourceMenuSelectable ? (
  <DataSourceMenu
    {...sharedProps}
    componentType={'DataSourceSelectable'}
    componentConfig={{
      ...sharedProps.componentConfig,
      savedObjects: props.savedObjectsMDSClient.client,
      notifications: props.notifications,
      onSelectedDataSources: onSelectedDataSource,
    }}
  />
) : (
  <DataSourceMenuView
    {...sharedProps}
    componentType={'DataSourceView'}
  />
);
```

public/components/trace_analytics/home.tsx
```diff
@@ -133,6 +138,10 @@ export const Home = (props: HomeProps) => {
     DataSourceSelectableConfig
   >();
 
+  const DataSourceMenuView = props.dataSourceManagement?.ui?.getDataSourceMenu<
```
RyanL1997, line 141: Just for my knowledge, under what condition the `dataSourceManagement` is optional here?

RyanL1997 (approved):
Hi @TackAdam , thanks for putting this together. Just left some minor comments

public/components/trace_analytics/home.tsx
TackAdam: When data source is disabled. 
#data_source.enabled: false

public/components/trace_analytics/home.tsx
TackAdam: Looks cleaner, thanks applied change!

public/components/trace_analytics/home.tsx
RyanL1997: I think the datasource management is always enabled for now. I think we should check the dataSource plugin instead.

RyanL1997: In other word, the registration of datasource management is not following the MDS feature flag anymore. The dataSource plugin is the one to follow if we would like to check the MDS feature flag status here.

RyanL1997 (changes requested):
> When data source is disabled.
#data_source.enabled: false

~~Hi @TackAdam, If im understanding this correct, we are still relying on dataSource management's registration for checking the MDS feature flag status. Lets change the check from datasource management check into dataSource plugin, since the MDS status is not coming from data source management plugin's registration anymore.~~

RyanL1997 (approved):
Transferring the sync up with @TackAdam here: 

`  const DataSourceMenuView = props.dataSourceManagement?.ui?.getDataSourceMenu<`
The above is just for getting the picker. So I think we can just keep like this for now.

public/plugin.tsx
```diff
@@ -452,6 +455,21 @@ export class ObservabilityPlugin
...
+    // redirect trace URL to current URL under observability
+    // if (window.location.hash.includes('#/traces/')) {
+    //   window.location.assign(convertTraceUrl(window.location));
+    // }
+
+    // // redirect service analytics URL to current URL under observability
+    // if (window.location.hash.includes('#/services/')) {
+    //   window.location.assign(convertServiceUrl(window.location));
+    // }
+
```
ps48, line 467: Can remove commented code if not used.

TackAdam: Removed, missed that one. Thanks

common/constants/shared.ts
```diff
@@ -61,13 +61,13 @@ export const observabilityMetricsTitle = 'Metrics';
...
 export const observabilityTracesNewNavID = 'observability-traces-nav';
-export const observabilityTracesNewNavURL = observabilityTracesNewNavID + '#/traces';
+export const observabilityTracesNewNavURL = observabilityTracesNewNavID;
 export const observabilityTracesID = 'observability-traces';
 export const observabilityTracesTitle = 'Traces';
 export const observabilityTracesPluginOrder = 5093;
 
 export const observabilityServicesNewNavID = 'observability-services-nav';
-export const observabilityServicesNewNavURL = observabilityServicesNewNavID + '#/services';
+export const observabilityServicesNewNavURL = observabilityServicesNewNavID;
```
ps48, line 70: If we are assigning the `observabilityTracesNewNavURL = observabilityTracesNewNavID;` Can we just use the observabilityTracesNewNavID directly?

ps48: nit. Not sure why we have `ID` capitalized should be `Id` for at all places.

public/components/overview/components/dashboard_controls.tsx
```diff
@@ -48,15 +50,29 @@ export function DashboardControls() {
...
+        <EuiFlexGroup alignItems="center" gutterSize="s">
+          <EuiFlexItem grow={false}>
+            <EuiText size="m" className="obsOverviewDashboardHeader">
+              <h4>{dashboardState?.dashboardTitle}</h4>
+            </EuiText>
+          </EuiFlexItem>
+          <EuiFlexItem grow={false}>
+            <EuiToolTip
+              content={
+                <FormattedMessage id="dashboard.popout.tooltip" defaultMessage="Go to dashboard" />
```
ps48, line 64: Should we use `observability.dashboard.popout.tooltip`? dashboard is too generic.

public/components/trace_analytics/components/common/__tests__/legacy_route_helpers.test.tsx
```diff
@@ -0,0 +1,164 @@
...
+  it('should convert legacy URL correctly', () => {
+    const location = {
+      pathname: '/app/trace-analytics-dashboards',
+      hash: '#/traces',
+      search: '?param1=value1',
+    } as Location;
+
+    const result = convertLegacyTraceAnalyticsUrl(location);
+    expect(result).toBe(`/app/${observabilityTracesID}#/traces?param1=value1`);
+  });
```
ps48, line 27: Can we add more test cases for the legacy URL redirection with tracesId and servicename. E.g of lagacy trace URL with traceID: https://playground.opensearch.org/app/trace-analytics-dashboards#/traces/02feb3a4f611abd81f2a53244d1278ae

public/components/trace_analytics/components/services/service_view.tsx
```diff
@@ -118,16 +114,31 @@ export function ServiceView(props: ServiceViewProps) {
...
-            href: `#/services/${encodeURIComponent(props.serviceName)}`,
+            href: (() => {
+              const dataSourceId = props.dataSourceMDSId[0].id;
+              if (dataSourceId && dataSourceId !== '') {
+                return `#/services?datasourceId=${encodeURIComponent(
+                  dataSourceId
+                )}&serviceId=${encodeURIComponent(props.serviceName)}`;
+              } else {
+                return `#/services?serviceId=${encodeURIComponent(props.serviceName)}`;
+              }
```
ps48, line 125: Can this be simplified further?

```
const dataSourceId = props.dataSourceMDSId[0].id;
const url = `#/services?serviceId=${encodeURIComponent(props.serviceName)}`;

if(dataSourceId && dataSourceId !== '')
  return `${url}&datasourceId=${encodeURIComponent(dataSourceId)}`

return  url;

File: public/components/trace_analytics/components/services/service_view.tsx [hunk 8]
```diff
@@ -118,16 +114,31 @@ export function ServiceView(props: ServiceViewProps) {
...
   const redirectToServicePage = (service: string) => {
-    window.location.href = `#/services/${service}`;
+    const dataSourceId = props.dataSourceMDSId[0].id;
+    if (dataSourceId && dataSourceId !== '') {
+      window.location.href = `#/services?datasourceId=${encodeURIComponent(
+        dataSourceId
+      )}&serviceId=${encodeURIComponent(service)}`;
+    } else {
+      window.location.href = `#/services?datasourceId=&serviceId=${encodeURIComponent(service)}`;
+    }
```
ps48, line 141: if we are duplicating code might as well use a function for this.

public/components/trace_analytics/components/common/legacy_route_helpers.ts
```diff
@@ -12,3 +17,72 @@ export const convertLegacyTraceAnalyticsUrl = (location: Location) => {
...
+      } else {
+        window.location.assign(
+          `/app/${observabilityTracesID}#/services?datasourceId=&serviceId=${encodeURIComponent(
+            serviceId
+          )}`
+        );
+      }
+      return;
+    }
+  }
```
ps48, line 46: URL query parameters are not ordered by default. Browsers, servers, and other systems generally treat query parameters as unordered key-value pairs.

```
https://example.com/page?param1=value1&param2=value2
https://example.com/page?param2=value2&param1=value1
```
With that said we don't need to see if MDS is enabled. The url parser on load should default to `dataSource=""` if datasource is not defined.

TackAdam: This check is for if a id is provided. The new navigation check is to handle the different urls used for both
observabilityServicesNewNavID vrs observabilityTracesID
This isn't affected by if MDS is enabled, as the legacy re-direction should only be for local host id's.

common/constants/shared.ts
TackAdam: Looks like all ID's are in this file Ill do a separate PR to fix as it will affect a lot of files and don't want the other changes to get overshadowed by it.

public/components/overview/components/dashboard_controls.tsx
TackAdam: Done

public/components/trace_analytics/components/common/__tests__/legacy_route_helpers.test.tsx
TackAdam: Added four test to cover these conditions for new navigation on and off.

public/components/trace_analytics/components/services/service_view.tsx
TackAdam: Done

public/components/trace_analytics/components/services/service_view.tsx
TackAdam: export const generateServiceUrl = (service: string, dataSourceId: string) => {
  const url = `#/services?serviceId=${encodeURIComponent(service)}`;

  if (dataSourceId && dataSourceId !== '') {
    return `${url}&datasourceId=${encodeURIComponent(dataSourceId)}`;
  }

  return `${url}&datasourceId=`;
};

ps48 (approved):

PR #2187, opened 2024-10-01
Add compile step before Cypress runs in CI

Swiddis:
### Description
The current cypress runs are failing with the error `Timeout for 600 seconds reached. OpenSearch Dashboards did not finish compiling.`. By far, the longest step of OSD startup is compilation. A little poking around upstream reveals that there's a script that can frontload this compilation, so we can separate the server startup from the compile step. This PR applies this, and lowers the timeout since it no longer needs to be so high. It also adds some caching to reduce build times, although there's a bit more that needs to happen to make build times fast (namely, caching all the /target folders, see #2188).

### Issues Resolved
See above

### Check List

### DIFF
--- a/.github/workflows/integration-tests-workflow.yml
+++ b/.github/workflows/integration-tests-workflow.yml
@@ -149,18 +149,7 @@ jobs:
       - name: Run OpenSearch Dashboards server
         run: |
           cd OpenSearch-Dashboards
-          nohup yarn start --no-base-path --no-watch | tee dashboard.log &
-
-      - name: Check If OpenSearch Dashboards Is Ready
-        if: ${{ runner.os == 'Linux' }}
-        run: |
-          cd ./OpenSearch-Dashboards
-          if timeout 60 grep -q "[OpenSearchDashboards][http] http server running at" <(tail -n0 -f dashboard.log); then
-            echo "OpenSearch Dashboards started successfully."
-          else
-            echo "Timeout of 60 seconds reached. OpenSearch Dashboards did not start successfully."
-            exit 1
-          fi
+          nohup yarn start --no-base-path --no-watch &
 
       - name: Install Cypress
         run: |
--- a/.github/workflows/integration-tests-workflow.yml
+++ b/.github/workflows/integration-tests-workflow.yml
@@ -140,7 +140,7 @@ jobs:
         uses: actions/cache@v4
         with:
           path: OpenSearch-Dashboards/**/node_modules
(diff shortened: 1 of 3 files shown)


.github/workflows/integration-tests-workflow.yml
```diff
@@ -150,10 +155,10 @@ jobs:
         if: ${{ runner.os == 'Linux' }}
         run: |
           cd ./OpenSearch-Dashboards
-          if timeout 600 grep -q "bundles compiled successfully after" <(tail -n0 -f dashboard.log); then
-            echo "OpenSearch Dashboards compiled successfully."
+          if timeout 60 grep -q "[OpenSearchDashboards][http] http server running at" <(tail -n0 -f dashboard.log); then
```
RyanL1997, line 158: we only need 1 min for time out here?

RyanL1997 (approved):
Hi @Swiddis , thanks for taking this on! Just left a minor question about the timeout

.github/workflows/integration-tests-workflow.yml
Swiddis: Once compilation is done, startup is really fast (about 5 seconds from a cold start on my machine).

RyanL1997: I see, that make sense since we separated out the compilation step. LGTM!

Swiddis: Ended up needing to refactor this to tail the whole file because the server was starting up _too_ fast, and so `tail` missed the "server running" line.

ps48 (approved):

.github/workflows/integration-tests-workflow.yml
```diff
@@ -150,12 +165,12 @@ jobs:
...
-          if timeout 600 grep -q "bundles compiled successfully after" <(tail -n0 -f dashboard.log); then
-            echo "OpenSearch Dashboards compiled successfully."
+          if timeout 60 grep -q "http server running" <(tail -n +1 -f dashboard.log); then
+            echo "OpenSearch Dashboards started successfully."
           else
-            echo "Timeout for 600 seconds reached. OpenSearch Dashboards did not finish compiling."
+            echo "Timeout of 60 seconds reached. OpenSearch Dashboards did not start successfully."
             exit 1
-          fi
+          fi&
```
joshuali925, line 173: if you put to background it will continue to next step even if `http server running` didn't appear in `dashboard.log` yet, right? and if 60 seconds passed, will `exit 1` be able to abort the workflow if it has already moved to the next step?

PR #2196, opened 2024-10-09
Update workflows to use real build cache

### LINKED ISSUE #2188: [FEATURE] [CI] Cache build artifacts to reduce build times
Coming from #2187, build times are very slow because of time taken to repeatedly compile the same modules. These build artifacts are stored in various `/target` directories around the OSD directory. A quick search shows 4 general glob patterns (relative to `/OpenSearch-Dashboards`):
- `src/core/target`
- `src/plugins/*/target`
- `packages/*/target`
- `plugins/*/target` <-- We are here

Somehow, we should [cache](https://github.com/actions/cache) these build artifacts.

The tricky thing about a global cache that matches all of these directories' source is that, _in aggregate_, these directories change very often, multiple times per day. This means trying to put all of them in one cache is going to be mostly fruitless. But _individually_, these different packages don't get changed that often: lots of the core plugins and packages have been stable for months or sometimes years, and the "hotspots" for code changes drift over time.

Ideally, we would be able to store one separate cache for each of these modules. This isn't supported very well by the cache action, however. I generally see two options in this direction:
- Write a script to code-generate a large cache action YML that will cache all of these directories individually
- Try to write a bash script directly in the action (I tried it, but it didn't work for reasons I haven't fully debugged #2187)

- No caching: what we've been doing so far. It's worked historically, but the build times are continuing to rack up. The last time I tried to cache builds our builds were at 5 minutes, now (just over a year later) they're at 12.
- One mega-cache: may work in principle, just inefficient, and the cache would be missed more often than not (i.e., for every code change). The issue I saw last time from trying this is that eventually so many differences accumulate from the cache that building all the changes ends up taking longer than just a clean build (#965).
- Only cache the core targets (skip `plugins/*`): also works in principle, but only marginally more efficient than the above since core also changes very often.

N/A

Swiddis:
### Description
Following https://github.com/opensearch-project/OpenSearch-Dashboards/pull/8472 and #2188, this PR switches the cache to the actual build target directories to speed up caching dramatically.

### Issues Resolved
Resolves #2188 
Avenges #925 

### Check List

### DIFF
--- a/.github/workflows/dashboards-observability-test-and-build-workflow.yml
+++ b/.github/workflows/dashboards-observability-test-and-build-workflow.yml
@@ -39,23 +39,6 @@ jobs:
         with:
           path: OpenSearch-Dashboards/plugins/dashboards-observability
 
-      - name: Get yarn cache dir
-        id: setup-yarn
-        shell: bash
-        run: |
-          cd ./OpenSearch-Dashboards
-          source $NVM_DIR/nvm.sh
-          nvm use
-          echo "yarn-cache-dir=$(yarn cache dir)" >> $GITHUB_OUTPUT
-
-      - name: Yarn Cache
-        uses: actions/cache@v4
-        with:
-          path: ${{ steps.setup-yarn.outputs.yarn-cache-dir }}
-          key: ${{ runner.OS }}-yarn-${{ hashFiles('OpenSearch-Dashboards/**/yarn.lock') }}
-          restore-keys: |
-            ${{ runner.OS }}-yarn-
-
       - name: Plugin Bootstrap
         uses: nick-fields/retry@v2
         with:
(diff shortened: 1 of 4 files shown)


.github/workflows/dashboards-observability-test-and-build-workflow.yml
```diff
@@ -135,20 +118,6 @@ jobs:
...
       - run: yarn -v
 
-      - name: Get yarn cache dir
-        id: setup-yarn
-        shell: bash
-        run: |
-          echo "yarn-cache-dir=$(yarn cache dir)" >> $GITHUB_OUTPUT
-
-      - name: Yarn Cache
-        uses: actions/cache@v4
```
joshuali925, line 145: why removing yarn cache? do they actually work? i remember trying to set it up long time ago but didn't work for some reason

joshuali925 (approved):

.github/workflows/dashboards-observability-test-and-build-workflow.yml
Swiddis: In the unit tests they have no performance diff still -- the compilation is implicit to yarn test

TackAdam (approved):
LGTM

.github/workflows/dashboards-observability-test-and-build-workflow.yml
joshuali925: there are two caches, npm packages and transpiled/optimized javascript. does cache for npm packages work in this workflow?

Swiddis: You would expect so, but it doesn't look like it. Bootstrap also doesn't have a noticeable speedup by caching packages.

joshuali925: yeah i think i saw similar results when i tried it, not sure why

RyanL1997 (approved):

PR #2230, opened 2024-10-25
Service map updates

TackAdam:
### Description
Update the service table functionality
1. Adjust the focus table to have a popover that list all the options and is searchable.
2. Make the focus on selection enlarge the node following the filter behavior (also clear when reset)
3. Add reset button to the filter that recenters the tree and clears focus
4. Adjust zoomSpeed: 0.5
5. Set max and min zoom using  moveTo function.
6. If max zoom is exceeded recenter the page so the tree does not get lost.
7. Loading state while rendering
8. Darker borders / thicker on each node
9. No movement after rendering
10. Focus now filters to only display connected nodes
11. Consistent diagram using seed for physics generation
12. Bug fixes for handling switching between custom source and data prepper as well as flyout opening.

Filter Focus:

https://github.com/user-attachments/assets/ba04727f-0346-4f08-8235-b6765ebcff05

Zoom constraints:

https://github.com/user-attachments/assets/b5345a43-c14f-4199-a4b1-690596a65475

Dynamic filter:
Screenshot 2024-10-25 at 4 24 04 PM

Testing with 300 nodes:
Screenshot 2024-11-08 at 11 55 09 AM

Testing Dense Nodes
Screenshot 2024-11-08 at 11 54 27 AM

Screenshot 2024-11-08 at 11 54 39 AM

Service flyout:
ServiceFlyout

### Check List


public/components/trace_analytics/components/common/plots/service_map.tsx
```diff
@@ -124,9 +132,21 @@ export function ServiceMap({
...
+      const visibleNodes = items.graph.nodes.map((node) => node.label);
+      const options = Object.keys(serviceMap)
+        .filter((key) => visibleNodes.includes(serviceMap[key].serviceName))
+        .map((key) => ({
+          label: serviceMap[key].serviceName,
+          value: serviceMap[key].serviceName,
+        }));
+      setSelectableOptions(options);
+    }
+  }, [items.graph, serviceMap]);
```
paulstn, line 146: if `items` is undefined or doesn't have the `graph` field inside it, wouldn't this throw an error that would crash the page? i see that `items` is initialized to an empty object, so it seems like this might be a possibility.

public/components/trace_analytics/components/common/plots/service_map.tsx
```diff
@@ -302,11 +420,13 @@ export function ServiceMap({
               <div style={{ position: 'relative' }}>
                 {items?.graph && (
                   <Graph
+                    key={graphKey}
                     graph={items.graph}
                     options={options}
                     events={events}
                     getNetwork={(networkInstance: any) => {
```
paulstn, line 427: does networkInstance have a type that can be set here? making sure that this has a type might also require setZoomLimits to set the input variable type as well, so ideally that function should have a type for networkInstance

public/components/trace_analytics/components/common/plots/service_map.tsx
```diff
@@ -163,6 +185,23 @@ export function ServiceMap({
...
+    const initialPosition = networkInstance.getViewPosition();
+
+    networkInstance.on('zoom', (params) => {
+      const zoomLevel = params.scale;
+
+      if (zoomLevel < 0.25 && zoomLevel < lastZoomLevel) {
+        networkInstance.moveTo({ scale: 0.25, position: initialPosition });
+      } else if (zoomLevel > 1.75) {
+        networkInstance.moveTo({ scale: 1.75 });
+      }
```
paulstn, line 199: Just to get an idea of whats happening here, we're adding a boundary to the scale of the screen so that if these limits are reached, it'll change the scale back, but what does `zoomLevel < lastZoomLevel` do here and why isn't it included in the conditional when we go too big? My first assumption was that it was to allow for the screen to gradually zoom back out, but why wouldn't we want it for `zoomLevel > 1.75`?
Also, why is position set in one condition and not the other, what does it do?

public/components/trace_analytics/components/common/plots/service_map.tsx
```diff
@@ -204,9 +243,27 @@ export function ServiceMap({
...
+        graph: { ...items.graph, nodes: resetNodes },
+      });
+      if (networkInstance) networkInstance.fit(); // Adjust the view if needed
       setInvalid(false);
     } else if (serviceMap[service]) {
       if (!networkInstance) networkInstance = network;
+
+      // Enlarge the focused node and reset others
+      const updatedNodes = items.graph.nodes.map((node) =>
+        node.label === service ? { ...node, size: 30 } : { ...node, size: 15 }
```
paulstn, line 259: just wondering, is there some global or environmental variable that can be used for the size here to get rid of the magic numbers?

public/components/trace_analytics/components/common/plots/service_map.tsx
```diff
@@ -271,16 +328,77 @@ export function ServiceMap({
...
+                      onFocus(''); // Clear node focus when input is cleared
+                    }
+                  }}
+                  isInvalid={query.length > 0 && invalid}
+                  append={
+                    <EuiButtonIcon
+                      iconType="refresh"
+                      size="s"
+                      onClick={() => {
+                        setGraphKey((prevKey) => prevKey + 1);
```
paulstn, line 354: what's the purpose of graph key?

public/components/trace_analytics/components/common/plots/service_map.tsx
TackAdam: Because there are no zoom limits available to bet set we are handling this using the moveTo function. This functionality sets a limit on if the user zooms in to much if they reach the limit it stops it, as a side effect there is a little movement if the user continuously zooms in while at the limit. 
The zoom out limit has the added functionality to reset the node cluster to the center to prevent the user from getting lost. This is only present on the zoom out as we would not want to recenter and lose position if the user accidentally zoomed in to much.

Video example of behavior at limits:

https://github.com/user-attachments/assets/df088c21-196a-4292-a144-207ae5037f2a

public/components/trace_analytics/components/common/plots/service_map.tsx
TackAdam: I think this would require a additional dependency of @types/vis-network

ps48:
Still looking at the code. Just a question in the service flyout when users focus on a node, does the flyout details, other red metrics graphs also change to show the details of this particular node?

TackAdam:
> Still looking at the code. Just a question in the service flyout when users focus on a node, does the flyout details, other red metrics graphs also change to show the details of this particular node?

Adjusted to hide the focus while in the service fly-out. Ideally the user is utilizing the fly-out for a quick overview, and using the main page to investigate. Adjusted the functionality of the main search to allow user to quickly investigate deeper into connected nodes by clicking on them to filter (showcased in first video of description 'Filter Focus:').

public/components/trace_analytics/components/common/plots/service_map.tsx
paulstn: ah that makes sense thanks

paulstn (approved):

public/components/trace_analytics/components/common/plots/service_map_node_details.tsx
```diff
@@ -18,7 +18,7 @@ import { ServiceNodeDetails } from '../../../../../../common/types/trace_analyti
 interface ServiceMapNodeDetailsProps {
   selectedNodeDetails: ServiceNodeDetails | null;
   setSelectedNodeDetails: React.Dispatch<React.SetStateAction<ServiceNodeDetails | null>>;
-  addServiceFilter: (selectedServiceName: string) => void;
+  addServiceFilter: (selectedServiceName: string | null) => void;
```
mengweieric, line 21: are we expecting to have a null passed in for addServiceFilter?

test/constants.ts
```diff
@@ -209,8 +209,11 @@ export const TEST_SERVICE_MAP_GRAPH = {
         label: 'order',
         size: 15,
         title: 'order\n\n Average duration: 90.1ms \n Error rate: 4.17% \n Request rate: 48',
-        borderWidth: 0,
-        color: 'rgba(158, 134, 192, 1)',
+        borderWidth: 3,
```
mengweieric, line 212: question: are these changes being signed off from UX/product?

public/components/trace_analytics/components/common/plots/service_map_node_details.tsx
TackAdam: No, added in the null check to handle switching between custom and data prepper to clear selections and prevent existing bugs where the page would fail to render.

mengweieric (approved):

test/constants.ts
TackAdam: Synched offline for sign off.

PR #2242, opened 2024-11-01
[Bug fix] Metrics datasource

TackAdam:
### Description
While datasource is enabled, set initial condition to null and wait to render page until the selection/default is registered.

Currently if a localhost is not defined the call will trigger the search on a non valid datasource.
NoLocalHost

This change prevents the search call from being triggered until the datasource has been read.

https://github.com/user-attachments/assets/75e8559e-07dc-46d3-b9bf-fe127f83ee65

### Check List

### DIFF
--- a/public/components/metrics/index.tsx
+++ b/public/components/metrics/index.tsx
@@ -55,7 +55,7 @@ export const Home = ({
   // Side bar constants
   const [selectedDataSource, setSelectedDataSource] = useState<OptionType[]>([]);
   const [selectedOTIndex, setSelectedOTIndex] = useState([]);
-  const [dataSourceMDSId, setDataSourceMDSId] = useState<string>('');
+  const [dataSourceMDSId, setDataSourceMDSId] = useState<string | null>(null);
   const [reloadSidebar, setReloadSidebar] = useState<boolean>(false);
 
   useEffect(() => {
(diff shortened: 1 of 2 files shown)


public/components/metrics/sidebar/sidebar.tsx
```diff
@@ -95,6 +97,11 @@ export const Sidebar = ({
   }, [selectedMetrics, selectedMetricsIds, dataSourceMDSId]);
 
   useEffect(() => {
+    // If data source is enabled but is still null prevent the invalid call
+    if (dataSourceEnabled && dataSourceMDSId === null) {
```
joshuali925, line 101: if local cluster is selected, what would `dataSourceMDSId` be?

joshuali925 (approved):

public/components/metrics/sidebar/sidebar.tsx
TackAdam: Local cluster is just an empty string "" when selected it get sets and is no longer null

ps48 (approved):

public/components/metrics/index.tsx
```diff
@@ -77,7 +77,7 @@ export const Home = ({
   const dispatch = useDispatch();
 
   const onSelectedDataSource = async (dataSources: DataSourceOption[]) => {
-    const id = dataSources[0] ? dataSources[0].id : '';
+    const id = dataSources[0] ? dataSources[0].id : null;
```
ps48, line 80: Only one question: What's the condition when `dataSources[0]` will be undefined and we have to set the id as `null`?

TackAdam: The only condition this should be triggered is when a workspace is created and its datasource is removed putting it in a state where it has no data sources available at all. In this case metrics won't work and we don't want to display options. 
We should probably have an empty state for this such as " No connected data source with redirection to associate one".

PR #2264, opened 2024-11-27
[Services map] Update redirection/Focus field rework

TackAdam:
### Demo:

https://github.com/user-attachments/assets/0230c798-8524-4ccc-850d-696d3b83e9ef

### Description
Updated the focus field to no longer apply a filter, it now just renders the connected services and allows the user to investigate using the focus window. The drop down now reflects the connected viewable services to switch to. The reset icon now clears the existing filter if it matches the focus field and is currently applied. The focus field is now disabled while a filter of an existing service is active.

https://github.com/user-attachments/assets/e14b2a20-f04c-4d14-95fa-9a3924392717

Clicking a service currently causes the title clicked to be set as a filter.

https://github.com/user-attachments/assets/0acd9974-5ac6-4dc6-a5e7-976dce758732

Updated behavior now re-directs the user to the corresponding service view page

https://github.com/user-attachments/assets/cd27d3ca-88c6-4866-bf3f-4dcc4a5e0524

### Check List


public/components/trace_analytics/components/services/services_table.tsx
```diff
@@ -65,20 +68,15 @@ export function ServicesTable(props: ServicesTableProps) {
...
 
   const nameColumnAction = (serviceName: string) => {
-    addFilter({
-      field: mode === 'jaeger' ? 'process.serviceName' : 'serviceName',
-      operator: 'is',
-      value: serviceName,
-      inverted: false,
-      disabled: false,
-    });
+    window.location.href = generateServiceUrl(serviceName, dataSourceMDSId[0].id);
```
ps48, line 79: The change looks good to me. Can you please add cypress tests and / or unit tests for this? Thanks!

TackAdam: Added jest test 
it('redirects to the correct URL when the service link is clicked',

public/components/trace_analytics/components/services/service_view.tsx
```diff
@@ -76,6 +77,18 @@ export function ServiceView(props: ServiceViewProps) {
...
 
+  let serviceId: string | null = null;
+
+  try {
+    const location = useLocation();
+    const params = new URLSearchParams(location?.search || '');
+    serviceId = params.get('serviceId');
+  } catch (error) {
+    serviceId = null;
+  }
```
ps48, line 88: Can we use hooks and state here to make it cleaner?

```
  const location = useLocation();

  useEffect(() => {
    try {
      const params = new URLSearchParams(location.search);
      const id = params.get('serviceId');
      setServiceId(id);
    } catch (error) {
      setServiceId(null);
    }
  }, [location]);
  ```

File: public/components/trace_analytics/components/services/services_table.tsx (hunk 1 above)
Comment (ps48) [inline public/components/trace_analytics/components/services/services_table.tsx:79]: Have we test the `generateServiceUrl` with workspaces enabled?

Comment (TackAdam) [reply public/components/trace_analytics/components/services/services_table.tsx:79]: Yes tested with enabled and disabled. Video demo above is in a workspace.

File: public/components/trace_analytics/components/services/service_view.tsx (hunk 2 above)
Comment (TackAdam) [reply public/components/trace_analytics/components/services/service_view.tsx]: Done. Thanks for the feedback =)

File: public/components/trace_analytics/components/common/plots/service_map.tsx [hunk 3]
```diff
@@ -81,6 +83,10 @@ export function ServiceMap({
   filterByCurrService?: boolean;
   includeMetricsCallback?: () => void;
   mode?: string;
+  filters: FilterType[];
+  setFilters:
+    | React.Dispatch<React.SetStateAction<FilterType[]>>
+    | ((filters: FilterType[]) => void);
```
joshuali925, line 89: curious why are both needed?

public/components/trace_analytics/components/common/plots/service_map.tsx
```diff
@@ -328,33 +291,36 @@ export function ServiceMap({
...
-          disabled: false,
-        });
+      if (focusedService !== service) {
+        const filteredGraph = getServiceMapGraph(
+          serviceMap,
+          idSelected,
+          ticks,
+          service,
+          serviceMap[service]?.relatedServices,
+          true // Enable filtering to focus on connected nodes
```
joshuali925, line 319: i didn't look into `getServiceMapGraph`, but seems `service` and `serviceMap[service]?.relatedServices` are used for filtering, and they will be absent if we don't want filtering. Then is this boolean required?

public/components/trace_analytics/components/common/plots/service_map.tsx
```diff
@@ -404,14 +366,16 @@ export function ServiceMap({
...
         idSelected,
         calculatedTicks,
-        currService,
+        focusedService ?? currService,
         serviceMap[currService!]?.relatedServices,
-        filterByCurrService
+        showRelatedServices
       )
     );
   }, [serviceMap, idSelected]);
```
joshuali925, line 381: do we need `focusedService` and `filterByCurrService` in dependency array?

public/components/trace_analytics/components/services/service_view.tsx
```diff
@@ -76,6 +77,18 @@ export function ServiceView(props: ServiceViewProps) {
   const [redirect, setRedirect] = useState(false);
   const [actionsMenuPopover, setActionsMenuPopover] = useState(false);
 
+  let serviceId: string | null = null;
+
+  try {
+    const location = useLocation();
```
joshuali925, line 83: > Location is causing the service view page to not render as it is not defined, so it needs to be within the "Try".

do you have more details on what the error is? is it happening when calling `useLocation`?

public/components/trace_analytics/components/services/service_view.tsx
```diff
@@ -76,6 +77,18 @@ export function ServiceView(props: ServiceViewProps) {
   const [redirect, setRedirect] = useState(false);
   const [actionsMenuPopover, setActionsMenuPopover] = useState(false);
 
+  let serviceId: string | null = null;
```
joshuali925, line 80: should this be a state?

public/components/trace_analytics/components/common/plots/service_map.tsx
TackAdam: The focus functionality is not applying a filter so for the new experience these are required to change the options that show up using the focus functionality.

Without:

https://github.com/user-attachments/assets/d560e8d1-83c6-4c98-ad40-6b1024dce7f5

With: 

https://github.com/user-attachments/assets/b5fa8fef-eab6-416d-bf5c-b53cfc80607c

public/components/trace_analytics/components/common/plots/service_map.tsx
TackAdam: I don't believe so

https://github.com/user-attachments/assets/d97625b4-69e0-4390-a828-a088c6d682da

public/components/trace_analytics/components/services/service_view.tsx
ps48: ➕

public/components/trace_analytics/components/services/__tests__/service_view.test.tsx
```diff
@@ -7,39 +7,51 @@ import { configure, shallow } from 'enzyme';
 import Adapter from 'enzyme-adapter-react-16';
 import React from 'react';
 import { ServiceView } from '..';
-import { coreStartMock } from '../../../../../../test/__mocks__/coreMocks';
+
+jest.mock('../../../../../../test/__mocks__/coreMocks', () => ({
+  coreStartMock: {
+    chrome: { setBreadcrumbs: jest.fn() },
+    http: { post: jest.fn() },
```
ps48, line 14: Can we please add mocks with some actual data (both empty and non-empty response)?

public/components/trace_analytics/components/common/plots/service_map.tsx
```diff
@@ -443,35 +407,57 @@ export function ServiceMap({
...
+                            if (focusedService) {
+                              setGraphKey((prevKey) => prevKey + 1);
+                              setQuery('');
+                              onFocus(focusedService);
+                            } else {
+                              setGraphKey((prevKey) => prevKey + 1);
+                              setQuery('');
+                              onFocus('');
+                            }
+                          }
```
ps48, line 440: May be move repeated line outside this `if`:

```suggestion
                          if (newValue === '') {
                           setGraphKey((prevKey) => prevKey + 1);
                           setQuery('');
                           focusedService ?  onFocus(focusedService) : onFocus('');
                          }
```

public/components/trace_analytics/components/common/plots/__tests__/service_map.test.tsx
```diff
@@ -3,25 +3,243 @@
...
+// Mock crypto.getRandomValues
+const crypto = {
+  getRandomValues: jest.fn((arr) => arr.fill(0)), // Fill with consistent values
+};
+Object.defineProperty(global, 'crypto', { value: crypto });
+
+const mockContext = ({
+  canvas: document.createElement('canvas'),
+  fillRect: jest.fn(),
+  clearRect: jest.fn(),
```
ps48, line 38: bigger mocks can be moved to `test/contants`

public/components/trace_analytics/components/common/plots/service_map.tsx
TackAdam: Thought i needed to cover other conditions, removed it and seems to work fine.
Thanks

public/components/trace_analytics/components/services/service_view.tsx
TackAdam: Resolved with moving the flyout component into router.

public/components/trace_analytics/components/services/service_view.tsx
TackAdam: Changed to a state.

public/components/trace_analytics/components/common/plots/__tests__/service_map.test.tsx
TackAdam: Moved to MOCK_CANVAS_CONTEXT in constants file

public/components/trace_analytics/components/common/plots/service_map.tsx
TackAdam: Changed too     
if (newValue === '') {
                            setGraphKey((prevKey) => prevKey + 1);
                            setQuery('');
                            onFocus(focusedService || '');
                          }

public/components/trace_analytics/components/services/__tests__/service_view.test.tsx
TackAdam: Cypress added for service map 
Tests focus functionality in Service map
Testing the re-renders and filtering

Added 
import toJson from 'enzyme-to-json';
To service map test

TackAdam:
Updated applications to no longer filter on name click. Instead it opens the fly-out for details. Also removed the actions column when in the applications page.

https://github.com/user-attachments/assets/d6ac64f5-a936-4119-9d9f-baed40821e1f

ps48:
> Updated applications to no longer filter on name click. Instead it opens the fly-out for details. Also removed the actions column when in the applications page.
> Screen.Recording.2024-12-06.at.11.25.15.AM.mov

Does the Applications trace navigation work as expected?

public/components/trace_analytics/components/common/plots/service_map.tsx
ps48: Thanks!

TackAdam:
> > Updated applications to no longer filter on name click. Instead it opens the fly-out for details. Also removed the actions column when in the applications page.
> > Screen.Recording.2024-12-06.at.11.25.15.AM.mov
> 
> Does the Applications trace navigation work as expected?

Clicking Trace applies the filter and switches to the Traces and Spans  tab.



https://github.com/user-attachments/assets/bd1e19f5-5995-4f6f-a2c1-fa82b72a935e

public/components/trace_analytics/components/common/plots/__tests__/service_map.test.tsx
```diff
@@ -98,9 +99,18 @@ describe('ServiceMap Component', () => {
     jest.clearAllMocks();
   });
 
-  it('renders service map component', () => {
+  it('renders service map component', async () => {
     const wrapper = mount(<ServiceMap {...defaultProps} />);
-    expect(wrapper).toMatchSnapshot();
+    wrapper.update();
+
```
ps48, line 105: Should we wait for the canvas to appear, so that map renders in the snapshot?

ps48:
Also, looks like this test `1) "before each" hook for "Verifies tree view expand arrow functionality"` is flaky we should solve this immediately in a follow up PR.

ps48:
I see one failure in test because the nodes changes position. We do have a set seed which makes sure the graph has nodes on the position. Need to check what's causing the change cc: @TackAdam

TackAdam:
> I see one failure in test because the nodes changes position. We do have a set seed which makes sure the graph has nodes on the position. Need to check what's causing the change cc: @TackAdam

The position is consistent but it seems cypress environment used here might be causing the physics to behave slightly different than we see testing locally on original layout. As the point of this test is to ensure the canvas is rendering the nodes my solution is to first focus on a node to cause less to be rendered and ensure that it matches the layout we see locally. This still ensures the nodes are displayed and are intractable. Ran multiple times and seems to be working with my fix.

.cypress/integration/trace_analytics_test/trace_analytics_services.spec.js
```diff
@@ -149,7 +149,7 @@ describe('Testing service view', () => {
     cy.contains('Spans (1)').should('exist');
   });
 });
-
+//adam
```
ps48, line 152: Should remove this.

.cypress/integration/trace_analytics_test/trace_analytics_services.spec.js
```diff
@@ -171,7 +172,75 @@ describe('Testing Service map', () => {
...
+      .and('include', 'touch-action: none')
+      .and('include', 'user-select: none')
+      .and('include', 'width: 100%')
+      .and('include', 'height: 100%');
+
+    cy.get('.vis-network canvas').should('have.attr', 'width').and('not.eq', '0');
+    cy.get('.vis-network canvas').should('have.attr', 'height').and('not.eq', '0');
+  });
+
+  it.only('Click on a node to see the details', () => {
```
ps48, line 196: Should remove `only` here, this is causing only 1 test to run: https://github.com/opensearch-project/dashboards-observability/actions/runs/12216555045/job/34080417940#step:26:118

.cypress/integration/trace_analytics_test/trace_analytics_services.spec.js
```diff
@@ -171,7 +172,75 @@ describe('Testing Service map', () => {
...
+    // Focus on "order" by selecting the first option
+    cy.get('.euiFormLabel.euiFormControlLayout__prepend').contains('Focus on').should('exist');
+    cy.get('[placeholder="Service name"]').click();
+    cy.get('.euiSelectableList__list li').eq(0).click();
+
+    // ensure rendering is complete before node click, replace wait
+    cy.get('[data-text="Errors"]').click();
+    cy.contains('60%');
+    cy.get('[data-text="Duration"]').click();
+    cy.contains('100');
```
ps48, line 209: I have two question on this: 

1) We usually follow one test per it. Looks like we are mixing the focus test and the node click to view details test. This helps us to pin point the issue in the code quickly. Should we keep them independent?

2) Without focus, Why is the click on node with the `x, y` coordinates not working? Is there a possibility in our current code that the graph rendering can change shape on page refresh?

TackAdam: 1. I don't think this can be avoided for this particular test as we need the behavior to work locally and on this test unit. If the canvas fails to render the nodes, having a focus will also have the same error so this test still serves its purpose. If focus was broken then this would also fail but the error would still be identifiable.

2. Without focus it is failing because the environment cypress is using is applying the seed differently. (Below screenshot showing its layout vrs the local). The click fails because there is no node at the location in the layout that the cypress version non local is running. The service map will always be the same for the user regardless of refresh. My change works as cypress and local have the same layout.

Local:
LocalTest

Cypress (Taken from recording of the failure):
CypressTest

Local and Cypress with Focus
Screenshot 2024-12-07 at 4 21 29 PM

ps48: > Without focus it is failing because the environment cypress is using is applying the seed differently. 

Are we assuming that the cypress issue won't be replicated on user's end?  We need to understand why is cypress seeing a different graph.

ps48 (approved):

public/components/trace_analytics/components/common/plots/service_map.tsx
paulstn: We should include every dependency in the dependency array even if it seems to pointlessly re-render as to avoid 'stale' states that might lead to bugs that would be difficult to fix.
React useEffect doc: https://react.dev/reference/react/useEffect

joshuali925 (approved):

paulstn (approved):

PR #2298, opened 2024-12-21
Traces custom source - Bug Fixes

TackAdam:
### Description
1. Handle a bug that was causing a page crash if the filter was selected twice.
2. Now removes the filter when clicking.
3. Set a default metric when page loading for custom source. (Was defaulting to no selection)
4. Handle detection of start time by distinguishing between mili and nano seconds based on the sort field length.
5. Added toast messages for 503 returns from traces and services request handlers so the UI reflects issues.
6. Added trace-mode to the url so that new tabs and re-directs/favorites will work. (Previously it defaulted to data_prepper causing jaeger and custom_data_prepper links to break as session storage was not available in the new tabs)

Before: (Bug crashing page when a selected filter was clicked again)

https://github.com/user-attachments/assets/807c112e-fc2e-4ee0-b8c0-bb96f4473bf0

After: (No longer crashes, now removes the filter when clicked again)

https://github.com/user-attachments/assets/91b9afd2-7ee0-4b76-8d44-bde8e101cb6a

Before: (When the custom traces sources used date for the start time mapping instead of date_nanos the times would all start at 0)
Before1

After: (Now the time conversion is determined by the length of the sort metric. This example uses the mapping of start time "date", the indexes themselves have some in mili seconds and some in nanos but because the mapping of "date" is used the nanosecond precision is lost.)
After1

Toast message example (manually setting buckets to 1 to cause failure)

https://github.com/user-attachments/assets/eceb6e4d-4a24-4530-a0b8-de3bbf1e246c

URL redirection working:

https://github.com/user-attachments/assets/e544e915-0135-4a4b-8b92-e06a381d8196

### Check List

### DIFF
--- a/public/components/trace_analytics/components/common/helper_functions.tsx
+++ b/public/components/trace_analytics/components/common/helper_functions.tsx
@@ -140,6 +140,27 @@ export const getTimestampPrecision = (timestamp: number): 'millis' | 'micros' | 
   }
 };
 
+export const appendModeToTraceViewUri = (
+  traceId: string,
+  getTraceViewUri?: (traceId: string) => string,
+  traceMode?: string | null
+): string => {
+  const baseUri = getTraceViewUri ? getTraceViewUri(traceId) : '';
+  const isHashRouter = baseUri.includes('#');
+  const separator = isHashRouter
+    ? baseUri.includes('?')
+      ? '&'
+      : '?'
+    : baseUri.includes('?')
+    ? '&'
+    : '?';
+
+  if (traceMode) {
+    return `${baseUri}${separator}mode=${encodeURIComponent(traceMode)}`;
+  }
+  return baseUri;
+};
+
 export function microToMilliSec(micro: number) {
   if (typeof micro !== 'number') return 0;
   return micro / 1000;
--- a/public/components/trace_analytics/components/common/helper_functions.tsx
+++ b/public/components/trace_analytics/components/common/helper_functions.tsx
@@ -122,7 +122,6 @@ export function nanoToMilliSec(nano: number) {
 export const getTimestampPrecision = (timestamp: number): 'millis' | 'micros' | 'nanos' => {
   if (!timestamp) return 'millis';
 
-  // Convert to string and get total length
   const digitCount = timestamp.toString().length;
 
   // For Unix timestamps:
--- a/public/components/trace_analytics/components/common/helper_functions.tsx
+++ b/public/components/trace_analytics/components/common/helper_functions.tsx
@@ -146,14 +146,7 @@ export const appendModeToTraceViewUri = (
   traceMode?: string | null
 ): string => {
   const baseUri = getTraceViewUri ? getTraceViewUri(traceId) : '';
-  const isHashRouter = baseUri.includes('#');
-  const separator = isHashRouter
-    ? baseUri.includes('?')
-      ? '&'
-      : '?'
-    : baseUri.includes('?')
-    ? '&'
-    : '?';
+  const separator = baseUri.includes('?') ? '&' : '?';
 
   if (traceMode) {
     return `${baseUri}${separator}mode=${encodeURIComponent(traceMode)}`;
--- a/public/components/trace_analytics/components/common/helper_functions.tsx
+++ b/public/components/trace_analytics/components/common/helper_functions.tsx
@@ -119,6 +119,28 @@ export function nanoToMilliSec(nano: number) {
   return nano / 1000000;
 }
 
+export const getTimestampPrecision = (timestamp: number): 'millis' | 'micros' | 'nanos' => {
+  if (!timestamp) return 'millis';
+
+  // Convert to string and get total length
+  const digitCount = timestamp.toString().length;
+
+  // For Unix timestamps:
+  // 13 digits = milliseconds (e.g., 1703599200000)
+  // 16 digits = microseconds (e.g., 1703599200000000)
+  // 19 digits = nanoseconds  (e.g., 1703599200000000000)
+  switch (digitCount) {
+    case 13:
+      return 'millis';
+    case 16:
+      return 'micros';
+    case 19:
+      return 'nanos';
+    default:
+      return 'millis';
+  }
+};
+
 export function microToMilliSec(micro: number) {
   if (typeof micro !== 'number') return 0;
   return micro / 1000;
--- a/public/components/trace_analytics/components/common/plots/service_map.tsx
+++ b/public/components/trace_analytics/components/common/plots/service_map.tsx
@@ -149,7 +149,7 @@ export function ServiceMap({
         includeMetricsCallback();
       }
     }
-  }, [selectableValue, setIdSelected]);
+  }, []);
 
   const removeFilter = (field: string, value: string) => {
     if (!setFilters) return;
--- a/public/components/trace_analytics/components/common/plots/service_map.tsx
+++ b/public/components/trace_analytics/components/common/plots/service_map.tsx
@@ -123,17 +123,6 @@ export function ServiceMap({
     );
   }, [filters, focusedService]);
 
-  const onChangeSelectable = (value: React.SetStateAction<Array<EuiSuperSelectOption<any>>>) => {
-    // if the change is changing for the first time then callback servicemap with metrics
-    if (selectableValue.length === 0 && value.length !== 0) {
-      if (includeMetricsCallback) {
-        includeMetricsCallback();
-      }
-    }
-    setIdSelected(value);
-    setSelectableValue(value);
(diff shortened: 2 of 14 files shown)


public/components/trace_analytics/components/common/plots/service_map.tsx
```diff
@@ -149,6 +138,19 @@ export function ServiceMap({
     },
   ];
 
+  // For the traces custom page
+  useEffect(() => {
+    if (!selectableValue || selectableValue.length === 0) {
+      // Set to the first option ("latency") and trigger the onChange function
+      const defaultValue = metricOptions[0].value;
+      setSelectableValue(defaultValue); // Update the state
+      setIdSelected(defaultValue); // Propagate the default to parent
```
joshuali925, line 147: what's the purpose of this useEffect? it's setting state for idSelected and selectableValue, but it also depends on them. When these states are updated, wouldn't it trigger useEffect again?

public/components/trace_analytics/components/dashboard/mode_picker.tsx
```diff
@@ -62,6 +62,17 @@ export function DataSourcePicker(props: {
     );
   };
 
+  const updateUrlWithMode = (key: TraceAnalyticsMode) => {
+    const currentUrl = window.location.href.split('?')[0];
+    const queryParams = new URLSearchParams(window.location.search);
```
joshuali925, line 67: this will only get the standard search params, but not the search param we use with hash router. for example in playground https://playground.opensearch.org/app/observability-traces#/traces?datasourceId=44cc4c50-6f80-11ed-a8ec-958f39714b93

`window.location.search` is `''`, and `currentUrl` is `'https://playground.opensearch.org/app/observability-traces#/traces'`. The `dataSourceId` information becomes lost?

public/components/trace_analytics/components/traces/services_list.tsx
```diff
@@ -58,14 +80,14 @@ export const ServicesList = ({
   );
 
   useEffect(() => {
+    // Update selectable options based on the current filtered service
     setOptions(
-      Object.keys(serviceMap).map((key) => {
-        return filteredService === key
-          ? { label: key, checked: 'on', bordered: false }
-          : { label: key, bordered: false };
```
joshuali925, line 65: we don't need `bordered`?

public/components/trace_analytics/components/traces/trace_table_helpers.tsx
```diff
@@ -73,28 +73,44 @@ export const getTableColumns = (
...
-          {(copy) => (
-            <EuiButtonIcon aria-label="Copy trace id" iconType="copyClipboard" onClick={copy} />
-          )}
-        </EuiCopy>
-      </EuiFlexItem>
-    </EuiFlexGroup>
-  );
+  const renderTraceLinkField = (item: string) => {
+    // Extract the current mode from the URL or session storage
+    const queryParams = new URLSearchParams(window.location.search);
```
joshuali925, line 78: does this work correctly with hash router?

public/components/trace_analytics/components/traces/traces_table.tsx
```diff
@@ -54,6 +54,18 @@ export function TracesTable(props: TracesTableProps) {
...
+    const traceMode = queryParams.get('mode') || sessionStorage.getItem('TraceAnalyticsMode');
+
+    const updatedGetTraceViewUri = (traceId: string) => {
+      const baseUri = getTraceViewUri ? getTraceViewUri(traceId) : '';
+      if (traceMode) {
+        const separator = baseUri.includes('?') ? '&' : '?';
+        return `${baseUri}${separator}mode=${encodeURIComponent(traceMode)}`;
+      }
+      return baseUri;
+    };
```
joshuali925, line 67: this logic is being duplicated, can it be extracted? might also need unit tests

public/components/trace_analytics/requests/traces_request_handler.ts
```diff
@@ -141,6 +152,13 @@ export const handleTracesRequest = async (
       const percentileRanges =
         percentileRangesResult.status === 'fulfilled' ? percentileRangesResult.value : {};
       const response = responseResult.value;
+
+      // Check if the mode hasn't been set first
+      // if (mode === 'jaeger' && !response?.aggregations?.service_type?.buckets) {
+      //   console.warn('No traces or aggregations found.');
+      //   return [];
+      // }
```
joshuali925, line 160: is this needed?

public/components/trace_analytics/components/traces/trace_table_helpers.tsx
```diff
@@ -73,28 +73,44 @@ export const getTableColumns = (
...
+    const traceMode = queryParams.get('mode') || sessionStorage.getItem('TraceAnalyticsMode');
+
+    // Update the href to include the mode if available
+    const updatedGetTraceViewUri = (traceId: string) => {
```
joshuali925, line 82: "update get trace view uri" is not very clear, maybe
```suggestion
    const appendModeToTraceViewUri = (traceId: string) => {
```

public/components/trace_analytics/components/traces/services_list.tsx
TackAdam: I think this setting is over-written and not needed. Checked with it set to true, false, and not included with no reflection in the UI.

public/components/trace_analytics/components/traces/trace_table_helpers.tsx
TackAdam: Updated too

```
const currentUrl = window.location.href;
const traceMode = new URLSearchParams(currentUrl.split('?')[1]).get('mode') || sessionStorage.getItem('TraceAnalyticsMode');
```

```
export const appendModeToTraceViewUri = (
  traceId: string,
  getTraceViewUri?: (traceId: string) => string,
  traceMode?: string | null
): string => {
  const baseUri = getTraceViewUri ? getTraceViewUri(traceId) : '';
  const isHashRouter = baseUri.includes('#');
  const separator = isHashRouter ? (baseUri.includes('?') ? '&' : '?') : (baseUri.includes('?') ? '&' : '?');

  if (traceMode) {
    return `${baseUri}${separator}mode=${encodeURIComponent(traceMode)}`;
  }
  return baseUri;
};
```

public/components/trace_analytics/requests/traces_request_handler.ts
TackAdam: Removed, became unnecessary after the url mode implementation.

public/components/trace_analytics/components/traces/trace_table_helpers.tsx
TackAdam: Updated, thank you.

public/components/trace_analytics/home.tsx
```diff
@@ -431,7 +432,7 @@ export const Home = (props: HomeProps) => {
                   chrome={props.chrome}
                   http={props.http}
                   traceId={decodeURIComponent(traceId)}
-                  mode={mode}
+                  mode={((traceMode as unknown) as TraceAnalyticsMode) || mode}
```
paulstn, line 435: Why can the mode in TraceView be potentially different from the mode used throughout this page? It looks like the state used throughout this page is set by sessionStorage and then traceMode here is from the url params?

TackAdam: Added it to the other calls. Thanks for noticing this (should be in all calls). 
Essentially if the url contains mode= we know the user is navigation through a shared link or opening a new tab. In these cases we want this to take precedence over the session storage that could potentially be empty or using the default value.

paulstn: Sounds good!

paulstn (approved):

public/components/trace_analytics/components/common/helper_functions.tsx
```diff
@@ -140,6 +140,27 @@ export const getTimestampPrecision = (timestamp: number): 'millis' | 'micros' |
...
+): string => {
+  const baseUri = getTraceViewUri ? getTraceViewUri(traceId) : '';
+  const isHashRouter = baseUri.includes('#');
+  const separator = isHashRouter
+    ? baseUri.includes('?')
+      ? '&'
+      : '?'
+    : baseUri.includes('?')
+    ? '&'
+    : '?';
```
joshuali925, line 156: nit. is this just `const separator = baseUri.includes('?') ? '&'
      : '?'

joshuali925 (approved):

public/components/trace_analytics/components/common/helper_functions.tsx
TackAdam: Fixed thank you.

PR #2310, opened 2025-01-13
Traces/Services - Query optimization / UI setting / Bugfix

TackAdam:
### Description
1. Remove the all_services aggregation field from getRelatedServiceQuery. The only purpose seemed to be in displaying the “greyed-out” connection in service map. Earlier code changes removed the full map being displayed with this grey-out functionality in preference of the “Focus on” search box being implemented. This field being presented seemed to require a 4x increase in the number of buckets.
* Removed the shading logic from getServiceMapGraph as we no longer enter that state based on related services.
2. Remove the traceGroupName aggregation field from getServiceNodesQuery. This appeared to be an old artifact and was not used in any visualizations.
3. Remove the toast messages after the first check of handleServiceMapRequest as they were indicating false positives as the map took time to render.
4. Add "TRACE_CUSTOM_MODE_DEFAULT_SETTING" option for users to enable it as the default landing page. Update the fly-out for custom source to include the option to enable this. When the mode is not presented and this setting is enabled it will direct the user to correct mode.
5. Update the error field for custom spans under the non "traces" field. It was originally using the >0 check used for error_count but this field uses status.code which is only an error when it is equal to 2. https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#set-status
* Added renderStatusCodeErrorsField() as it's behavior is different for custom traces checking the spans and needs to do the ===2 check similiar to https://github.com/opensearch-project/dashboards-observability/blob/main/public/components/trace_analytics/requests/traces_request_handler.ts#L403
6. Update the Services table link, and Expand view buttons to append the trace mode to allow it to be used as a favorite or shared link.
7. Fix a bug that was causing infinite refreshes if MDS was disabled and the service page - traces link was clicked as the datasource was appended as undefined.
8. Fix flaky cypress test in services by adding wait condition.

Before: (with all_services) 
```
PUT _cluster/settings
{ "persistent": { "search.max_buckets": 450 } }
```

https://github.com/user-attachments/assets/da44ef8d-0bcb-49c8-bdf7-7038118ccbdb

Required 460 buckets to avoid errors.
After: (removed all_services)

https://github.com/user-attachments/assets/c6596f12-4a96-410d-8d15-cd8059cc07c1

Required just 110 buckets to avoid errors.

TRACE_CUSTOM_MODE_DEFAULT_SETTING:
Screenshot 2025-01-14 at 2 00 22 PM
Screenshot 2025-01-14 at 2 00 47 PM
Example usage:

https://github.com/user-attachments/assets/c2fef915-3498-4fcf-9996-7ff7e68299e6

Service page infinite refresh while MDS disabled bug Before:

https://github.com/user-attachments/assets/2f30a84f-71f0-42d2-b8bb-b9263789c089

After (fixed):

https://github.com/user-attachments/assets/5a91fc38-26cf-4fbf-8d9a-db6e97544c39

### Issues Resolved

### Check List

### DIFF
--- a/public/components/trace_analytics/components/common/__tests__/helper_functions.test.tsx
+++ b/public/components/trace_analytics/components/common/__tests__/helper_functions.test.tsx
@@ -93,17 +93,6 @@ describe('Trace analytics helper functions', () => {
     expect(serviceMapGraph).toEqual(TEST_SERVICE_MAP_GRAPH);
   });
 
-  it('extracts correct target resources for a given service from its traceGroups', () => {
-    const traceGroups = TEST_SERVICE_MAP.order.traceGroups;
-
-    // Extract all target resources from the traceGroups
-    const targetResources = [].concat(...traceGroups.map((group) => group.targetResource));
-
-    // Verify the extracted resources match the expected list
-    const expectedResources = ['clear_order', 'update_order', 'get_order', 'pay_order'];
-    expect(targetResources).toEqual(expectedResources);
-  });
-
   it('calculates ticks', () => {
     const ticks = calculateTicks(500, 200);
     const ticks2 = calculateTicks(0, 200, 10);
--- a/public/components/trace_analytics/components/common/__tests__/helper_functions.test.tsx
+++ b/public/components/trace_analytics/components/common/__tests__/helper_functions.test.tsx
@@ -82,14 +82,11 @@ describe('Trace analytics helper functions', () => {
   });
 
   it('returns service map graph', () => {
-    const serviceMapGraph = getServiceMapGraph(TEST_SERVICE_MAP, 'latency', [
-      0,
-      50,
-      100,
-      150,
-      200,
-      250,
-    ]);
+    const serviceMapGraph = getServiceMapGraph({
+      map: TEST_SERVICE_MAP,
+      idSelected: 'latency',
+      ticks: [0, 50, 100, 150, 200, 250],
+    });
     expect(serviceMapGraph).toEqual(TEST_SERVICE_MAP_GRAPH);
   });
 
--- a/public/components/trace_analytics/components/common/__tests__/helper_functions.test.tsx
+++ b/public/components/trace_analytics/components/common/__tests__/helper_functions.test.tsx
@@ -32,7 +32,7 @@ import {
   renderBenchmark,
 } from '../helper_functions';
 
-describe('Helper functions', () => {
+describe('Trace analytics helper functions', () => {
   configure({ adapter: new Adapter() });
 
   it('renders panel title', () => {
@@ -93,10 +93,15 @@ describe('Helper functions', () => {
     expect(serviceMapGraph).toEqual(TEST_SERVICE_MAP_GRAPH);
   });
 
-  it('returns target resources for a service from the traceGroups', () => {
+  it('extracts correct target resources for a given service from its traceGroups', () => {
     const traceGroups = TEST_SERVICE_MAP.order.traceGroups;
+
+    // Extract all target resources from the traceGroups
     const targetResources = [].concat(...traceGroups.map((group) => group.targetResource));
-    expect(targetResources).toEqual(['clear_order', 'update_order', 'get_order', 'pay_order']);
+
+    // Verify the extracted resources match the expected list
+    const expectedResources = ['clear_order', 'update_order', 'get_order', 'pay_order'];
+    expect(targetResources).toEqual(expectedResources);
   });
 
   it('calculates ticks', () => {
--- a/public/components/trace_analytics/components/common/__tests__/helper_functions.test.tsx
+++ b/public/components/trace_analytics/components/common/__tests__/helper_functions.test.tsx
@@ -22,7 +22,6 @@ import {
   getPercentileFilter,
   getServiceMapGraph,
   getServiceMapScaleColor,
-  getServiceMapTargetResources,
   getTimestampPrecision,
   milliToNanoSec,
   minFixedInterval,
@@ -94,8 +93,9 @@ describe('Helper functions', () => {
     expect(serviceMapGraph).toEqual(TEST_SERVICE_MAP_GRAPH);
   });
 
-  it('returns target resources by service name', () => {
-    const targetResources = getServiceMapTargetResources(TEST_SERVICE_MAP, 'order');
+  it('returns target resources for a service from the traceGroups', () => {
+    const traceGroups = TEST_SERVICE_MAP.order.traceGroups;
+    const targetResources = [].concat(...traceGroups.map((group) => group.targetResource));
     expect(targetResources).toEqual(['clear_order', 'update_order', 'get_order', 'pay_order']);
   });
 
--- a/public/components/trace_analytics/components/common/custom_index_flyout.tsx
+++ b/public/components/trace_analytics/components/common/custom_index_flyout.tsx
@@ -5,6 +5,7 @@
 
 import {
   EuiCallOut,
+  EuiCheckbox,
   EuiCompressedFieldText,
   EuiDescribedFormGroup,
   EuiFlexGroup,
@@ -23,6 +24,7 @@ import React, { Fragment, useEffect, useState } from 'react';
 import {
   TRACE_CUSTOM_SERVICE_INDEX_SETTING,
   TRACE_CUSTOM_SPAN_INDEX_SETTING,
+  TRACE_CUSTOM_MODE_DEFAULT_SETTING,
 } from '../../../../../common/constants/trace_analytics';
 import { uiSettingsService } from '../../../../../common/utils';
 import { useToast } from '../../../common/toast';
@@ -39,6 +41,7 @@ export const CustomIndexFlyout = ({
   const { setToast } = useToast();
   const [spanIndices, setSpanIndices] = useState('');
   const [serviceIndices, setServiceIndices] = useState('');
+  const [customModeDefault, setCustomModeDefault] = useState(false);
   const [isLoading, setIsLoading] = useState(false);
 
   const onChangeSpanIndices = (e: { target: { value: React.SetStateAction<string> } }) => {
@@ -49,21 +52,27 @@ export const CustomIndexFlyout = ({
     setServiceIndices(e.target.value);
   };
 
+  const onToggleCustomModeDefault = (e: { target: { checked: boolean } }) => {
+    setCustomModeDefault(e.target.checked);
+  };
+
   useEffect(() => {
     setSpanIndices(uiSettingsService.get(TRACE_CUSTOM_SPAN_INDEX_SETTING));
     setServiceIndices(uiSettingsService.get(TRACE_CUSTOM_SERVICE_INDEX_SETTING));
+    setCustomModeDefault(uiSettingsService.get(TRACE_CUSTOM_MODE_DEFAULT_SETTING) || false);
   }, [uiSettingsService]);
 
-  const onSaveIndices = async () => {
+  const onSaveSettings = async () => {
     try {
       setIsLoading(true);
       await uiSettingsService.set(TRACE_CUSTOM_SPAN_INDEX_SETTING, spanIndices);
       await uiSettingsService.set(TRACE_CUSTOM_SERVICE_INDEX_SETTING, serviceIndices);
+      await uiSettingsService.set(TRACE_CUSTOM_MODE_DEFAULT_SETTING, customModeDefault);
       setIsLoading(false);
-      setToast('Updated trace analytics sources successfully', 'success');
+      setToast('Updated trace analytics settings successfully', 'success');
     } catch (error) {
       console.error(error);
-      setToast('Failed to update trace analytics sources', 'danger');
+      setToast('Failed to update trace analytics settings', 'danger');
     }
     setIsLoading(false);
   };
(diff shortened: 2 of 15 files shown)


public/components/trace_analytics/components/common/custom_index_flyout.tsx
```diff
@@ -134,6 +143,23 @@ export const CustomIndexFlyout = ({
               />
             </EuiFormRow>
           </EuiDescribedFormGroup>
+          <EuiDescribedFormGroup
+            title={<h3>Set default mode</h3>}
+            description={
+              <Fragment>
+                Enable this to set &quot;Custom source&quot; as the default mode for trace analytics
```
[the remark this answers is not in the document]
TackAdam: This fly-out is only accessible from custom sources. The user would have the context of this allowing them to pick their custom span / custom indices from the options above. This would just be allowing a user explicitly using the traces service for this experimental feature to return to the page without the default data-prepper option being selected. 
PM feedback has cleared the description so going to hold off on changing that.

Screenshot 2025-01-16 at 10 41 28 AM

public/components/trace_analytics/components/common/helper_functions.tsx
```diff
@@ -615,14 +600,25 @@ export const getServiceIndices = (mode: TraceAnalyticsMode) => {
...
 
-export const generateServiceUrl = (service: string, dataSourceId: string) => {
-  const url = `#/services?serviceId=${encodeURIComponent(service)}`;
+export const generateServiceUrl = (
+  service: string,
+  dataSourceId: string,
+  mode?: TraceAnalyticsMode
+): string => {
+  // Construct the base URL with the serviceId
+  let url = `#/services?serviceId=${encodeURIComponent(service)}`;
```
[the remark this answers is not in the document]
TackAdam: Will look into this as a follow-up as we are using encodeURIComponent in a lot of places that would need the change as well.

public/components/trace_analytics/components/traces/trace_table_helpers.tsx
```diff
@@ -115,6 +115,17 @@ export const getTableColumns = (
       'No'
     );
 
+  const renderStatusCodeErrorsField = (item: number) =>
+    item === undefined || item === null ? (
+      '-'
+    ) : item === 2 ? ( // Error status code is 2
```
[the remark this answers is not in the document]
RyanL1997: +1

public/components/trace_analytics/home.tsx
```diff
@@ -259,6 +263,32 @@ export const Home = (props: HomeProps) => {
...
+  useEffect(() => {
+    const urlParts = window.location.href.split('?');
+    const queryParams =
+      urlParts.length > 1 ? new URLSearchParams(urlParts[1].split('#')[0]) : new URLSearchParams();
+
+    const urlMode = queryParams.get('mode');
+    const isCustomModeEnabled = uiSettingsService.get(TRACE_CUSTOM_MODE_DEFAULT_SETTING) || false;
+
+    if (!urlMode && isCustomModeEnabled) {
+      // Add `mode=custom_data_prepper` to the URL
```
RyanL1997, line 275: could you elaborate more on this? Are we getting this from the URL first then put this into the state/session storage?

public/components/trace_analytics/home.tsx
```diff
@@ -295,12 +325,21 @@ export const Home = (props: HomeProps) => {
 
   const traceColumnAction = () => {
     const tracesPath = '#/traces';
+    const dataSourceId = dataSourceMDSId[0]?.id || ''; // Default to empty string if undefined
```
RyanL1997, line 328: nit: we can keep the same comment format.

public/components/trace_analytics/home.tsx
RyanL1997: If im understanding this correct, we can extract some of the logic into helper functions like this
```typescript
const updateUrlWithMode = (mode: string) => {
  const urlParts = window.location.href.split('?');
  const queryParams = new URLSearchParams(urlParts[1]?.split('#')[0] || '');
  queryParams.set('mode', mode);
  return `${urlParts[0]}?${queryParams.toString()}${
    urlParts[1]?.includes('#') ? `#${urlParts[1].split('#')[1]}` : ''
  }`;
};

const updateSessionStorageMode = (mode: string) => {
  sessionStorage.setItem('TraceAnalyticsMode', mode);
};
```
Then, in the `useEffect()`,
```typescript
useEffect(() => {
  const urlParts = window.location.href.split('?');
  const queryParams = new URLSearchParams(urlParts[1]?.split('#')[0] || '');
  const urlMode = queryParams.get('mode');
  const isCustomModeEnabled = uiSettingsService.get(TRACE_CUSTOM_MODE_DEFAULT_SETTING) || false;

  if (!urlMode && isCustomModeEnabled) {
    const newUrl = updateUrlWithMode('custom_data_prepper');
    window.history.replaceState(null, '', newUrl);
    setMode('custom_data_prepper');
    updateSessionStorageMode('custom_data_prepper');
  } else if (isValidTraceAnalyticsMode(urlMode)) {
    setMode(urlMode as TraceAnalyticsMode);
    updateSessionStorageMode(urlMode!);
  }
}, []);
```

RyanL1997 (commented):
Hi @TackAdam , thanks for putting this together. And I just left some comments.

public/components/trace_analytics/home.tsx
RyanL1997: And also for these well-known design, I think we can remove these comments

public/components/trace_analytics/components/traces/trace_table_helpers.tsx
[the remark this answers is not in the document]
TackAdam: Removed redundant comment.
Following the implementation of https://github.com/opensearch-project/dashboards-observability/blob/main/public/components/trace_analytics/requests/traces_request_handler.ts#L403 the error code is defined from https://github.com/open-telemetry/opentelemetry-proto/blob/724e427879e3d2bae2edc0218fff06e37b9eb46e/opentelemetry/proto/trace/v1/trace.proto#L275

public/components/trace_analytics/components/common/__tests__/helper_functions.test.tsx
```diff
@@ -94,8 +93,9 @@ describe('Helper functions', () => {
     expect(serviceMapGraph).toEqual(TEST_SERVICE_MAP_GRAPH);
   });
 
-  it('returns target resources by service name', () => {
-    const targetResources = getServiceMapTargetResources(TEST_SERVICE_MAP, 'order');
+  it('returns target resources for a service from the traceGroups', () => {
+    const traceGroups = TEST_SERVICE_MAP.order.traceGroups;
+    const targetResources = [].concat(...traceGroups.map((group) => group.targetResource));
     expect(targetResources).toEqual(['clear_order', 'update_order', 'get_order', 'pay_order']);
```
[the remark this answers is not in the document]
TackAdam: Updated test describe to "Trace analytics helper functions" for more clarity
Commented the test for more clarity and update it part to be 
'extracts correct target resources for a given service from its traceGroups'

server/plugin_helper/register_settings.ts
```diff
@@ -32,4 +33,15 @@ export const registerObservabilityUISettings = (uiSettings: UiSettingsServiceSet
...
     },
   });
+
+  uiSettings.register({
+    [TRACE_CUSTOM_MODE_DEFAULT_SETTING]: {
+      name: 'Trace analytics custom mode default',
+      value: false,
+      category: ['Observability'],
+      description:
+        '<strong>Experimental feature:</strong> Enable this to default to "custom_data_prepper" mode in the trace analytics plugin',
```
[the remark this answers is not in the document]
TackAdam: Added i18n to all name and descriptions for register settings.

public/components/trace_analytics/home.tsx
TackAdam: Moved the updateUrlWithMode() functionality above the useEffect for calrity.

This setting should only be invoked if the mode parameter is not present, this is so that if a favorited or shared link is provided such as "http://localhost:5601/app/observability-traces-nav#/traces?mode=data_prepper" the user should still land on the page with the mode set to data_prepper and not custom_data_prepper regardless of if they have the custom setting enabled.

The use case for this is that a user choses to enable this when they navigate to traces to begin working it lands with custom_data_prepper as the setting. With this enabled though we still want to support them going to the other modes and being able to refresh the page.

public/components/trace_analytics/home.tsx
TackAdam: Removed, thanks for feedback!

public/components/trace_analytics/home.tsx
TackAdam: I don't think the helper function for 
```
sessionStorage.setItem('TraceAnalyticsMode', 'custom_data_prepper');
```
is necessary as it is one line and is descriptive in the actions. I think the helper might just make it more confusing as the setItem is used in multiple places in the file.

Swiddis (approved):

public/components/trace_analytics/components/common/helper_functions.tsx
```diff
@@ -229,14 +229,10 @@ export function getServiceMapGraph(
         borderWidth: 3,
         color: {
           border: '#4A4A4A',
-          background:
-            relatedServices!.indexOf(service) >= 0 ? `rgba(${color}, 1)` : `rgba(${color}, 0.2)`,
```
joshuali925, line 233: since it's not using `relatedServices` anymore, remove the `relatedServices` in the `getServiceMapGraph()` parameter and where it's passed in to `getServiceMapGraph()`?

public/components/trace_analytics/components/common/__tests__/helper_functions.test.tsx
```diff
@@ -94,9 +93,15 @@ describe('Helper functions', () => {
...
+  it('extracts correct target resources for a given service from its traceGroups', () => {
+    const traceGroups = TEST_SERVICE_MAP.order.traceGroups;
+
+    // Extract all target resources from the traceGroups
+    const targetResources = [].concat(...traceGroups.map((group) => group.targetResource));
+
+    // Verify the extracted resources match the expected list
+    const expectedResources = ['clear_order', 'update_order', 'get_order', 'pay_order'];
+    expect(targetResources).toEqual(expectedResources);
   });
```
joshuali925, line 105: remove this test? it's not really testing anything

public/components/trace_analytics/components/traces/trace_table_helpers.tsx
```diff
@@ -115,6 +115,17 @@ export const getTableColumns = (
       'No'
     );
 
+  const renderStatusCodeErrorsField = (item: number) =>
+    item === undefined || item === null ? (
+      '-'
+    ) : item === 2 ? (
```
joshuali925, line 121: probably add a comment to avoid confusion in the future, https://github.com/opensearch-project/data-prepper/blob/41e3227e8d79c2fd271ecf1aa92d48be807bf53f/docs/schemas/trace-analytics/otel-v1-apm-span-index-template.md#fields and status.code === 2 means Error, 1 means OK, 0 means Unset

public/components/trace_analytics/requests/queries/services_queries.ts
```diff
@@ -225,24 +208,9 @@ export const getServiceMetricsQuery = (
   map: ServiceObject,
   mode: TraceAnalyticsMode
 ) => {
-  const traceGroupFilter = new Set(
-    DSL?.query?.bool.must
-      .filter((must: any) => must.term?.['traceGroup'])
-      .map((must: any) => must.term.traceGroup) || []
+  const targetResource = [].concat(
```
joshuali925, line 211: this logic no longer consider traceGroup filters in DSL anymore, does it change behavior of service metrics when filtering by trace group?

seems it would always consider all targetResources instead of the subset of targetResources that belong to services with traceGroups matching the user's traceGroup filter, which should cause a difference?

I'm ok if the difference it makes is not meaningful for users, just want to check if it's what we want

RyanL1997 (approved):
LGTM. Once we address @joshuali925 's comment we can proceed to merge

public/components/trace_analytics/components/common/helper_functions.tsx
TackAdam: Removed

public/components/trace_analytics/components/common/__tests__/helper_functions.test.tsx
TackAdam: Removed

public/components/trace_analytics/components/traces/trace_table_helpers.tsx
TackAdam: Added comment

joshuali925 (approved):
also CI has been failing?

TackAdam:
> also CI has been failing?

Added in 
```
interface ServiceMapParams {
  map: ServiceObject;
  idSelected: 'latency' | 'error_rate' | 'throughput';
  ticks: number[];
  currService?: string;
  filterByCurrService?: boolean;
}
```
To clean up the code readability when removing the related service param. Had to adjust the test as well I missed in last commit. Should pass now, will wait for run to complete.

public/components/trace_analytics/components/common/helper_functions.tsx
TackAdam: Removed

Issue #2334, opened 2025-02-03
[WIP] Trace Query & Page Load Optimization


ps48:
## **1. Overview**

This document outlines the proposed optimizations for trace queries and page load performance in OpenSearch Dashboards. The goal is to enhance query efficiency, reduce data load times, and improve the user experience (UX) for trace and service views.

## **2. Pages Impacted**

* **Trace View -** details of a single trace with span gantt charts and filtered service map
* **Trace Content -** overview of all the spans and traces in systems for the given time frame
* **Service View -** details of a particular service and it’s trends based on the selected time-frame
* **Service Content -** overview of all the services for the given time frame with a service map

### **3.1 List of Queries**

* Identify all the existing queries running on the affected pages.
* Categorize them by purpose and usage.

### **3.2 Query Initiators**

* Determine which components or user interactions trigger queries.
* Assess whether some queries can be optimized or eliminated.

### **3.3 Context of Queries**

* Identify where each query is used in the UI.
* Determine the semantic meaning of the queries for the user.

@ps48 &  @TackAdam  to add further details on investigations done on services and traces pages 

### **4.1 OpenSearch Index Optimization**

* Evaluate current storage structure.
* Optimize storage for scan queries and aggregation queries.
* Look into star tree structure for traces index and try to see if aggregations on services and tracegroups can be accelerated.
* https://opensearch.org/docs/latest/search-plugins/star-tree-index/

### **4.2 Query Optimization**

* Combine redundant queries to minimize requests.
* Identify slow queries and optimize their execution.

### **4.3 Pagination Implementation**

* Implement pagination for trace tables, span tables, and service tables.
* Reduce the volume of data retrieved per request.

### **4.4 Managing Multiple Queries**

* Optimize queries from server-side and initiators.
* Improve React state management to handle fewer redundant queries.

### **4.5 Data Load Reduction**

* Reduce the number of records retrieved per page.
* Optimize the backend response structure to minimize payload size.

### **4.6 UX Enhancements**

* Modify the UI to start with filtered views rather than full data loads.
* Provide users with quick access to relevant data through pre-applied filters.

### **4.7 Benchmarking and Testing**

* Benchmark OpenSearch performance with **180GB data per cluster**.
* Measure query response times and assess the impact of optimizations.
* Actions items:
    * Query benchmarking with OS benchmark - https://opensearch.org/docs/latest/benchmark/user-guide/working-with-workloads/creating-custom-workloads/
[...]

PR #2349, opened 2025-02-06
TraceView - Optimization of queries 

TackAdam:
### Description
1. Remove the query request for handleTraceViewRequest(), handleServicesPieChartRequest(), handleSpansGanttRequest() by utilizing the primary payload and parsing it on the UI level. This reduces the number of queries when moving from Trace content  to a specific trace went from 4 to 1 (Service map still generating 4 calls pending optimization).

* Testing on a OpenSearch cluster with 200gb of spans indicated that the overview query call was causing a 2x delay to retrieve the metrics already available in payload. Average time reduced by 50% for trace view page for non-cached data.

2. Remove handleSpansRequest() from Span List and Tree view to prevent unnecessary queries when switching between Timeline and table views.
* Span_detail_table will maintain a handleSpansRequest() to cover application analytics pending its depreciation as it does not have a payload request to use the new functionality.

3. Fixed a bug where Jaeger overview was not being populated with information through edge case handiling in the helper function getOverviewFields()

4. Fixed Jaeger Tree Span table not working

5. Add sorting to Jaeger Span Table.

6. Fix gantt chart displaying errors. The time text we added to the right was overriding the error field. We now check if it has an error and display that as priority over the time.

Queries Before 8 request + 1 per table change

https://github.com/user-attachments/assets/8a5264e6-282b-4abe-b2a6-433981adde61

Queries after 5 request + 0 per table change

https://github.com/user-attachments/assets/9cab519b-385d-47e4-8ef5-229dd4f85017

Jaeger Before:
JaegerBefore

Jaeger After:
Screenshot 2025-02-10 at 11 56 29 AM

Gantt chart error display fix:

hasErrorFix

### Issues Resolved
https://github.com/opensearch-project/dashboards-observability/issues/2334 - Trace View section

### Check List


public/components/trace_analytics/components/traces/trace_view_helpers.tsx
```diff
@@ -0,0 +1,128 @@
...
+    let computedLatency: number | null = null;
+    if (firstSpan.traceGroupFields && firstSpan.traceGroupFields.durationInNanos != null) {
+      computedLatency = Number(firstSpan.traceGroupFields.durationInNanos) / 1e6;
+    } else if (firstSpan.startTime && firstSpan.endTime) {
+      computedLatency = moment(firstSpan.endTime).diff(
+        moment(firstSpan.startTime),
+        'milliseconds',
+        true
+      );
+    }
```
ps48, line 47: Based on the community feedback, should we be showing a note or a caution icon, saying the values fall back to top-level values when the traceGroupFields is null?

public/components/trace_analytics/components/traces/trace_view_helpers.tsx
```diff
@@ -0,0 +1,128 @@
...
+
+export function normalizePayload(parsed: any): any[] {
+  if (Array.isArray(parsed)) {
+    return parsed;
+  }
+  if (parsed.hits && Array.isArray(parsed.hits.hits)) {
+    return parsed.hits.hits;
+  }
+  return [];
+}
```
ps48, line 16: Why do we need this function to `normalizePayload`? Since we are using just 1 request for all trace related details shouldn't we normalize this at source where we receive the request?

public/components/trace_analytics/components/traces/trace_view_helpers.tsx
```diff
@@ -0,0 +1,128 @@
...
+    '#c33d69',
+    '#2ea597',
+    '#8456ce',
+    '#e07941',
+    '#3759ce',
+    '#ce567c',
+    '#9469d6',
+    '#4066df',
+    '#da7596',
+  ];
```
ps48, line 108: Should move the colors to constants.

public/components/trace_analytics/components/traces/trace_view_helpers.tsx
```diff
@@ -0,0 +1,128 @@
...
+
+  const totalLatency = serviceBreakdownArray.reduce((sum, [, value]) => sum + value, 0);
```
ps48, line 110: We are looping on this multiple times. Could we merge these? 
* `hits.forEach((hit)` in line 76
* `.reduce((sum,` in line 110

public/components/trace_analytics/components/traces/trace_view_helpers.tsx
```diff
@@ -0,0 +1,128 @@
...
+      );
+    }
+
+    const computedLastUpdated =
+      firstSpan.traceGroupFields && firstSpan.traceGroupFields.endTime
+        ? firstSpan.traceGroupFields.endTime
+        : firstSpan.endTime || null;
+
+    const errorCount =
+      firstSpan.traceGroupFields && firstSpan.traceGroupFields.statusCode != null
```
ps48, line 55: Should we also replace ` firstSpan.traceGroupFields.endTime` with get(firstSpan, 'traceGroupFields.endTime'). We can also use vanilla js to reduce the Lodash usage and go with `firstSpan['traceGroupFields.endTime'] || null

ps48 (commented):
In general, since we are moving lot of computation from OpenSearch to browser we need to manually do some sanity checks, to see if the calculation before and after is the same. May be use our sanity data and check the values in 2.18 (playground) and now with this PR in your local.

public/components/trace_analytics/components/traces/__tests__/trace_view_helpers.test.tsx
```diff
@@ -0,0 +1,140 @@
...
+      {
+        _source: {
+          traceID: 'abc',
+          operationName: 'opA',
+          startTime: 1000,
+          duration: 2000,
+          tag: { error: false },
+          process: { serviceName: 'serviceX' },
+        },
+        sort: [1000 * 1000],
```
ps48, line 58: what is the sort field here?

public/components/trace_analytics/components/traces/span_detail_table.tsx
```diff
@@ -192,10 +195,53 @@ export function SpanDetailTable(props: SpanDetailTableProps) {
...
+
+      const hitsArray =
+        parsed.hits && Array.isArray(parsed.hits.hits)
+          ? parsed.hits.hits
+          : Array.isArray(parsed)
+          ? parsed
+          : [];
+
+      if (hitsArray.length === 0) {
+        console.warn('No hits found in payloadData');
```
ps48, line 221: Why are we doing this, the response should be the same as the request doesn't change.
```
const hitsArray =
        parsed.hits && Array.isArray(parsed.hits.hits)
          ? parsed.hits.hits
          : Array.isArray(parsed)
          ? parsed
          : [];
 ```
Also, Do we need `console.warn`?

File: public/components/trace_analytics/components/traces/span_detail_table.tsx [hunk 8]
```diff
@@ -253,32 +299,42 @@ export function SpanDetailTable(props: SpanDetailTableProps) {
...
-  }, [DSL, http, mode, dataSourceMDSId]);
+    if (!props.payloadData) return;
+    try {
+      const parsed = JSON.parse(props.payloadData);
+      const hitsArray =
+        parsed.hits && Array.isArray(parsed.hits.hits)
+          ? parsed.hits.hits
+          : Array.isArray(parsed)
+          ? parsed
+          : [];
```
ps48, line 317: same here

public/components/trace_analytics/components/traces/span_detail_panel.tsx
```diff
@@ -135,72 +153,83 @@ export function SpanDetailPanel(props: {
...
+      let hits: any[] = [];
+      if (parsed.hits && Array.isArray(parsed.hits.hits)) {
+        hits = parsed.hits.hits;
+      } else if (Array.isArray(parsed)) {
+        hits = parsed;
+      } else {
+        console.warn('Unexpected payload format:', parsed);
+        return;
+      }
+
```
ps48, line 173: Usually, we'd want to keep useEffect lean and put the logic inside a function.

public/components/trace_analytics/components/traces/span_detail_panel.tsx
```diff
@@ -135,72 +153,83 @@ export function SpanDetailPanel(props: {
...
+      hits = hits.map((hit) => {
+        if (mode === 'jaeger') {
+          if (!hit.sort || !hit.sort[0]) {
```
ps48, line 176: Why are we doing this check? hits won't have sort field by default?

public/components/trace_analytics/components/traces/span_detail_panel.tsx
```diff
@@ -15,18 +15,34 @@ import {
...
+  if (!match) {
+    throw new Error(`Invalid ISO timestamp: ${iso}`);
+  }
+  // Parse the base part using Date.parse (which gives ms)
+  const baseMs = new Date(match[1] + 'Z').getTime();
+  // Get the fractional part (if any), pad to 9 digits for nanosecond precision
+  let fraction = match[2] || '0';
+  fraction = fraction.padEnd(9, '0'); // ensure it has 9 digits
+  return baseMs * 1e6 + Number(fraction);
+}
```
ps48, line 45: May be move this to helper function?

public/components/trace_analytics/components/traces/trace_view_helpers.tsx
TackAdam: LatencyWarning
Added icon for if a fallback value is used
"Latency may not be accurate due to missing or unexpected duration data."

public/components/trace_analytics/components/traces/trace_view_helpers.tsx
TackAdam: Moved to traces_request_handler
and used in handlePayloadRequest
```
const normalizedData = normalizePayload(response);
setPayloadData(JSON.stringify(normalizedData, null, 2));
```

public/components/trace_analytics/components/traces/trace_view_helpers.tsx
TackAdam: Created constant file public/components/trace_analytics/components/common/constants.tsx
And moved color there.

public/components/trace_analytics/components/traces/trace_view_helpers.tsx
TackAdam: Updated too 
```
    let computedLatency: number | null = null;
    const tgDuration = get(firstSpan, 'traceGroupFields.durationInNanos');
    if (firstSpan.traceGroupFields && firstSpan.traceGroupFields.durationInNanos != null) {
      computedLatency = Number(firstSpan.traceGroupFields.durationInNanos) / NANOS_TO_MS;
    } else if (tgDuration != null) {
      computedLatency = Number(tgDuration) / NANOS_TO_MS;
    } else if (firstSpan.durationInNanos != null) {
      computedLatency = Number(firstSpan.durationInNanos) / NANOS_TO_MS;
      fallbackValueUsed = true;
    } else if (firstSpan.startTime && firstSpan.endTime) {
      computedLatency = moment(firstSpan.endTime).diff(
        moment(firstSpan.startTime),
        'milliseconds',
        true
      );
      fallbackValueUsed = true;
    }

    const computedLastUpdated = get(firstSpan, 'traceGroupFields.endTime') || firstSpan.endTime || null;
    const tgStatus = get(firstSpan, 'traceGroupFields.statusCode');
    const errorCount = tgStatus != null ? (Number(tgStatus) !== 0 ? 1 : 0) : 0;

```

public/components/trace_analytics/components/traces/__tests__/trace_view_helpers.test.tsx
TackAdam: Removed from test file

public/components/trace_analytics/components/traces/span_detail_table.tsx
TackAdam: Removed console.warn 
I think the fallback to prevent page crashing is helpful although the request will not be changing. Just an extra safe-guard

public/components/trace_analytics/components/traces/span_detail_panel.tsx
TackAdam: Moved too public/components/trace_analytics/components/common/helper_functions.tsx

TackAdam:
> In general, since we are moving lot of computation from OpenSearch to browser we need to manually do some sanity checks, to see if the calculation before and after is the same. May be use our sanity data and check the values in 2.18 (playground) and now with this PR in your local.

Comparing results
| NewCode | Playground 2.18 |
| ------- | --------------- |
| Before Image 1 | After Image 1 |
| Before Image 2 | After Image 2 |
| Before Image 3 | After Image 3 |
| Before Image 4 | After Image 4 |
| Before Image 5 | After Image 5 |

public/components/trace_analytics/components/traces/trace_view_helpers.tsx
```diff
@@ -0,0 +1,128 @@
...
+    return parsed.hits.hits;
+  }
+  return [];
+}
+
+export function getOverviewFields(parsed: any, mode: string) {
+  const hits = normalizePayload(parsed);
+  if (hits.length === 0) return null;
+  const firstSpan = hits[0]._source;
+  if (!firstSpan) return null;
```
ps48: Can we add a small check here, that first span doesn't have a `parentTraceId`? To account for cases we the OTel data might miss parent traces.

public/components/trace_analytics/components/traces/span_detail_panel.tsx
```diff
@@ -48,16 +49,16 @@ export function SpanDetailPanel(props: {
   const [spanFilters, setSpanFilters] = useState<Array<{ field: string; value: any }>>(
     storedFilters ? JSON.parse(storedFilters) : []
   );
-  const [DSL, setDSL] = useState<any>({});
+  const [DSL, _setDSL] = useState<any>({});
```
joshuali925, line 52: remove this? since it's no longer using DSL request

public/components/trace_analytics/components/traces/span_detail_panel.tsx
```diff
@@ -135,72 +135,98 @@ export function SpanDetailPanel(props: {
...
-            },
-          };
-    spanFilters.map(({ field, value }) => {
-      if (value != null) {
-        spanDSL.query.bool.filter.push({
-          term: {
-            [field]: value,
-          },
+        } else {
+          if (!hit.sort || !hit.sort[0]) {
```
joshuali925, line 164: nit. since this if is the same for both modes, it might be better to check `traceMode` inside this if

public/components/trace_analytics/components/traces/span_detail_panel.tsx
```diff
@@ -135,72 +135,98 @@ export function SpanDetailPanel(props: {
...
+            return {
+              ...hit,
+              sort: [parseIsoToNano(hit._source.startTime)],
+            };
+          }
+        }
+        return hit;
+      });
+
+      hits.sort((a, b) => b.sort[0] - a.sort[0]);
```
joshuali925, line 174: do we use `hit.sort` anywhere else? if not then probably just do `hits.sort((a, b) => b._source.startTime - a._source.startTime);` here and remove the inserting sort logic above

public/components/trace_analytics/components/traces/span_detail_panel.tsx
```diff
@@ -135,72 +135,98 @@ export function SpanDetailPanel(props: {
...
+      if (payloadSpanFilters.length > 0) {
+        hits = hits.filter((hit) => {
+          return payloadSpanFilters.every(({ field, value }) => {
+            let fieldVal;
+            if (traceMode === 'jaeger' && field.startsWith('process.')) {
+              fieldVal = hit._source?.process?.[field.split('.')[1]];
+            } else {
+              fieldVal = hit._source?.[field];
+            }
+            return fieldVal === value;
```
joshuali925, line 185: lodash can handle the nested and flattened fields when reading value
```suggestion
            return _.get(hit._source, field) === value;
```

public/components/trace_analytics/components/traces/span_detail_table.tsx
joshuali925: is there a reason that this is different from https://github.com/opensearch-project/dashboards-observability/blob/0fe071019eef38123533982bb9183a0b874d4521/public/components/trace_analytics/components/traces/span_detail_panel.tsx#L144-L149? and maybe extract the parsing into a util function?

public/components/trace_analytics/components/traces/span_detail_table.tsx
```diff
@@ -253,67 +295,123 @@ export function SpanDetailTable(props: SpanDetailTableProps) {
...
+      console.error('Error parsing payloadData in SpanDetailTableHierarchy:', error);
+    } finally {
+      setIsSpansTableDataLoading(false);
+    }
+  }, [props.payloadData, props.DSL, props.mode, props.dataSourceMDSId, props.filters]);
 
   type SpanMap = Record<string, Span>;
 
-  const buildHierarchy = (spans: Span[]): Span[] => {
+  const buildHierarchy = (spans: Span[], modeTree: TraceAnalyticsMode): Span[] => {
```
joshuali925, line 337: why is it `modeTree` and not using props.mode directly?

public/components/trace_analytics/components/traces/span_detail_table.tsx
```diff
@@ -253,67 +295,123 @@ export function SpanDetailTable(props: SpanDetailTableProps) {
...
+          : Array.isArray(parsed)
+          ? parsed
+          : [];
+
+      let spans = hitsArray.map((hit: any) => hit._source);
+
+      if (props.filters.length > 0) {
+        spans = spans.filter((span: any) => {
+          return props.filters.every(
+            ({ field, value }: { field: string; value: any }) => span[field] === value
```
joshuali925, line 320: will there be negate filters like `field a is not b`?

public/components/trace_analytics/components/traces/span_detail_table.tsx
```diff
@@ -253,67 +295,123 @@ export function SpanDetailTable(props: SpanDetailTableProps) {
...
+        // Data Prepper
+        if (span.parentSpanId && spanMap[span.parentSpanId]) {
+          spanMap[span.parentSpanId].children.push(spanMap[span[spanIdKey]]);
+        } else {
+          // No parentSpanId or no matching parent span, treat as root span
+          if (!alreadyAddedRootSpans.has(span[spanIdKey])) {
+            rootSpans.push(spanMap[span[spanIdKey]]);
+            alreadyAddedRootSpans.add(span[spanIdKey]);
+          }
+        }
```
joshuali925, line 386: this might need some unit tests

public/components/trace_analytics/components/traces/span_detail_panel.tsx
TackAdam: Removed.

public/components/trace_analytics/components/traces/span_detail_panel.tsx
TackAdam: Updated too 
```
      hits = hits.map((hit) => {
        if (!hit.sort || !hit.sort[0]) {
          const time = traceMode === 'jaeger' 
            ? Number(hit._source.startTime) * MILI_TO_SEC
            : parseIsoToNano(hit._source.startTime);
          
          return {
            ...hit,
            sort: [time],
          };
        }
        return hit;
      });
```

public/components/trace_analytics/components/traces/span_detail_panel.tsx
TackAdam: I believe the inserting sort logic is needed for handling jaeger vrs data prepper. Using just the hits.sort you show fails to render the gantt visualization.

public/components/trace_analytics/components/traces/span_detail_panel.tsx
TackAdam: Updated too
```
      if (payloadSpanFilters.length > 0) {
        hits = hits.filter((hit) => {
          return payloadSpanFilters.every(({ field, value }) => {
            return _.get(hit._source, field) === value;
          });
        });
      }
```

public/components/trace_analytics/components/traces/span_detail_table.tsx
TackAdam: Not currently supported the flyout only allows filters to be added as such:
Screenshot 2025-02-10 at 4 19 45 PM
Where the user can only clear. Intended use is for investigating based on the flyout 
Screenshot 2025-02-10 at 4 20 37 PM

public/components/trace_analytics/components/traces/span_detail_table.tsx
TackAdam: Added cypress test for jaeger tree view functionality 
 ```
  it('Checks tree view for specific traceId in Jaeger mode', () => {
    cy.contains('15b0b4004a651c4c').click();
    cy.get('[data-test-subj="globalLoadingIndicator"]').should('not.exist');

    cy.get('.euiButtonGroup').contains('Tree view').click();
    cy.get("[data-test-subj='treeExpandAll']").should('exist');
    cy.get("[data-test-subj='treeCollapseAll']").should('exist');

    // Waiting time for render to complete
    cy.get("[data-test-subj='treeExpandAll']").click();
    cy.get("[data-test-subj='treeCollapseAll']").click();
    
    cy.get("[data-test-subj='treeViewExpandArrow']").should('have.length', 1);
    cy.get("[data-test-subj='treeExpandAll']").click();
    cy.get("[data-test-subj='treeViewExpandArrow']").should('have.length', 6);
  });
  ```

public/components/trace_analytics/components/traces/trace_view_helpers.tsx
TackAdam: Updated too 
```
    let computedLatency: number | null = null;
    const tgDuration = get(firstSpan, 'traceGroupFields.durationInNanos');
    if (firstSpan.traceGroupFields && firstSpan.traceGroupFields.durationInNanos != null) {
      computedLatency = Number(firstSpan.traceGroupFields.durationInNanos) / NANOS_TO_MS;
    } else if (tgDuration != null) {
      computedLatency = Number(tgDuration) / NANOS_TO_MS;
    } else {
      const startTimes = parsed.map((span: any) => moment(span._source.startTime));
      const endTimes = parsed.map((span: any) => moment(span._source.endTime));
      
      const minStartTime = moment.min(startTimes);
      const maxEndTime = moment.max(endTimes);
      
      if (minStartTime && maxEndTime) {
        computedLatency = maxEndTime.diff(minStartTime, 'milliseconds', true);
        fallbackValueUsed = true;
      }
    }
```
So if the first span is not present we get a more accurate latency instead of 0 or a smaller number

Before - invalid date:
Before

After - invalid date:
After

Latency is now much more accurate when handling missing data

public/components/trace_analytics/components/traces/span_detail_panel.tsx
```diff
@@ -15,18 +15,18 @@ import {
   EuiSmallButton,
   EuiSpacer,
 } from '@elastic/eui';
-import debounce from 'lodash/debounce';
-import isEmpty from 'lodash/isEmpty';
 import React, { useCallback, useEffect, useMemo, useRef, useState } from 'react';
 import useObservable from 'react-use/lib/useObservable';
+import _ from 'lodash';
```
mengweieric, line 20: I thought we are avoiding using lodash?

public/components/trace_analytics/components/traces/span_detail_table.tsx
```diff
@@ -192,13 +195,72 @@ export function SpanDetailTable(props: SpanDetailTableProps) {
...
+    if (!props.payloadData) {
+      fetchData();
+    }
   }, [tableParams, props.DSL]);
 
   useEffect(() => {
-    if (props.setTotal) props.setTotal(total);
-  }, [total]);
+    if (!props.payloadData) {
+      console.warn('No payloadData provided in SpanDetailTable');
```
mengweieric, line 211: why warn?

public/components/trace_analytics/components/traces/span_detail_table.tsx
```diff
@@ -192,13 +195,72 @@ export function SpanDetailTable(props: SpanDetailTableProps) {
...
+      setTotal(spans.length);
+    } catch (error) {
+      console.error('Error parsing payloadData in SpanDetailTable:', error);
+    } finally {
+      setIsSpansTableDataLoading(false);
+    }
+  }, [props.payloadData, props.DSL, props.filters, tableParams]);
+
+  const applySorting = (spans: Span[]) => {
+    if (tableParams.sortingColumns.length > 0) {
```
mengweieric, line 241: We should move this condition check outside of this function. The caller should handle sorting and ensure the array is not empty before invoking this function to perform the sorting.

public/components/trace_analytics/components/traces/span_detail_table.tsx
```diff
@@ -253,50 +303,89 @@ export function SpanDetailTable(props: SpanDetailTableProps) {
...
     spans.forEach((span) => {
-      if (span.parentSpanId && spanMap[span.parentSpanId]) {
-        // If the parent span exists, add this span to its children array
-        spanMap[span.parentSpanId].children.push(spanMap[span.spanId]);
+      const spanIdKey = props.mode === 'jaeger' ? 'spanID' : 'spanId';
+      const references = span.references || [];
+
+      // Handle Jaeger references (FOLLOWS_FROM and CHILD_OF)
+      if (props.mode === 'jaeger') {
+        references.forEach((ref: any) => {
```
mengweieric, line 360: type for ref?

public/components/trace_analytics/components/traces/span_detail_table.tsx
```diff
@@ -253,50 +303,89 @@ export function SpanDetailTable(props: SpanDetailTableProps) {
...
+          if (ref.refType === 'CHILD_OF') {
+            const parentSpan = spanMap[ref.spanID];
+            if (parentSpan) {
+              parentSpan.children.push(spanMap[span[spanIdKey]]);
+            }
+          }
+          if (ref.refType === 'FOLLOWS_FROM') {
+            if (!alreadyAddedRootSpans.has(span[spanIdKey])) {
+              rootSpans.push(spanMap[span[spanIdKey]]);
```
mengweieric, line 369: we should be able to refactor these a little bit for example like extracting these duplicated root span handing to a function:

```
rootSpans.push(spanMap[span[spanIdKey]]);
alreadyAddedRootSpans.add(span[spanIdKey]);
```

and also for these nested if conditions, I don't see you are having any else if or else then why not to merge them like 

```
if (ref.refType === 'FOLLOWS_FROM') {
            if (!alreadyAddedRootSpans.has(span[spanIdKey])) {
...
```
to

```
if (ref.refType === 'FOLLOWS_FROM' && !alreadyAddedRootSpans.has(span[spanIdKey])) {
...
```

public/components/trace_analytics/components/traces/trace_view.tsx
```diff
@@ -142,6 +141,15 @@ export function TraceView(props: TraceViewProps) {
                     <EuiText className="overview-title">Latency</EuiText>
                     <EuiText size="s" className="overview-content">
                       {overviewFields.latency}
+                      {overviewFields.fallbackValueUsed && (
+                        <EuiIconTip
+                          aria-label="Warning"
+                          size="m"
+                          type="alert"
+                          color="warning"
+                          content="Latency may not be accurate due to missing or unexpected duration data."
```
mengweieric, line 150: i18n?

public/components/trace_analytics/requests/traces_request_handler.ts
```diff
@@ -472,6 +324,16 @@ const hitsToSpanDetailData = async (hits: any, colorMap: any, mode: TraceAnalyti
   return data;
 };
 
+export function normalizePayload(parsed: any): any[] {
```
mengweieric, line 327: type?

public/components/trace_analytics/requests/traces_request_handler.ts
```diff
@@ -370,7 +222,7 @@ export const handleSpansFlyoutRequest = (
     });
 };
 
-const hitsToSpanDetailData = async (hits: any, colorMap: any, mode: TraceAnalyticsMode) => {
+export const hitsToSpanDetailData = async (hits: any, colorMap: any, mode: TraceAnalyticsMode) => {
```
mengweieric, line 225: type?

public/components/trace_analytics/components/traces/trace_view_helpers.tsx
```diff
@@ -0,0 +1,125 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { get } from 'lodash';
```
mengweieric, line 6: avoid using lodash?

public/components/application_analytics/components/flyout_components/trace_detail_render.tsx
```diff
@@ -95,10 +96,29 @@ export const TraceDetailRender = ({
...
     handlePayloadRequest(traceId, http, payloadData, setPayloadData, mode);
   }, [traceId]);
 
+  useEffect(() => {
+    if (!payloadData) return;
+
+    try {
+      const parsedPayload = JSON.parse(payloadData);
+      const overview = getOverviewFields(parsedPayload, mode);
+      if (overview) {
```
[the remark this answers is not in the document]
TackAdam: That one specifically is for application analytics and the useEffect triggers on payload change. As long as it is a valid payload the overview will get the new information from it. If the payload is successful from handlePayloadRequest causing new data the overview will always be generated from that information.

public/components/trace_analytics/components/common/helper_functions.tsx
```diff
@@ -615,3 +616,39 @@ export const generateServiceUrl = (
...
+  // Get the fractional part (if any), pad to 9 digits for nanosecond precision
+  let fraction = match[2] || '0';
+  fraction = fraction.padEnd(9, '0'); // ensure it has 9 digits
+  return baseMs * NANOS_TO_MS + Number(fraction);
+}
+
+export const parseHits = (payloadData: string) => {
+  try {
+    const parsed = JSON.parse(payloadData);
+    let hits: any[] = [];
```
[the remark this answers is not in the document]
TackAdam: Updated too
```
interface Span {
  traceId: string;
  spanId: string;
  traceState: string;
  parentSpanId: string;
  name: string;
  kind: string;
  startTime: string;
  endTime: string;
  durationInNanos: number;
  serviceName: string;
  events: any[];
  links: any[];
  droppedAttributesCount: number;
  droppedEventsCount: number;
  droppedLinksCount: number;
  traceGroup: string;
  traceGroupFields: {
    endTime: string;
    durationInNanos: number;
    statusCode: number;
  };
  status: {
    code: number;
  };
  instrumentationLibrary: {
    name: string;
    version: string;
  };
}

interface ParsedHit {
  _index: string;
  _id: string;
  _score: number;
  _source: Span;
  sort?: any[];
}

export const parseHits = (payloadData: string): ParsedHit[] => {
  try {
    const parsed = JSON.parse(payloadData);
    let hits: ParsedHit[] = [];

    if (parsed.hits && Array.isArray(parsed.hits.hits)) {
      hits = parsed.hits.hits;
    } else if (Array.isArray(parsed)) {
      hits = parsed;
    }

    return hits;
  } catch (error) {
    console.error('Error processing payloadData:', error);
    return [];
  }
};
```

public/components/trace_analytics/components/traces/span_detail_panel.tsx
TackAdam: Removed, replaced with 
```
      if (payloadSpanFilters.length > 0) {
        hits = hits.filter((hit) => {
          return payloadSpanFilters.every(({ field, value }) => {
            const fieldValue = field.split('.').reduce((acc, part) => acc?.[part], hit._source);
            return fieldValue === value;
          });
        });
      }
```

public/components/trace_analytics/components/traces/span_detail_table.tsx
TackAdam: Removed.

public/components/trace_analytics/components/traces/span_detail_table.tsx
TackAdam: Updated, moved the check to the useEffect calling it.

public/components/trace_analytics/components/traces/span_detail_table.tsx
TackAdam: added
```
  interface SpanReference {
    refType: 'CHILD_OF' | 'FOLLOWS_FROM';
    spanID: string;
  }
```

public/components/trace_analytics/components/traces/span_detail_table.tsx
TackAdam: Updated too
```
interface SpanReference {
    refType: 'CHILD_OF' | 'FOLLOWS_FROM';
    spanID: string;
  }
  
  const addRootSpan = (
    spanId: string,
    spanMap: SpanMap,
    rootSpans: Span[],
    alreadyAddedRootSpans: Set<string>
  ) => {
    if (!alreadyAddedRootSpans.has(spanId)) {
      rootSpans.push(spanMap[spanId]);
      alreadyAddedRootSpans.add(spanId);
    }
  };
  
  const buildHierarchy = (spans: Span[]): Span[] => {
    const spanMap: SpanMap = {};
  
    spans.forEach((span) => {
      const spanIdKey = props.mode === 'jaeger' ? 'spanID' : 'spanId';
      spanMap[span[spanIdKey]] = { ...span, children: [] };
    });
  
    const rootSpans: Span[] = [];
    const alreadyAddedRootSpans: Set<string> = new Set(); // Track added root spans
  
    spans.forEach((span) => {
      const spanIdKey = props.mode === 'jaeger' ? 'spanID' : 'spanId';
      const references: SpanReference[] = span.references || [];
  
      if (props.mode === 'jaeger') {
        references.forEach((ref: SpanReference) => {
          if (ref.refType === 'CHILD_OF') {
            const parentSpan = spanMap[ref.spanID];
            if (parentSpan) {
              parentSpan.children.push(spanMap[span[spanIdKey]]);
            }
          }
  
          if (ref.refType === 'FOLLOWS_FROM' && !alreadyAddedRootSpans.has(span[spanIdKey])) {
            addRootSpan(span[spanIdKey], spanMap, rootSpans, alreadyAddedRootSpans);
          }
        });
  
        if (references.length === 0 || references.every((ref) => ref.refType === 'FOLLOWS_FROM')) {
          addRootSpan(span[spanIdKey], spanMap, rootSpans, alreadyAddedRootSpans);
        }
      } else {
        // Data Prepper handling
        if (span.parentSpanId && spanMap[span.parentSpanId]) {
          spanMap[span.parentSpanId].children.push(spanMap[span[spanIdKey]]);
        } else {
          addRootSpan(span[spanIdKey], spanMap, rootSpans, alreadyAddedRootSpans);
        }
      }
    });
  
    return rootSpans;
  };
  ```

File: public/components/trace_analytics/components/traces/trace_view.tsx (hunk 25 above)
Comment (TackAdam) [reply public/components/trace_analytics/components/traces/trace_view.tsx]: Updated too
```
                        <EuiIconTip
                          aria-label={i18n.translate('tracesOverview.iconTip.ariaLabel', {
                            defaultMessage: 'Warning',
                          })}
                          size="m"
                          type="alert"
                          color="warning"
                          content={i18n.translate('tracesOverview.iconTip.content', {
                            defaultMessage: 'Latency may not be accurate due to missing or unexpected duration data.',
                          })}
                        />
```

File: public/components/trace_analytics/requests/traces_request_handler.ts (hunk 26 above)
Comment (TackAdam) [reply public/components/trace_analytics/requests/traces_request_handler.ts]: Updated too 
```
interface Hit {
  _index: string;
  _id: string;
  _score: number;
  _source: any;
  sort?: any[];
}

interface ParsedResponse {
  hits?: {
    hits: Hit[];
  };
  [key: string]: any;
}

export function normalizePayload(parsed: ParsedResponse): Hit[] {
  if (Array.isArray(parsed)) {
    return parsed;
  }
  if (parsed.hits && Array.isArray(parsed.hits.hits)) {
    return parsed.hits.hits;
  }
  return [];
}
```

File: public/components/trace_analytics/components/traces/trace_view_helpers.tsx (hunk 28 above)
Comment (TackAdam) [reply public/components/trace_analytics/components/traces/trace_view_helpers.tsx]: Removed and refactored to work without it.

File: public/components/trace_analytics/requests/traces_request_handler.ts [hunk 31]
```diff
@@ -472,6 +324,31 @@ const hitsToSpanDetailData = async (hits: any, colorMap: any, mode: TraceAnalyti
...
+
+export function normalizePayload(parsed: ParsedResponse): Hit[] {
+  if (Array.isArray(parsed)) {
+    return parsed;
+  }
+  if (parsed.hits && Array.isArray(parsed.hits.hits)) {
+    return parsed.hits.hits;
+  }
+  return [];
+}
```
ps48, line 350: Why do we still need this? If we know one check is for Jaeger and other is data prepper can we just use mode based if conditions instead?

public/components/trace_analytics/components/common/helper_functions.tsx
```diff
@@ -615,3 +616,77 @@ export const generateServiceUrl = (
...
+  status: {
+    code: number;
+  };
+  instrumentationLibrary: {
+    name: string;
+    version: string;
+  };
+}
+
+interface ParsedHit {
```
ps48, line 668: Can we please move these to common constants?

TackAdam: Moved to constants

public/components/trace_analytics/requests/traces_request_handler.ts
TackAdam: Updated too 
```
export function normalizePayload(parsed: ParsedResponse): Hit[] {
  if (parsed.hits && Array.isArray(parsed.hits.hits)) {
    return parsed.hits.hits;
  }
  return [];
}
```
Jaeger and data prepper should be under parsed.hits

public/components/trace_analytics/components/traces/trace_view_helpers.tsx
joshuali925: do we have a replacement for lodash? some of them like `get` are useful and not trivial in plain js
https://github.com/you-dont-need/You-Dont-Need-Lodash-Underscore?tab=readme-ov-file#_get

ps48 (approved):

joshuali925 (approved):

PR #2368, opened 2025-02-28
Trace Groups Optimization - Remove duplicate filters

TackAdam:
### Description
1. TraceGroups was unnecessarily triggering a re-query of the main table when opened. This was due to the use-effect inside dashboard_content setting the lock for valid filters. Upon investigating this call seemed to be a repeat and unnecessary as it is already handled in:
 https://github.com/opensearch-project/dashboards-observability/blob/main/public/components/trace_analytics/components/common/filters/filters.tsx#L79
 and by the trace content page itself.
 
 Removing these unnecessary filter settings led the queries to stop firing unnecessarily when opening the trace group panel.
 
 2. Fix a bug where a trace group filter being applied was not refreshing the trace group table due to the re-direction flag being set.
 
Before (7 Queries (2 unnecessary))
  
Before

After (5 Queries)
AfterRemoving

Filter behavior before:

https://github.com/user-attachments/assets/ac17312d-daa3-4c57-be69-12a1e012c623

After (unchanged with code removal):

https://github.com/user-attachments/assets/7f5c1680-5946-47e6-a761-9541aa5da6ed

2. Before (filter not applied):

https://github.com/user-attachments/assets/86f644ed-68b8-4bca-b399-60e15ef04281

After (filter applied):

https://github.com/user-attachments/assets/129d0d18-53bd-40e4-97a0-13f75bfcafe8

### Issues Resolved
https://github.com/opensearch-project/dashboards-observability/issues/2334 - Trace Group Optimizations

### Check List

### DIFF
--- a/.github/workflows/ftr-e2e-dashboards-observability-test.yml
+++ b/.github/workflows/ftr-e2e-dashboards-observability-test.yml
@@ -151,12 +151,19 @@ jobs:
         if: ${{ runner.os == 'Linux' }}
         run: |
           cd ./OpenSearch-Dashboards
-          if timeout 60 grep -q "http server running" <(tail -n +1 -f dashboard.log); then
+          if timeout 1200 grep -q "http server running" <(tail -n +1 -f dashboard.log); then
             echo "OpenSearch Dashboards started successfully."
           else
-            echo "Timeout of 60 seconds reached. OpenSearch Dashboards did not start successfully."
+            echo "Timeout of 1200 seconds reached. OpenSearch Dashboards did not start successfully."
             exit 1
           fi&
+
+      - name: Upload Dashboards logs
+        uses: actions/upload-artifact@v4
+        if: failure()
+        with:
+          name: dashboard.logs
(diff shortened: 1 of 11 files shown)


ps48:
Just one issue when we click on number of traces from the traceGroup table, filter doesn't pushed down to tracegroup table Only the traces table is updated. Not sure if this is an existing bug.

ps48 (approved):

TackAdam:
> Just one issue when we click on number of traces from the traceGroup table, filter doesn't pushed down to tracegroup table Only the traces table is updated. Not sure if this is an existing bug.
> 
> Also, opening the tracegroup table reloads the trace table. This is the call that we wanted to remove I believe

TraceGroup not triggering the re-load confirmed offline and in attached videos.

Number of traces not filtering when clicked was an existing but that I just added a fix for 
```
if (!['app', 'traces'].includes(props.page)) {
```
Dashboards was previously a seperate page so the re-direction flag was causing it to not refresh.

sumukhswamy (approved):

PR #2369, opened 2025-03-03
[Bug] Traces redirection while QA enabled

TackAdam:
### Description
Re-direction from Traces to Explorer to view logs is being over written when QA is enabled. 

Before:

https://github.com/user-attachments/assets/56f094ea-ad4a-4794-9e61-376ae3cee33a

After:

https://github.com/user-attachments/assets/ec6a3c34-c3e4-4b1e-8b33-4c4d7ca92338

### Check List

### DIFF
--- a/public/components/common/search/search.tsx
+++ b/public/components/common/search/search.tsx
@@ -259,7 +259,6 @@ export const Search = (props: any) => {
       setSelectedIndex(reduxIndex);
       // sets the editor text and populates sidebar field for a particular index upon initialization
       const indexQuery = `source = ${reduxIndex[0].label}`;
-      handleQueryChange(indexQuery);
       getAvailableFields(indexQuery);
     }
     if (queryRedux.ollyQueryAssistant.length > 0) {
--- a/public/components/common/search/query_area.tsx
+++ b/public/components/common/search/query_area.tsx
@@ -38,7 +38,6 @@ export function QueryArea({
   const memoizedHandleQueryChange = useMemo(() => handleQueryChange, []);
   useEffect(() => {
     const indexQuery = `source = ${selectedIndex[0]?.label || ''}`;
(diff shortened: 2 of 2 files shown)


joshuali925 (approved):
does it affect query assist feature?

TackAdam:
> does it affect query assist feature?

Not that I can tell, as soon as a question is prompted it switched the source to the selected index and overrides what is present from the re-direction.

public/components/common/search/search.tsx
```diff
@@ -259,7 +259,6 @@ export const Search = (props: any) => {
       setSelectedIndex(reduxIndex);
       // sets the editor text and populates sidebar field for a particular index upon initialization
       const indexQuery = `source = ${reduxIndex[0].label}`;
-      handleQueryChange(indexQuery);
```
paulstn, line 262: If I'm remembering correctly, this method was used to set the query inside of the "query area" from a redirection related to query assistant. Did we know what this was used for beforehand and if it remains unaffected with this change?
One of these redirections I believe were from the welcome page with sample indices and sample natural language queries, could start with testing that interaction.

ps48: @paulstn Do you know was it some custom welcome page or released in a particular OS version? We are not able to find this on our side..

ps48 (approved):

public/components/common/search/search.tsx
TackAdam: Tested the 2.13 new home page setting which just provides links to documentation, no workflow identified that landed on the explorer page with an example query.

paulstn: Regardless of being able to find the welcome page, the purpose of this line was to populate the query area with some placeholder text. If we're fine removing that placeholder then there isn't an issue.

Issue #2141, opened 2024-09-06
[Meta] RFC: Enhancements to Trace Analytics Plugin 


ps48:
## Overview
The Trace Analytics plugin in OpenSearch Dashboards provides users with an intuitive interface to analyze and visualize trace data. Initially, the plugin supported integration with Data Prepper, and later extended support to the Jaeger schema. Since its introduction in ODFE 1.12 and OpenSearch 1.0, the plugin has functioned as a read-only solution. It currently does not store metadata and relies on browser session storage to maintain modes. This RFC outlines new features and enhancements aimed at improving storage capabilities, UI design, query performance, and seamless integration with other OpenSearch Dashboards plugins.

### Plugin as it exists today
- The plugin enables users to explore and analyze trace data from span and service indices through a user-friendly interface.
- Initially built with support for Data Prepper, it has evolved to also support Jaeger schema for broader trace collection compatibility.
- First introduced in Open Distro for Elasticsearch (ODFE) 1.12 and OpenSearch 1.0.
- The plugin operates in a read-only mode, without the ability to persist configurations across sessions.
- No trace metadata is stored, and configurations are temporarily saved in browser session storage only.
- Lacks support for advanced features like custom indices, cross-cluster queries, or configuration persistence.

## Proposed Enhancements
The proposed changes address the following requirements:
1. **Storage Layer**  
   - Introduce a connection to the storage layer by saving configurations in the `.kibana` index.
   - Store trace and service index configurations.
   - Support for custom indices with wildcards and cross-cluster queries.

2. **UI Enhancements**  
   - A dedicated Traces and Services page in the new navigation system.
   - A refreshed look and feel for a more efficient, sleeker, and consistent UI.
   - An updated service map graph UI that can display in full-screen mode with easy navigation.
   - Componentization of existing UI widgets, such as the traces table and service map.

3. **Query Optimization**  
   - Optimize DSL queries for better aggregation performance, especially at scale.
   - Leverage pre-aggregated indices from Data Prepper to reduce the load on OpenSearch during aggregation operations.

4. **Seamless Correlation**  
   - Implement schema-based correlation between logs, traces, and metrics using OTEL standards.
   - Improve navigation and correlation between Trace Analytics and other OpenSearch Dashboards plugins like Discover, Anomaly Detection, and Alerting.

## Architecture
The enhancements will involve the following architectural changes:
1. **Storage Layer**  
[...]

PR #2375, opened 2025-03-05
Support custom logs correlation

ps48:
### Description
Many users of trace analytics have different ways of defining their logs index. Many of them do not want set an alias on their existing indices and fields. Hence, we are adding this experimental feature to enable them to add custom log source used for correlation of spans & services to logs.  

Redirection to Explorer:

https://github.com/user-attachments/assets/2387592b-22a5-4a9b-9c69-81bd9110fa29

Redirection to Discover:
NOTE: We see an error in discover to load data from specific indexes using PPL. This looks to be a discover only issue.

https://github.com/user-attachments/assets/8a17e3d0-a137-4dd9-bd6a-74b4c1095d08

### Issues Resolved
https://github.com/opensearch-project/dashboards-observability/issues/2141

### Check List

### DIFF
--- a/common/constants/trace_analytics.ts
+++ b/common/constants/trace_analytics.ts
@@ -34,7 +34,8 @@ export const DEFAULT_SS4O_LOGS_INDEX = 'ss4o_logs-*';
 export const DEFAULT_CORRELATED_LOGS_FIELD_MAPPINGS = `
 {
   "serviceName": "serviceName",
-  "spanId": "spanId"
+  "spanId": "spanId",
+  "timestamp": "time"
 }`;
 
 export enum TRACE_TABLE_TITLES {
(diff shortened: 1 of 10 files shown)


TackAdam:
Are we intentionally applying the log “Correlated log indicies” to links from “Data Prepper” or should it be isolated to only affecting Traces while under the “Custom Source” tab. It is currently applying to both which seems to be useful but just wanted to make sure that its intended.

TackAdam:
“View associated logs” redirection seems broken under the new navigation when “Explorer” is not present as it cannot land on the page. Don’t think this was introduced by this PR but restricts the ability to test the functionality with new home enabled.

ps48:
> “View associated logs” redirection seems broken under the new navigation when “Explorer” is not present as it cannot land on the page. Don’t think this was introduced by this PR but restricts the ability to test the functionality with new home enabled.

Did you have MDS enabled and query_enhancements enabled? -> We need this to make sure these are enabled, when explorer is missing.

paulstn (commented):
Changes look good to me, could we get tests?

TackAdam (approved):

paulstn (approved):

ps48:
Note: We need to fixate of default timestamp in the next upcoming version.

PR #2380, opened 2025-03-07
Database selector in "Set Up Integration" page

aalva500-prog:
### Description
Currently, when an Integration is created through the "Set Up Integration" page, the default database is hardcoded, as the table template is created as `${dataSource}.default.${tableName}`. In order to avoid hard coding the database name, this feature allows the user to specify the database, and add their value in the template.

### Issues Resolved
This feature will allow the user to specify the database, and add their value in the template.

### Check List
  - [x] New functionality has javadoc added


public/components/integrations/components/setup_integration.tsx
```diff
@@ -151,92 +185,123 @@ const addIntegration = async ({
...
-    )) {
-      // Skip any queries that have conditional workflows but aren't enabled
-      if (query.workflows && !query.workflows.some((w) => config.enabledWorkflows.includes(w))) {
-        continue;
-      }
-
-      const queryStr = prepareQuery(query.query, config);
+  try {
+    // First create the database if one is specified
+    if (config.databaseName) {
```
Swiddis, line 190: issue: It's not guaranteed that this will only happen for S3 integrations.

This works for clean state because the DB is set to undefined by default (see the other comment), but will incorrectly create the DB if the user switches to S3, sets a DB, and then switches back to Index.

As-is we have separate install processes between index and S3, this should go under S3.

For readability, I think this is actually a good time to move the index install and S3 install to their own dedicated methods. Then the body of addIntegration can look like this, and the processes are less likely to get confused in the future.

```ts
if (config.connectionType === 'index') {
  installNativeIntegration(...);
} else if (config.connectionType === 's3') {
  installFlintIntegration(...);
} else {
  // Invalid data source
}
```

Swiddis (changes requested):
Thanks for the PR!

public/components/integrations/components/setup_integration.tsx
```diff
@@ -108,8 +113,19 @@ const runQuery = async (
...
+ *
+ * @param config - The integration setup configuration object
+ * @param config.connectionDataSource - The data source connection name
+ * @param config.databaseName - The database name (defaults to 'default' if not provided)
+ * @param config.connectionTableName - The table name
+ * @returns A string representing the fully qualified table name in the format: dataSource.database.table
+ */
 const makeTableName = (config: IntegrationSetupInputs): string => {
-  return `${config.connectionDataSource}.default.${config.connectionTableName}`;
+  return `${config.connectionDataSource}.${config.databaseName || 'default'}.${
```
Swiddis, line 126: quibble: let's remove the `default` fallback

Once there's a default value per other comments, databaseName can never be empty, so the fallback value will never trigger.

aalva500-prog: Suggestion Done!
```suggestion
  return `${config.connectionDataSource}.${config.databaseName}.${
```

public/components/integrations/components/setup_integration.tsx
```diff
@@ -28,12 +28,17 @@ import { coreRefs } from '../../../framework/core_refs';
 import { addIntegrationRequest } from './create_integration_helpers';
 import { SetupIntegrationFormInputs } from './setup_integration_inputs';
 
+/**
+ * Configuration inputs for integration setup
```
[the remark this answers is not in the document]
aalva500-prog: np!

public/components/integrations/components/setup_integration_inputs.tsx
```diff
@@ -321,6 +338,19 @@ export function IntegrationQueryInputs({
...
         />
       </EuiCompressedFormRow>
+      {/* Add the new database name field */}
+      <EuiCompressedFormRow
+        label="Database Name"
+        helpText="Enter the name of the database to store your data. The 'default' database will be used if no input is provided."
+      >
+        <EuiCompressedFieldText
+          placeholder="default"
+          value={config.databaseName}
```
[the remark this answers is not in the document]
aalva500-prog: Done!
```
const [integConfig, setConfig] = useState({
    displayName: `${integration} Integration`,
    connectionType: forceConnection?.type ?? 'index',
    connectionDataSource: forceConnection?.name ?? '',
    connectionLocation: '',
    checkpointLocation: '',
    connectionTableName: integration,
    databaseName: 'default',
    enabledWorkflows: [],
  });
```

public/components/integrations/components/setup_integration.tsx
```diff
@@ -151,92 +185,123 @@ const addIntegration = async ({
...
+      }
+      if (!res) {
+        setLoading(false);
+      }
+    } else {
+      console.error('Invalid data source type');
     }
-  } else {
-    console.error('Invalid data source type');
+  } catch (error) {
```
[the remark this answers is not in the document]
aalva500-prog: I think you are right about that, the original code doesn't have the outer try/catch block. The reason I added this outer try/catch block was to be able to catch any errors that could occur during the whole integration creation process. I think is better to remove it.

public/components/integrations/components/setup_integration.tsx
aalva500-prog: This is a good point, I was actually thinking about improving the readability of the "addIntegration" method by separating the index and S3 install to their own dedicated methods.

Suggestions done:
- Moved the DB creation to the S3 install section.
- Improved the readability of the "addIntegration" method by separating the index and S3 install to their own dedicated methods.

public/components/integrations/components/setup_integration.tsx
```diff
@@ -148,95 +177,225 @@ const addIntegration = async ({
...
+      sessionId,
+      dataSourceMDSId
+    );
+
+    if (!result.ok) {
+      setLoading(false);
+      setCalloutLikeToast('Failed to create database', 'danger', result.error.message);
+      return;
+    }
+    sessionId = result.value.sessionId;
```
Swiddis, line 329: suggestion (non-blocking,readability): This could be moved to its own function instead of inlined

public/components/integrations/components/setup_integration.tsx
```diff
@@ -148,95 +177,225 @@ const addIntegration = async ({
...
+    const result = await runQuery(
+      queryStr,
+      config.connectionDataSource,
+      sessionId,
+      dataSourceMDSId
+    );
+
+    if (!result.ok) {
       setLoading(false);
+      setCalloutLikeToast('Failed to add integration', 'danger', result.error.message);
```
Swiddis, line 357: thought: At some point we need to register all these strings to i18n, but do we need to do that now?

@ps48 / @TackAdam any context on i18n timelines, or is there a doc that describes the migration process? I can make an issue to do it later.

public/components/integrations/components/setup_integration.tsx
```diff
@@ -132,6 +146,21 @@ const prepareQuery = (query: string, config: IntegrationSetupInputs): string =>
   return queryStr;
 };
 
+/**
+ * Handles the integration setup process based on the connection type.
+ *
+ * @param {Object} params - The parameters object
```
Swiddis, line 152: suggestion (non-blocking): There's a lot of duplication with all of these param objects.

We could consider making a dedicated `interface` for these params, and moving all the comments to all of the fields of the params object. That will make the code more clear as the function signature looks like `{ config, integrations, ... }: IntegrationInstallParams`.

Swiddis (approved):
Looks good! Just leaving some additional thoughts. @sumukhswamy can you give a second review?

public/components/integrations/components/setup_integration.tsx
aalva500-prog: Good point, I've implemented it, as suggested.

public/components/integrations/components/setup_integration.tsx
TackAdam: Not blocking, is good to have but repo needs a pass to be uniform in its implementation.

TackAdam (approved):

PR #2383, opened 2025-03-10
Service Content/View Optimizations

TackAdam:
### Description
Service content and service view page called the following functions handleServicesRequest() and handleServiceViewRequest(). Both of these functions utilized handleServiceMapRequest() to determine connected services and metrics. 
Services content and services view had an additional call directly to handleServiceMapRequest() for generating the service map displayed at the bottom of the page. By removing these calls and passing setServiceMap into handleServicesRequest() and handleServiceViewRequest() we are able to remove 2 unnecessary queries from both pages.

Before (Service Content: 5 queries, Service View: 7 queries)

https://github.com/user-attachments/assets/635a2423-64a0-4bbb-9afa-b5adf9de2e33

After (Service Content: 3 queries, Service View: 5 queries)

https://github.com/user-attachments/assets/51886d97-fa74-45fa-99c6-dba9aa152a91

This change also reduces the calls made in application analytics
Before (Services: 5 queries, Flyout: 11 queries)

https://github.com/user-attachments/assets/6f3fe145-777a-4578-8c69-4ef26b6a4b8a

After (Services: 3 queries, Flyout: 7 queries)

https://github.com/user-attachments/assets/e336f2d9-2e89-47b7-a52e-7471cc9efcfa

### Issues Resolved
https://github.com/opensearch-project/dashboards-observability/issues/2334 - Service Content/View Optimizations

### Check List


public/components/trace_analytics/components/services/service_view.tsx
```diff
@@ -106,25 +103,19 @@ export function ServiceView(props: ServiceViewProps) {
...
       props.serviceName,
       props.http,
       DSL,
       setFields,
       mode,
+      setServiceMap,
       props.dataSourceMDSId[0].id
-    ).finally(() => setIsServiceOverviewLoading(false));
-
-    if (mode === 'data_prepper' || mode === 'custom_data_prepper') {
```
paulstn, line 118: I'm understanding that the content of this condition has the service map call you're removing, but is the actual condition here (i.e. `mode === <something>`) being removed problematic to leave out? handleServiceViewRequest doesn't have that same condition above.

TackAdam: This was previously just preventing the call if the mode was set to jaeger, as the service map is not displayed in this mode. As the existing logic already calls handleServiceMapRequest(). 
The previous logic was actually still calling this function just once instead of twice for the jaeger mode.
The condition is not necessary in the new code as the setServiceMap is only utilized in displaying the service map which is not present in jaeger.

public/components/application_analytics/components/flyout_components/service_detail_flyout.tsx
```diff
@@ -135,8 +132,7 @@ export function ServiceDetailFlyout(props: ServiceFlyoutProps) {
 
     setIsServicesDataLoading(true);
     Promise.all([
-      handleServiceViewRequest(serviceName, http, serviceDSL, setFields, mode),
-      handleServiceMapRequest(http, serviceDSL, mode, '', setServiceMap),
+      handleServiceViewRequest(serviceName, http, serviceDSL, setFields, mode, setServiceMap),
     ]).finally(() => setIsServicesDataLoading(false));
```
joshuali925, line 136: it's one promise, you can remove the Promise.all

public/components/trace_analytics/components/services/services_content.tsx
```diff
@@ -124,13 +123,17 @@ export function ServicesContent(props: ServicesProps) {
...
+      setServiceMap,
+      dataSourceMDSId[0].id
+    ).finally(() => setIsServicesTableDataLoading(false));
 
     setIsServicesDataLoading(true);
     Promise.all([
-      handleServiceMapRequest(http, serviceMapDSL, mode, dataSourceMDSId[0].id, setServiceMap),
       isServiceTrendEnabled
         ? handleServiceTrendsRequest(http, '1h', setServiceTrends, mode, [], dataSourceMDSId[0].id)
         : Promise.resolve(null),
```
joshuali925, line 139: same here, it can just be if isServiceTrendEnabled, then `handleServiceTrendsRequest(http, '1h', setServiceTrends, mode, [], dataSourceMDSId[0].id).finally(() => setIsServicesDataLoading(false))`

joshuali925 (approved):

public/components/trace_analytics/components/services/services_content.tsx
TackAdam: Adjusted too
```
    setIsServicesDataLoading(true);
    if (isServiceTrendEnabled) {
      handleServiceTrendsRequest(http, '1h', setServiceTrends, mode, [], dataSourceMDSId[0].id)
        .finally(() => setIsServicesDataLoading(false));
    } else {
      setIsServicesDataLoading(false);
    }
```
Thanks

public/components/application_analytics/components/flyout_components/service_detail_flyout.tsx
TackAdam: Removed, thanks!

paulstn (approved):

PR #2397, opened 2025-04-01
Custom Traces- Sorting/Toast

TackAdam:
### Description
1. Add a default sorting of descending time while in the custom trace mode for spans/traces.
2. Fix the sorting of the tracing table under custom source to use the paginated results for 'trace_group', 'percentile_in_trace_group', 'trace_id' as they can not be ran in the query. (Matching behavior of old memory table)
3. Add early return to the traces/custom traces request handlers to prevent toast messages on empty data.
4. Update cypress testing for the custom source to account for the results being sorted now.

Before with invalid/empty span
Before
After
InvalidSpan

Invalid sorting that required it to be local only:
InvalidSort

### Check List

### DIFF
--- a/public/components/trace_analytics/components/traces/traces_content.tsx
+++ b/public/components/trace_analytics/components/traces/traces_content.tsx
@@ -4,7 +4,6 @@
  */
 /* eslint-disable react-hooks/exhaustive-deps */
 
-import datemath from '@elastic/datemath';
 import {
   EuiAccordion,
   EuiFlexGroup,
@@ -25,7 +24,7 @@ import {
 } from '../../requests/traces_request_handler';
 import { getValidFilterFields } from '../common/filters/filter_helpers';
 import { Filters, FilterType } from '../common/filters/filters';
-import { filtersToDsl, processTimeStamp } from '../common/helper_functions';
+import { filtersToDsl, isUnderOneHourRange, processTimeStamp } from '../common/helper_functions';
 import { ServiceMap, ServiceObject } from '../common/plots/service_map';
 import { SearchBar } from '../common/search_bar';
 import { DashboardContent } from '../dashboard/dashboard_content';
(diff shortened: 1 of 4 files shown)


joshuali925 (approved):

public/components/trace_analytics/components/traces/traces_content.tsx
```diff
@@ -95,43 +101,40 @@ export function TracesContent(props: TracesProps) {
 
     setSortingColumns(sortColumns);
 
+    const localOnlyFields = ['trace_group', 'percentile_in_trace_group', 'trace_id'];
```
joshuali925, line 104: maybe add some comments to explain what local only means, for future reference

public/components/trace_analytics/components/traces/traces_content.tsx
```diff
@@ -273,12 +277,57 @@ export function TracesContent(props: TracesProps) {
...
+    const timeFilterDSL = filtersToDsl(
+      mode,
+      [],
+      '',
+      processTimeStamp(startTime, mode),
+      processTimeStamp(endTime, mode),
+      page
+    );
+
+    const isUnderOneHour = datemath.parse(endTime)?.diff(datemath.parse(startTime), 'hours')! < 1;
```
joshuali925, line 315: util functions like this can be extracted, i see the same logic is also used in other places

public/components/trace_analytics/components/traces/traces_content.tsx
TackAdam: Added comment
// The columns that can not be used in a query, only rendered on page

public/components/trace_analytics/components/traces/traces_content.tsx
TackAdam: added helper function isUnderOneHourRange
```
export const isUnderOneHourRange = (startTime: string, endTime: string): boolean => {
  const start = dateMath.parse(startTime);
  const end = dateMath.parse(endTime);

  if (!start || !end) return false;

  return end.diff(start, 'hours')! < 1;
};
```

ps48 (approved):

PR #2407, opened 2025-04-08
Traces - Update custom source toast/error/sorting

TackAdam:
### Description
1. Change all "Trace ID" text for tables to "Trace Id"
2. Add empty/error state for trace view and service view.
3. Remove sorting from fields that didn't make sense such as hashed values { Span Id, Trace Id, Trace Group, ect.. }
4. Adjust "... Load more" based on UX feedback, maintain sorting and page on query. Only show load more option when there is more to load and user is on the last page.
5. Add handling for custom traces source when there are less than "maxTraces"
6. Adjust right hand indicator to reflect the number of root spans on the traces page accurately.

"Trace Id" Change:
Screenshot 2025-04-14 at 3 50 26 PM

Traces error state:

https://github.com/user-attachments/assets/17107dd9-d1c6-4b27-bd40-42ce695f3580

Services error state:

https://github.com/user-attachments/assets/6f5784ac-ca2c-44b3-8a86-9a13b315e20c

Sorting removed from hash values:
Screenshot 2025-04-14 at 3 51 15 PM

Load More:

https://github.com/user-attachments/assets/96b25479-75b3-43ff-9525-d36bf385bee2

Toast for bucket exception:

https://github.com/user-attachments/assets/efc820ad-1870-4b29-b1ff-835ba161a8dc

Adjust right hand indicator to reflect the number of root spans on the traces page accurately.
Screenshot 2025-04-14 at 4 06 55 PM

### Check List


public/components/trace_analytics/components/common/shared_components/component_helper_functions.tsx
```diff
@@ -51,18 +52,16 @@ export const useInjectElementsIntoGrid = (
...
       const pagination = document.querySelector<HTMLElement>('.euiDataGrid__pagination');
+      const existingLoadMoreButton = pagination?.querySelector('.trace-table-load-more');
 
-      if (tracesTableMode !== 'traces') {
-        if (pagination) {
-          const existingLoadMoreButton = pagination.querySelector('.trace-table-load-more');
-          if (existingLoadMoreButton) {
-            existingLoadMoreButton.remove();
-          }
+      if (tracesTableMode !== 'traces' || !pagination || !loadMoreHandler || !isLastPage) {
```
paulstn, line 57: Is this change in the logic intentional? if so, you could remove the `pagination`, `loadMoreHandler`, and `isLastPage` from line 64, since not having one of those expressions as true would have an early return. 
Whereas I believe they were there initially because this statement here wouldn't necessarily early return for not having all those expressions/flags.

public/components/trace_analytics/components/common/shared_components/custom_datagrid.tsx
```diff
@@ -137,14 +137,38 @@ export const RenderCustomDataGrid: React.FC<RenderCustomDataGridParams> = ({
...
+    }
+  }, [rowCount, maxTraces, tracesTableMode]);
+
   useInjectElementsIntoGrid(
     rowCount,
     MAX_DISPLAY_ROWS,
     tracesTableMode ?? '',
     () => {
-      setMaxTraces((prevMax: number) => Math.min(prevMax + 500, MAX_DISPLAY_ROWS));
+      // Add 500 more traces at a time untill all traces present
```
paulstn, line 164: nit: typo

TackAdam: Good catch, thanks!
Fixed spelling to "until"

public/components/trace_analytics/components/common/shared_components/component_helper_functions.tsx
TackAdam: Yes this change is intentional, I have removed the redundant check on line 64 for 
```
if (pagination && loadMoreHandler && isLastPage)
```

public/components/trace_analytics/components/services/service_view.tsx
```diff
@@ -98,7 +101,23 @@ export function ServiceView(props: ServiceViewProps) {
       props.serviceName,
       props.http,
       DSL,
-      setFields,
+      (data) => {
```
paulstn, line 104: Having this much logic within the parameter of a function seems like a code smell to me. 
Is there an easy way to bring the code either outside of this function or within this function? Since you're setting some state it might be better to define this outside of the function and pass that in. 
Ultimately not a big deal and its completely up to you, just some thoughts.

TackAdam: Agreed this can make the readability difficult. Updated to extract this to a function call right above the refresh
for services:
```
  const handleServiceDataResponse = (data: any) => {
    try {
      if (!data || Object.keys(data).length === 0) {
        setServiceIdError(true);
        setFields({});
        setServiceMap({});
      } else {
        setServiceIdError(false);
        setFields(data);
      }
    } catch (e) {
      console.error('Failed to parse service response:', e);
      setServiceIdError(true);
      setFields({});
      setServiceMap({});
    }
  };
  
    handleServiceViewRequest(
      props.serviceName,
      props.http,
      DSL,
      handleServiceDataResponse,
      mode,
      setServiceMap,
      props.dataSourceMDSId[0].id
    )
```
for traces:
```
  const handlePayloadResponse = (data: string) => {
    setPayloadData(data);
    try {
      const parsed = JSON.parse(data);
      if (!Array.isArray(parsed) || parsed.length === 0) {
        setTraceIdError(true);
        setFields({});
      } else {
        setTraceIdError(false);
      }
    } catch {
      setTraceIdError(true);
      setFields({});
    }
  };
  
      handlePayloadRequest(
      props.traceId,
      props.http,
      payloadData,
      handlePayloadResponse,
      mode,
      props.dataSourceMDSId[0].id
    ).
```

paulstn (approved):

RyanL1997 (approved):

public/components/trace_analytics/requests/traces_request_handler.ts
```diff
@@ -389,7 +418,7 @@ export const handlePayloadRequest = (
     .catch((error) => {
       console.error('Error in handlePayloadRequest:', error);
       coreRefs.core?.notifications.toasts.addError(error, {
-        title: `Failed to retrieve payload for trace ID: ${traceId}`,
+        title: `Failed to retrieve payload for trace Id: ${traceId}`,
```
ps48, line 421: We have toast messaged issued here, why aren't we seeing the toast pop up already when the user lands onto a wrong URL on trace view page today?

public/components/trace_analytics/requests/traces_request_handler.ts
```diff
@@ -157,8 +158,37 @@ export const handleTracesRequest = async (
         })
       : Promise.resolve({});
 
-  return Promise.allSettled([responsePromise, percentileRangesPromise])
-    .then(([responseResult, percentileRangesResult]) => {
+  let rootSpansTotalHitsPromise: Promise<number> | undefined;
+
+  if (setTotalHits && (mode === 'data_prepper' || mode === 'custom_data_prepper')) {
+    const rootSpansDSL = cloneDeep(timeFilterDSL);
+    rootSpansDSL.query.bool.filter.push({
```
ps48, line 165: Looks like we are adding a `_search` request, this should be just `_count` request as we need just the hits here.

public/components/trace_analytics/requests/traces_request_handler.ts
```diff
@@ -157,8 +158,37 @@ export const handleTracesRequest = async (
         })
       : Promise.resolve({});
 
-  return Promise.allSettled([responsePromise, percentileRangesPromise])
-    .then(([responseResult, percentileRangesResult]) => {
+  let rootSpansTotalHitsPromise: Promise<number> | undefined;
+
+  if (setTotalHits && (mode === 'data_prepper' || mode === 'custom_data_prepper')) {
+    const rootSpansDSL = cloneDeep(timeFilterDSL);
```
ps48, line 164: Question: Why do we need cloneDeep? I might have missed something here.

public/components/trace_analytics/components/traces/traces_content.tsx
```diff
@@ -232,9 +203,30 @@ export function TracesContent(props: TracesProps) {
...
+    refreshTracesTableData(sort, pageIndex, pageSize);
+  }, [maxTraces]);
+
+  useEffect(() => {
+    setPageIndex(0);
+  }, [tracesTableMode]);
+
+  useEffect(() => {
+    if (mode !== 'custom_data_prepper') {
+      setMaxTraces(500);
```
ps48, line 226: nit. Let's put this in constants

.cypress/integration/trace_analytics_test/trace_analytics_traces.spec.js
```diff
@@ -137,7 +137,7 @@ describe('Testing traces table', () => {
   });
```
ps48, line 137: Would be great, if we can add tests to capture the error states that we add in this PR.

public/components/trace_analytics/components/traces/trace_table_helpers.tsx
```diff
@@ -219,7 +219,7 @@ export const getTableColumns = (
         field: 'percentile_in_trace_group',
         name: 'Percentile in trace group',
         align: 'right',
-        sortable: true,
+        sortable: false,
```
ps48, line 222: Question: Why is percentile not filterable?

public/components/trace_analytics/components/traces/trace_view.tsx
```diff
@@ -204,6 +206,22 @@ export function TraceView(props: TraceViewProps) {
...
 
+  const handlePayloadResponse = (data: string) => {
+    setPayloadData(data);
+    try {
+      const parsed = JSON.parse(data);
+      if (!Array.isArray(parsed) || parsed.length === 0) {
+        setTraceIdError(true);
+        setFields({});
+      } else {
+        setTraceIdError(false);
```
ps48, line 217: Do we need the else part here? as the page starts with traceIdError: false

public/components/trace_analytics/components/common/shared_components/custom_datagrid.tsx
```diff
@@ -137,14 +137,38 @@ export const RenderCustomDataGrid: React.FC<RenderCustomDataGridParams> = ({
       )
     : [];
 
+  const isLastPage =
+    tracesTableMode === 'traces' && pagination?.pageSize && maxTraces != null
```
ps48, line 141: `maxTraces` is a number when will it be null?

public/components/trace_analytics/components/common/shared_components/custom_datagrid.tsx
```diff
@@ -137,14 +137,38 @@ export const RenderCustomDataGrid: React.FC<RenderCustomDataGridParams> = ({
...
+  }, [rowCount, maxTraces, tracesTableMode]);
+
   useInjectElementsIntoGrid(
     rowCount,
     MAX_DISPLAY_ROWS,
     tracesTableMode ?? '',
     () => {
-      setMaxTraces((prevMax: number) => Math.min(prevMax + 500, MAX_DISPLAY_ROWS));
+      // Add 500 more traces at a time until all traces present
+      const nextIncrement = Math.min(500, rowCount - maxTraces);
```
ps48, line 165: may be we can declare this function first and then add it as a param to the `useInjectElementsIntoGrid` function

public/components/trace_analytics/requests/traces_request_handler.ts
TackAdam: https://github.com/user-attachments/assets/cf041379-2f67-4ba5-ba16-5628602d563c

I believe this is not considered a "error" as it just returns 0. I think our implementation prevents that to stop the toast message spam that was occuring on the landing page before anything was set up by the user.

ps48: In that case, may be remove this dead code, if it is not reachable

TackAdam: Removed.

public/components/trace_analytics/requests/traces_request_handler.ts
TackAdam: The size 0 part prevents the more intensive aggregation part and acts similar to count.
Keeping current implementation for the time being.

public/components/trace_analytics/requests/traces_request_handler.ts
TackAdam: Clone deep is needed for the query to prevent any possible mutation as we adjust it to account for parentSpanId being empty.

public/components/trace_analytics/components/traces/trace_table_helpers.tsx
TackAdam: The percentile field is only sort-able with the currently displayed on page values as the query can not be fired off with this field. The sorting also doesn't distinguish between the trace groups. Disabling it as the sorting could be misleading as it performs differently then all the others and doesn't provide accurate information.

public/components/trace_analytics/components/traces/trace_view.tsx
TackAdam: handlePayloadResponse is part of the refresh so I think it is safer to maintain as the payload could be re-triggered if the mode changed.

public/components/trace_analytics/components/common/shared_components/custom_datagrid.tsx
TackAdam: Updated to
```
  const incrementTraceCount = () => {
    const nextIncrement = Math.min(500, rowCount - maxTraces);
    if (nextIncrement > 0 && maxTraces < MAX_DISPLAY_ROWS) {
      setMaxTraces((prevMax) => Math.min(prevMax + nextIncrement, MAX_DISPLAY_ROWS));
    }
  };

  useInjectElementsIntoGrid(
    rowCount,
    MAX_DISPLAY_ROWS,
    tracesTableMode ?? '',
    incrementTraceCount,
    maxTraces,
    isLastPage
  );
  ```

File: .cypress/integration/trace_analytics_test/trace_analytics_traces.spec.js (hunk 8 above)
Comment (TackAdam) [reply .cypress/integration/trace_analytics_test/trace_analytics_traces.spec.js:163]: Added cypress for the empty state / invalid url
Services: (Note it haded to be added to the existing {it} the test suite fails random test if any additional describes/its are added. I believe this is due to memory issues from the upgrade, will look into breaking dashboards/services/traces into separate folders to prevent this in future PR)
```
describe('Testing service view empty state and invalid url', () => {
  it('Renders service view empty state and invalid url', () => {
    cy.visit(`app/observability-traces#/services?serviceId=${SERVICE_NAME}`, {
      onBeforeLoad: (win) => {
        win.sessionStorage.clear();
      },
    });
    cy.get('[data-test-subj="globalLoadingIndicator"]').should('not.exist');
    cy.contains('frontend-client').should('exist');
    cy.contains('No matches').should('exist');

    // Renders service view invalid url state
    cy.visit(`app/observability-traces#/services?serviceId=${INVALID_URL}`, {
      onBeforeLoad: (win) => {
        win.sessionStorage.clear();
      },
    });
    cy.get('[data-test-subj="globalLoadingIndicator"]').should('not.exist');
    cy.contains(`${INVALID_URL}`).should('exist');
    cy.get('.euiCallOut.euiCallOut--danger')
      .should('exist')
      .within(() => {
        cy.get('.euiCallOutHeader__title')
          .should('contain.text', `Error loading service: ${INVALID_URL}`);
        cy.get('p')
          .should(
            'contain.text',
            'The service name is invalid or could not be found. Please check the URL or try again.'
          );
      });
    cy.contains('No matches').should('exist');
  });
});
```
Traces:
```
describe('Testing trace view invalid url', () => {
  beforeEach(() => {
    cy.visit(`app/observability-traces#/traces?traceId=${INVALID_URL}`, {
      onBeforeLoad: (win) => {
        win.sessionStorage.clear();
      },
    });
  });

  it('Handles a invalid trace url', () => {
    cy.get('[data-test-subj="globalLoadingIndicator"]').should('not.exist');
    cy.contains(`${INVALID_URL}`).should('exist');
    cy.get('.euiCallOut.euiCallOut--danger')
      .should('exist')
      .within(() => {
        cy.get('.euiCallOutHeader__title')
          .should('contain.text', `Error loading Trace Id: ${INVALID_URL}`);
        cy.get('p')
          .should(
            'contain.text',
            'The Trace Id is invalid or could not be found. Please check the URL or try again.'
          );
      });
  });
});
```

TackAdam:
PR updated to included empty state for Services, and differentiate between an empty call due to time filter and an invalid service name.


https://github.com/user-attachments/assets/e727aa58-db46-4d1b-a024-b461d4b90c36

ps48 (approved):

PR #2435, opened 2025-04-30
Traces - Add traces unique count based on cardinality aggregation

sthobis:
### Description
- Count unique traces using a cardinality aggregation on `traceId`.
- Use the unique trace count values for both `data_prepper` and `custom_data_prepper`.
- Display "warning" in the `data_prepper` view if there are more trace records than are currently shown.

https://github.com/user-attachments/assets/15403ce6-e6be-4a74-bb67-2e3a21060c7c

### Check List

### DIFF
--- a/public/components/trace_analytics/components/common/constants.tsx
+++ b/public/components/trace_analytics/components/common/constants.tsx
@@ -8,6 +8,8 @@ export const NANOS_TO_MS = 1e6;
 
 export const MILI_TO_SEC = 1000;
 
+export const MAX_DISPLAY_ROWS = 10000;
+
 export const pieChartColors = [
   '#7492e7',
   '#c33d69',
--- a/public/components/trace_analytics/components/common/shared_components/custom_datagrid.tsx
+++ b/public/components/trace_analytics/components/common/shared_components/custom_datagrid.tsx
@@ -22,7 +22,7 @@ import {
   EuiTextColor,
 } from '@elastic/eui';
 import { i18n } from '@osd/i18n';
-import React, { useEffect, useMemo, useState } from 'react';
+import React, { useMemo, useState } from 'react';
 import { NoMatchMessage } from '../helper_functions';
 import {
   TRACE_TABLE_OPTIONS,
@@ -91,8 +91,9 @@ interface RenderCustomDataGridParams {
   isTableDataLoading?: boolean;
   tracesTableMode?: string;
   setTracesTableMode?: (mode: string) => void;
-  maxTraces?: number;
-  setMaxTraces?: (max: number) => void;
+  maxTraces: number;
+  setMaxTraces: React.Dispatch<React.SetStateAction<number>>;
+  uniqueTraces: number;
 }
 
 export const RenderCustomDataGrid: React.FC<RenderCustomDataGridParams> = ({
@@ -112,6 +113,7 @@ export const RenderCustomDataGrid: React.FC<RenderCustomDataGridParams> = ({
   setTracesTableMode,
   maxTraces,
   setMaxTraces,
+  uniqueTraces,
 }) => {
   const defaultVisibleColumns = useMemo(() => {
     return columns
@@ -127,7 +129,9 @@ export const RenderCustomDataGrid: React.FC<RenderCustomDataGridParams> = ({
   const [isPopoverOpen, setIsPopoverOpen] = useState(false);
 
   const displayedRowCount =
-    tracesTableMode === 'traces' ? maxTraces : Math.min(rowCount, MAX_DISPLAY_ROWS);
+    tracesTableMode === 'traces'
+      ? Math.min(maxTraces, uniqueTraces, MAX_DISPLAY_ROWS)
+      : Math.min(rowCount, MAX_DISPLAY_ROWS);
 
   const isDarkMode = uiSettingsService.get('theme:darkMode');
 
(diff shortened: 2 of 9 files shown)


TackAdam:
Thank you for your contribution! Investigating an edge case on large data set where the cardinality count is differing between root spans and traces with this change {Root spans reading 57,794,167: Trace reading: 88,867,865}. Will update tomorrow.

Root spans:
RootSpans
Traces:
Traces

synhershko:
> Thank you for your contribution! Investigating an edge case on large data set where the cardinality count is differing between root spans and traces with this change {Root spans reading 57,794,167: Trace reading: 88,867,865}. Will update tomorrow.

@TackAdam probably due to this? https://www.elastic.co/docs/reference/aggregations/search-aggregations-metrics-cardinality-aggregation#_counts_are_approximate

I don't see a point in providing an actual number tbh. You could show an actual number if it's below 40k (the maximum you can get accurately enough) and otherwise show 10k+.

synhershko:
BTW, is there a reason why the UI components and features used are different between the Data Prepper and Custom Source views? It looks like Custom Sources is enjoying the use of the updated components also used in Discovery etc. "Data Prepper" should be a specific case of a "Custom Source", I don't see a reason to differentiate the two?

TackAdam:
> BTW, is there a reason why the UI components and features used are different between the Data Prepper and Custom Source views? It looks like Custom Sources is enjoying the use of the updated components also used in Discovery etc. "Data Prepper" should be a specific case of a "Custom Source", I don't see a reason to differentiate the two?

Custom source is an experimental feature introduced in version 2.17. It allowed us to iterate rapidly on the user experience, even with breaking changes. The team is actively working to bring this feature out of experimental status, at which point we can update and align the Data Prepper experience accordingly.

TackAdam:
> @TackAdam probably due to this? https://www.elastic.co/docs/reference/aggregations/search-aggregations-metrics-cardinality-aggregation#_counts_are_approximate
> 
> I don't see a point in providing an actual number tbh. You could show an actual number if it's below 40k (the maximum you can get accurately enough) and otherwise show 10k+.

After meeting with UX, we agreed that displaying the exact count is not particularly useful and can be replaced with a "10k+" indicator. This change enables the use of the more efficient cardinality aggregation. As a result, the existing check here: https://github.com/opensearch-project/dashboards-observability/blob/main/public/components/trace_analytics/requests/traces_request_handler.ts#L161 can safely be removed, since the exact count is no longer required.

Next steps:
1. Update the warning message to indicate that "10k+" results were retrieved for Traces / All Spans / Root Spans / Service Entry Spans.
2. Remove the redundant count aggregation.

cc: @sthobis

synhershko:
I think the message is confusing still. The paranthesis showing 500, and "500 out of ..." is incorrect. The page size is 10 not 500. We should just say "Traces ({exact number})" or "Traces (10,000+)" and remove the yellow text.

TackAdam:
@sthobis

Pending discussion on above, adjusting wording to “Results out of {number under 10,000}” and “Results out of 10,000+” per above comment.

Few minor bugs left to fix, but it’s looking good.
1. The Data Prepper page is showing the exact result not the 10k+ indicator.
2. Switching from Data Prepper to Custom Source (If Traces were selected earlier causes the “No Matches” indicator to show up beneath it, and the “Load more” functionality to not be rendered when reaching the last page) (Can also reproduce by going to Custom source “Traces” then switching to services page. Then back to Traces page.)

Video example of bugs:

https://github.com/user-attachments/assets/409d5719-23fb-4b6d-8d0e-c4f53797a09c


https://github.com/user-attachments/assets/e5454e15-52fe-41d1-8964-a596f2851a98

sthobis:
Updated 
- Update warning message wordings
- Match data_prepper warning message to custom_data_prepper
- Fix datagrid pagination/empty state on traces mode


https://github.com/user-attachments/assets/68c1eaf4-8855-43c2-9c63-985370f5af12

TackAdam:
@sthobis Great work! All bugs fixed.
Last adjustment is just for the text. We want to remove the row count from the start of the warning message for both cases {Custom Source {Spans/Traces} and DataPrepper}.
The reasoning is that we are already displaying the row count next to the mode. 
Screenshot 2025-05-06 at 10 16 04 AM
Screenshot 2025-05-06 at 10 16 14 AM


IE fixing
 {`${rowCount} results out of ${totalCountText}`}
 to {`Results out of ${totalCountText}`}
 
 Should be good to merge after those two minor changes!

TackAdam (approved):

public/components/trace_analytics/components/common/constants.tsx
```diff
@@ -8,6 +8,8 @@ export const NANOS_TO_MS = 1e6;
 
 export const MILI_TO_SEC = 1000;
 
+export const MAX_DISPLAY_ROWS = 10000;
```
ps48, line 11: Thanks for moving this to constants

public/components/trace_analytics/components/common/shared_components/custom_datagrid.tsx
```diff
@@ -91,10 +91,23 @@ interface RenderCustomDataGridParams {
   isTableDataLoading?: boolean;
   tracesTableMode?: string;
   setTracesTableMode?: (mode: string) => void;
-  maxTraces?: number;
-  setMaxTraces?: (max: number) => void;
+  maxTraces: number;
+  setMaxTraces: React.Dispatch<React.SetStateAction<number>>;
+  uniqueTraces: number;
```
ps48, line 96: I think these should remain optional as they are not used in the spans table

ps48: @sthobis this is the only change needed for the PR rest all looks good to me.

ps48 (approved):

public/components/trace_analytics/components/common/shared_components/custom_datagrid.tsx
ps48: Thanks for the change!

Issue #2506, opened 2025-09-25
[Enhancement] Time range selection for Integration Creation Flow


RyanL1997:
### Description
For OpenSearch Integration, we should a add a time range for first job refresh in integration creation flow.

### Current Behavior
Current behavior is when we create the integration assets, it is a single button clicking experience, and user just needs to enter:
- customized name of the table
- S3 bucket location
- S3 checkpoint location
- the selection of assets.

### Proposed behavior
- The there are 2 behaviors
  - default (without user provided time range): 1 day
  - following after the S3 checkpoint location, add a customized field if user wants to enter a customized time range.

The benefit of doing this is to avoid first refresh timeout for MV's first refresh job run when user has a huge dataset with long history and enhanced the creation experience, and keep leaving the flexibility for user to choose the time range if they want.

PR #2546, opened 2025-12-19
[Enhancement][Integration] Implement 1st MV Refresh Window For Integration

RyanL1997:
### Description
Implement 1st MV Refresh Window For Integration

This PR changes the original behavior for Integration creation flow:
- **Original Behavior**: When user create the integration, it will first listing all the data in S3 dir and fetch the data into the MV index that depending on the MV creation query. Since the MV creation query previously DOES NOT have any filter clause to limit the data range, so that it will always fetching from the first record from beginning

- **Enhanced Behavior**: Adding a UI component (date picker) in the integration creation flow that provide the flexibility for user to limit the data fetching range for their MV index, and it has 2 options:
  - No limit (same as old behavior, and it will fetch from the beginning)
  - Customized date (User can pick up a past date as initial range for MV fetch)

### Issues Resolved
* Resolve https://github.com/opensearch-project/dashboards-observability/issues/2506

### ScreenShot

https://github.com/user-attachments/assets/73e3efe8-8354-4e45-9852-6e61fbfe52f9

### Check List

### DIFF
--- a/public/components/integrations/components/integration_timefield_strategies.ts
+++ b/public/components/integrations/components/integration_timefield_strategies.ts
@@ -23,6 +23,10 @@
  * // Returns: "" (no filtering)
  */
 export function generateTimestampFilter(refreshRangeDays: number): string {
+  if (!Number.isInteger(refreshRangeDays) || refreshRangeDays < 0) {
+    throw new Error('refreshRangeDays must be a non-negative integer');
+  }
+
   // No filtering if refresh range is 0 (no limit)
   if (refreshRangeDays === 0) {
     return '';
--- a/public/components/integrations/components/__tests__/integration_timefield_strategies.test.ts
+++ b/public/components/integrations/components/__tests__/integration_timefield_strategies.test.ts
@@ -79,20 +79,31 @@ describe('generateTimestampFilter', () => {
   });
 
   describe('edge cases', () => {
-    it('should handle negative values by returning a future date filter', () => {
-      // This tests the mathematical behavior - negative days would result in future dates
-      // The function doesn't explicitly guard against this
-      const result = generateTimestampFilter(-1);
-
-      // -1 day means 1 day in the future: 2026-01-23 12:00:00
-      expect(result).toBe("WHERE `@timestamp` >= '2026-01-23 12:00:00'");
-    });
-
     it('should handle large values', () => {
       const result = generateTimestampFilter(365);
 
       // 365 days before 2026-01-22 = 2025-01-22
       expect(result).toBe("WHERE `@timestamp` >= '2025-01-22 12:00:00'");
+    });
+  });
+
+  describe('input validation', () => {
+    it('should throw an error for negative values', () => {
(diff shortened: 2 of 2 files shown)


public/components/integrations/components/integraiton_timefield_strategies.ts
```diff
@@ -0,0 +1,42 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
```
TackAdam: The filename is misspelled: integraiton_timefield_strategies.ts should be integration_timefield_strategies.ts
  - The actual file
  - The test file import
  - The import in setup_integration.tsx

public/components/integrations/components/integraiton_timefield_strategies.ts
```diff
@@ -0,0 +1,42 @@
...
+  if (refreshRangeDays === 0) {
+    return '';
+  }
+
+  // Calculate the start date based on refresh range
+  const now = new Date();
+  const startDate = new Date(now.getTime() - refreshRangeDays * 24 * 60 * 60 * 1000);
+
+  // Format as ISO-like string for SQL TIMESTAMP comparison
+  const isoString = startDate
```
TackAdam: Timezone Handling - public/components/integrations/components/integraiton_timefield_strategies.ts:36-39

  The code converts to ISO string (which is UTC) but the format might not align with how Spark SQL interprets timestamp strings. Spark SQL typically interprets timestamp literals in the session timezone or UTC depending on configuration.

  Recommendation: Add a comment explaining the timezone assumption or explicitly specify UTC in the query.

public/components/integrations/components/integration_timefield_strategies.ts
```diff
@@ -0,0 +1,42 @@
...
+  if (refreshRangeDays === 0) {
+    return '';
+  }
+
+  // Calculate the start date based on refresh range
+  const now = new Date();
+  const startDate = new Date(now.getTime() - refreshRangeDays * 24 * 60 * 60 * 1000);
+
+  // Format as ISO-like string for SQL TIMESTAMP comparison
+  const isoString = startDate
```
Swiddis: in sql you should use `to_timestamp_utc` to enforce this, but EMR serverless should generally be in UTC by default anyway. no reason to change the session TZ in EMR-S.

public/components/integrations/components/setup_integration.tsx
```diff
@@ -559,6 +565,7 @@ export function SetupIntegrationForm({
     connectionTableName: integration,
     databaseName: 'default',
     enabledWorkflows: [],
+    refreshRangeDays: 7,
```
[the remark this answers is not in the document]
RyanL1997: On the UI, it should show up aa default selection on calendar. 

So something like this:
Screenshot 2026-01-22 at 2 57 14 PM


So I think it is good to leave as is.

PR #2653, opened 2026-04-21
[Feature] Alert Manager

TackAdam:
# [Feature] Alert Manager — Revision 2 (PR comments addressed)

## Description

Migrates the alert-manager plugin into `dashboards-observability` as a new Alert Manager feature area, gated behind a yml config flag. Ships three tabs — **Alerts**, **Rules**, **Routing** — backed by both OpenSearch Alerting and Prometheus/Alertmanager. Fully integrated with OSD's scoped-client, multi-data-source, breadcrumb, toast, and delete-modal infrastructure. Jest unit-test coverage included from day one.

Alert1
Altert2

### Revision 2 — changes since original PR submission

This revision addresses **all 20 PR review comments** from @lezzago, @joshuali925, and @ps48. Key changes:

- **Feature-flag migration** — moved from uiSetting to yml config: `observability.alertManager.enabled: true` (default `false`).
- **Datasource persistence refactor** — deleted `InMemoryDatasourceService` (the 30-second wipe/rebuild cache); now reads directly from `data-source` + `data-connection` saved-object types. Addresses correctness issues around in-flight-request reference invalidation (Comments 8, 11).
- **Route count reduction** — 24 → 18 custom routes. 6 datasource CRUD routes deleted (replaced by client-side saved-object reads). 4 mutation routes moved to dedicated `mutations/` subfolder. Feature-flag gated at registration time (Comment 6).
- **Frontend transport migration to APM pattern** — deleted `AlarmsApiClient` (377 LOC); replaced with query services + React hooks using `coreRefs.http` (Comment 19).
- **Type safety** — all 8 `client: any` signatures retyped to `AlertingOSClient`; `_defaultDs` mutable singleton removed from `DirectQueryPrometheusBackend` (`ds: Datasource` now passed per call) (Comments 9, 10).
- **File splits** — `alert_service.ts` (1,507 LOC) split into 4 files; 4 oversized UI files split into per-page folders (30 new files total, all ≤500 LOC except one accepted overage) (Comments 16, 20.4).
- **Types split** — `common/types/alerting/types.ts` (777 LOC) split into `opensearch_types.ts`, `prometheus_types.ts`, `unified_types.ts` (Comment 17).
- **UI/UX cleanups** — AI Monitor Wizard renamed to Monitor Template Wizard (it's a static template picker, not AI); placeholder `runbooks.example.com` URLs stripped; AI summary block removed from monitor detail flyout; beta indicator consolidated to chrome-level setBadge; ECharts tooltip XSS hardening via `escapeHtml` (Comments 2, 4, 12, 14).
- **Manifest hygiene** — `opensearch-alerting` removed from `requiredOSDataSourcePlugins` so clusters without the alerting plugin aren't blocked from other observability features (Comment 13).

See **"Per-comment resolution"** table below for the full mapping.

---

## Feature flag

```yml
# opensearch_dashboards.yml
observability.alertManager.enabled: true
```

Default: `false`. When off, zero alerting routes register, no nav entries appear, no UI mounts. Users explicitly opt in via yml config — this is a yml config, **not** a uiSetting.

Two real uiSettings remain (gated by the yml flag):

| Key | Default | Purpose |
|---|---|---|
| `observability:alertManagerMaxDatasources` | `5` (1–20) | Cap applied to the Alerts/Rules datasource facet |
| `observability:alertManagerSelectedDatasources` | `[]` | Free-string array pre-selecting datasources (resolved by name / id / directQueryName / mdsId, case-insensitive) |

Feature gating is a yml config: `observability.alertManager.enabled` (default `false`).

User datasource selection persists across refresh via localStorage. Resolution precedence: localStorage → setting → first discovered datasource. Selecting past the cap fires a warning toast with an in-toast link into the Advanced Settings row.

---

### Multi-backend alerting

A single UI operating across two backends with unified list/detail views:

- **OpenSearch Alerting backend** (`HttpOpenSearchBackend`) — CRUD of query-level monitors, alert acknowledgement, destinations management. Typed with `AlertingOSClient` throughout (no more `client: any`).
- **Prometheus / Alertmanager backend** (`DirectQueryPrometheusBackend`) — rule groups, alerts, metadata (metrics / labels / label values), Alertmanager config proxy. `ds: Datasource` passed as a per-call parameter — no mutable singleton state.

Routing picks the correct backend per datasource (OS vs Prom) via `MultiBackendAlertService`. Cross-backend aggregation for unified list views; per-backend dispatch for mutations and detail fetches.

### Multi-datasource support (saved-object backed)

- OSD data sources discovered from saved objects: `type: 'data-source'` (OS) + `type: 'data-connection'` (Prom).
- **Client-side**: `useDatasources()` hook calls `savedObjectsClient.find()` on mount; reuses the existing saved-object infrastructure (no new SO type registered).
- **Server-side**: `SavedObjectDatasourceService` adapter implements the legacy `DatasourceService` interface using `ctx.core.savedObjects.client`. A `bindRequestDatasources(ctx)` helper at the top of each route handler swaps a per-request adapter into the shared alert service via `setDatasourceService()`.
- **Stable IDs**: saved-object IDs survive restarts and discovery passes (unlike the old `ds-N` auto-increment scheme). In-flight requests with cached IDs remain valid.
- **Per-request scoped client**: every backend call resolves a per-request `AlertingOSClient` through OSD's scoped-client factory. No env-var credentials, no TLS bypass, no global HTTP client.
- **Local-cluster dedupe**: when a user's MDS entry points at localhost, its user-given title wins over the hardcoded "Local Cluster" seed.
- **Prometheus Routing-tab source selector** keyed off the stable `directQueryName`.

### Frontend transport — APM pattern

Following the established APM pattern in this repo:

- **Query services** (`public/components/alerting/query_services/`):
  - `AlertingOpenSearchService` — no-arg constructor, uses `coreRefs.http`. Methods: `listAlerts`, `listRules`, `getAlertDetail`, `getRuleDetail`.
  - `AlertingPromResourcesService` — constructor takes a connection id. Methods: `listMetricNames`, `listLabelNames`, `listLabelValues`, `getMetricMetadata`.
  - `AlertmanagerAdminService` — `getConfig(dsId?)`.
- **React hooks** (`public/components/alerting/hooks/`): 8 hooks (`useAlerts`, `useRules`, `useAlertDetail`, `useRuleDetail`, `useAlertmanagerConfig`, `useMonitorMutations`, `useDatasources`, `usePrometheusMetadata`). Each memoizes its service, fetches in `useEffect`, returns `{ data, isLoading, error, refetch }`.
- **Mutation client** (`public/components/alerting/mutations/`): thin `MonitorMutationsClient` for the 4 write endpoints only.

No `AlarmsApiClient`, no prop-drilled HTTP client, no custom HTTP abstraction.

### Server routes

**18 total routes** (down from 24), feature-flag gated:

- **14 inline** in `server/routes/alerting/index.ts`:
  - Unified views (2): `GET /unified/alerts`, `GET /unified/rules`
  - OS reads (3): `GET /opensearch/{dsId}/monitors`, `GET /opensearch/{dsId}/monitors/{monitorId}`, `GET /opensearch/{dsId}/alerts`
  - Prom reads (2): `GET /prometheus/{dsId}/rules`, `GET /prometheus/{dsId}/alerts`
  - Detail flyouts (2): `GET /rules/{dsId}/{ruleId}`, `GET /alerts/{dsId}/{alertId}`
  - Alertmanager (1): `GET /alertmanager/config`
  - Prom metadata (4): `GET /prometheus/{dsId}/metadata/{metrics,labels,label-values/{label},metric-metadata}`
- **4 mutations** in `server/routes/alerting/mutations/index.ts`:
  - `POST /opensearch/{dsId}/monitors` (create)
  - `PUT /opensearch/{dsId}/monitors/{monitorId}` (update)
  - `DELETE /opensearch/{dsId}/monitors/{monitorId}` (delete)
  - `POST /opensearch/{dsId}/monitors/{monitorId}/acknowledge` (acknowledge alerts)

When `observability.alertManager.enabled: false`, `registerAlertingRoutes` is never invoked — all 18 endpoints return 404.

---

## Pure-logic layer in `common/`

Business logic extracted from React and server into testable pure modules:

- `form_transforms.ts` — form state → API payload transforms (logs + metrics monitors)
- `filter.ts` — filter/query builders over alert rows
- `serializer.ts` — monitor/alert (de)serialization with round-trip contract
- `validators.ts` — form validation predicates
- `promql_validator.ts` — PromQL string validation

No Node APIs, no React, no `coreRefs` — these modules are unit-testable in isolation.

**Types split into 3 files** (Comment 17):

- `common/types/alerting/opensearch_types.ts` — OS monitors, alerts, destinations, raw API shapes.
- `common/types/alerting/prometheus_types.ts` — Prom rules, alerts, workspaces, Alertmanager types.
- `common/types/alerting/unified_types.ts` — cross-cutting: unified summaries, pagination, backend/service interfaces.

---

## Test coverage

**142 alerting tests across 31 files, 100% passing on Node 22**:

| Directory | Files | Tests |
|---|---|---|
| Client-side (`public/components/alerting/__tests__/`) | 18 | 48 |
| Server routes (`server/routes/alerting/__tests__/`) | 5 | 32 |
| Server services (`server/services/alerting/__tests__/`) | 8 | 62 |
| **Total alerting** | **31** | **142** |

Plus shared-logic coverage in `common/services/alerting/__tests__/` (filter, form transforms, promql validator, serializer, validators) and `common/constants/__tests__/` (alerting settings).

Verification command (from repo root):

```bash
nvm use 22 && yarn test \
  common/constants/__tests__/ \
  common/services/alerting/__tests__/ \
  public/components/alerting/__tests__/ \
  public/components/alerting/hooks/__tests__/ \
  server/routes/alerting/__tests__/ \
  server/services/alerting/__tests__/ \
  server/plugin_helper/register_settings.test.ts \
  public/plugin.test.tsx \
  --silent
```

**Test tiers:**

- **Tier 1 — pure logic** (8 suites): primitives, form transforms, validators, filters, serializer, route utils, errors, timeout error.
- **Tier 2 — API + routes + services** (13 suites): query services + hooks, 5 server route handlers, 8 server services. `alert_service.ts` split into `.routing` and `.mutations` test files for focus.
- **Tier 3 — React components** (18 suites): smoke render + 1–2 focused interactions per component; scoped discipline on large files.

Shared fixtures at `common/services/alerting/__tests__/fixtures/index.ts` — 9 domain-type literal fixtures (OS monitor/alert/destination, Prom workspace/alert/rule group, alert history, notification routing, suppression rule) plus Prom metadata samples. Tests extend existing `public/plugin.test.tsx` and `server/plugin_helper/register_settings.test.ts` rather than duplicating; no snapshots; **zero `as any` / `@ts-ignore` / `@ts-expect-error`**.

---

### `common/` — shared between browser and server (35 files)

```
common/
├── constants/
│   ├── alerting_settings.ts         # uiSettings keys + default constants
│   ├── shared.ts                    # + observabilityAlertingID/Title/PluginOrder
│   └── __tests__/
│       └── alerting_settings.test.ts
├── services/
│   └── alerting/
│       ├── filter.ts                # pure query/filter builders
│       ├── form_transforms.ts       # form state → API payload (logs + metrics)
│       ├── index.ts                 # barrel
│       ├── promql_validator.ts      # PromQL syntax checks
│       ├── serializer.ts            # monitor/alert (de)serialization
│       ├── validators.ts            # form validation predicates
│       └── __tests__/
│           ├── filter.test.ts
│           ├── form_transforms.test.ts
│           ├── promql_validator.test.ts
│           ├── serializer.test.ts
│           ├── validators.test.ts
│           └── fixtures/
│               └── index.ts         # shared test fixtures
└── types/
    └── alerting/
        ├── index.ts                 # barrel re-exporting the 3 split files
        ├── opensearch_types.ts      # OS monitors, alerts, destinations
        ├── prometheus_types.ts      # Prom rules, alerts, Alertmanager types
        └── unified_types.ts         # unified summaries, interfaces
```

### `public/` — React UI (72 files)

```
public/
├── components/
│   ├── app.tsx                              # registers alerting in page registry
│   └── alerting/
│       ├── alarms_page.tsx                  # main page (Alerts/Rules/Routing tabs)
│       ├── alert_detail_flyout.tsx          # single alert detail flyout
│       ├── alerts_charts.tsx                # alert volume / severity / timeline charts
│       ├── alerts_dashboard.tsx             # Alerts tab dashboard
│       ├── alerts_summary_cards.tsx         # top summary stat cards
│       ├── create_logs_monitor/             # split folder (9 files)
│       │   ├── index.tsx
│       │   ├── create_logs_monitor_constants.ts
│       │   ├── create_logs_monitor_types.ts
│       │   └── sections/
│       │       ├── monitor_details.tsx
│       │       ├── monitor_type.tsx
│       │       ├── query.tsx
│       │       ├── schedule.tsx
│       │       ├── trigger_item.tsx
│       │       └── triggers.tsx
│       ├── create_metrics_monitor.tsx       # metrics-monitor form
│       ├── create_monitor/                  # split folder (4 files)
│       │   ├── index.tsx
│       │   ├── create_monitor_types.ts
│       │   ├── opensearch_form_section.tsx
│       │   └── prometheus_form_section.tsx
│       ├── echarts_render.tsx               # thin ECharts wrapper
│       ├── facet_filter_panel.tsx           # left-side facet filter
│       ├── home.tsx                         # HashRouter mount (setBadge sets chrome Beta indicator)
│       ├── hooks/                           # 8 React hooks
│       │   ├── use_alert_detail.ts
│       │   ├── use_alerts.ts
│       │   ├── use_alertmanager_config.ts
│       │   ├── use_datasources.ts
│       │   ├── use_monitor_mutations.ts
│       │   ├── use_prometheus_metadata.ts
│       │   ├── use_rule_detail.ts
│       │   └── use_rules.ts
│       ├── metric_browser.tsx
│       ├── monitor_detail_flyout.tsx        # (AI summary block removed)
│       ├── monitor_form_components.tsx
│       ├── monitor_template_wizard/         # renamed + split folder (4 files)
│       │   ├── index.tsx                    # (was AiMonitorWizard → MonitorTemplateWizard)
│       │   ├── alert_templates.ts           # (runbooks.example.com URLs stripped)
│       │   ├── metric_discovery.ts
│       │   └── severity_helpers.ts
│       ├── monitors_table/                  # split folder (8 files)
│       │   ├── index.tsx
│       │   ├── monitors_eui_table.tsx
│       │   ├── monitors_filters_panel.tsx
│       │   ├── monitors_main_panel.tsx
│       │   ├── monitors_table_columns.tsx
│       │   ├── monitors_table_filters.tsx
│       │   ├── monitors_table_helpers.ts
│       │   └── resizable_columns.ts
│       ├── mutations/
│       │   └── monitor_mutations_client.ts  # thin wrapper for 4 write endpoints
│       ├── notification_routing_panel.tsx   # Routing tab (uses AlertmanagerAdminService)
│       ├── promql_editor.tsx
│       ├── query_services/                  # NEW — APM-pattern transport layer
│       │   ├── alerting_opensearch_service.ts
│       │   ├── alerting_prom_resources_service.ts
│       │   └── alertmanager_admin_service.ts
│       ├── shared_constants.ts              # UI constants + escapeHtml
│       └── __tests__/                       # 18 component test files
├── plugin.tsx                               # app registration (feature-gated)
├── plugin.test.tsx                          # + Alert Manager gate tests
└── plugin_helpers/
    └── plugin_nav.tsx                       # + alerting in DEFAULT_NAV_GROUPS.all (icon side nav)
```

### `server/` — Node/Hapi (33 files)

```
server/
├── plugin_helper/
│   ├── register_settings.ts                 # + 2 Alert Manager uiSettings (gated by yml)
│   └── register_settings.test.ts
├── plugin.ts                                # observabilityConfig.create hoisted above setupRoutes
├── routes/
│   ├── index.ts                             # wires registerAlertingRoutes (flag-gated)
│   └── alerting/
│       ├── alertmanager_handlers.ts         # Alertmanager handlers (ds: Datasource threaded through)
│       ├── handlers.ts                      # read/detail handlers (dead datasource handlers deleted)
│       ├── index.ts                         # thin registrar with SO-backed resolvers
│       ├── metadata_handlers.ts             # Prom metadata route handlers
│       ├── mutations/                       # NEW — 4 mutation routes in subfolder
│       │   ├── handlers.ts
│       │   └── index.ts
│       ├── route_utils.ts                   # toErrorBody (single source)
│       └── __tests__/
│           ├── alertmanager_handlers.test.ts
│           ├── handlers.test.ts
│           ├── index.test.ts
│           ├── metadata_handlers.test.ts
│           └── route_utils.test.ts
└── services/
    └── alerting/
        ├── alert_service.ts                 # MultiBackendAlertService (split: 599 LOC, was 1499)
        ├── alert_detail.ts                  # NEW — 4 detail functions
        ├── alert_preview.ts                 # NEW — 6 preview/timeseries functions
        ├── alert_utils.ts                   # NEW — pure utils + 11 OS/Prom mappers
        ├── constants.ts
        ├── directquery_prometheus_backend.ts  # (_defaultDs removed, ds per-call)
        ├── errors.ts
        ├── index.ts                         # barrel (no InMemoryDatasourceService)
        ├── monitor_mutation_service.ts      # NEW — absorbs 4 OS mutations
        ├── opensearch_backend.ts            # (client: AlertingOSClient throughout)
        ├── prometheus_metadata_service.ts
        ├── saved_object_datasource_service.ts # NEW — SO-backed DatasourceService adapter
        ├── timeout_error.ts
        └── __tests__/
            ├── alert_service.mutations.test.ts
            ├── alert_service.routing.test.ts
            ├── directquery_prometheus_backend.test.ts
            ├── errors.test.ts
            ├── monitor_mutation_service.test.ts
            ├── opensearch_backend.test.ts
            ├── prometheus_metadata_service.test.ts
            └── timeout_error.test.ts
```

**Files deleted in this revision:**
- `server/services/alerting/datasource_service.ts` — `InMemoryDatasourceService` + 30-second wipe/rebuild loop (Comments 8, 11)
- `public/components/alerting/services/alarms_client.ts` — 377 LOC custom HTTP client (Comment 19)
- `public/components/alerting/ai_monitor_wizard.tsx` — replaced by `monitor_template_wizard/` folder (Comment 14)

---

## Other integration points

| File | Change |
|---|---|
| `opensearch_dashboards.json` | `opensearch-alerting` removed from `requiredOSDataSourcePlugins`; `supportedOSDataSourceVersions >= 2.9.0` (unchanged from repo default) |
| `package.json` | Declares `js-yaml ^4.1.1` + `@types/js-yaml ^4.0.5` (previously resolved transitively). Other dep changes in `package.json` landed in separate PR #2652 (TypeScript 6.0.2 migration). |
| `yarn.lock` | Lockfile update for js-yaml |

---

## Notable architectural decisions

- **Scoped client throughout** — every backend call resolves a per-request `AlertingOSClient` through OSD's scoped-client factory. No `HttpClient` class, no env-var credentials, no TLS-bypass defaults.
- **Types and pure logic in `common/`** — nothing in `server/` is imported by `public/`. Public components go through `common/types/alerting`.
- **Summary / Full type split** — list views type against `UnifiedAlertSummary` / `UnifiedRuleSummary`; detail flyouts re-fetch the full shape (`UnifiedAlert` / `UnifiedRule`) with backend-specific `raw` field. Prevents list code from reaching into detail-only fields.
- **Framework-agnostic handler results** — handlers return `{ error: string, status: number }` or `{ ok: ..., body: ... }` literals; `route_utils.toErrorBody` adapts to OSD's `ResponseError` shape at the route boundary.
- **Saved-object-backed datasources** — no in-memory cache; saved-object IDs are stable across restarts.
- **Per-request `DatasourceService` adapter** — `SavedObjectDatasourceService` constructed per-request with `ctx.core.savedObjects.client`, swapped into the shared alert service via `setDatasourceService()` at handler entry (`bindRequestDatasources(ctx)` helper).
- **`ds: Datasource` per-call for Prometheus** — `_defaultDs` mutable singleton removed; all 7 Alertmanager backend methods take `ds` as a parameter.
- **`client: AlertingOSClient` throughout** — zero `client: any` in `opensearch_backend.ts` (8 methods retyped).
- **Explicit `dsId` on mutations** — `createMonitor` / `updateMonitor` / `deleteMonitor` / `importMonitors` / `exportMonitors` require `dsId` (silent `'ds-1'` fallback removed). Call sites surface "select a datasource" toasts on empty selection.
- **Route schema hardening** — `monitorBodySchema`, URL-scheme restriction on datasource create/update (rejects `file://`, `javascript:`, etc.), 1000-element cap on acknowledge-alerts arrays, unknown top-level keys rejected.
- **Tab-scoped breadcrumbs** — `setNavBreadCrumbs` called per-tab; reads **Observability / Alert Manager / <tab>** (or **Alert Manager / <tab>** in nav-group mode).
- **Chrome-level Beta indicator** — `home.tsx` calls `coreRefs.chrome.setBadge({ text: 'Beta', tooltip: …, iconType: 'beaker' })` rather than embedding `{beta}` in the title constant (Comment 12).
- **ECharts tooltip XSS hardening** — `escapeHtml` wraps all user-supplied strings interpolated into tooltip HTML (Comment 4).
- **Reused Observability infrastructure** — `useToast`, `DeleteModal` (type-to-confirm destructive actions), `coreRefs.http` (no prop drilling). Matches APM / Notebooks / Traces patterns already in the repo.

---

## Per-comment resolution

| # | Comment | Phase | Status |
|---|---|---|---|
| 1 | PR desc: wrong OS version + uiSetting mis-claim | 7 | ✅ Corrected to `>=2.9.0` and yml config |
| 2 | Remove AI summary block | 7 | ✅ Block removed from `monitor_detail_flyout.tsx` |
| 3 | Missing icon side-nav group entry | 7 | ✅ Added to `DEFAULT_NAV_GROUPS.all` in `plugin_nav.tsx` |
| 4 | ECharts tooltip raw HTML | 7 | ✅ `escapeHtml` wraps interpolated strings |
| 5 | New interface with deprecated fields | 2 | ✅ `DirectQueryConfig` interface deleted |
| 6 | 24 custom routes — route-count reduction | 3 + 5 | ✅ 24 → 18, scope clarified via Slack |
| 7 | `toErrorBody` duplicated | 2 | ✅ Single source in `route_utils.ts` |
| 8 | Wipe-and-rebuild datasource IDs | 3 + 5 | ✅ SO-backed service; stable IDs |
| 9 | `_defaultDs` mutable singleton race | 2 | ✅ Removed; `ds: Datasource` per call |
| 10 | `client: any` on 8 methods | 2 | ✅ All retyped to `AlertingOSClient` |
| 11 | Auto-incrementing IDs reset | 3 + 5 | ✅ SO IDs stable across restarts |
| 12 | `{beta}` template var | 7 | ✅ Stripped; chrome-level badge remains |
| 13 | `opensearch-alerting` forced-required | 7 | ✅ Removed from `requiredOSDataSourcePlugins` |
| 14 | `runbooks.example.com` URLs + AI misnomer | 6 | ✅ 8 URLs stripped; wizard renamed |
| 15 | js-yaml v4 safety comment | 2 | ✅ Comment added above `yaml.load` |
| 16 | `alert_service.ts` 1,499 LOC | 6 | ✅ Split into 4 files (max 599 LOC) |
| 17 | `types.ts` 777 LOC | 1 | ✅ Split into 3 files |
| 18 | Unrelated package.json cleanups | — | ✅ Resolved by timeline (merged PR #2652) |
| 19 | `alarms_client.ts` should not exist | 4 | ✅ Deleted; APM-pattern services replace |
| 20 | Review summary | All | ✅ See individual comments |

---

## Quality markers

- **Lint**: 0 errors across all alerting code (`yarn lint:es public/components/alerting/ server/routes/alerting/ server/services/alerting/ common/types/alerting/ server/plugin.ts server/routes/index.ts`).
- **TypeScript**: 0 new errors introduced. Two pre-existing errors remain in non-alerting files (`common/constants/shared.ts:156` CSS.Properties, `monitor_detail_flyout.tsx:478` + `alerts_dashboard.tsx:170` ReactNode version mismatch) — these predate this PR and will clear when the repo's `@types/react` is reconciled with OSD core.
- **Tests**: 142 alerting tests / 31 files, all passing on Node 22 (repo's declared version per `.nvmrc`).
- **Strictness**: zero `as any`, zero `@ts-ignore`, zero `@ts-expect-error` in this PR's additions.
- **Dependencies**: no new npm deps beyond declaring `js-yaml` (previously resolved transitively from the OSD root).
- **Source conventions**: OSD file headers, `__tests__/` directory form, `data-test-subj` attribute convention, RTL (no Enzyme), EUI components, `EuiDescribedFormGroup` for settings.

---

## Follow-up work (out of scope for this PR)

- **Upstream `query_enhancements` extension** — adding `STATUS`, `RECEIVERS`, `SILENCES`, `ALERTS` resource types to `prometheus_manager.ts` would let us delete the surviving `GET /api/alerting/alertmanager/config` custom route. Scoped to a follow-up PR in `src/plugins/query_enhancements/`.
- **Pre-existing ReactNode type mismatches** — three files with `@types/react` version conflicts will clear when the root package's type packages are reconciled.

### Issues Resolved

### Check List


lezzago:
PR description mismatches code:                                                                                                         
  - `supportedOSDataSourceVersions` is >=2.9.0 in the manifest, not 2.13.0.
  - `observability:alertManagerEnabled` is not a uiSetting — it is a yaml config.

lezzago:
Should we remove the ai summary piece or just have it disabled in the code for the monitor details flyout since we dont have any content to put there now?

lezzago (approved):

public/plugin_helpers/plugin_nav.tsx
```diff
@@ -26,7 +27,8 @@ import { AppPluginStartDependencies } from '../types';
 export function registerAllPluginNavGroups(
```
joshuali925, line 27: where's the new registerIconSideNavGroups?

public/components/alerting/shared_constants.ts
```diff
@@ -0,0 +1,267 @@
...
+  if (seconds < 1) return `${Math.round(seconds * 1000)}ms`;
+  return `${seconds.toFixed(2)}s`;
+}
+
+// ============================================================================
+// HTML escaping
+// ============================================================================
+
+/**
+ * Escape HTML special characters to prevent XSS in ECharts tooltip formatters
```
joshuali925, line 240: why does echarts tooltip render raw html?

server/services/alerting/directquery_prometheus_backend.ts
```diff
@@ -0,0 +1,775 @@
...
+ * previously constructed with a DirectQueryConfig. The backend now relies on
+ * the OSD scoped client for auth and TLS, so only the logger is used.
+ */
+export interface DirectQueryConfig {
+  /** @deprecated No longer used — auth is supplied by OSD scoped client. */
+  opensearchUrl?: string;
+  /** @deprecated No longer used — auth is supplied by OSD scoped client. */
+  auth?: { username: string; password: string };
+  /** @deprecated No longer used — TLS is handled by OSD scoped client. */
+  rejectUnauthorized?: boolean;
```
joshuali925: why a new interface with deprecated fields?

server/routes/alerting/index.ts
```diff
@@ -0,0 +1,789 @@
...
+): { message: string; attributes?: Record<string, unknown> } {
+  const { error, message, ...rest } = body as { error?: unknown; message?: unknown };
+  const text =
+    typeof error === 'string' ? error : typeof message === 'string' ? message : 'An error occurred';
+  return Object.keys(rest).length > 0
+    ? { message: text, attributes: rest as Record<string, unknown> }
+    : { message: text };
+}
+
+export function registerAlertingRoutes(
```
ps48: 🔴 **Critical: These routes should be in the data plugin or query enhancements, not here.**

This function registers 24 custom `/api/alerting/*` routes with a bespoke handler→service→backend stack. The established pattern in this repo (see APM's `PPLSearchService` and `PromQLSearchService`) is to route all data reads through `search/ppl` and `search/promql` endpoints in query enhancements.

For this feature to be portable to OSD core later, reads should go through the existing query infrastructure. Only mutations (create/update/delete monitor, acknowledge alerts) are acceptable as custom routes since there's no shared mutation abstraction.

**Suggested approach:**
- OpenSearch alerting reads → `search/dsl` or `search/ppl`
- Prometheus rules/alerts/metadata → `search/promql`
- Datasource management → saved objects / data-source plugin
- Keep mutations as custom routes, clearly isolated

server/routes/alerting/index.ts
```diff
@@ -0,0 +1,789 @@
...
+} from './metadata_handlers';
+import type { PrometheusMetadataService } from '../../services/alerting/prometheus_metadata_service';
+import { handleGetAlertmanagerConfig } from './alertmanager_handlers';
+
+/**
+ * Adapt handler-result body ({ error: '...' } or arbitrary data) to OSD's
+ * ResponseError shape ({ message, attributes }). `error` → `message`, and the
+ * remaining fields become `attributes`.
+ */
+function toErrorBody(
```
ps48: 🟡 **Duplicate**: `toErrorBody` is already defined in `route_utils.ts`. This inline copy should be replaced with an import:
```ts
import { toErrorBody } from './route_utils';
```

server/routes/alerting/index.ts
```diff
@@ -0,0 +1,789 @@
...
+    if (Date.now() - lastDiscoveryTs < DISCOVERY_TTL_MS && existing.length > 0) return;
+    if (inflight) {
+      await inflight;
+      return;
+    }
+
+    inflight = (async () => {
+      try {
+        // Clear everything; we'll rebuild the list from scratch below.
+        for (const ds of existing) {
```
ps48: 🔴 **Correctness bug: Wipe-and-rebuild destroys in-flight references.**

Every 30s this deletes all datasources and re-creates them with new auto-incremented IDs (`ds-1`, `ds-2`, ...). Any request in progress that holds an old `dsId` will fail. The localStorage selection cached on the client also goes stale.

The workaround (matching by name/directQueryName) is fragile. Either:
1. Use stable IDs (e.g., deterministic hash of `name+type`), or
2. Use OSD saved objects for datasource persistence (preferred — this is what the data-source plugin already provides)

server/services/alerting/directquery_prometheus_backend.ts
```diff
@@ -0,0 +1,775 @@
...
+  // =========================================================================
+  // Alertmanager — via direct query resource APIs
+  // =========================================================================
+
+  // Alertmanager methods use the first available Prometheus datasource since
+  // they are global (not per-datasource). A `_defaultDs` is resolved lazily
+  // from whatever datasource was last used in getRuleGroups/getAlerts, or
+  // the caller can set it via setDefaultDatasource().
+
+  private _defaultDs?: Datasource;
```
ps48: 🔴 **Race condition: Mutable state on a singleton service.**

`DirectQueryPrometheusBackend` is instantiated once in `server/routes/index.ts` and shared across all requests. `setDefaultDatasource()` is called from the Alertmanager config route handler right before `handleGetAlertmanagerConfig()`. If two users hit this route concurrently with different Prometheus datasources, one user's request could execute against the other's datasource.

**Fix**: Remove `_defaultDs` mutable state. Pass the resolved datasource as a parameter to every Alertmanager method:
```ts
async getAlertmanagerStatus(client: AlertingOSClient, ds: Datasource): Promise<AlertmanagerStatus>
```
This matches how `getRuleGroups` and `getAlerts` already work — they take `ds` as a parameter.

server/services/alerting/opensearch_backend.ts
```diff
@@ -0,0 +1,349 @@
...
+    try {
+      await this.req(client, 'DELETE', `/_plugins/_alerting/monitors/${monitorId}`);
+      return true;
+    } catch (err) {
+      if (this.is404(err)) return false;
+      throw err;
+    }
+  }
+
+  async runMonitor(client: any, monitorId: string, dryRun?: boolean): Promise<unknown> {
```
ps48: 🟡 **`client: any` on 8 methods.** The PR description claims "zero `as any`" but these methods regress from the properly typed `AlertingOSClient` used in the first half of the file. `runMonitor`, `searchQuery`, `getAlerts`, `acknowledgeAlerts`, `getDestinations`, `createDestination`, `deleteDestination`, and `req` (line 251) all use `client: any`.

Should be `client: AlertingOSClient` throughout.

server/services/alerting/datasource_service.ts
```diff
@@ -0,0 +1,120 @@
...
+  AlertingOSClient,
+  Datasource,
+  DatasourceService,
+  PrometheusBackend,
+  Logger,
+} from '../../../common/types/alerting/types';
+
+export class InMemoryDatasourceService implements DatasourceService {
+  private datasources: Map<string, Datasource> = new Map();
+  private counter = 0;
```
ps48: 🟡 **Auto-incrementing counter produces unstable IDs.** `ds-${++this.counter}` resets on every server restart and every discovery wipe-rebuild cycle. Downstream code (client localStorage, in-flight requests, Routing tab selector) all cache these IDs which go stale within 30 seconds.

Consider using deterministic IDs:
```ts
const id = crypto.createHash('sha256').update(`${input.name}:${input.type}`).digest('hex').slice(0, 12);
```
Or better yet, use OSD saved objects for datasource persistence.

common/constants/shared.ts
```diff
@@ -78,6 +78,10 @@ export const observabilityNotebookID = 'observability-notebooks';
 export const observabilityNotebookTitle = 'Notebooks';
 export const observabilityNotebookPluginOrder = 5094;
 
+export const observabilityAlertingID = 'observability-alerting';
+export const observabilityAlertingTitle = 'Alert manager {beta}';
```
ps48: 🟡 `{beta}` looks like an unresolved template variable. Should this be `(Beta)` as display text, or should the beta indicator be rendered via `EuiBetaBadge` in the UI component instead of embedding it in the constant string?

opensearch_dashboards.json
```diff
@@ -33,6 +33,7 @@
   "supportedOSDataSourceVersions": ">=2.9.0",
   "requiredOSDataSourcePlugins": [
     "opensearch-sql",
-    "opensearch-observability"
+    "opensearch-observability",
+    "opensearch-alerting"
```
ps48: 🟡 **This makes `opensearch-alerting` required for the entire observability plugin's data-source compatibility**, not just the Alert Manager feature. Clusters without the alerting plugin will have other observability features (traces, metrics, notebooks) potentially blocked during data-source version checks.

Since Alert Manager is gated behind a yml flag that defaults to `false`, this dependency should be conditional or moved to `optionalPlugins` so it doesn't affect users who haven't enabled the feature.

public/components/alerting/ai_monitor_wizard.tsx
```diff
@@ -0,0 +1,1079 @@
...
+        condition: '> 85',
+        severity: 'high',
+        forDuration: '5m',
+        evaluationInterval: '1m',
+        labels: { severity: 'high', category: 'infrastructure', otel_signal: 'metrics' },
+        annotations: {
+          summary: 'High CPU usage on {{ $labels.instance }}',
+          description:
+            'CPU usage is above 85% for more than 5 minutes on {{ $labels.instance }}. Current value: {{ $value }}%',
+          runbook_url: 'https://runbooks.example.com/host/high-cpu',
```
ps48: 🟡 **Hardcoded `runbooks.example.com` URLs** will confuse users who see these in their alert annotations. These placeholder URLs should either be removed, left blank, or the template should clearly mark them as placeholders that need to be replaced.

Also: despite the "AI" name, this is a static template picker — not LLM-powered. Consider renaming to `MonitorTemplateWizard` or similar to set correct expectations.

server/routes/alerting/alertmanager_handlers.ts
```diff
@@ -0,0 +1,251 @@
...
+    }
+    const status = await promBackend.getAlertmanagerStatus(client);
+    const rawYaml = status.config?.original || '';
+
+    let parsedConfig: Record<string, unknown> | undefined;
+    let configParseError: string | undefined;
+
+    if (rawYaml) {
+      try {
+        const parsed = yaml.load(rawYaml) as Record<string, unknown> | null;
```
ps48, line 219: ℹ️ **Minor**: `yaml.load` in js-yaml v4 defaults to `DEFAULT_SCHEMA` (safe) — this is fine. But since js-yaml v3's `yaml.load` was notoriously unsafe, a brief comment here noting the v4 safety guarantee would prevent a future reviewer from flagging this as a vulnerability:
```ts
// js-yaml v4: yaml.load() defaults to safe DEFAULT_SCHEMA (no code execution)
```

server/services/alerting/alert_service.ts
```diff
@@ -0,0 +1,1499 @@
+/*
```
ps48, line 1: 🟡 **1,499 lines — should be split.** This file contains routing logic, preview data fetching, OS-specific detail resolution, Prom-specific detail resolution, unified aggregation, and Mustache template handling. The test suite already acknowledges this by splitting into `alert_service.routing.test.ts` and `alert_service.mutations.test.ts` — the implementation should be split similarly.

Suggested split:
- `alert_service.ts` — unified views + routing
- `alert_preview.ts` — preview/timeseries helpers
- `alert_detail.ts` — OS/Prom detail resolution
- `alert_utils.ts` — `extractTimestampField`, `stripRangeFilters`, `substituteMustacheTemplates`

common/types/alerting/types.ts
```diff
@@ -0,0 +1,777 @@
+/*
```
ps48: 🟡 **777 lines mixing multiple domains.** This single file contains OpenSearch alerting types, Prometheus types, unified view types, backend interfaces, service interfaces, and raw API response shapes. For maintainability, consider splitting:
- `opensearch_types.ts` — OS monitors, alerts, destinations, raw API shapes
- `prometheus_types.ts` — Prom rules, alerts, workspaces, Alertmanager types
- `unified_types.ts` — unified summaries, pagination, backend/service interfaces

package.json
```diff
@@ -32,6 +32,7 @@
     "antlr4": "4.8.0",
     "antlr4ts": "^0.5.0-alpha.4",
     "isomorphic-dompurify": "~2.26.0",
+    "js-yaml": "^4.1.1",
```
ps48, line 35: ℹ️ This diff also removes `eslint`, `@types/react-test-renderer`, and several `resolutions` entries (`**/eslint/cross-spawn`, `**/eslint/ajv`, `**/table/ajv`). These are unrelated cleanups that should be in a separate PR to keep this feature PR reviewable and bisectable.

public/components/alerting/services/alarms_client.ts
```diff
@@ -0,0 +1,377 @@
+/*
```
ps48: 🔴 **This client should not exist in its current form.** Instead of building a custom HTTP client that talks to 24 custom `/api/alerting/*` routes, this should follow the APM pattern:

1. Create a `PPLSearchService`-style class for OpenSearch alerting reads (routing through `search/ppl` or `search/dsl`)
2. Create a `PromQLSearchService`-style class for Prometheus reads (routing through `search/promql`)
3. Keep a thin mutation client only for create/update/delete/acknowledge operations

See:
- `public/components/apm/query_services/ppl_search_service.ts`
- `public/components/apm/query_services/promql_search_service.ts`

This is the key change needed to make the feature portable to query enhancements in OSD core.

ps48 (commented):
## Code Review: [Feature] Alert Manager

### Overview

This PR adds an Alert Manager feature with three tabs (Alerts, Rules, Routing) backed by OpenSearch Alerting and Prometheus/Alertmanager. 24,084 lines across 97 files with 265 tests in 41 suites. Feature-gated behind `observability.alertManager.enabled` in yml config.

---

### 🔴 CRITICAL: API Routing Pattern — Must Use `search/ppl` and `search/promql` Endpoints

This is the primary architectural concern. The PR registers **24 custom server routes** under `/api/alerting/*` with its own handler layer, service layer, and backend abstractions. **Zero** of these route through the established `search/ppl` or `search/promql` endpoints.

**How APM does it (the established pattern):**
- APM registers **zero custom backend routes**
- PPL queries go through `POST /api/enhancements/search/ppl` via `PPLSearchService`
- PromQL queries go through `POST /api/enhancements/search/promql` via `PromQLSearchService`
- Configuration is stored as saved objects (no custom CRUD endpoints)
- Data flow: `React Hook → Service Class → /api/enhancements/search/{ppl|promql} → Query Enhancements Plugin → OpenSearch/Prometheus`

**What this PR does instead:**
```
React Component → AlarmsApiClient → /api/alerting/* (24 custom routes) → handlers.ts → 
alert_service.ts → opensearch_backend.ts / directquery_prometheus_backend.ts → client.transport.request(...)
```

This creates a parallel, feature-specific backend stack that bypasses shared query infrastructure entirely. **These routes should be extended in the data plugin or query enhancements plugin**, not added here in the observability plugin. The Prometheus backend correctly uses `/_plugins/_directquery/_resources/{name}/...` under the hood, but wraps it in custom OSD routes — this should go through query enhancements instead.

**Recommendation for portability to OSD core:**

| Route Category | Count | Should Use |
|---|---|---|
| Datasource CRUD | 6 | Saved objects / data-source plugin |
| OpenSearch monitor/alert reads | 6 | Extend `search/dsl` or `search/ppl` in data plugin |
| Prometheus rules/alerts/metadata | 6 | Extend `search/promql` in query enhancements |
| Unified views | 2 | Compose client-side from above |
| Alertmanager config | 1 | Extend `search/promql` or keep (admin-only) |
| Mutations (create/update/delete/ack) | 3 | Acceptable as custom routes — isolate clearly |

Only mutations lack a shared abstraction and are acceptable as custom routes. Everything else should extend routes in the data plugin or query enhancements — matching the APM pattern — so this feature can be cleanly moved into OSD core later.

---

### 🔴 Mutable Singleton Race Condition

`DirectQueryPrometheusBackend._defaultDs` is mutable state on a singleton service. The Alertmanager config route calls `setDefaultDatasource()` then `handleGetAlertmanagerConfig()` — concurrent requests from different users can race and swap the datasource between those two lines. Pass datasource as a parameter to each method instead.

---

### 🟡 Design Issues

1. **InMemoryDatasourceService with wipe-and-rebuild** — Every 30s, all datasources are deleted and re-created with new auto-incremented IDs. In-flight requests referencing old IDs will fail. localStorage selections go stale. Should use stable IDs or saved objects.

2. **`client: any` in `opensearch_backend.ts`** — PR description claims "zero `as any`" but `HttpOpenSearchBackend` uses `client: any` on 8 methods. The first half properly types `AlertingOSClient`, but later methods regress.

3. **`opensearch-alerting` in `requiredOSDataSourcePlugins`** — Makes alerting plugin required for the entire observability plugin's data-source compatibility. Clusters without alerting installed could have other features blocked. Should be conditional or in `optionalPlugins`.

4. **Monster files** — Several files significantly exceed ~500 line OSD convention: `alert_service.ts` (1,499), `monitors_table.tsx` (1,469), `create_logs_monitor.tsx` (1,313), `create_monitor.tsx` (1,294), `ai_monitor_wizard.tsx` (1,079).

5. **777-line types file** — `common/types/alerting/types.ts` mixes OS types, Prometheus types, unified types, backend interfaces, service interfaces, and raw API response shapes. Should be split.

6. **`toErrorBody` duplicated** — Defined in `route_utils.ts` and also inlined in `server/routes/alerting/index.ts`.

7. **Package.json cleanup mixed in** — Unrelated removals of `eslint`, `@types/react-test-renderer`, and resolution entries should be a separate PR.

---

### 🟢 What Works Well

- Clean separation between common/public/server layers
- Feature gating via yml config (not just uiSettings)
- Scoped client usage throughout — no env-var credentials, no TLS bypass
- URL scheme validation, PromQL injection prevention, array size caps
- `auth`/`tls` stripped from datasource responses
- 265 tests with good three-tier structure
- Pure logic in `common/` with no React/Node dependencies

---

### Summary

The feature is well-implemented at the code level but architecturally misaligned with the codebase. **The 24 custom routes should be replaced by extending routes in the data plugin and query enhancements** — matching the APM pattern — so this feature can be cleanly moved into OSD core later. The mutable singleton and ephemeral datasource IDs are correctness bugs that should be fixed before merge.

server/routes/alerting/index.ts
lezzago: For `OpenSearch alerting reads → search/dsl or search/ppl`, should we have something a new route for alerting since its calling alerting api? For any part that needs ppl/dsql queries to be executed should go through `search/dsl or search/ppl`.

public/plugin_helpers/plugin_nav.tsx
TackAdam: Added. The alerting app is now registered to both `DEFAULT_NAV_GROUPS.observability` (left side nav) and `DEFAULT_NAV_GROUPS.all` (icon side nav), matching how `observabilityIntegrationsID` and `observabilityNotebookID` are wired.

- `public/plugin_helpers/plugin_nav.tsx` — added `observabilityAlertingID` to `DEFAULT_NAV_GROUPS.all` registration inside the existing `if (alertManagerEnabled)` block.

public/components/alerting/shared_constants.ts
TackAdam: `escapeHtml` is now applied to user-supplied strings in the tooltip formatter at `monitor_detail_flyout.tsx:137`. `shared_constants.ts` already exports an `escapeHtml` helper; the tooltip formatter there now wraps both `p.axisValue` and the formatted value. `alerts_charts.tsx` was already safe (already used `escapeHtml(p.name)` and the numeric `p.value` does not need escaping).

- `public/components/alerting/monitor_detail_flyout.tsx` — L137 formatter wraps interpolated strings in `escapeHtml(...)`.
- `public/components/alerting/shared_constants.ts` — unchanged; already exports `escapeHtml` at L240.

server/services/alerting/directquery_prometheus_backend.ts
TackAdam: Removed. The `@deprecated DirectQueryConfig` interface with its 3 deprecated fields (`opensearchUrl`, `auth`, `rejectUnauthorized`) is deleted from `directquery_prometheus_backend.ts` along with its JSDoc block and the companion `export type { DirectQueryConfig }` re-export in the alerting barrel. Grep confirms zero remaining references.

- `server/services/alerting/directquery_prometheus_backend.ts` — interface + JSDoc deleted at L51-64.
- `server/services/alerting/index.ts` — stale re-export removed.

server/routes/alerting/index.ts
TackAdam: Discussed, the clarified direction is to keep the alerting routes in `dashboards-observability` (the alerting-dashboards plugin may be deprecated, so we shouldn't depend on its routes), and focus on reducing the number of custom routes rather than relocating them. Applied that guidance:

- **6 datasource CRUD routes deleted** (Phase 3) — the client now reads datasources from the existing `'data-source'` + `'data-connection'` saved-object types via a new `useDatasources` hook; no bespoke datasource API remains.
- **4 mutation routes moved** to a dedicated `server/routes/alerting/mutations/` sub-folder (Phase 5) with a typed `MonitorMutationService` that takes `client: AlertingOSClient` throughout.
- **Feature-flag gating** added: `registerAlertingRoutes` is now wrapped in `if (alertManagerEnabled)` in `setupRoutes`; when the yml flag is `false` (the default), zero alerting routes register.

Final route count: **24 → 18** (14 reads/AM-config inline + 4 mutations).

- `server/routes/alerting/index.ts` — 581 LOC, 14 inline routes.
- `server/routes/alerting/mutations/index.ts` + `handlers.ts` — 4 mutation routes.
- `server/routes/index.ts` — `alertManagerEnabled` gate added.
- `server/plugin.ts` — `observabilityConfig.create()` hoisted above `setupRoutes` to plumb the flag.

server/routes/alerting/index.ts
TackAdam: Fixed. The inline `function toErrorBody(...)` definition is removed from `server/routes/alerting/index.ts`. `route_utils.ts` is now the single source of truth, and `index.ts` imports it at the top. All 12 existing call sites resolve against the imported symbol.

- `server/routes/alerting/route_utils.ts` — exports `toErrorBody`.
- `server/routes/alerting/index.ts` — imports instead of inlining.

server/routes/alerting/index.ts
TackAdam: Fixed. `InMemoryDatasourceService` (and its 30-second setInterval wipe/rebuild loop) is deleted entirely. Datasources are now read directly from the existing `'data-source'` and `'data-connection'` saved-object types registered by `src/plugins/data_source/`:

- **Client-side:** new `useDatasources()` hook in `public/components/alerting/hooks/use_datasources.ts` calls `savedObjectsClient.find({ type: 'data-source' })` and `.find({ type: 'data-connection' })` on mount.
- **Server-side:** new `SavedObjectDatasourceService` adapter in `server/services/alerting/saved_object_datasource_service.ts` implements the `DatasourceService` interface using the OSD saved-objects client; a `bindRequestDatasources(ctx)` helper at the top of each route handler swaps a per-request adapter into the shared alert service via `setDatasourceService()`.

**Saved-object IDs are stable across restarts**, which addresses the in-flight-request concern directly.

- `server/services/alerting/datasource_service.ts` — DELETED.
- `server/services/alerting/saved_object_datasource_service.ts` — NEW.
- `public/components/alerting/hooks/use_datasources.ts` — NEW.

server/services/alerting/directquery_prometheus_backend.ts
TackAdam: Fixed. The `private _defaultDs?: Datasource` field, `setDefaultDatasource()` setter, and `requireDefaultDs()` helper are all removed from `DirectQueryPrometheusBackend`. `ds: Datasource` is now a required second parameter on all 7 Alertmanager methods:

```ts
async getAlertmanagerStatus(client: AlertingOSClient, ds: Datasource): Promise<AlertmanagerStatus>
```

The `PrometheusBackend` interface in `common/types/alerting/prometheus_types.ts` was updated to match, and all 7 handler functions in `alertmanager_handlers.ts` now accept + forward `ds`. The single external caller in `server/routes/alerting/index.ts` resolves the datasource via `resolvePrometheusDatasource(ctx, req.query.dsId)` and passes it in per-request — no shared mutable state.

- `server/services/alerting/directquery_prometheus_backend.ts` — field + setter + helper removed; 7 methods take `ds`.
- `server/routes/alerting/alertmanager_handlers.ts` — all 7 handlers thread `ds` through.
- `common/types/alerting/prometheus_types.ts` — interface updated.

server/services/alerting/opensearch_backend.ts
TackAdam: Fixed. All 8 occurrences are now `client: AlertingOSClient`:

- `runMonitor` (L153)
- `searchQuery` (L166)
- `getAlerts` (L179)
- `acknowledgeAlerts` (L203)
- `getDestinations` (L217)
- `createDestination` (L226)
- `deleteDestination` (L236)
- `req` private helper (L251)

`grep -c "client: any" server/services/alerting/opensearch_backend.ts` returns `0`.

- `server/services/alerting/opensearch_backend.ts` — 8 signatures retyped.

server/services/alerting/datasource_service.ts
TackAdam: Resolved via the Comment 8 fix. The `ds-${++counter}` auto-increment scheme is gone — the `InMemoryDatasourceService` that owned it was deleted entirely. Datasource IDs are now saved-object IDs, which are stable across server restarts and discovery passes. See reply to Comment 8 for implementation detail.

common/constants/shared.ts
TackAdam: Fixed, added icon per discussion under new nav. Screenshot in description

opensearch_dashboards.json
TackAdam: Fixed. `opensearch-alerting` is removed from `requiredOSDataSourcePlugins` in `opensearch_dashboards.json`. The list is now exactly 2 entries (`opensearch-sql`, `opensearch-observability`) — the plugins that are genuinely required for non-alerting observability features. Runtime availability of the alerting plugin is no longer tied to the manifest; it is handled by the existing yml feature flag (`observability.alertManager.enabled`, default `false`).

Note: `opensearch-alerting` is not moved to `optionalPlugins` because that field is for OSD plugins, not OS-cluster plugins; the OSD manifest has no equivalent `optionalOSDataSourcePlugins` field, so removal is the correct path.

- `opensearch_dashboards.json` — one array entry removed.

public/components/alerting/ai_monitor_wizard.tsx
TackAdam: Both issues fixed.

- **Runbook URLs:** All 8 `runbook_url: 'https://runbooks.example.com/...',` lines are stripped from the `APPLICATION_CATALOG` array. `grep runbook_url` returns 0 in the split-out `alert_templates.ts` file.
- **Rename:** `AiMonitorWizard` → `MonitorTemplateWizard`, file moved to `public/components/alerting/monitor_template_wizard/` folder (split concurrent with Phase 6), test file `git mv` to `monitor_template_wizard.test.tsx`, and 3 consumer import sites updated. User-visible rename extended to: flyout title "AI Monitor Setup" → "Monitor Template Setup" + the "AI-assisted" badge in `create_monitor/index.tsx:280-283` → "From template" (both `onClickAriaLabel` and visible text).

- `public/components/alerting/monitor_template_wizard/` — new folder (4 files).
- `public/components/alerting/ai_monitor_wizard.tsx` — DELETED.
- Test file renamed from `ai_monitor_wizard.test.tsx` → `monitor_template_wizard.test.tsx`.

server/routes/alerting/alertmanager_handlers.ts
TackAdam: Added. The comment sits directly above the `yaml.load(rawYaml)` call in `handleGetAlertmanagerConfig`:

```ts
// js-yaml v4: yaml.load() defaults to safe DEFAULT_SCHEMA (no code execution).
const parsed = yaml.load(rawYaml) as Record<string, unknown> | null;
```

- `server/routes/alerting/alertmanager_handlers.ts` — comment at L212.

server/services/alerting/alert_service.ts
TackAdam: Split. `alert_service.ts` is now 4 files (composition split, acyclic import graph):

| File | LOC | Purpose |
|---|---|---|
| `alert_service.ts` | 599 | `MultiBackendAlertService` class only (routing + unified views + fetch helpers) |
| `alert_detail.ts` | 258 | 4 detail functions (`getRuleDetail`, `getOSRuleDetail`, `getPromRuleDetail`, `getAlertDetail`) |
| `alert_preview.ts` | 372 | 6 preview/timeseries functions |
| `alert_utils.ts` | 486 | 6 pure utilities + `requireDatasource` + 11 OS/Prom mapping helpers |

Public API of `MultiBackendAlertService` is unchanged (call sites in the route layer are untouched) — detail/preview methods on the class now thin-delegate to the sibling files. The 11 mapping helpers landed in `alert_utils.ts` rather than `alert_service.ts` to keep the import DAG acyclic (`alert_service → alert_detail → alert_utils`, with `alert_service → alert_utils` direct).

- `server/services/alerting/{alert_service,alert_detail,alert_preview,alert_utils}.ts` — new 4-file structure.

common/types/alerting/types.ts
TackAdam: Split. `common/types/alerting/types.ts` is gone; the 3 suggested files exist:

- `common/types/alerting/opensearch_types.ts` — 18 OS interfaces (monitors, alerts, destinations, raw API shapes).
- `common/types/alerting/prometheus_types.ts` — 23 Prom interfaces (rules, alerts, workspaces, Alertmanager types, metadata).
- `common/types/alerting/unified_types.ts` — 17 cross-cutting interfaces (unified summaries, pagination, backend/service interfaces).

`common/types/alerting/index.ts` now re-exports all three as a single barrel. 38 consumer files were updated to import through the barrel; zero duplicate exports across the three new files.

package.json
TackAdam: Resolved by timeline — those cleanups actually landed in a separate merged PR (#2652, commit `d39434fb` "Migrate plugin to TypeScript 6.0.2 compatibility"), not in this Alert Manager PR. `git diff HEAD package.json` on this branch is empty. No change needed here.

public/components/alerting/services/alarms_client.ts
TackAdam: Done. `public/components/alerting/services/alarms_client.ts` (377 LOC) and the entire `services/` directory are deleted. The frontend now follows the APM transport pattern:

**Query services** (under `public/components/alerting/query_services/`):
- `AlertingOpenSearchService` — no-arg constructor, uses `coreRefs.http`. Methods: `listAlerts`, `listRules`, `getAlertDetail`, `getRuleDetail`.
- `AlertingPromResourcesService` — constructor takes a connection id. Methods: `listMetricNames`, `listLabelNames`, `listLabelValues`, `getMetricMetadata`.
- `AlertmanagerAdminService` — `getConfig(dsId?)`.

**React hooks** (under `public/components/alerting/hooks/`):
- `useAlerts`, `useRules`, `useAlertDetail`, `useRuleDetail`, `useAlertmanagerConfig`, `useMonitorMutations`, `useDatasources` — each memoizes its service, fires in `useEffect`, and returns `{ data, isLoading, error, refetch }`.

**Mutation client** (under `public/components/alerting/mutations/`):
- `MonitorMutationsClient` — thin wrapper for the 4 write endpoints only (create/update/delete monitor + acknowledge alert). Mutations are the one area where the APM pattern allows custom routes — there is no shared mutation abstraction.

17 call sites across 5 components migrated. `grep -rn "AlarmsApiClient"` returns zero outside handoff docs.

- `public/components/alerting/query_services/` — NEW.
- `public/components/alerting/hooks/` — 8 hooks.
- `public/components/alerting/mutations/monitor_mutations_client.ts` — NEW.
- `public/components/alerting/services/alarms_client.ts` — DELETED.

**Follow-up fix (post-migration regression caught during manual QA):** The Phase 4 migration initially introduced a fake pagination contract on `AlertingOpenSearchService.listAlerts`/`listRules` — the client sent `page` + `pageSize` query params and expected a `PaginatedResponse<T>` envelope, but the underlying server endpoints (`/api/alerting/unified/alerts` + `/unified/rules`) have always returned a `ProgressiveResponse<T>` (`{ results, datasourceStatus, totalDatasources, completedDatasources, fetchedAt }`) with a schema that only accepts `dsIds` / `timeout` / `maxResults`. OSD's Joi schema validation rejected the unknown `page` key with a 400 "definition for this key is missing" error at runtime. Fixed by aligning the client contract to the real server shape: dropped `page` / `pageSize` parameters, switched the return type to `ProgressiveResponse<T>`, and derived per-datasource warnings from `res.datasourceStatus.filter(s => s.status === 'error')` in the caller (`alarms_page.tsx`). No server-side change required — the API was correct all along.

- `public/components/alerting/query_services/alerting_opensearch_service.ts` — updated.
- `public/components/alerting/alarms_page.tsx` — fetchAlerts / fetchRules updated.

ps48 (commented):
## Code Review — Critical & High-Priority Issues

### 🔴 Critical: `setDatasourceService()` — Shared mutable state race condition under concurrent requests

**Files:** `server/services/alerting/alert_service.ts` (lines 68–71), `server/services/alerting/prometheus_metadata_service.ts` (lines 56–58), `server/routes/alerting/index.ts` (`bindRequestDatasources` helper)

`MultiBackendAlertService` and `PrometheusMetadataService` are **singletons** — constructed once in `setupRoutes` and shared across all concurrent requests. `bindRequestDatasources()` mutates them at the start of each request handler:

```ts
function bindRequestDatasources(ctx) {
  const ds = new SavedObjectDatasourceService(ctx.core.savedObjects.client, logger);
  alertService.setDatasourceService(ds);        // ← mutates shared singleton
  metadataService.setDatasourceService(ds);      // ← mutates shared singleton
  return ds;
}
```

If two requests arrive concurrently, Request B's `setDatasourceService()` overwrites Request A's adapter **before** Request A finishes reading from `this.datasourceService`. This causes Request A to read datasources using Request B's saved-objects client — and therefore **Request B's auth context / permissions**.

**Recommendation:** Instead of mutating the singleton, either:
1. Pass the `DatasourceService` as a parameter to service methods that need it, or
2. Construct the service instances per-request (backends are stateless and can be shared):

```ts
// In each handler:
const ds = new SavedObjectDatasourceService(ctx.core.savedObjects.client, logger);
const alertSvc = new MultiBackendAlertService(ds, logger);
alertSvc.registerOpenSearch(osBackend);   // stateless, safe to share
alertSvc.registerPrometheus(promBackend); // stateless, safe to share
```

This also eliminates the fragile `null as unknown as any` bootstrap cast in `server/routes/index.ts:93–96`.

---

### 🔴 Critical: Duplicated logic between `MonitorMutationService` and `HttpOpenSearchBackend`

**Files:** `server/services/alerting/monitor_mutation_service.ts`, `server/services/alerting/opensearch_backend.ts`

`MonitorMutationService` duplicates verbatim from `HttpOpenSearchBackend`:
- `req()` helper
- `mapMonitor()` + `mapTrigger()`
- `is404()`
- `createMonitor()`, `updateMonitor()`, `deleteMonitor()`, `acknowledgeAlerts()`

If monitor mapping logic changes, it must be updated in both places. Either have `MonitorMutationService` delegate to `HttpOpenSearchBackend` (the pattern `MultiBackendAlertService` already uses for reads), or extract the shared helpers into a common utility module.

---

### 🟠 High: `monitorId` URL path parameter not validated against path traversal

**Files:** `server/routes/alerting/index.ts`, `server/routes/alerting/mutations/index.ts`

`monitorId` is validated as `schema.string()` with no pattern restriction, then interpolated directly into OS API paths:

```ts
`/_plugins/_alerting/monitors/${monitorId}`
`/_plugins/_alerting/monitors/${monitorId}/_acknowledge/alerts`
```

A `monitorId` containing `/` or `..` could cause unexpected routing through the OS transport layer. Consider adding `encodeURIComponent(monitorId)` at the transport call sites, or restricting the schema: `schema.string({ pattern: /^[a-zA-Z0-9_-]+$/ })`.

---

### 🟠 High: ECharts tooltip XSS — `escapeHtml` defined but appears unused in chart code

The PR description claims XSS hardening via `escapeHtml` in ECharts tooltips (Comment 4). The function is defined in `shared_constants.ts`, but `alerts_charts.tsx` uses `trigger: 'axis'` with the **default** formatter — no custom `formatter` function that would call `escapeHtml`. If user-supplied data (alert names, label values) appears in axis labels or series names, the default ECharts formatter could still render unsanitized HTML. Verify no tooltip formatters interpolate user data as raw HTML, or add `escapeHtml` wrapping where they do.

---

### 🟠 High: `package.json` removes `eslint` and related resolution overrides

The diff removes `"eslint": "^6.8.0"` from devDependencies and several eslint-related overrides (`**/eslint/cross-spawn`, `**/eslint/ajv`, `**/table/ajv`). It also removes `"@types/react-test-renderer": "^18.0.0"`. The PR description says these were resolved by PR #2652 (TypeScript migration) — please confirm these removals are safe if this PR merges independently of #2652, and that nothing still imports from `react-test-renderer`.

TackAdam:
Fixed new findings with merge from @lezzago and verified manually.

  🔴 Critical #1 — Shared-state race (setDatasourceService)
  Fixed. server/routes/alerting/index.ts:204-220 now has buildRequestServices(ctx) which constructs SavedObjectDatasourceService, MultiBackendAlertService, and PrometheusMetadataService fresh per request. Every handler calls it at entry. The setDatasourceService setter is removed from both singletons — grep returns zero production usages, and alert_service.routing.test.ts:169-183 is an explicit @ts-expect-error regression guard that fails compilation if the setter comes back. The null as unknown as any bootstrap in server/routes/index.ts is also gone.

  🔴 Critical #2 — MonitorMutationService duplication
  Fixed. server/services/alerting/monitor_mutation_service.ts is now a ~60-line thin delegate — each of createMonitor/updateMonitor/deleteMonitor/acknowledgeAlerts forwards straight to the injected HttpOpenSearchBackend. No more duplicated req/mapMonitor/mapTrigger/is404.

  🟠 High #1 — monitorId path traversal
  Fixed. server/routes/alerting/schema_helpers.ts adds alertingIdSchema (/^[A-Za-z0-9_-]+$/, max 128) and every dsId/monitorId/ruleId/alertId path/query param in server/routes/alerting/index.ts and server/routes/alerting/mutations/index.ts now uses it. . and / are rejected at the schema layer before ever hitting the upstream URL.

  🟠 High #2 — ECharts tooltip XSS
  Addressed. shared_constants.ts:240-272 now documents exactly which charts are safe and why. alerts_charts.tsx uses trigger: 'axis' with the default formatter, and both the axis labels (toLocaleTimeString output) and series names ('critical'/'high'/...) are code-generated, not user-supplied. The only custom formatter that interpolates data is in monitor_detail_flyout.tsx:133-138, which wraps axisValue and numeric value in escapeHtml.

  🟠 High #3 — package.json removals
  Safe. The ESLint and react-test-renderer removals came from PR #2652 (commit d39434fb), which is already merged into main — they're not part of this branch's diff. A full-repo grep for react-test-renderer returns zero hits.

  Tests
  - Server alerting: 14/14 suites, 122/122 tests pass.
  - UI alerting: 20/20 suites, 59/59 tests pass.
  - TypeScript: clean (yarn typecheck → no errors).

ps48 (approved):

sumukhswamy (approved):

PR #2661, opened 2026-05-01
fix: exclude text-only fields from default pattern field selection

### LINKED ISSUE #2645: [BUG] Logs explorer "seems" broken (version 3.5)
**What is the bug?**
Trying to run a `PPL` query in `Logs Explorer` does not show any results. It is reporting, that it cannot find any data.

But in browser debugger, I'm able to verify, that actually something is found. Running the same query in `Query Workbench` works

1. Install opensearch and opensearch-dashboards 3.5
2. Have an otel-demo running and sending to opensearch via data-prepper
3. Search for a log, that contains a `body` and a `traceId`
4. Go log Log Explorer and try to execute `source=ss4o_logs-* | where traceId='<yourtrace>'`

The actually found results should be shown

 - OS: Linux
 - Version `3.5`
 - Plugins: (defaults)

Image

- acutally returning something

Image

Image

- but then starts to break

Image

Image

Also, `body` is just a `text` field in my case. So additionally, I got this error message in those 2 that failed

```
{\n  \"error\": {\n    \"reason\": \"Error occurred in OpenSearch engine: all shards failed\",\n    \"details\": \"Shard[0]: java.lang.IllegalArgumentException: Text fields are not optimised for operations that require per-document field data like aggregations and sorting, so these operations are disabled by default. Please use a keyword field instead. Alternatively, set fielddata=true on [body] in order to load field data by uninverting the inverted index. Note that this can use significant memory.\\n\\nFor more details, please send request for Json format to see the raw response from OpenSearch engine.\",\n    \"type\": \"SearchPhaseExecutionException\"\n  },\n  \"status\": 400\n}
```

I was not yet able to check, when this started to break. But I am able to see this in a different instance, running Opensearch 2.19.x, where it seems to work

joshuali925-osdbot:
### Description
When selecting the default pattern field in Logs Explorer, the code previously picked the string field with the longest value. For SS4O logs, this often selected `body`, which is a `text` field. Text fields do not support fielddata operations like aggregations (`stats`, `take`), causing the patterns query to fail with:

> Text fields are not optimised for operations that require per-document field data like aggregations and sorting

This fix uses the PPL `describe` command to identify text-only fields and excludes them from the default pattern field selection. If no keyword fields are found, it falls back to all string fields. Also adds null safety when accessing field values.

### Issues Resolved
Resolves #2645

### Check List

### DIFF
--- a/public/components/event_analytics/hooks/use_fetch_patterns.test.ts
+++ b/public/components/event_analytics/hooks/use_fetch_patterns.test.ts
@@ -1,0 +1,266 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { renderHook, act } from '@testing-library/react';
+import { useFetchPatterns } from './use_fetch_patterns';
+import { SELECTED_PATTERN_FIELD } from '../../../../common/constants/explorer';
+
+const mockDispatch = jest.fn();
+const mockFetchEvents = jest.fn();
+
+jest.mock('react-redux', () => ({
+  ...jest.requireActual('react-redux'),
+  useDispatch: () => mockDispatch,
+  useSelector: () => ({}),
+  batch: (fn: () => void) => fn(),
+}));
+
+jest.mock('./use_fetch_events', () => ({
+  useFetchEvents: () => ({
+    isEventsLoading: false,
+    fetchEvents: mockFetchEvents,
+  }),
+}));
+
+describe('setDefaultPatternsField', () => {
+  const defaultParams = {
+    pplService: {} as any,
+    mlCommonsRCFService: {} as any,
+    requestParams: { tabId: 'tab-1' },
+  };
+
+  beforeEach(() => {
+    jest.clearAllMocks();
+    // By default, fetchEvents invokes the handler with the provided response
+    mockFetchEvents.mockImplementation((_query, _format, handler) => {
+      return Promise.resolve(handler?.({}));
+    });
+  });
+
+  it('excludes text fields and selects keyword field', async () => {
+    // describe response: body is text, traceId is not
+    mockFetchEvents.mockImplementationOnce((_q, _f, handler) => {
+      return Promise.resolve(
+        handler({
+          schema: [
+            { name: 'COLUMN_NAME', type: 'string' },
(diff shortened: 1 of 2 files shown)


ps48:
@joshuali925-osdbot can you please add a screenshot to the PR to make sure the UI loads on running a query

ps48:
@joshuali925-osdbot What change caused this regression? Was it some change in dashboards-observability, OSD core or SQL/PPL opensearch plugin?

TackAdam:
@joshuali925-osdbot 

  Please address two follow-up items on this PR:

  1. Add unit tests for setDefaultPatternsField in use_fetch_patterns.ts.

  The PR description checks the "New functionality includes testing" box, but no test file was added. Create
  public/components/event_analytics/hooks/use_fetch_patterns.test.ts (or extend an existing test if one exists) with tests that mock useFetchEvents / fetchEvents and cover at minimum:

  - describe response includes a text field (e.g. body) and a keyword field (e.g. traceId) → the text field is excluded, and the keyword field is selected as defaultPatternField.
  - describe response has no text fields → behavior matches the original (longest string field wins).
  - describe call fails or returns malformed schema → falls back gracefully without throwing, and a string field is still selected from source = ... | head 1.
  - res.jsonData[0][field.name] is undefined / null / non-string → no crash, field is skipped (treated as length 0).
  - All string fields are text-only → the fallback branch runs and picks the longest one from all string fields.
  - Verifies dispatch(changeQuery(...)) is called with the resolved SELECTED_PATTERN_FIELD.

  Follow existing testing patterns in this repo — Jest, mock react-redux useDispatch/useSelector, and mock ./use_fetch_events so fetchEvents is a jest.fn() you can script per-call (first call = describe response, second call = head 1 response).

  2. Fix the misleading comment on line 233.

  The current comment reads:
  // Fall back to all string fields if no keyword fields found

  This is inaccurate — the primary filter excludes only text-typed fields, not "non-keyword" fields (it still keeps ip,
  dates-as-strings, etc.). Change it to:

  // Fall back to including text fields if no non-text string fields were found

  Please keep the DCO sign-off on the new commit.

TackAdam (approved):

ps48 (approved):

PR #2668, opened 2026-05-06
fix: register Alert Manager in side nav when enabled

lezzago:
## Summary

`registerAllPluginNavGroups` was passed the `alertManager` flag from `plugin.tsx` but its signature didn't accept a 4th parameter, so `registerDefaultNavGroups` always fell back to its `alertManagerEnabled = false` default and the alerting nav link was never added to any nav group. The app still loaded via direct URL (because `application.register` ran unconditionally), but it was invisible in the side nav.

- Thread the flag through `registerAllPluginNavGroups` → `registerIconSideNavGroups` / `registerDefaultNavGroups`.
- Register the alerting nav link in the icon side nav branch too (previously only the default branch had the block, and even that one was unreachable).

## Test plan
- [ ] Set `observability.alertManager.enabled: true` in `opensearch_dashboards.yml`, run `yarn start`, verify "Alert manager" appears in the Observability side nav.
- [ ] Toggle the flag off, restart, verify the nav entry is gone.
- [ ] Verify in both the default nav and the icon side nav modes.

### DIFF
--- a/public/plugin_helpers/plugin_nav.tsx
+++ b/public/plugin_helpers/plugin_nav.tsx
@@ -45,7 +45,7 @@ function registerIconSideNavGroups(
         id: observabilityAlertingID,
         category: DEFAULT_APP_CATEGORIES.observabilityTools,
         showInAllNavGroup: true,
-        order: 500,
+        order: 150,
         euiIconType: 'beaker',
       },
     ]);
@@ -53,7 +53,7 @@ function registerIconSideNavGroups(
       {
         id: observabilityAlertingID,
         category: DEFAULT_APP_CATEGORIES.visualizeAndReport,
-        order: 500,
+        order: 150,
       },
     ]);
   }
--- a/public/plugin_helpers/plugin_nav.tsx
+++ b/public/plugin_helpers/plugin_nav.tsx
@@ -26,7 +26,8 @@ import { AppPluginStartDependencies } from '../types';
 
 function registerIconSideNavGroups(
   core: CoreSetup<AppPluginStartDependencies>,
-  apmEnabled: boolean
+  apmEnabled: boolean,
+  alertManagerEnabled: boolean = false
 ) {
   // Notebooks → Tools category
   core.chrome.navGroup.addNavLinksToGroup(DEFAULT_NAV_GROUPS.observability, [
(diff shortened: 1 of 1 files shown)


lezzago:
- "Parameter Order Change" (reviewer bot) — False positive. applicationMonitoringCategory was already the 3rd parameter of registerAllPluginNavGroups before this PR; only the 4th alertManagerEnabled parameter is new. Existing 3-arg callers remain valid via the default. 

- "Verify registerDefaultNavGroups handles alertManagerEnabled" (suggestions bot) — False positive. registerDefaultNavGroups already accepts alertManagerEnabled (added in #2653, commit b6e8576a). The bot only sees the diff, not the rest of the file. The bug this PR fixes is precisely that the flag wasn't being threaded through to that already-wired parameter.
 
- "Wrong Icon (beaker)" (reviewer + suggestions bots) — Intentional. beaker is already used when the app itself is registered (public/plugin.tsx:523) and Alert Manager is still in beta — there's a "Beta" chrome badge set inside the app at public/components/alerting/home.tsx:21-35. Changing the icon in only the nav link would create inconsistency; changing everywhere is a separate decision once Alert Manager exits beta.

- "Inconsistent Category" (reviewer bot) — Matches existing convention. The icon side nav uses observabilityTools / visualizeAndReport; the default nav uses undefined. This asymmetry mirrors the Notebooks registration in the same file (icon nav at line ~35 uses observabilityTools, default nav at line ~176 uses visualizeAndReport for the "all" group). I followed the Notebooks pattern deliberately to keep the two apps consistent.

- Separately, the linkchecker job failure is pre-existing flake against opentelemetry.io in files this PR doesn't touch (see #2664 for the last fix).

TackAdam:
SideNav
Current location

TackAdam (approved):
LGTM, thanks for the fix

joshuali925 (approved):

public/plugin_helpers/plugin_nav.tsx
```diff
@@ -37,6 +38,25 @@ function registerIconSideNavGroups(
       euiIconType: 'notebookApp',
     },
   ]);
+
+  if (alertManagerEnabled) {
```
ps48, line 42: does this work with and without the `opensearchDashboards.enableIconSideNav: true`

ps48: also can we keep this right below dashboards, that's the expectation from the product side.

lezzago: > does this work with and without the `opensearchDashboards.enableIconSideNav: true`

This has been verified.



> also can we keep this right below dashboards, that's the expectation from the product side.

updated

PR #2675, opened 2026-05-11
Alert Manager — Time Range Selector on the Alerts page

TackAdam:
### Description
Adds a time range selector (`EuiSuperDatePicker`) to the **Alerts** tab of Alert Manager so users can constrain the visible alerts, summary cards, and timeline chart to an arbitrary window. Picker is threaded end-to-end through the existing APM-pattern transport: `useAlerts` hook → `AlertingOpenSearchService` → `/api/alerting/unified/alerts` → `MultiBackendAlertService` → both backends.

The core design constraint was **uniform, predictable behavior across backends**. OpenSearch Alerting stores alert history in an index, so post-fetch filtering is straightforward. Prometheus is trickier — `/api/v1/alerts` only returns currently-firing alerts, and historical data has to be reconstructed from the `ALERTS{}` time series via `queryRange`. To keep row counts stable regardless of query step size or Prometheus retention, the Prometheus backend queries matrix history AND live alerts in parallel and merges them through a rule-identity dedupe.

Behind the existing `observability.alertManager.enabled` beta flag — zero impact when the flag is off.

Follow-up to #2653 (the original Alert Manager feature). No changes to mutation routes, Rules tab, Routing tab, monitor-detail flyout, or alert-preview flow.

AlertTimePickeer

---

#### State ownership and persistence

- **State** — date-math strings (`'now-24h'`, `'now'`) owned on `AlarmsPage` via two `useState` hooks. Matches the Trace Analytics `Home` precedent: strings travel through the system, epoch resolution happens once at a defined boundary.
- **Persistence** — `sessionStorage` keys `AlertManagerStartTime` / `AlertManagerEndTime`, hydrated lazily on mount, written on every `onTimeChange` and `onRefresh`. Survives page reload within the same tab; a new tab starts at the defaults.
- **Defaults** — `now-24h` → `now`. Tuned for alerts velocity: alerts are longer-lived signals than traces (APM uses `now-5m`), and 24 hours gives users a useful initial view on first load.
- **URL sync** — not implemented. Matches APM precedent; URL params are reserved for datasource selection.
- **Crash-safety and self-healing** — `parseDateMathMs` is wrapped in `useMemo` + try/catch. On corruption (manual DevTools edit, extension interference) the memo falls back to defaults, an effect resets component state + `sessionStorage` to the defaults, and the hook stops forwarding garbage to the backend. One `console.warn`, then silent recovery.

#### Propagation

Pure prop drilling: `AlarmsPage` → `AlertsDashboard` → `AlertTimeline`. No new context, no redux. `startTime` / `endTime` are forwarded as date-math strings for transport; resolved `startMs` / `endMs` are computed once at the page level. The memo depends on `refreshToken` too, so clicking Refresh on a `now`-relative range re-resolves the wall clock and the chart window slides correctly.

One resolution point, no chance of drift between the chart window and the server query.

#### `useAlerts` hook

`AlarmsPage` drives the Alerts tab entirely through `useAlerts({ dsIds, startTime, endTime, refreshToken })`. The hook uses a monotonic request-id ref so rapid picker changes can't let an older in-flight response overwrite a newer one. Matches APM-pattern transport and consolidates the loading/error/refetch lifecycle.

#### UI chrome

`EuiSuperDatePicker` + its built-in Refresh/Update button share the same flex row as the tabs, right-aligned, **only on the Alerts tab**. Rules and Routing tabs render just the tabs — no picker space wasted when the range is irrelevant. The picker's `onRefresh` captures the resolved `{ start, end }` snapshot and persists it, so a refresh can't drift against a stale hydrated value.

The picker's native button (`showUpdateButton={true}`) renders persistent text (`Update` or `Refresh` depending on whether the range is relative to `now`), matching Discover / Dashboards / the rest of OSD. APM's icon-only variant was considered but rejected as inconsistent with the rest of the product.

### API changes (additive, backward compatible)

All three alerts routes accept the range for schema-shape uniformity; only the unified route currently honors it downstream (the per-backend routes' response shape is raw `OSAlert[]` / `PromAlert[]`, not `UnifiedAlertSummary[]`, so historical reconstruction is surfaced exclusively through `/api/alerting/unified/alerts`).

| Route | New query params | Validation |
|---|---|---|
| `GET /api/alerting/unified/alerts` | `startTime?: string`, `endTime?: string` | per-field `validateDateMath` + cross-field `validateTimeRangeQuery` (rejects one-sided and inverted ranges) |
| `GET /api/alerting/opensearch/{dsId}/alerts` | same | same |
| `GET /api/alerting/prometheus/{dsId}/alerts` | same (accepted, no-op on this route — logs a debug warn when passed) | same |

Types (`common/types/alerting/unified_types.ts`):

```ts
export type DatasourceFetchFallback = 'prometheus-alerts-current-only';

interface UnifiedFetchOptions {
  // ...existing...
  /** Date-math string (e.g. "now-1h"). */
  startTime?: string;
  /** Date-math string (e.g. "now"). */
  endTime?: string;
}

interface DatasourceFetchResult<T> {
  // ...existing...
  /** OS hit the 1000-alert cap for this range (per-datasource). */
  truncated?: boolean;
  /** Historical reconstruction unavailable; using currently-firing alerts only. */
  fallback?: DatasourceFetchFallback;
  error?: string;
}
```

### OpenSearch backend — interval overlap + cap

Upstream `GET /_plugins/_alerting/monitors/alerts` does not accept a time filter, so we post-fetch filter using an **interval-overlap** predicate:

```ts
const effectiveEnd = alert.end_time ?? windowEnd; // active alerts have no end_time
const include      = alert.start_time <= windowEnd
                  && effectiveEnd      >= windowStart;
```

`windowEnd` is the resolved time window (computed once at service entry), not a fresh `Date.now()` per iteration — this keeps the filter deterministic across multi-page scans. An alert is included if its active interval intersects the picked window:

| Alert timing (window `[2pm, 4pm]`) | Included? |
|---|---|
| 1pm – 3pm (spans windowStart) | yes |
| 3pm – 5pm (spans windowEnd) | yes |
| 3pm – ongoing (no `end_time`) | yes |
| 12pm – 1pm (entirely before) | no |
| 5pm – 6pm (entirely after) | no |

Pagination is capped at **1000 post-overlap alerts per datasource**. If the cap is reached, the response carries `truncated: true` and the UI renders an `EuiCallOut`: *"Results capped at 1000 alerts. Narrow the time range for complete results."* This bounds worst-case traversal when a monitor produces tens of thousands of historical alerts. End-of-stream is detected via `pageAlerts.length < PAGE_SIZE`; we deliberately don't early-exit on the upstream `totalAlerts` count, which can be stale under heavy ingest and cut the loop before the real last page.

### Prometheus backend — historical + live, merged

Prometheus has two distinct alert sources: the `ALERTS{}` time series (history) and `/api/v1/alerts` (currently firing). For consistent results regardless of window size or cluster retention, the backend queries **both in parallel on every fetch** and merges them by rule identity.

```ts
class DirectQueryPrometheusBackend {
  async getHistoricalAlerts(
    client, ds, startEpochSec, endEpochSec, stepSec, endIsNow,
  ): Promise<{
    alerts: UnifiedAlertSummary[];
    fallback?: 'prometheus-alerts-current-only';
    error?: string;
  }>;

  // New multi-series parser for the ALERTS{} matrix:
  async queryRangeMatrix(client, ds, query, start, end, step): Promise<PromSeriesMatrix[]>;
}
```

**Matrix reconstruction** — scans each series for contiguous `value === 1` runs. The upstream query is scoped to `ALERTS{alertstate="firing"}` so pending transitions (a transient grace-period state whose sample count flaps with step size) never reach the parser — row counts are step-independent.

Edge-tolerance at the window boundaries is 1.5× step:

- **First firing sample at matrix index 0 AND within tolerance of windowStart** → `truncatedStart: true`, `startMs` clamped to windowStart. Alert was already firing before the window opened.
- **Last firing sample within tolerance of windowEnd** → `stillActiveAtRangeEnd: true`, `endMs` clamped to windowEnd, `state: 'active'`.
- **Any other case** → real sample timestamps used, `state: 'resolved'`.

Without this tolerance every episode would appear to span the full window (every alert's first matrix sample is at index 0, regardless of when the alert actually started).

**Merge step** — a live alert is suppressed only if a historical episode with the same rule identity is already flagged `stillActiveAtRangeEnd`. Rule identity is an FNV-1a hash of the full label map minus `__name__` and `alertstate` (which differ between matrix samples and the live endpoint). Live alerts in `state: 'pending'` are dropped, mirroring the matrix-side filter. Errors are isolated: a matrix failure doesn't hide live alerts, and a live failure doesn't hide historical data.

**`fallback: 'prometheus-alerts-current-only'`** is set when the matrix returned empty but live alerts exist — retention gap. Range fully in the past with no matrix data is a legitimate empty result, no banner.

**Alert ids** are `${dsId}-${alertname}-${labelHash}-${startMs}` where `labelHash` is an 8-char FNV-1a hash of the sorted full label map. Two series with the same alertname but different `service_name` / `job` / etc. produce distinct ids; the hash is deterministic and order-independent.

**Step sizing** (shared helper):

```ts
step = clamp(
  Math.max(Math.floor((endSec - startSec) / 500), evalIntervalSec ?? 15),
  15,
  300,
);
```

Target: ≤500 points per series, floor 15s (typical scrape interval), ceiling 300s. Caller computes step; backend does not re-derive it. `queryRangeMatrix` also enforces a defensive 50k-points-per-series cap to bound memory on misbehaving exporters.

### Shared helper

`common/services/alerting/time_range.ts` — pure, framework-free, consumed by both the server (`MultiBackendAlertService` at the service boundary) and the UI (`AlarmsPage` for chart-prop resolution):

```ts
parseDateMath(expr, isEndTime)       // epoch seconds, roundUp for end
parseDateMathMs(expr, isEndTime)     // epoch ms
dateMathToDSLString(expr)            // pass-through for OS DSL range
computeStep(startSec, endSec)  // formula above
validateDateMath(expr)               // per-field route validator
validateTimeRangeQuery({start, end}) // cross-field: one-sided + inverted
```

Tests pin `process.env.TZ = 'UTC'` before `@elastic/datemath` imports (TZ-dependent rounding otherwise breaks assertions on non-UTC runners).

### AlertTimeline — variable range bucketing

`AlertTimeline` accepts `startMs`/`endMs` and computes:

```ts
const targetBucketMs = 5 * 60_000;
const bucketCount    = clamp(Math.ceil(rangeMs / targetBucketMs), 12, 24);
const bucketDuration = rangeMs / bucketCount;
```

Label format adapts to range:

| Range | x-axis label |
|---|---|
| ≤ 24h | `HH:mm` |
| ≤ 7d | `MM-DD HH:mm` |
| > 7d | `MM-DD` |

### Testing

| Path | Suites | Tests |
|---|---|---|
| `common/services/alerting` | 6 | 45 |
| `public/components/alerting` | 22 | 98 |
| `server/routes/alerting` | 5 | 51 |
| `server/services/alerting` | 10 | 176 |
| **Total** | **43** | **370** |

All pass on Node 22. `yarn lint` clean across touched files.

TZ-robustness verified by re-running `common/services/alerting` under `TZ=America/Los_Angeles` — all pass.

### Manual smoke

With `observability.alertManager.enabled: true`:

1. Navigate to **Alert manager** → **Alerts** tab.
2. Confirm `EuiSuperDatePicker` renders to the right of the tabs with a built-in Refresh button.
3. Change the range → alerts table, summary cards, and timeline chart all refetch and the chart x-axis adapts.
4. Reload the page → selection persists in-tab via sessionStorage.
5. Open a new tab → selection resets to `now-24h` / `now`.
6. Switch to Rules or Routing → picker is not rendered (no clutter).
7. Corrupt `AlertManagerStartTime` in DevTools → page does not crash; state + sessionStorage silently reset to defaults.
8. For a Prometheus datasource whose retention is shorter than the picked range → per-datasource fallback banner renders; currently-firing alerts still show.
9. For an OS datasource with >1000 historical alerts in the range → `truncated` callout renders.
10. Picking a range with `endTime < startTime` or only one side supplied → route returns 400 before the handler runs.

### Dependencies

- Prometheus historical alerts require the `ALERTS` time series to be recorded. This is on by default in stock Prometheus / Cortex builds. Clusters that disable recording rules or remap `ALERTS` still show currently-firing alerts through the live-merge path, with the fallback banner noting the coverage limit.

### Check List

### DIFF
--- a/common/services/alerting/time_range.ts
+++ b/common/services/alerting/time_range.ts
@@ -106,6 +106,13 @@ export function computeStep(
  * Returns `true` for any string `@elastic/datemath` can parse (both as a
  * start AND as an end — we check both since some expressions are only
  * valid in one rounding mode).
+ *
+ * Known leniency: `@elastic/datemath` falls through to `moment()` for
+ * non-date-math strings, which accepts partial ISO forms (`'2024'`,
+ * `'2024-06'`, `'Jan 15'`, etc). This matches the permissiveness of the
+ * OpenSearch query DSL and `EuiSuperDatePicker`, so we do not tighten it
+ * here — the UI surfaces malformed-input errors through `parseDateMath*`
+ * on the handler side.
  */
 export function validateDateMath(expr: string): boolean {
   if (typeof expr !== 'string' || expr.length === 0) return false;
@@ -116,4 +123,44 @@ export function validateDateMath(expr: string): boolean {
   } catch {
     return false;
   }
+}
+
+/**
+ * Cross-field validator for a route query schema carrying optional
+ * `startTime`/`endTime` date-math strings. Called from the `validate`
+ * option on `schema.object(...)` — returns a string to reject, `undefined`
+ * to accept. Enforces two invariants beyond per-field `validateDateMath`:
+ *
+ *   1. Both fields are supplied together or neither. One-sided ranges are
+ *      ambiguous — the backend would silently treat the missing side as
+ *      "open", which is usually not what the client meant.
+ *   2. When both are supplied and resolvable, `endTime` is not strictly
+ *      before `startTime`. An inverted range passes the post-fetch filter
+ *      as "zero matches" rather than a client error, which is a footgun
+ *      when debugging an empty dashboard.
+ *
+ * Malformed inputs are already rejected by the per-field validator; a
+ * resolve failure here is treated as a no-op (let the per-field error
+ * propagate with its own message).
+ */
+export function validateTimeRangeQuery(query: {
+  startTime?: string;
+  endTime?: string;
+}): string | undefined {
+  const hasStart = typeof query.startTime === 'string' && query.startTime.length > 0;
+  const hasEnd = typeof query.endTime === 'string' && query.endTime.length > 0;
+  if (hasStart !== hasEnd) {
+    return 'startTime and endTime must be supplied together';
+  }
+  if (!hasStart || !hasEnd) return undefined;
+  try {
+    const startMs = parseDateMathMs(query.startTime as string, false);
+    const endMs = parseDateMathMs(query.endTime as string, true);
+    if (endMs < startMs) {
+      return 'endTime must be on or after startTime';
+    }
+  } catch {
+    // Per-field validator owns the malformed-input message; stay silent here.
+  }
+  return undefined;
 }
--- a/common/services/alerting/time_range.ts
+++ b/common/services/alerting/time_range.ts
@@ -1,0 +1,119 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * Shared time-range helpers — consumed by both the server (alerting routes,
+ * `MultiBackendAlertService`) and the browser (`AlarmsPage`, `useAlerts`).
+ *
+ * Pure, framework-free — no React, no http, no OSD imports. `@elastic/datemath`
+ * is already bundled in both server and browser builds (it's a direct
+ * dependency of OSD core), so this module is safe to import from either side.
+ *
+ * Design notes:
+ *   - We accept date-math strings (e.g. `"now-1h"`, `"now-7d/d"`) and absolute
+ *     ISO timestamps. The caller decides whether the string is meant to
+ *     express the start or the end of the window via the `isEndTime` flag;
+ *     date-math rounding (e.g. `"now/d"`) depends on which side of the window
+ *     we're on (start ⇒ floor, end ⇒ ceil).
+ *   - `parseDateMath*` throw on invalid input. Route-layer validators should
+ *     call `validateDateMath` first to reject malformed input with a 400
+ *     instead of crashing a handler.
+ *   - `computeStep` derives the Prometheus `query_range` step in seconds from
+ *     the window size, with a floor + ceiling + rule-eval-interval preference.
+ */
+import dateMath from '@elastic/datemath';
+
+/**
+ * Parse a date-math expression (e.g. `"now-1h"`) into an epoch-seconds value.
+ *
+ * @param expr       Date-math string or absolute ISO timestamp.
+ * @param isEndTime  When `true`, rounds UP (end-of-day semantics, so
+ *                   `"now/d"` → end-of-today); when `false`, rounds DOWN
+ *                   (`"now/d"` → start-of-today). Matches `EuiSuperDatePicker`
+ *                   semantics for the two ends of a picked window.
+ * @returns Epoch seconds (integer).
+ * @throws If the expression cannot be parsed.
+ */
+export function parseDateMath(expr: string, isEndTime: boolean): number {
+  return Math.floor(parseDateMathMs(expr, isEndTime) / 1000);
+}
+
+/**
+ * Same as {@link parseDateMath} but returns epoch milliseconds — useful for
+ * the OpenSearch backend's post-fetch filter where `start_time` / `end_time`
+ * are epoch ms, and for `new Date(ms).toISOString()` conversions.
+ *
+ * @throws If the expression cannot be parsed.
+ */
+export function parseDateMathMs(expr: string, isEndTime: boolean): number {
+  const parsed = dateMath.parse(expr, { roundUp: isEndTime });
+  if (!parsed || !parsed.isValid()) {
+    throw new Error(`Invalid date-math expression: ${expr}`);
+  }
+  return parsed.valueOf();
+}
+
+/**
+ * Pass-through helper for OpenSearch DSL `range` clauses. OpenSearch's
+ * query DSL accepts date-math natively (`"gte": "now-1h"`), so the common
+ * case is simply forwarding the string. Declared as a function for
+ * call-site clarity and to reserve a seam if we later need to strip
+ * whitespace / normalize.
+ */
+export function dateMathToDSLString(expr: string): string {
+  return expr;
+}
+
+/**
+ * Derive the Prometheus `query_range` step (seconds) for the given window.
+ *
+ * Formula:
+ *   `clamp(max(floor((end - start) / 500), evalIntervalSec ?? 15), 15, 300)`
+ *
+ * Bounds rationale:
+ *   - 15s floor: Prometheus default scrape interval; a finer step wastes
+ *     resolution on empty points.
+ *   - 300s (5min) ceiling: caps the point count per series to keep
+ *     payload + plot-render cost predictable on 30d+ ranges.
+ *   - 500-point denominator: approximately fills a `AlertTimeline` chart at
+ *     typical bucket-counts (12..24) with a few sub-pixels of dither.
+ *   - `evalIntervalSec` preference: when the rule's evaluation interval is
+ *     known, never ask Prometheus for a finer step than the source rule is
+ *     evaluated at — finer would return stale repeats.
+ */
+export function computeStep(
+  startEpochSec: number,
+  endEpochSec: number,
+  evalIntervalSec?: number
+): number {
+  const span = Math.max(0, endEpochSec - startEpochSec);
+  const pointBased = Math.floor(span / 500);
+  const eval_ = evalIntervalSec ?? 15;
+  const preferred = Math.max(pointBased, eval_);
+  // clamp to [15, 300]
+  if (preferred < 15) return 15;
+  if (preferred > 300) return 300;
+  return preferred;
+}
+
+/**
+ * Lightweight predicate used by route-layer schema validators to reject
+ * malformed date-math input at the HTTP boundary (400) instead of letting
+ * it crash a handler later.
+ *
+ * Returns `true` for any string `@elastic/datemath` can parse (both as a
+ * start AND as an end — we check both since some expressions are only
+ * valid in one rounding mode).
+ */
+export function validateDateMath(expr: string): boolean {
+  if (typeof expr !== 'string' || expr.length === 0) return false;
+  try {
+    const parsedStart = dateMath.parse(expr, { roundUp: false });
+    const parsedEnd = dateMath.parse(expr, { roundUp: true });
+    return !!(parsedStart && parsedStart.isValid() && parsedEnd && parsedEnd.isValid());
+  } catch {
+    return false;
+  }
+}
--- a/common/services/alerting/time_range.ts
+++ b/common/services/alerting/time_range.ts
@@ -70,7 +70,7 @@ export function dateMathToDSLString(expr: string): string {
  * Derive the Prometheus `query_range` step (seconds) for the given window.
  *
  * Formula:
- *   `clamp(max(floor((end - start) / 500), evalIntervalSec ?? 15), 15, 300)`
+ *   `clamp(floor((end - start) / 500), 15, 300)`
  *
  * Bounds rationale:
  *   - 15s floor: Prometheus default scrape interval; a finer step wastes
@@ -79,23 +79,14 @@ export function dateMathToDSLString(expr: string): string {
  *     payload + plot-render cost predictable on 30d+ ranges.
  *   - 500-point denominator: approximately fills a `AlertTimeline` chart at
  *     typical bucket-counts (12..24) with a few sub-pixels of dither.
- *   - `evalIntervalSec` preference: when the rule's evaluation interval is
- *     known, never ask Prometheus for a finer step than the source rule is
- *     evaluated at — finer would return stale repeats.
  */
-export function computeStep(
-  startEpochSec: number,
-  endEpochSec: number,
-  evalIntervalSec?: number
-): number {
+export function computeStep(startEpochSec: number, endEpochSec: number): number {
   const span = Math.max(0, endEpochSec - startEpochSec);
   const pointBased = Math.floor(span / 500);
-  const eval_ = evalIntervalSec ?? 15;
-  const preferred = Math.max(pointBased, eval_);
   // clamp to [15, 300]
-  if (preferred < 15) return 15;
-  if (preferred > 300) return 300;
-  return preferred;
+  if (pointBased < 15) return 15;
+  if (pointBased > 300) return 300;
+  return pointBased;
 }
 
 /**
--- a/public/components/alerting/alerts_charts.tsx
+++ b/public/components/alerting/alerts_charts.tsx
@@ -52,9 +52,10 @@ const TARGET_BUCKET_MS = 5 * 60 * 1000;
 const MIN_BUCKETS = 12;
 const MAX_BUCKETS = 24;
 
-/** Clamp `value` to `[min, max]`. Standard util; kept local to avoid a */
-/** lodash import for a two-line helper. */
-function clamp(min: number, max: number, value: number): number {
+/** Clamp `value` to `[min, max]`. Matches lodash's `(value, min, max)`
+ *  argument order so callers don't have to think twice. Kept local to
+ *  avoid a lodash import for a two-line helper. */
+function clamp(value: number, min: number, max: number): number {
   if (value < min) return min;
   if (value > max) return max;
   return value;
@@ -100,9 +101,9 @@ export const AlertTimeline: React.FC<AlertTimelineProps> = ({ alerts, startMs, e
     // produce transient oddities on first mount.
     const rangeMs = Math.max(1, endMs - startMs);
 
-    // bucketCount = clamp(12, 24, ceil(rangeMs / targetBucketMs))
+    // bucketCount = clamp(ceil(rangeMs / targetBucketMs), 12, 24)
     const rawBucketCount = Math.ceil(rangeMs / TARGET_BUCKET_MS);
-    const bucketCount = clamp(MIN_BUCKETS, MAX_BUCKETS, rawBucketCount);
+    const bucketCount = clamp(rawBucketCount, MIN_BUCKETS, MAX_BUCKETS);
     const bucketDuration = rangeMs / bucketCount;
 
     const buckets: Array<{
--- a/public/components/alerting/alerts_charts.tsx
+++ b/public/components/alerting/alerts_charts.tsx
@@ -7,7 +7,10 @@
  * Alerts Charts — ECharts visualizations for the Alerts dashboard.
  *
  * Currently exposes:
- *  - AlertTimeline: stacked bar chart of alerts over 24h time buckets
+ *  - AlertTimeline: stacked bar chart of alerts over a variable-range time
+ *    window. The parent owns range resolution (date-math → epoch ms) and
+ *    passes pre-resolved `startMs` / `endMs` props so this component never
+ *    re-resolves date-math on every render.
  *
  * Other breakdown panels (SeverityDonut, StateBreakdown, AlertsByDatasource,
  * AlertsByMonitor) were removed in a UI cleanup pass — the facet filter panel
@@ -35,13 +38,73 @@ const SEVERITY_COLORS: Record<string, string> = {
 // AlertTimeline — stacked bar chart by time buckets
 // ============================================================================
 
-export const AlertTimeline: React.FC<{ alerts: UnifiedAlertSummary[] }> = ({ alerts }) => {
+/** Common range thresholds (ms). */
+const ONE_HOUR_MS = 60 * 60 * 1000;
+const ONE_DAY_MS = 24 * ONE_HOUR_MS;
+const SEVEN_DAYS_MS = 7 * ONE_DAY_MS;
+
+/** Target bucket width — 5 minutes. With a 1h range, this yields exactly 12 */
+/** buckets (ceil(60m / 5m) = 12), matching the fixed bucketCount used before. */
+const TARGET_BUCKET_MS = 5 * 60 * 1000;
+
+/** Minimum / maximum bucket counts. Within this clamp the X-axis stays */
+/** readable across ranges from 5 minutes up to 30 days. */
+const MIN_BUCKETS = 12;
+const MAX_BUCKETS = 24;
+
+/** Clamp `value` to `[min, max]`. Standard util; kept local to avoid a */
+/** lodash import for a two-line helper. */
+function clamp(min: number, max: number, value: number): number {
+  if (value < min) return min;
+  if (value > max) return max;
+  return value;
+}
+
+/** Zero-pad to two digits. */
+function pad2(n: number): string {
+  return n < 10 ? `0${n}` : String(n);
+}
+
+/**
+ * Format a bucket-start timestamp based on the overall range length:
+ *  - `HH:mm` for ranges ≤ 24h
+ *  - `MM-DD HH:mm` for ranges ≤ 7d
+ *  - `MM-DD` otherwise
+ */
+function formatBucketLabel(ts: number, rangeMs: number): string {
+  const d = new Date(ts);
+  const hh = pad2(d.getHours());
+  const mm = pad2(d.getMinutes());
+  const mo = pad2(d.getMonth() + 1);
+  const da = pad2(d.getDate());
+  if (rangeMs <= ONE_DAY_MS) return `${hh}:${mm}`;
+  if (rangeMs <= SEVEN_DAYS_MS) return `${mo}-${da} ${hh}:${mm}`;
+  return `${mo}-${da}`;
+}
+
+export interface AlertTimelineProps {
+  alerts: UnifiedAlertSummary[];
+  /** Range start in epoch ms (resolved by the parent). */
+  startMs: number;
+  /** Range end in epoch ms (resolved by the parent). */
+  endMs: number;
+}
+
+export const AlertTimeline: React.FC<AlertTimelineProps> = ({ alerts, startMs, endMs }) => {
   const spec = useMemo(() => {
     if (alerts.length === 0) return null;
 
-    const now = Date.now();
-    const bucketCount = 12;
-    const bucketDuration = (24 * 60 * 60 * 1000) / bucketCount;
+    // Defend against inverted or zero-length ranges — pick a minimum 1ms
+    // window so division below doesn't blow up. In practice the parent
+    // should never send this, but the picker+sessionStorage cycle can
+    // produce transient oddities on first mount.
+    const rangeMs = Math.max(1, endMs - startMs);
+
+    // bucketCount = clamp(12, 24, ceil(rangeMs / targetBucketMs))
+    const rawBucketCount = Math.ceil(rangeMs / TARGET_BUCKET_MS);
+    const bucketCount = clamp(MIN_BUCKETS, MAX_BUCKETS, rawBucketCount);
+    const bucketDuration = rangeMs / bucketCount;
+
     const buckets: Array<{
       label: string;
       critical: number;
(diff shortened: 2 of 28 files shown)


common/services/alerting/time_range.ts
```diff
@@ -0,0 +1,166 @@
...
+ *   - 500-point denominator: approximately fills a `AlertTimeline` chart at
+ *     typical bucket-counts (12..24) with a few sub-pixels of dither.
+ *   - `evalIntervalSec` preference: when the rule's evaluation interval is
+ *     known, never ask Prometheus for a finer step than the source rule is
+ *     evaluated at — finer would return stale repeats.
+ */
+export function computeStep(
+  startEpochSec: number,
+  endEpochSec: number,
+  evalIntervalSec?: number
```
lezzago: `evalIntervalSec` seems to never be used

public/components/alerting/alerts_charts.tsx
```diff
@@ -52,12 +116,9 @@ export const AlertTimeline: React.FC<{ alerts: UnifiedAlertSummary[] }> = ({ ale
...
     for (let i = 0; i < bucketCount; i++) {
-      const bucketStart = now - (bucketCount - i) * bucketDuration;
+      const bucketStart = startMs + i * bucketDuration;
       const bucketEnd = bucketStart + bucketDuration;
-      const label = new Date(bucketStart).toLocaleTimeString([], {
-        hour: '2-digit',
-        minute: '2-digit',
-      });
+      const label = formatBucketLabel(bucketStart, rangeMs);
       const inBucket = alerts.filter((a) => {
```
lezzago: **Chart bucketing by `startTime` only disagrees with the backend overlap filter**

The OS backend's interval-overlap predicate (`opensearch_backend.ts:~265`) returns any alert whose active interval overlaps the picked window — including alerts that started before `windowStart` and are still firing (or resolved inside the window). Those show up in the table and summary-cards counts.

The chart bucket predicate here is strict start-time containment:

```ts
const t = new Date(a.startTime).getTime();
return t >= bucketStart && t < bucketEnd;
```

So an alert that started at 8am and resolved at 11am, viewed through a 10am→noon picker, is returned by the backend (overlap) but contributes to zero buckets here. A user sees "1 alert" in the summary cards and "0 bars" on the timeline and assumes the chart is broken.

Suggested fix: clamp each alert's effective start to `max(windowStart, a.startTime)` for bucketing purposes. Pre-window alerts get credited to the first bucket; post-window alerts stay excluded. This also folds in a perf win — precompute the clamped bucket-start once per alert instead of re-parsing per-bucket:

```ts
const alertBucketStart = alerts.map((a) =>
  Math.max(startMs, new Date(a.startTime).getTime())
);
// ...inside the loop:
const inBucket = alerts.filter((_, idx) =>
  alertBucketStart[idx] >= bucketStart && alertBucketStart[idx] < bucketEnd
);
```

Worth a test pinning the pre-window-start case — `alerts_charts.test.tsx` already has the bucketing scaffolding.

server/services/alerting/alert_utils.ts
```diff
@@ -301,6 +301,144 @@ export function promAlertToUnified(a: PromAlert, dsId: string): UnifiedAlertSumm
...
+ * UI can render a badge without needing to extend `UnifiedAlertSummary`.
+ */
+export function promEpisodeToUnified(ep: PromAlertEpisode, dsId: string): UnifiedAlertSummary {
+  const name = ep.labels.alertname || 'Unknown';
+  // Plan override: historical episodes without `labels.severity` fall back
+  // to `'medium'` rather than `'info'` (the live-alert default). When
+  // severity IS present, defer to the existing `promSeverityFromLabels`
+  // mapping so `warning`/`page`/etc aliases stay consistent with live alerts.
+  const finalSeverity =
+    ep.labels.severity === undefined ? 'medium' : promSeverityFromLabels(ep.labels);
```
lezzago: **Severity fallback is inconsistent between `undefined` and `""`**

The block comment above promises:

> historical episodes without `labels.severity` fall back to `'medium'` rather than `'info'` (the live-alert default)

But the guard is strict `=== undefined`:

```ts
const finalSeverity =
  ep.labels.severity === undefined ? 'medium' : promSeverityFromLabels(ep.labels);
```

`ep.labels.severity === ""` (empty-string label, not uncommon in Prometheus recording rules) falls through to `promSeverityFromLabels`, which does `labels.severity || ''`, matches none of the aliases, and returns `'info'`. Same for any other unrecognized non-empty string. So two series off the same rule with `severity=""` vs. `severity=critical` sort in different classes than the comment describes.

Suggested fix — one line:

```ts
const finalSeverity = ep.labels.severity
  ? promSeverityFromLabels(ep.labels)
  : 'medium';
```

And tighten the comment to match actual behavior:

```ts
// Historical episodes with no `labels.severity` (missing key OR empty
// value) fall back to `'medium'` rather than `'info'` (the live-alert
// default). When severity IS a non-empty string, defer to
// `promSeverityFromLabels` so `warning` / `page` aliases stay consistent
// with live alerts — an unrecognized non-empty severity still falls
// through to `'info'` there, matching live-alert behavior.
```

`alert_utils.test.ts` already covers `promEpisodeToUnified`; worth one case pinning the empty-string fallback.

TackAdam: Good catch — applied the one-liner you suggested:

```ts
const finalSeverity = ep.labels.severity
  ? promSeverityFromLabels(ep.labels)
  : 'medium';
```

…and tightened the block comment to match actual behavior (missing key OR empty value ⇒ `'medium'`; non-empty but unrecognized severity still falls through to `promSeverityFromLabels` which returns `'info'`, matching live-alert behavior).

Added a test case to `alert_utils.test.ts` (`empty-string severity ⇒ fallback "medium" (treated same as missing)`) pinning the fallback for the empty-string case so recording-rule series don't regress into `'info'` again.

public/components/alerting/alerts_charts.tsx
TackAdam: Applied the suggested fix verbatim — precomputed `alertBucketStart` outside the loop (parses each `startTime` once instead of per-bucket), then bucket membership is `alertBucketStart[idx] >= bucketStart && alertBucketStart[idx] < bucketEnd`. Pre-window alerts now land in bucket 0, post-window alerts still drop (their clamped start is still `>= endMs`).

Added a dedicated test (`AlertTimeline: pre-window alerts are credited to the first bucket (matches backend overlap)`) pinning the new behavior, and updated the existing "falls outside the window" test to reflect that only post-window alerts are dropped now. Block comment in the component calls out why we clamp so future readers see the invariant.

common/services/alerting/time_range.ts
TackAdam: The param was wired in before we stopped passing rule-eval intervals from the single caller in `alert_service.ts`. Removed the param (and the `max(pointBased, eval_)` branch), simplified the formula to the plain `clamp(floor(span / 500), 15, 300)` that the call site actually exercises, and updated the docstring + tests to match. The three evalInterval test cases in `time_range.test.ts` are gone since the param no longer exists.

ps48 (commented):
## Review Comments

Overall this is well-engineered with strong test coverage (370 tests) and thoughtful handling of the dual-backend challenge. A few items to address:

---

### 1. Reuse APM's existing time-range utilities

APM already has a suite of time-range helpers in `public/components/apm/shared/utils/time_utils.ts` and a `TimeRangePicker` component in `public/components/apm/shared/components/time_filter/time_range_picker.tsx` that overlap significantly with what this PR introduces:

| APM already has | Alert Manager reinvents |
|---|---|
| `parseTimeRange({ from, to })` — parses date-math with `roundUp` for end | `parseDateMathMs(expr, isEndTime)` in `common/services/alerting/time_range.ts` |
| `getTimeInSeconds(time)` — epoch seconds from Date | `parseDateMath(expr, isEndTime)` |
| `getTimeAxisConfig(startTime, endTime)` — x-axis label format + tick intervals based on range duration (handles `<15m`, `<1h`, `<6h`, `<24h`, `<7d`, `>7d`) | `formatBucketLabel(ts, rangeMs)` in `alerts_charts.tsx` (only 3 tiers: `≤24h`, `≤7d`, `>7d`) |
| `formatDisplayTimestamp(time, uiSettingsService)` — respects `dateFormat:tz` advanced setting | `formatBucketLabel` using local `getHours()` (no timezone awareness) |
| `TimeRangePicker` component — wraps `EuiSuperDatePicker` with `onChange`/`onRefresh`/`compressed` | Inline `EuiSuperDatePicker` in `alarms_page.tsx` |
| `calculateTimeRangeDuration(start, end)` — PromQL duration string | `computeStep(startSec, endSec)` (different purpose but related) |

Can we consolidate? Specifically:
- Move the shared parsing helpers (`parseDateMathMs`, `validateDateMath`, `validateTimeRangeQuery`, `computeStep`) to a shared location that both APM and Alert Manager import — `common/services/alerting/time_range.ts` is already isomorphic, but it duplicates the core parsing that APM's `time_utils.ts` already does.
- Reuse `getTimeAxisConfig` (or a generalized version) instead of the custom `formatBucketLabel` — it has more granular tiers and is already tested.
- Consider reusing (or extracting to shared) APM's `TimeRangePicker` component rather than inlining the `EuiSuperDatePicker` setup.

---

### 2. `server/services/alerting/opensearch_backend.ts` — Pagination loop unbounded

The pagination loop fetches ALL alerts from the OS alerting API before the post-filter cap kicks in. On a cluster with 100k+ historical alerts, this could issue 1000+ sequential network requests (100 per page). Consider adding a hard ceiling on pages fetched (e.g., `startIndex >= 10_000 → break`) to bound worst-case network cost even when the 1000-alert post-filter cap hasn't been reached.

---

### 3. `public/components/alerting/alerts_charts.tsx` — Bucket labels use local timezone

`formatBucketLabel` uses `new Date(ts).getHours()` / `.getDate()` etc., which render in the browser's local timezone. The rest of OSD (Discover, Dashboards) and APM (via `formatDisplayTimestamp`) respect the `dateFormat:tz` advanced setting. This will show different times for users in different timezones looking at the same data.

---

### 4. `public/components/alerting/alarms_page.tsx` — Acknowledge lost optimistic update

The acknowledge flow now bumps `refreshToken` to trigger a full refetch instead of the previous optimistic local state update. With the Prometheus historical reconstruction path potentially being slow, this introduces noticeable latency between clicking "Acknowledge" and seeing the state change. Consider either keeping a local optimistic patch on the alerts array, or showing a per-row loading indicator during the refetch.

---

### 5. `public/components/alerting/hooks/use_alerts.ts` — No request cancellation on unmount

The hook uses a monotonic request ID to ignore stale responses (good), but doesn't abort in-flight requests when the component unmounts or deps change. With the historical reconstruction path being potentially expensive, the backend continues doing work for abandoned requests. Consider threading an `AbortController` signal through to the HTTP transport.

---

### 6. Comment verbosity in `directquery_prometheus_backend.ts`

Several multi-paragraph block comments explain WHAT the code does rather than WHY (e.g., the episode reconstruction algorithm doc, the `closeRun` closure, the `edgeToleranceMs` explanation). Per project conventions, consider moving algorithm descriptions to the PR description and keeping inline comments to non-obvious constraints only.

---

### 7. `public/components/alerting/alarms_page.tsx` — `warnings` state naming

There are now two warning sources: `warnings` (state, used for Rules) and `alertsWarnings` (derived from `useAlerts`). Consider renaming the state variable to `rulesWarnings` to make the split explicit.

---

### 8. PR description — `computeStep` signature mismatch

Nit: The PR description shows `computeStep(startSec, endSec, eval)` with an `evalIntervalSec` parameter, but the actual implementation only takes `(startEpochSec, endEpochSec)`. Worth updating the description to match.

TackAdam:
1. Reuse APM's existing time-range utilities

Took a closer look at the actual overlap and I don't think the helpers are as duplicated as the table suggests:

- parseTimeRange({ from, to }) (APM) returns { startTime: Date, endTime: Date } from a TimeRange object — different signature, different return shape, and consumed only inside APM. parseDateMathMs(expr, isEndTime) (alerting) takes a single string + a side flag and returns epoch ms, because the alerting routes need to validate one side at a time at the HTTP boundary and the OS backend needs ms for its post-fetch interval-overlap filter.
- getTimeAxisConfig (APM) returns ECharts label-format strings ('{HH}:{mm}') that ECharts evaluates per-tick on a time-typed axis. formatBucketLabel (alerting) returns precomputed strings used as values on a category-typed axis (because the alert timeline buckets alerts into a fixed N-bucket grid and renders stacked bars by bucket index, not by timestamp). They serve different chart types.
- formatDisplayTimestamp is a per-row table formatter for full timestamps; the bucket label is a chart axis tick. Different output (full date+time vs. tick label) and different consumer.
- calculateTimeRangeDuration returns a PromQL duration like "15m". computeStep returns the query_range step in seconds. The comment correctly notes these are different.
- TimeRangePicker (APM): wraps EuiSuperDatePicker with a transient loading state and an onRefresh that just fires the parent callback. The Alerts picker has different behavior: it persists the picker state per-tab in sessionStorage, syncs the picker's resolved {start,end} back into local state on refresh (so a now-relative range re-resolves the right edge instead of staying pinned to mount time), and bumps a refreshToken that the data hook treats as an effect dep. Wrapping it in APM's TimeRangePicker would require either threading those concerns through APM's component or losing them.

What I did do for 3 (timezone): bucket labels now resolve dateFormat:tz via uiSettingsService and use moment.tz(...), which is the same pattern formatDisplayTimestamp uses, so the underlying mechanism is shared even if the output isn't.

Happy to do a deeper consolidation pass as a follow-up if you want — would mean extracting a lower-level shared module that both APM's parseTimeRange and alerting's parseDateMathMs delegate to, plus migrating callers — but that's broader than this PR.

---

2. Pagination loop unbounded

Added a hard SCAN_CAP = 10_000 ceiling in opensearch_backend.ts. When startIndex reaches the cap we set truncated = true and break, so worst-case is 100 sequential page fetches regardless of cluster size.

---

3. Bucket labels use local timezone

Fixed. formatBucketLabel now reads dateFormat:tz from uiSettingsService (resolving 'Browser' to the runtime's IANA zone, same as APM's formatDisplayTimestamp) and formats with moment.tz(...) instead of Date.getHours()/getDate(). Users in different timezones will now see consistent labels.

---

4. Acknowledge lost optimistic update

Restored the optimistic flip via an ackOverrides keyed map layered on top of the hook's data. Clicking Acknowledge immediately renders the row as acknowledged, then the refetch reconciles and the override drops off (either the server confirms or the row falls out of the result set). The hook's data remains the single source of truth — overrides are render-only.

---

5. No request cancellation on unmount

Added an AbortController per effect run in use_alerts.ts; the cleanup aborts. AlertingOpenSearchService.listAlerts/listRules now accept and forward signal to http.get. Aborts are caught in the catch block (DOMException("AbortError")) and skip the error commit so picker changes don't flash a callout. Added a test that asserts the signal is forwarded and fires on unmount.

---

6. Comment verbosity

Trimmed the multi-paragraph block comments in directquery_prometheus_backend.ts. Kept the non-obvious "why" notes inline (1.5×step edge tolerance rationale, the matrix-vs-live dedupe key, why pending is filtered at the PromQL level for step-stability) and let the file-level docstring + PR description carry the algorithm description.

---

7. warnings state naming

Renamed to rulesWarnings. The Alerts-tab equivalent is already alertsWarnings (derived from the hook), so the split now reads explicitly.

---

8. PR description — computeStep signature

Updated. The implementation only takes (startEpochSec, endEpochSec); the evalIntervalSec parameter was an earlier draft that didn't ship.

server/services/alerting/alert_service.ts
```diff
@@ -53,6 +55,62 @@ import {
...
+ */
+function resolveRangeMsFromOptions(options?: {
+  startTime?: string;
+  endTime?: string;
+}): ResolvedRange | undefined {
+  if (!options?.startTime || !options.endTime) return undefined;
+  return {
+    startMs: parseDateMathMs(options.startTime, /* isEndTime */ false),
+    endMs: parseDateMathMs(options.endTime, /* isEndTime */ true),
+    endIsNow: /\bnow\b/.test(options.endTime),
```
lezzago: **`endIsNow` regex matches any `now`-prefixed expression, not only windows that include the present**

```ts
endIsNow: /\bnow\b/.test(options.endTime),
```

`\bnow\b` matches `'now'`, `'now-5m'`, `'now-1h'`, `'now-7d'` — every `now`-relative value. The doc comment above explicitly intends that behavior, but I think the premise is off: a window ending at `'now-1h'` does **not** include the present, yet we'll still call `/api/v1/alerts` (which ignores time entirely and returns whatever is firing right now) and merge those into the result.

Two consequences for `endTime: 'now-1h'`:
1. `getHistoricalAlerts` (`directquery_prometheus_backend.ts:710`) fires the live fetch and merges live additions into a window they don't belong to — alerts firing at the wall-clock moment surface in a window that ended an hour ago.
2. When the matrix is legitimately empty (no alerts in the picked past window) but the cluster has live alerts, the result carries `fallback: 'prometheus-alerts-current-only'` and the UI banner reads "historical alert data unavailable; showing currently active alerts instead" — both untrue for a past-only window.

Suggested fix — use the resolved `endMs` instead of the source string so absolute timestamps that happen to equal "now" also work:

```ts
const endMs = parseDateMathMs(options.endTime, true);
const NOW_TOLERANCE_MS = 60_000;
return {
  startMs: parseDateMathMs(options.startTime, false),
  endMs,
  endIsNow: Math.abs(endMs - Date.now()) <= NOW_TOLERANCE_MS,
};
```

Worth adding a test in `alert_service.routing.test.ts` for `endTime: 'now-1h'` asserting the live `getAlerts` is **not** called — current routing tests only cover `endTime: 'now'`.

server/services/alerting/opensearch_backend.ts
```diff
@@ -221,14 +250,65 @@ export class HttpOpenSearchBackend implements OpenSearchBackend {
...
+      // totalAlerts` when filtering — `totalAlerts` is the server's raw
+      // index count, not the filtered count. If the upstream total is
+      // stale (a common thing during heavy ingest) or differs from the
+      // actual number of alerts we'd see paginating, cutting the loop
+      // based on it can terminate BEFORE we've seen the real last page,
+      // dropping matches silently. `pageAlerts.length < PAGE_SIZE` is the
+      // authoritative end-of-stream signal; worst case we make one extra
+      // empty request on an exact PAGE_SIZE multiple, which is cheap.
       startIndex += PAGE_SIZE;
+      if (startIndex >= SCAN_CAP) {
```
lezzago, line 297: **`truncated` flag conflates two distinct backstops; UI banner text is misleading for one of them**

Two separate paths set `truncated = true`:

- **Filter cap** (line 274) — pagination found ≥ 1000 alerts overlapping the window. "Narrow the range" is the right advice.
- **Scan cap** (line 297, here) — pagination walked 10,000 raw rows and stopped, regardless of how many actually matched. The result count could be 3.

The UI in `alerts_dashboard.tsx:692` always renders the same callout:

> **"Results capped at 1000. Narrow the range."**
> "One or more datasources returned more alerts than can be shown. Pick a shorter time range or refine your filters to see all matches."

For a cluster with 80k historical alerts where only 3 fall inside `[now-15m, now]`, the user sees 3 rows in the table and a banner claiming results were capped at 1000. The advice to narrow the range also doesn't help — the range is already 15 minutes.

Two ways to disambiguate:

**(a)** Two flags on the result — e.g. `truncated` (filter cap) and `scanLimitReached` (scan cap). UI renders different copy.

**(b)** Keep one flag, but neutralize the banner copy:
> **"Search incomplete — too many alerts to scan."**
> "Narrow the time range or refine your filters and try again."

Existing test at `opensearch_backend.test.ts:3854` only covers the filter-cap path; a sparse-overlap test that hits `SCAN_CAP` would protect whichever fix lands.

lezzago (commented):
Two correctness/UX concerns flagged inline below — both worth addressing before merge. Otherwise the architecture and test coverage look solid.

server/services/alerting/alert_service.ts
TackAdam: Good catch — fixed in the latest push. Replaced /\bnow\b/.test(options.endTime) with the suggested resolved-endMs comparison and a 60s tolerance:

const startMs = parseDateMathMs(options.startTime, /* isEndTime */ false);
const endMs = parseDateMathMs(options.endTime, /* isEndTime */ true);
```
return {
  startMs,
  endMs,
  endIsNow: Math.abs(endMs - Date.now()) <= NOW_TOLERANCE_MS,
};
```
Past-only windows (now-1h, now-7d, absolute past timestamps) now resolve to endIsNow: false, so getHistoricalAlerts skips the live fetch entirely and the misleading prometheus-alerts-current-only banner stays off for past-only windows.

Updated the doc comment to explain why the simple "is now-relative" check would be wrong here. Added two routing tests in alert_service.routing.test.ts:
- endTime: 'now' ⇒ endIsNow=true (live merge enabled)
- endTime: 'now-1h' ⇒ endIsNow=false (regression guard)

server/services/alerting/opensearch_backend.ts
TackAdam: Took option (b) — kept the single flag and neutralized the banner copy in alerts_dashboard.tsx:

Search incomplete — too many alerts to scan  
Narrow the time range or refine your filters and try again.

That copy is correct for both paths (filter cap and scan cap) and the advice (narrow / refine) helps in either case. Avoided splitting into two flags because the user remediation is the same — keeping one boolean keeps the result type and the unified-fetch propagation simpler.

Added a sparse-overlap test in opensearch_backend.test.ts: 100 pages of 100 alerts that all fall outside the window. Asserts the result list is empty (filter cap never trips), truncated: true (scan cap fires), and exactly 100 transport calls (loop stops at SCAN_CAP).

lezzago (approved):

server/services/alerting/alert_utils.ts
```diff
@@ -301,6 +301,145 @@ export function promAlertToUnified(a: PromAlert, dsId: string): UnifiedAlertSumm
...
+ * `${key}\x00${value}\x01` separators are bytes that can't appear in
+ * Prometheus label names / values so `{a: "b", c: "d"}` and
+ * `{ab: "", cd: ""}` can't collide.
+ *
+ * Not cryptographic — we need a cheap deterministic discriminator for the
+ * client-side `UnifiedAlertSummary.id` field, nothing more.
+ */
+/* eslint-disable no-bitwise -- FNV-1a is defined in terms of 32-bit XOR and
+   shift operations; the whole function is bitwise by design. */
+function hashLabels(labels: Record<string, string>): string {
```
joshuali925, line 397: why do we need hash?

joshuali925 (approved):

server/services/alerting/alert_utils.ts
TackAdam: The hash exists because Prometheus alerts come from two different endpoints with different label shapes, and we have to merge them without double-counting.

1. ALERTS{} matrix samples (historical reconstruction in getHistoricalAlerts) — every series carries Prometheus-injected transport labels: __name__="ALERTS" and alertstate="firing".

2. /api/v1/alerts (currently-firing endpoint) — no __name__/alertstate; just alertname + the user-defined labels.

Both sources can describe the same rule firing right now — the matrix shows it as a still-active episode at the right edge of the window, and /api/v1/alerts shows it as a live alert. Without dedupe, that single firing rule appears twice in the unified table.

So we need an equality predicate on "same rule instance" that works across both shapes. The hash is just that key:

- Strip the transport-only labels (__name__, alertstate)
- Sort the remaining (key, value) pairs (so key order doesn't change the result)
- Hash to a stable 8-char hex string

Then the merge step in directquery_prometheus_backend.ts:getHistoricalAlerts builds a Set<hash> of episodes still firing at the window's right edge, and any live alert whose hash matches one of those is suppressed (already represented).

Two reasons it's a hash and not the raw label map as a Map key:

- Set<string> lookup is O(1) and trivial; Set<Map> doesn't work at all (reference equality), and emitting a canonical-JSON string per label set would burn allocations on every comparison in a hot loop.
- The hashed string also doubles as a stable component of UnifiedAlertSummary.id for the historical episodes (the alert ID needs to be deterministic across refetches so React keys stay stable).

The function is non-crypto (FNV-1a 32-bit) on purpose — collision resistance isn't the requirement here, just deterministic equality across the two label-shape variants.

sumukhswamy (approved):

PR #2676, opened 2026-05-13
feat(slo): SLO/SLI foundation (PR 1 — behind feature flag)

lezzago:
## Summary

PR 1 of an 8-PR SLO/SLI train. Lands the **minimum foundation** with a
forward-compatible saved-object schema so later PRs (workspace
plumbing, status aggregator, rule-health, reconciler, adoption,
services-home, polish) layer in without schema churn. Surfaces ship
**dark** by default — opt in via `observability.slo.enabled: true` in
`opensearch_dashboards.yml`, mirroring the existing
`alertManager.enabled` pattern.

> **For train-level context — read this first:**
> [SLO/SLI tracking issue](https://github.com/lezzago/dashboards-observability/issues/1)
>
> Covers the AMP-compatibility invariant, the schema design choices,
> the 8-PR roadmap, the deliberately-unwired API surface in PR 1, and
> the cross-cutting constraints every PR in the train respects. The PR
> body below is scoped to **what changes in this diff**.

Screenshot 2026-05-13 at 3 36 33 PM
Screenshot 2026-05-13 at 3 37 01 PM
Screenshot 2026-05-13 at 3 36 47 PM

## What ships in PR 1

- **Saved-object schema** designed for forward compatibility so later
  PRs add features without migration. `spec.workspaceId` is
  server-stamped (no `datasourceId` fallback);
  `PrometheusProvisioning.{recordingFingerprints, alertGroupName}` are
  required (empty map / empty string valid);
  `rule-ref:<ws>:<ds>:<fpv>:<fp>` SO id embeds fingerprintVersion so a
  future algorithm bump produces a disjoint id namespace;
  `adoptionSource` lives on `SloPersistedStatus` (outer) so it stays
  backend-agnostic; `pendingRedeploy?: { reason; queuedAt }` is carved
  out as a future-migration escape hatch (PR 1 never writes it).
  Rationale for each decision is in the [tracking issue](https://github.com/lezzago/dashboards-observability/issues/1).
- **AMP-compat invariant plumbing** — every rule group for workspace `W`
  lands in `slo-generated-<W>`. Service-layer callers resolve through
  `sloRulerNamespaceFor(workspaceId)`, which validates the workspace id
  against a path-safe regex and is the single composer of the namespace
  string. Two-workspace integration test in
  `common/slo/__tests__/slo_workspace_isolation.integration.test.ts`
  asserts the contract.
- **Read-boundary normalizer** `normalizeSloSpec` backfills missing
  `alarms.*` keys with defaults so a future alarm type lands as a type +
  default-filler change, no SO migration. Write-boundary `normalizeSpec`
  strips unknown top-level keys via `SLO_SPEC_KEYS` allow-list before
  persistence.
- **Defensive PromQL validation** on
  `customExpr.{goodQuery,totalQuery,errorRatioQuery}` — paren / brace /
  bracket balance after string-literal stripping, trailing-backslash +
  control-char rejection, 8 KiB size cap. Surfaces malformed input at
  wizard-save instead of later at ruler-write.
- **CRUD surface**: list / create / get / update / delete / enable /
  disable / preview / statuses routes, a minimal one-template wizard, a
  listing page with `EuiBasicTable` pagination + counter + refresh, and
  a skeleton detail page with `EuiDescriptionList` metadata + spec code
  block. All user-facing strings i18n-wrapped.
- **Feature flags**: `observability.slo.enabled` (default `false`) and
  `observability.slo.ruleDedup.enabled` (default `true`). Both
  registered in `server/index.ts`. `ruleAdoption.enabled` and
  `recordingGraceMs` are intentionally *not* registered — they land
  with their consumers in PRs 5/6.

## What's deliberately **not** in PR 1

The [tracking issue](https://github.com/lezzago/dashboards-observability/issues/1)
has the full deferral table. Highlights for this review:

- The route layer hard-codes `workspaceId: 'default'` (`FIXME(pr-2)` at
  `server/routes/slo/index.ts:339, :380`) — PR 2 plumbs
  `core.workspace.resolveRequest(req)` into the deploy context.
- `createdBy` uses an `'osd-user'` sentinel (`FIXME(pr-2)` at
  `server/routes/slo/index.ts:265`) — PR 2 wires the real resolver.
- **The listing page's "State" column will read `no_data` for every
  active SLO** until PR 3 wires the status aggregator. This is
  expected — the service falls through to a stub that returns
  `disabled` / `no_data` until a real `SloStatusAggregator` is set on
  the service. Worth knowing while reviewing the listing code.
- `SloService.repair`, `getFingerprintRefcounts`, `writeTombstone`, and
  the `SloStatusAggregator` / `SloTombstoneStoreLite` / `SloRuleHealthProbe`
  interfaces exist on the service but **no route consumes them and no
  store is wired in `server/plugin.ts`**. They're structural surface so
  PRs 3–6 can plug in real implementations without moving code across
  the `common/server` boundary. See the "Intentionally-unwired surface"
  table in the tracking issue.

## Security notes

- **PromQL validation** closes input shapes that confuse downstream
  parsers (unbalanced delimiters, trailing backslashes, control chars).
  YAML injection via `customExpr` was already defended by the
  generator's literal-block indent discipline (`expr: |` blocks
  terminated by dedent, every line 6-space-indented); the new checks
  surface malformed PromQL earlier.
- **Schema allow-list**: `normalizeSpec` strips unknown top-level keys
  via `SLO_SPEC_KEYS` before the spec lands in
  `SavedObjectSloStore.save`. Route-level `@osd/config-schema` stays
  lenient (`unknowns: 'allow'`) for client forward-compat, but the
  plugin never round-trips arbitrary JSON into saved-object
  attributes — blocks `__proto__` injection and unknown-attr
  smuggling.
- **Workspace id shape**: `sloRulerNamespaceFor` validates the id
  against `/^[a-zA-Z0-9][a-zA-Z0-9_-]{0,62}$/` and throws on
  path-traversal, URL-special, overlong, or empty input. Today
  reachable only via PR 1's `'default'` sentinel; live once PR 2
  plumbs real ids.

## Test plan

- [x] **Jest**: 1999 pass / 0 fail / 3 skips across 280 suites. 203
  SLO-specific tests across 13 suites — fingerprint stability,
  generator split, validators + custom-PromQL, provenance, service
  CRUD, service dedup, normalizer, cross-workspace isolation
  integration, rule-ref id format, ruler-namespace validation, ruler
  client, refstore.
- [x] **AMP audit**: every `ruler.upsertRuleGroup` /
  `deleteRuleGroup` caller resolves the namespace through
  `sloRulerNamespaceFor(...)` or via the persisted
  `provisioning.rulerNamespace` (itself written by that helper). No
  cross-namespace enumeration; no second namespace axis.
- [x] **Live OFF state**: route 404, nav absent, app shows
  "Application Not Found".
- [x] **Live ON state** (`observability.slo.enabled: true`): listing
  renders with pagination (35 seeded rows), counter shows correct
  total, target column tabular-nums right-aligned, detail page
  renders metadata + spec code block, create stamps `workspaceId:
  "default"` + `recordingFingerprints` + `rulerNamespace:
  "slo-generated-default"`, delete returns `{deleted: true}` and
  counter updates.
- [x] **Cypress smoke**
  (`.cypress/integration/slo_test/slo_crud_smoke.spec.js`): passes in
  ~11s against a local observability-stack cluster with a registered
  DirectQuery datasource. Guards on `Cypress.env('sloEnabled') !==
  false` so CI runs without the stack skip cleanly. Not yet wired
  into the CI pipeline (PR 5 will once the stack ships as a CI
  prereq).
- [x] **Reviewer**: toggle `observability.slo.enabled` off/on in
  `opensearch_dashboards.yml` and confirm the app
  appears/disappears cleanly. Yml changes require an OSD restart —
  the flag is captured once at plugin setup.

### DIFF
--- a/.cypress/integration/slo_test/slo_crud_smoke.spec.js
+++ b/.cypress/integration/slo_test/slo_crud_smoke.spec.js
@@ -1,0 +1,61 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/// <reference types="cypress" />
+
+/*
+ * PR-1 SLO CRUD smoke — create → listing row → delete → row gone.
+ *
+ * Runs against a locally-running dev cluster + observability plugin. Use:
+ *
+ *   yarn cypress:run-without-security --spec \
+ *     .cypress/integration/slo_test/slo_crud_smoke.spec.js
+ *
+ * Requires a registered Prometheus-backed DirectQuery datasource. The spec
+ * picks up the datasource id from the `sloDatasourceId` Cypress env.
+ */
+
+const DATASOURCE_ID = Cypress.env('sloDatasourceId') || 'prom';
+
+const uniqueSuffix = () => Math.random().toString(36).slice(2, 8);
+
+describe('SLO — CRUD smoke (PR 1)', () => {
+  it('creates, lists, and deletes an SLO through the wizard', () => {
+    const name = `Smoke SLO ${uniqueSuffix()}`;
+    const service = `smoke-svc-${uniqueSuffix()}`;
+
+    // Land on the listing page. The sidebar entry is registered when APM is
+    // enabled (observability-apm-slo).
+    cy.visit('/app/observability-apm-slo#/slos');
+    cy.get('[data-test-subj="sloCreateBtn"], [data-test-subj="sloCreateBtnEmpty"]')
+      .first()
+      .click();
+
+    // Fill the wizard.
+    cy.get('[data-test-subj="sloWizardName"]').type(name);
+    cy.get('[data-test-subj="sloWizardService"]').type(service);
+    cy.get('[data-test-subj="sloWizardTeam"]').type('platform');
+    cy.get('[data-test-subj="sloWizardDatasource"]').type(DATASOURCE_ID);
+    cy.get('[data-test-subj="sloWizardMetric"]')
+      .clear()
+      .type('http_requests_total');
+    cy.get('[data-test-subj="sloWizardTarget"]').clear().type('99.9');
+    cy.get('[data-test-subj="sloWizardSubmit"]').click();
+
+    // Back to listing — row should appear.
+    cy.location('hash').should('eq', '#/slos');
+    cy.contains('[data-test-subj="sloRowName"]', name).should('be.visible');
+
+    // Delete.
+    cy.contains('[data-test-subj="sloRowName"]', name)
+      .parents('tr')
+      .find('[data-test-subj="sloRowDelete"]')
+      .click();
+    cy.get('[data-test-subj="sloDeleteConfirm"]').find('button.euiButton--danger').click();
+
+    // Gone.
+    cy.contains('[data-test-subj="sloRowName"]', name).should('not.exist');
+  });
+});
--- a/.cypress/integration/slo_test/slo_crud_smoke.spec.js
+++ b/.cypress/integration/slo_test/slo_crud_smoke.spec.js
@@ -6,30 +6,41 @@
 /// <reference types="cypress" />
 
 /*
- * PR-1 SLO CRUD smoke — create → listing row → delete → row gone.
+ * PR-1 SLO CRUD smoke — create → detail → listing row → delete → row gone.
  *
- * Runs against a locally-running dev cluster + observability plugin. Use:
+ * Prereqs:
+ *   - Dev stack running with `observability.slo.enabled: true` in
+ *     `opensearch_dashboards.yml`.
+ *   - A registered Prometheus-backed DirectQuery datasource.
+ *   - An OSD workspace in which the APM + SLO apps are visible. The spec
+ *     picks up the workspace id from `workspaceId` env (see cypress.config).
  *
- *   yarn cypress:run-without-security --spec \
- *     .cypress/integration/slo_test/slo_crud_smoke.spec.js
+ * Run with (from plugin dir):
+ *   yarn cypress:run \
+ *     --spec .cypress/integration/slo_test/slo_crud_smoke.spec.js \
+ *     --env workspaceId=<WS_ID>,sloDatasourceId=<DS_ID>,baseUrl=http://localhost:5602
  *
- * Requires a registered Prometheus-backed DirectQuery datasource. The spec
- * picks up the datasource id from the `sloDatasourceId` Cypress env.
+ * The baseUrl env overrides `cypress.config.js` for dev servers on :5602.
  */
 
+const WORKSPACE_ID = Cypress.env('workspaceId');
 const DATASOURCE_ID = Cypress.env('sloDatasourceId') || 'prom';
 
 const uniqueSuffix = () => Math.random().toString(36).slice(2, 8);
+const prefix = WORKSPACE_ID ? `/w/${WORKSPACE_ID}` : '';
+const sloApp = `${prefix}/app/observability-apm-slo`;
 
 describe('SLO — CRUD smoke (PR 1)', () => {
   it('creates, lists, and deletes an SLO through the wizard', () => {
     const name = `Smoke SLO ${uniqueSuffix()}`;
     const service = `smoke-svc-${uniqueSuffix()}`;
 
-    // Land on the listing page. The sidebar entry is registered when APM is
-    // enabled (observability-apm-slo).
-    cy.visit('/app/observability-apm-slo#/slos');
-    cy.get('[data-test-subj="sloCreateBtn"], [data-test-subj="sloCreateBtnEmpty"]')
+    // Land on the listing page.
+    cy.visit(`${sloApp}#/slos`);
+    // Wait for the listing to render — either an empty prompt or the table.
+    cy.get(
+      '[data-test-subj="sloCreateBtn"], [data-test-subj="sloCreateBtnEmpty"]'
+    )
       .first()
       .click();
 
@@ -38,24 +49,34 @@ describe('SLO — CRUD smoke (PR 1)', () => {
     cy.get('[data-test-subj="sloWizardService"]').type(service);
     cy.get('[data-test-subj="sloWizardTeam"]').type('platform');
     cy.get('[data-test-subj="sloWizardDatasource"]').type(DATASOURCE_ID);
-    cy.get('[data-test-subj="sloWizardMetric"]')
-      .clear()
-      .type('http_requests_total');
+    cy.get('[data-test-subj="sloWizardMetric"]').clear().type('http_requests_total');
     cy.get('[data-test-subj="sloWizardTarget"]').clear().type('99.9');
     cy.get('[data-test-subj="sloWizardSubmit"]').click();
 
-    // Back to listing — row should appear.
-    cy.location('hash').should('eq', '#/slos');
-    cy.contains('[data-test-subj="sloRowName"]', name).should('be.visible');
+    // After create the wizard redirects to the detail page for the new SLO.
+    cy.location('hash', { timeout: 15000 }).should('match', /#\/slos\/[0-9a-f-]{36}/);
+    // Detail page shows the SLO name as the chrome h1 (via breadcrumb).
+    cy.contains('h1', name, { timeout: 15000 }).should('be.visible');
 
-    // Delete.
+    // Navigate back to the listing and locate the new row. Tolerate either
+    // page (new SLOs sort last; pageSize=100 guarantees visibility even with
+    // a seeded cluster).
+    cy.visit(`${sloApp}#/slos`);
+    cy.get('[data-test-subj="tablePaginationPopoverButton"]').click();
+    cy.get('.euiContextMenuItem').contains('100 rows').click();
+    cy.contains('[data-test-subj="sloRowName"]', name, { timeout: 15000 })
+      .should('be.visible');
+
+    // Trigger delete via the row's trash action.
     cy.contains('[data-test-subj="sloRowName"]', name)
-      .parents('tr')
+      .closest('tr')
       .find('[data-test-subj="sloRowDelete"]')
       .click();
-    cy.get('[data-test-subj="sloDeleteConfirm"]').find('button.euiButton--danger').click();
+    cy.get('[data-test-subj="sloDeleteConfirm"]')
+      .find('button.euiButton--danger')
+      .click();
 
-    // Gone.
+    // Row is gone (toast confirms success; assert on the list for durability).
     cy.contains('[data-test-subj="sloRowName"]', name).should('not.exist');
   });
 });
--- a/.cypress/integration/slo_test/slo_crud_smoke.spec.js
+++ b/.cypress/integration/slo_test/slo_crud_smoke.spec.js
@@ -30,7 +30,18 @@ const uniqueSuffix = () => Math.random().toString(36).slice(2, 8);
 const prefix = WORKSPACE_ID ? `/w/${WORKSPACE_ID}` : '';
 const sloApp = `${prefix}/app/observability-apm-slo`;
 
+// FIXME(pr-5): wire this spec into the CI pipeline once the observability-
+// stack dev cluster ships as a CI prereq. Today it's operator-run only:
+// CI doesn't have a Prometheus-backed DirectQuery datasource available.
+// The `sloEnabled` env gate below lets anyone drop it in a CI run that
+// doesn't have the stack up without failing the whole suite.
+const SLO_ENABLED = Cypress.env('sloEnabled') !== false;
+
 describe('SLO — CRUD smoke (PR 1)', () => {
+  before(function () {
+    if (!SLO_ENABLED) this.skip();
+  });
+
   it('creates, lists, and deletes an SLO through the wizard', () => {
     const name = `Smoke SLO ${uniqueSuffix()}`;
     const service = `smoke-svc-${uniqueSuffix()}`;
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,11 @@ coverage/
 .cypress/downloads
 common/query_manager/antlr/output
 .eslintcache
+
+# Local tool caches
+.claude/
+.playwright-mcp/
+
+# SLO audit evidence (screenshots, handoff notes) — kept locally, not shipped
+slo-bugbash-evidence/
+.slo-audit-*/
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ common/query_manager/antlr/output
 # SLO audit evidence (screenshots, handoff notes) — kept locally, not shipped
 slo-bugbash-evidence/
 .slo-audit-*/
+slo-*.png
--- a/.gitignore
+++ b/.gitignore
@@ -7,12 +7,3 @@ coverage/
 .cypress/downloads
 common/query_manager/antlr/output
 .eslintcache
-
-# Local tool caches
-.claude/
-.playwright-mcp/
-
-# SLO audit evidence (screenshots, handoff notes) — kept locally, not shipped
-slo-bugbash-evidence/
-.slo-audit-*/
-slo-*.png
--- a/common/constants/apm.ts
+++ b/common/constants/apm.ts
@@ -20,3 +20,8 @@ export const observabilityApmApplicationMapPluginOrder = 5101;
 export const observabilityApmApplicationConfigID = 'observability-apm-application-config';
 export const observabilityApmApplicationConfigTitle = 'Configuration';
 export const observabilityApmApplicationConfigPluginOrder = 5102;
+
+// APM application IDs - Service Level Objectives (PR 1)
+export const observabilityApmSloID = 'observability-apm-slo';
+export const observabilityApmSloTitle = 'SLOs';
+export const observabilityApmSloPluginOrder = 5103;
--- a/common/slo/__tests__/slo_promql_generator.test.ts
+++ b/common/slo/__tests__/slo_promql_generator.test.ts
@@ -1,0 +1,340 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import {
+  DEFAULT_MWMBR_TIERS,
+  RECORDING_WINDOWS,
+  generateSloRuleGroup,
+  parseDurationToMs,
+  findClosestRecordingWindow,
+  ruleSuffix,
+  slugifySloObjective,
+} from '../slo_promql_generator';
+import type { SloDocument } from '../slo_types';
+
+function baseSlo(overrides: Partial<SloDocument['spec']> = {}): SloDocument {
+  return {
+    id: 'slo-test-001',
+    spec: {
+      datasourceId: 'prom-ds-001',
+      name: 'API Availability',
+      enabled: true,
+      mode: 'active',
+      service: 'api-gateway',
+      owner: { teams: ['platform'] },
+      sli: {
+        type: 'single',
+        definition: {
+          backend: 'prometheus',
+          type: 'availability',
+          calcMethod: 'events',
+          metric: 'http_requests_total',
+          goodEventsFilter: 'status_code!~"5.."',
+        },
+        dimensions: [
+          { name: 'service', value: 'api-gateway' },
+          { name: 'handler', value: '/api/v1/users' },
+        ],
+      },
+      objectives: [{ name: 'availability-99-9', target: 0.999 }],
+      budgetWarningThresholds: [
+        { threshold: 0.5, severity: 'warning' },
+        { threshold: 0.2, severity: 'critical' },
+      ],
+      window: { type: 'rolling', duration: '30d' },
+      alerting: { strategy: 'mwmbr', burnRates: DEFAULT_MWMBR_TIERS.map((t) => ({ ...t })) },
+      alarms: {
+        sliHealth: { enabled: false },
+        attainmentBreach: { enabled: false },
+        budgetWarning: { enabled: true },
+        noData: { enabled: false, forDuration: '10m' },
+        resolved: { enabled: false },
+      },
+      exclusionWindows: [],
+      labels: {},
+      annotations: {},
+      ...overrides,
+    },
+    status: {
+      version: 1,
+      createdAt: '2026-04-22T00:00:00.000Z',
+      createdBy: 'test',
+      updatedAt: '2026-04-22T00:00:00.000Z',
+      updatedBy: 'test',
+      provisioning: {
+        backend: 'prometheus',
+        alertGroupName: 'slo:api_availability_abcd1234',
+        rulerNamespace: 'slo-generated',
+      },
+    },
+  };
+}
+
+describe('parseDurationToMs', () => {
+  it('parses common Prometheus durations', () => {
+    expect(parseDurationToMs('5m')).toBe(5 * 60_000);
+    expect(parseDurationToMs('1h')).toBe(3_600_000);
+    expect(parseDurationToMs('3d')).toBe(3 * 86_400_000);
+    expect(parseDurationToMs('1w')).toBe(604_800_000);
+  });
+  it('returns 0 for unparseable input', () => {
+    expect(parseDurationToMs('bogus')).toBe(0);
+    expect(parseDurationToMs('')).toBe(0);
+  });
+});
+
+describe('findClosestRecordingWindow', () => {
+  it('returns the first window >= target', () => {
+    expect(findClosestRecordingWindow('5m')).toBe('5m');
+    expect(findClosestRecordingWindow('45m')).toBe('1h');
+    expect(findClosestRecordingWindow('1d')).toBe('1d');
+  });
+  it('returns the longest recording window when the target exceeds all windows (design §6.4)', () => {
+    expect(findClosestRecordingWindow('30d')).toBe('3d');
+    expect(findClosestRecordingWindow('7d')).toBe('3d');
+  });
+});
+
+describe('ruleSuffix', () => {
+  it('is deterministic', () => {
+    const a = ruleSuffix('ws1', 'slo-1', 'obj-a');
+    const b = ruleSuffix('ws1', 'slo-1', 'obj-a');
+    expect(a).toBe(b);
+  });
+  it('distinguishes workspaces, SLOs, and objectives', () => {
+    const a = ruleSuffix('ws1', 'slo-1', 'obj-a');
+    const b = ruleSuffix('ws2', 'slo-1', 'obj-a');
+    const c = ruleSuffix('ws1', 'slo-2', 'obj-a');
+    const d = ruleSuffix('ws1', 'slo-1', 'obj-b');
+    expect(new Set([a, b, c, d]).size).toBe(4);
+  });
+  it('is 8 hex characters', () => {
+    const s = ruleSuffix('x', 'y', 'z');
+    expect(s).toMatch(/^[0-9a-f]{8}$/);
+  });
+
+  // Pinned-hash guard: design §6.3 / §13.1 commits to sha256(workspace:sloId:objective).slice(0,8).
+  // Value computed with:
+  //   node -e "console.log(require('crypto').createHash('sha256')
+  //     .update('ws-1:slo-abc:availability').digest('hex').slice(0,8))"
+  // If this value changes, external dashboards / Alertmanager silences /
+  // GitOps manifests pinning rule names will break. Coordinate a migration
+  // before touching this.
+  it('matches the pinned sha256 prefix for a known triple (§13.1 commitment)', () => {
+    expect(ruleSuffix('ws-1', 'slo-abc', 'availability')).toBe('3e048ec6');
+  });
+});
+
+describe('slugifySloObjective', () => {
+  it('lowercases and collapses non-alphanum to underscores', () => {
+    expect(slugifySloObjective('API Gateway', 'availability-99-9')).toBe(
+      'api_gateway_availability_99_9'
+    );
+  });
+});
+
+describe('generateSloRuleGroup — availability, single objective', () => {
+  it('generates 7 recording + 4 burn-rate + 2 budget warnings = 13 rules (design §6.4)', () => {
+    const group = generateSloRuleGroup(baseSlo());
+    expect(group.rules).toHaveLength(13);
+
+    const recording = group.rules.filter((r) => r.type === 'recording');
+    const alerting = group.rules.filter((r) => r.type === 'alerting');
+    expect(recording).toHaveLength(7);
+    expect(alerting).toHaveLength(6);
+
+    const burnRates = alerting.filter((r) => r.labels.slo_alarm_type === 'burn_rate');
+    const warnings = alerting.filter((r) => r.labels.slo_alarm_type === 'error_budget_warning');
+    expect(burnRates).toHaveLength(4);
+    expect(warnings).toHaveLength(2);
+  });
+
+  it('emits one recording rule per RECORDING_WINDOWS entry', () => {
+    const group = generateSloRuleGroup(baseSlo());
+    const recordingWindows = group.rules
+      .filter((r) => r.type === 'recording')
+      .map((r) => r.labels.slo_window)
+      .sort();
+    expect(recordingWindows).toEqual([...RECORDING_WINDOWS].sort());
+  });
+
+  it('attaches slo_id, slo_name, slo_objective, slo_service on every rule', () => {
+    const group = generateSloRuleGroup(baseSlo());
+    for (const rule of group.rules) {
+      expect(rule.labels.slo_id).toBe('slo-test-001');
+      expect(rule.labels.slo_name).toBe('API Availability');
+      expect(rule.labels.slo_objective).toBe('availability-99-9');
+      expect(rule.labels.slo_service).toBe('api-gateway');
+      expect(rule.labels.slo_owner_team).toBe('platform');
+    }
+  });
+
+  it('propagates spec.labels as slo_label_<key> (design §10.1)', () => {
+    const doc = baseSlo({ labels: { compliance: 'pci', region: ['us-west-2', 'us-east-1'] } });
+    const group = generateSloRuleGroup(doc);
+    const rule = group.rules[0];
+    expect(rule.labels.slo_label_compliance).toBe('pci');
+    expect(rule.labels.slo_label_region).toBe('us-west-2,us-east-1');
+  });
+
+  it('does not propagate spec.annotations (design §10.3)', () => {
+    const doc = baseSlo({ annotations: { runbook: 'https://wiki/slo/api' } });
+    const group = generateSloRuleGroup(doc);
+    for (const rule of group.rules) {
+      expect(rule.labels.slo_label_runbook).toBeUndefined();
+      expect(rule.labels.runbook).toBeUndefined();
+      expect(rule.annotations?.runbook).toBeUndefined();
+    }
+  });
+
+  it('burn-rate thresholds match burnRateMultiplier * errorBudget', () => {
+    const group = generateSloRuleGroup(baseSlo());
+    const burnRates = group.rules.filter(
+      (r) => r.type === 'alerting' && r.labels.slo_alarm_type === 'burn_rate'
+    );
+    // 99.9% target → 0.001 error budget → page-quick threshold = 14.4 × 0.001 = 0.0144
+    const pageQuick = burnRates.find((r) => r.labels.slo_burn_rate_multiplier === '14.4')!;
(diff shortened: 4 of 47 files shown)


joshuali925:
### Code review

Found 2 issues:

1. `refStore` key uses inconsistent forms of `datasourceId`. `incrementRef`/`decrementRef` are called with `deploy.datasource.id` (registry id like `ds-4`), but `getFingerprintRefcounts` reads back with `doc.spec.datasourceId` (canonical name like `ObservabilityStack_Prometheus`). The two halves use different keys for the same `(workspaceId, datasourceId, fingerprint)` tuple, so the lookup will always miss — `Shared with N other SLOs` will read 0 and `wasZero` is unreliable.

   Increment site: https://github.com/lezzago/dashboards-observability/blob/2b115b375c363be6be133911bb55ac7d4a85f1f4/common/slo/slo_service.ts#L697-L706

   Read site: https://github.com/lezzago/dashboards-observability/blob/2b115b375c363be6be133911bb55ac7d4a85f1f4/common/slo/slo_service.ts#L470-L477

2. `SloRuleRefStore.listStaleZero` pagination break condition compares filtered `results.length` against unfiltered `response.total`, so it stops early when many page-N entries are skipped by the `zeroSinceAt` + cutoff filter. The sibling `listByDatasource` correctly uses `results.length >= response.total` because it pushes every SO. Stale-zero entries past the early-exit point are silently left unswept.

   https://github.com/lezzago/dashboards-observability/blob/2b115b375c363be6be133911bb55ac7d4a85f1f4/server/services/slo/slo_rule_ref_store.ts#L178-L188

TackAdam:
Code Review:

## CRITICAL — must address before merge

### C-1 · `PUT /api/observability/v1/slos/{id}` accepts an unvalidated `spec`

`server/routes/slo/index.ts:244-247`

```ts
const updateBody = schema.object({
  version: schema.number({ min: 1 }),
  spec: schema.object({}, { unknowns: 'allow' }),
});
```

The strong `sloSpecSchema` (lines 195-237) is enforced on POST (`createBody`) but the PUT body schema is wide open. A client can update an SLO with **any** spec shape — missing `objectives`, missing `owner.teams`, `dimensions: undefined`, etc. The service-layer validator (`validateSloSpec` in `common/slo/slo_validators.ts`) does run inside `SloService.update`, so most malformed shapes are rejected there with `SloValidationError`. **However**:

- `previewBody` at line 251 has the same wide-open shape and reaches the same validator path — but in preview the body is never persisted, so the impact there is limited.
- The validator is shape-aware but not type-safe; the service-layer code **does** dereference `merged.objectives.length`, `merged.sli.type`, `merged.owner.teams[0]` (`slo_service.ts:822, 1791-1822`, etc.) before the validator runs in some helper code paths (e.g. `assertNameUnique` at `:822`, `dedupAlertGroupName(merged.name, …)` at `:925`).

**Action:** either reuse `sloSpecSchema` directly for PUT, or define an explicit `Partial<sloSpecSchema>` allowing only the fields that are legitimately patchable, with proper inner shapes. `previewBody.spec` should also be tightened to the full spec schema since the route already gates by feature flag.

### C-2 · `projectAttributes` crashes on partially-formed SLI when `dimensions` is missing

`server/services/slo/slo_saved_object_store.ts:47-48`

```ts
const dimensionNames = single?.dimensions.map((d) => d.name) ?? [];
const dimensionValues = single?.dimensions.map((d) => d.value) ?? [];
```

The route's `sliNodeSchema` declares `dimensions: schema.maybe(schema.arrayOf(...))` (route layer, line 133) — i.e. `dimensions` is **optional** at the boundary. TypeScript's `SingleSli.dimensions: Dimension[]` is non-optional, but that's a type assertion, not a runtime guarantee. If `single` is non-null but `single.dimensions` is `undefined` (legal per the route schema, and reachable for `sli.type === 'single'` + `definition.type === 'custom'`, where `validateSloSpec` skips the dimensions check at `slo_validators.ts:202-217`), the projection throws `Cannot read properties of undefined (reading 'map')` *during the SO `save()`* — which happens **after** the ruler write succeeded. That's a dual-write divergence (ruler has rules, SO never persisted), exactly the failure mode the rest of the service is hardened against.

**Action:** coerce defensively: `single?.dimensions?.map(...) ?? []`, and audit other lines in `projectAttributes` for the same pattern (`doc.spec.owner.teams[0]` at line 66, `doc.spec.objectives.reduce` at line 42 — both will throw if the spec is partial).

### C-3 · `updateDedup` reuses `existing.spec.workspaceId` but also re-derives namespace from the **deploy** workspace; mismatch silently routes writes to the wrong namespace

`common/slo/slo_service.ts:909-914`

```ts
const namespace =
  existing.status.provisioning.backend === 'prometheus' &&
  existing.status.provisioning.rulerNamespace
    ? existing.status.provisioning.rulerNamespace
    : sloRulerNamespaceFor(deploy.workspaceId);
```

Plus, `dedupAlertGroupName(merged.name, deploy.workspaceId, existing.id)` at line 925 uses `deploy.workspaceId`, *not* `existing.spec.workspaceId`. If those ever differ (and PR 2 will plumb real workspace ids that may diverge from the persisted value because clients can spoof `workspaceId` in the wire body — see C-4 below), the `alertGroupName` is recomputed against the deploy workspace while the namespace falls back to the persisted one. Update would write the alert group to a name that can collide with another workspace's SLO sharing the same `existing.id` (impossible in normal operation, but possible across import/export). The `ruleSuffix` hash includes `workspaceId`, which protects collision, but the *intent* of the AMP-compat invariant (spec §"namespace derivation") is that **the same workspaceId is used for the namespace and the group name**. Today they're decoupled.

**Action:** define one source of truth: either (a) always use `existing.spec.workspaceId ?? deploy.workspaceId`, or (b) always use `deploy.workspaceId` and reject the request when `deploy.workspaceId !== existing.spec.workspaceId`. The latter is consistent with the "workspace is immutable after create" invariant in the schema doc.

### C-4 · `workspaceId` is in the route schema (`sloSpecSchema`, line 199) but the service stamp only happens on **create**, not on **update**

`server/routes/slo/index.ts:199` declares `workspaceId: schema.maybe(schema.string())` with the comment "server-stamped; clients may send it but it's overwritten." The create path enforces this in `slo_service.ts:523`. The **update** path does not call `sloRulerNamespaceFor(input.spec.workspaceId)` at all — but it accepts `input.spec.workspaceId` in the merged spec and, at `slo_service.ts:813-817`:

```ts
const merged: SloSpec = normalizeSpec({
  ...existing.spec,
  ...input.spec,
  workspaceId: existing.spec.workspaceId,
});
```

Good — `existing.spec.workspaceId` overwrites client input. But `normalizeSpec` then strips unknown top-level keys from a copy that is `{...existing.spec, ...input.spec, workspaceId: …}`, so a malicious client cannot smuggle `workspaceId`. Verified.

Where the bug is: `validateSloSpec(merged)` (line 819) does not validate `merged.workspaceId` shape. Then `dedupAlertGroupName(merged.name, deploy.workspaceId, existing.id)` (line 925) uses **the deploy context's** workspaceId, while `existing.status.provisioning.rulerNamespace` carries the *persisted* one. If a future PR allows `deploy.workspaceId` to differ from `existing.spec.workspaceId` (PR 2 absolutely will, on cross-workspace move), the alert group is named under the wrong workspace's hash domain.

**Action:** the merge is correct but inconsistent with how the namespace is derived. Pick one workspace value for the entire update and assert at function entry: `if (deploy.workspaceId !== existing.spec.workspaceId) throw …`.

### C-5 · `goodEventsFilter` regex permits matcher injection

`common/slo/slo_validators.ts:154-157`

```ts
if (prom.goodEventsFilter && /[{}()\n\\]/.test(prom.goodEventsFilter)) {
  errors['spec.sli.definition.goodEventsFilter'] =
    'Good events filter must not contain braces, parentheses, backslashes, or newlines';
}
```

The blocklist excludes `,` and `"`. The generator splices the user's filter into a label-matcher list:

`common/slo/slo_promql_generator.ts:166-173`

```ts
const pairs: string[] = sli.dimensions.map(...);
if (includeGoodFilter && sli.definition.backend === 'prometheus') {
  const good = sli.definition.goodEventsFilter?.trim();
  if (good) pairs.push(good);
}
return pairs.join(', ');
```

A filter `status="x",pwn="y` passes validation — Cortex ultimately rejects the malformed PromQL at deploy time, so the user-impact today is rule-write failures instead of silent data exfil; but this is exactly the class of input the validator exists to block. A more careful filter `status="x",extra_label="legitimate value"` is **fully valid PromQL** and would be silently smuggled into every recording rule, mutating the metric series the SLO computes against.

**Action:** require the filter to match a single matcher shape, e.g. `^[a-zA-Z_][a-zA-Z0-9_]*\s*(=|!=|=~|!~)\s*"([^"\\]|\\.)*"$`, or wrap the user's filter in its own paren group at emit time *and* document that nesting matchers via `,` is unsupported. Also reject embedded literal LF/CR (which would terminate the YAML mapping line).

### C-6 · YAML emitter passes raw control characters (LF/CR/TAB) from `spec.name`, `spec.service`, `spec.tier`, team names, etc., into a double-quoted YAML scalar

`common/slo/slo_promql_generator.ts:559-561, 563-590`

```ts
function escapeYaml(s: string): string {
  return s.replace(/\\/g, '\\\\').replace(/"/g, '\\"');
}
…
lines.push(`      ${k}: "${escapeYaml(v)}"`);
```

`escapeYaml` only escapes `\` and `"`. The route schema only enforces `minLength`/`maxLength` on `name`, `service`, `tier`, `owner.teams[*]`. The validator (`slo_validators.ts:120-138`) checks length and trim but **not** character class. A user with `spec.name = "ok\nname"` produces YAML like:

```yaml
      slo_name: "ok
name"
```

— a literal LF inside a double-quoted scalar, which the YAML 1.2 spec forbids in flow scalars. Cortex returns `RULER_VALIDATION_FAILED` *after* the SO is already written by the dual-write order. Same risk for `\t`, `\r`, `\x00`, and unicode line/paragraph separators (`U+2028` / `U+2029`).

**Action:** either (a) use `js-yaml`'s `dump` (which `ruler_client.ts:ruleGroupToYaml` already does for the same shape — picking the homemade emitter is the bug), or (b) extend `escapeYaml` to escape every character `< 0x20` plus `\x7F`, `U+2028`, `U+2029`. Option (a) is the cleaner fix and would let `slo_promql_generator.ts:rulesToYaml` (lines 563-590) and `ruler_client.ts:ruleGroupToYaml` (lines 110-117) share one path. Note that `validators.ts:UNSAFE_LABEL_VALUE_RE` already rejects newlines on user-supplied label values — extend the same regex to `name`/`service`/`tier`/`owner.teams[*]`.

### C-7 · `SloRulerError.rawBody` is sent verbatim to the client

`server/routes/slo/handlers.ts:48-62` returns `rawBody: e.rawBody` in the response body. `rawBody` is whatever Cortex sent — possibly internal hostnames, tenant ids, or a stack trace. Any authenticated OSD user (not just the SLO author) who triggers a ruler error sees that text in their toast (the wizard renders it with `notifications.toasts.addDanger({ text: ruler.rawBody })` at `slo_wizard_page.tsx:192`). For a multi-tenant deployment this is information leakage.

**Action:** truncate `rawBody` to a small budget (e.g. 256 chars) and strip ANSI escape sequences; log the full body server-side at `info` for ops, but only return a short, sanitized excerpt to the client. The `code` discriminator (`RULER_VALIDATION_FAILED`/`RULER_AUTH_FAILED`/`RULER_UNREACHABLE`) is already enough for the wizard to render a meaningful message.

---

## HIGH

### H-1 · `slo_saved_object_store.list` pagination terminates by `results.length >= response.total`, but `results` excludes malformed docs

`server/services/slo/slo_saved_object_store.ts:131-145`. If any SO is malformed (`fromAttributes` throws and the doc is skipped at `:138`), `results.length` permanently lags `response.total`, so the loop falls through to the `response.saved_objects.length === 0` exit — fetching one extra empty page. Not infinite (the agent's first read was wrong), but every list call to a large index pays one extra round-trip whenever any malformed SO exists.

**Action:** track `processed += response.saved_objects.length` separately and exit on `processed >= response.total`.

### H-2 · `slo_rule_ref_store.listStaleZero` pagination accumulator is wrong

`server/services/slo/slo_rule_ref_store.ts:180-184`

```ts
if (
  response.saved_objects.length === 0 ||
  results.length + (page - 1) * perPage >= response.total
)
  break;
```

`results` only includes refs whose `zeroSinceAt` is past the cutoff (post-filtered after the SO `find`); the formula mixes filtered and unfiltered counts. With a 1000-row dataset where 500 pass the secondary check, the loop terminates correctly only because the SO index runs out of pages. With a 10000-row dataset, it can terminate **before the last page** is fetched (when `results.length + (page-1)*perPage` happens to exceed `total` on a cross-page boundary), silently missing stale refs and skipping reconciliation. PR 5's reconciler depends on this being correct.

**Action:** same fix as H-1. Track `processed = page * perPage` (or `processed += response.saved_objects.length`) and exit when `processed >= response.total`.

### H-3 · `incrementRef` retry loop has no jitter or backoff

`server/services/slo/slo_rule_ref_store.ts:206-300, 302-363`. Three retries back-to-back with no `setImmediate`/`setTimeout`. Under contention from a workspace-wide bulk-create (PR 6 adoption flow will do this), conflict probability is high enough that 3 retries, all in the same microtask, are likely to all see the same stale version. Even a `Math.random() * 50ms` between attempts would make this materially better.

**Action:** add a small randomized delay between retries; consider raising `MAX_RETRIES` to 5; document the contention model in the docstring.

### H-4 · `slo_promql_generator.ts:rulesToYaml` open-codes its own YAML, while `ruler_client.ts:ruleGroupToYaml` uses `js-yaml`

The PromQL generator's preview output (`slo_promql_generator.ts:563-590`) goes to the client unchanged — this is what powers the wizard's preview. The ruler write path (`ruler_client.ts:110-117`) re-serializes via `js-yaml`. So **the YAML the user sees is not the YAML Cortex receives**. Different escape rules → different rejection surfaces. The PR description claims "preview and deployment share this function so what the user sees matches what the ruler gets" (slo_promql_generator.ts:13), which is no longer true once the ruler client switched to `js-yaml`.

**Action:** unify on one serializer. `js-yaml` is the right choice (dependency exists, and it handles control chars correctly).

### H-5 · `customExpr` PromQL balance check counts braces/brackets inside backtick-quoted raw strings

`common/slo/slo_validators.ts:55`

```ts
const stripped = expr.replace(/"(?:\\.|[^"\\])*"|'(?:\\.|[^'\\])*'/g, '');
```

PromQL also accepts backtick raw strings (`` `…` ``) where escapes don't apply. A query containing `` `{` `` is rejected as unbalanced. PromQL `# …` line comments are also not stripped — a `# ))` comment trips the check.

**Action:** strip backtick raw strings (`/`(?:[^`])*`/g`) and `#…\n` comment runs before counting.

### H-6 · Sequential `sloService.get(id)` then `tryBuildDeployContext(existing.spec.datasourceId, …)` is a TOCTOU window in PUT/DELETE/enable/disable

`server/routes/slo/index.ts:565-590, 608-626, 643-668, 683-707`. The route fetches the SO, extracts the datasource id from `spec`, then builds the deploy context. Between the `get` and the actual mutation, a concurrent PUT can change `datasourceId`. The deploy context built off the stale read targets the old datasource; the service-layer update reads the new spec; ruler write goes to the wrong tenant.

**Action:** push the deploy-context resolution down into the service. The service already loads `existing` for the version check; it can resolve the datasource off the post-merge spec atomically.

### H-7 · DELETE returns 400 ("datasource not registered") when the datasource has been removed but the SO still exists

`server/routes/slo/index.ts:608-626`. If the user deleted the datasource out-of-band, this path returns 400 from `tryBuildDeployContext`, leaving the orphaned SO in place with no API to delete it. The user is forced into saved-object management UI to clean up.

**Action:** delete should be tolerant — if the datasource is gone, log warn and skip the ruler call (the SO record is what the user wants gone). The reconciler (PR 5) handles dangling rule groups; it cannot help when the SO itself can't be deleted.

### H-8 · GET routes lose structured error attributes; only the top-level `message` is forwarded

`server/routes/slo/index.ts:464-471, 535, 551, 672, 711`. POST/PUT/DELETE include `attributes: result.body` so the wizard can read `code`, `httpStatus`, etc. GET endpoints (list, statuses, get-by-id, enable, disable) only forward `message`. That means a `SloRulerError` raised during enable/disable is reduced to a string toast, dropping the diagnostic the wizard already knows how to render.

**Action:** always set `attributes` consistently across all routes.

### H-9 · `coerceRuleGroupList` silently treats Cortex error envelopes as empty namespaces

`server/services/slo/ruler_client.ts:434-469`. When the response is `{ status: 'error', errorType: '…', error: '…', data: null }`, none of the four shapes match cleanly: `data` is non-array, the array-flatten loop finds no array fields, single-group coercion fails (no `name`), and the function returns `[]`. The PR-5 reconciler will then think the namespace is empty and start tearing down rules.

**Action:** when `record.status === 'error'`, throw `SloRulerError('RULER_VALIDATION_FAILED', extracted_status, body)` rather than coercing to empty.

### H-10 · `isWrappedEmptyNamespaceError` substring sniff is fragile

`server/services/slo/ruler_client.ts:343-357`. Case-insensitive substring match on `"no rule groups found"` plus `"code: 404"` + `"ruler request failed"` — any unrelated error message that happens to contain one of those substrings (verbose stack traces, generic 404 templates) reclassifies a genuine 400 as "empty namespace" and returns `[]`.

**Action:** parse the JSON envelope and pin to the exact `details` field; avoid stringification-then-substring.

### H-11 · `slo_rule_ref_store.incrementRef` returns `attributes: { ...prior, ...updated.attributes, ...nextAttrs }`

`server/services/slo/slo_rule_ref_store.ts:280-287, 347-355`. The triple-spread reorder lets `nextAttrs` (computed pre-write, locally) overwrite `updated.attributes` (returned by the SO client, post-write). Today the only writer is this store, so impact is bounded — but PR 5's reconciler will write `zeroSinceAt`. The next concurrent decrement will overwrite the reconciler's `zeroSinceAt` with `undefined` (the local pre-write value).

**Action:** trust the SO client's response: `attributes: updated.attributes`.

### H-12 · Wizard form has no `onSubmit` handler and no Enter-to-submit prevention

`public/components/apm/pages/slos/slo_wizard_page.tsx:218`

```tsx
<EuiForm component="form">
```

The form element has no `onSubmit`, so pressing Enter inside a text field triggers the native form submission, which causes a full-page reload (no `event.preventDefault`). Keyboard-only users cannot submit reliably; mouse users may lose the form on accidental Enter.

**Action:** add `onSubmit={(e) => { e.preventDefault(); onSubmit(); }}` or set `component="div"` and rely on the submit button.

### H-13 · Wizard never warns about unsaved changes on navigation away

`public/components/apm/pages/slos/slo_wizard_page.tsx:279`. The Cancel button calls `history.push('/slos')` and clicking the OSD chrome breadcrumb leaves silently — both discard a fully-typed form. With required fields name/service/team/datasource/metric/target, accidental loss is real.

**Action:** add `<Prompt when={dirty} message="…" />` from `react-router-dom`.

### H-14 · Wizard `onSubmit` is not unmount-guarded; `setSubmitting(false)` after unmount logs a React warning

`public/components/apm/pages/slos/slo_wizard_page.tsx:178-203`. The listing page already has the `generationRef` cancellation pattern (`slo_listing_page.tsx:135, 142-167`). The wizard does not.

**Action:** mirror the listing page's pattern.

### H-15 · `EuiFieldNumber` for target collapses partial input

`public/components/apm/pages/slos/slo_wizard_page.tsx:260-268`. `Number(e.target.value)` returns `0` for `""` (empty input), which flunks the `target >= 50` validity check at line 211 — but more importantly **silently changes the user's input to `0` mid-edit**, so the user can't clear the field to retype.

**Action:** store the raw string in form state and parse-on-submit, or special-case `e.target.value === '' ? '' : Number(...)` and treat empty as "incomplete."

---

## MEDIUM

### M-1 · `body: yaml-string` may be JSON-wrapped by the OSD transport

`server/services/slo/ruler_client.ts:181-188`. The comment claims the SQL plugin sets Content-Type on the upstream hop, but the inner OSD `client.transport.request({ body: yamlString })` typically JSON-encodes the body when Content-Type is `application/json` (the default). I cannot verify without a live trace; the existing test suite mocks `transport.request` so it doesn't catch this. **Action:** verify with a curl trace against a real cluster, or pass `headers: { 'content-type': 'application/yaml' }` in the request options if the underlying transport honors them.

### M-2 · `parseFloat(n.toPrecision(N)).toString()` produces scientific notation for small/large numbers

`common/slo/slo_promql_generator.ts:323-328` (`formatLatencyBoundLe`), and similar at `:392-394` (`roundThreshold`). For `n = 1e-7` the output is `"1e-7"`. PromQL accepts scientific notation in expressions, but the same string is also used as the **string label** `slo_budget_threshold` and as the histogram bucket `le` label. Bucket labels must literally match the producer's `le=` annotation; scientific-vs-decimal mismatch would silently miss the bucket. **Action:** use `toFixed` with a fixed precision, or document a minimum representable bound.

### M-3 · `computeSliFingerprint` doesn't normalize the latency threshold's numeric form

`common/slo/slo_sli_fingerprint.ts:54`. `objective.latencyThreshold` is included raw. `0.1 + 0.2 = 0.30000000000000004` and `0.3` are different inputs but produce the same `formatLatencyBoundLe` output (`"0.3"`) — i.e. equivalent recording rules with non-equal fingerprints. Two SLOs that *should* dedup will fail to. **Action:** funnel through the same `formatLatencyBoundLe`/`toPrecision` before hashing.

### M-4 · `normalizeGoodEventsFilter` strips internal whitespace inside string literals

`common/slo/slo_sli_fingerprint.ts:78-87`. `replace(/\s+/g, ' ')` collapses whitespace everywhere, including **inside** matcher string literals. `status=" 5xx "` and `status="5xx"` are semantically distinct in PromQL but produce the same fingerprint. **Action:** strip whitespace outside string literals only, or skip whitespace normalization entirely (the canonicalize step is the dedup unit, not whitespace).

### M-5 · `slo_promql_generator.ts:slugifySloObjective` — slug truncation at 40 char + the alert group naming uses `slugifySloObjective(spec.name, 'group')` — the `'group'` literal does not depend on the objective name

`slo_promql_generator.ts:640-642, 700-704`. Two SLOs whose `spec.name` shares the first 40 chars and have the same `sloId` would collide on the alert group name — but `sloId` is unique (UUID or validated slug per `validateSloId`), and `ruleSuffix` includes `sloId`. So the group name is unique by construction. Verified safe; flag as a sharp edge for future readers.

### M-6 · `dedupAlertGroupName(specName, …)` uses `specName` while `dedupRecordingGroupName(fp)` does not

`slo_promql_generator.ts:683-705`. The alert group name depends on `spec.name`; the recording group does not. If the rename of `spec.name` is ever supported (PR 6 adoption can rebuild from provenance), the alert group name changes but the recording group stays. Rename = delete-and-recreate of alert group. The dedup test suite doesn't cover rename. **Action:** flag for PR 6 review.

### M-7 · `validateSloSpec` reports only the **last** offending value when multiple values fail per-key

`slo_validators.ts:319-330`. The same key produces a single error message. Inconsistent with the per-index dimensions reporting at `:209-215`. Cosmetic but trips up users who fix the message they see and resubmit, only to discover another value also failed.

### M-8 · `validateCustomPromQL` control-char regex doesn't include `\n`/`\r`/`\t`, but the comment claims "control chars (other than tab/newline)"

`slo_validators.ts:79`. Comment says newline + tab are excluded; regex `/[\x00-\x08\x0B\x0C\x0E-\x1F\x7F]/` is consistent with that. So `\n` and `\t` and `\r` are *all* permitted in the user expression. The literal-block emitter `expr: |` handles `\n` correctly, but `\r` and `\t` produce parser-defined behavior across YAML libs. **Action:** reject `\r` explicitly; consider rejecting `\t` for diff determinism.

### M-9 · `escapeYaml` and label/annotation emit have no `null`/non-string guard

`slo_promql_generator.ts:559-561, 576-587`. Trusts callers to pass strings. `commonLabels` builds a `Record<string, string>` from typed sources, so today this is safe — but the provenance annotation embeds `JSON.stringify(spec)` from `slo_rule_provenance.ts:127-130`, where nested values are caller-controlled. **Action:** add a defensive `String(v)` cast at the emit site.

### M-10 · Wizard submit redirects to detail page even though `apiClient.create()` may have thrown after the SO was saved (race)

`public/components/apm/pages/slos/slo_wizard_page.tsx:178-203`. If the network blip between server's 201 and client's response handler causes a fetch error, the user sees a "Failed to create" toast, but the SLO was actually created. Listing-page refresh would surface it; wizard does not. **Action:** on failure, refresh the listing and double-check the toast text.

### M-11 · Internal app routing does not encode SLO ids

`slo_listing_page.tsx:194` (`to={`/slos/${slo.id}`}`), `slo_wizard_page.tsx:186` (`history.push(`/slos/${doc.id}`)`), `slo_detail_page.tsx:128-129` (`href: `#/slos/${id}``). The api client encodes on the wire (slo_api_client.ts:117, 125, 131), but in-app router links/redirects do not. SLO ids today are UUID v4 or validated slugs (`/^[a-z][a-z0-9-]{2,62}$/`), so practically there's no path-special character — but the listing page renders **whatever the server returned**, so a future migration / import / clone-with-custom-id flow could break. **Action:** encodeURIComponent consistently.

### M-12 · `data-test-subj="sloRow-${row.id}"` violates the camelCase convention

`slo_listing_page.tsx:278`. CLAUDE.md: "`data-test-subj` values should be camelCase." `sloRow-...` mixes a hyphen and an arbitrary id (which contains hyphens for slugs). **Action:** `sloRow_${row.id}` or just `sloRow` and disambiguate by row index in tests.

### M-13 · `parentBreadcrumb` is a fresh object per render in `app.tsx`, so the breadcrumb effect re-runs on every render

`public/components/app.tsx` constructs the prop inline. `slo_*.tsx` `useEffect` deps array contains `parentBreadcrumb` — fires on every render. Cosmetic (re-setting the same breadcrumb is idempotent), but wasteful. **Action:** memoize in `app.tsx` with `useMemo`.

### M-14 · `SloListingPage` declares `http` prop but doesn't consume it

`slo_listing_page.tsx:43-49, 118-123`. `apiClient` already wraps `http`. Dead prop; the wrapper at `slos_page.tsx:52` passes it. **Action:** remove from the interface.

### M-15 · `EuiFormRow`s lack `isInvalid` / `error` props

`slo_wizard_page.tsx:219-277`. The form gates the submit button via `valid` (`:205-212`) but never tells the user *why*. Required-field hints are needed for screen readers and discoverability.

### M-16 · `LABEL_NAME_RE` permits `__proto__`

`slo_validators.ts:16`. `__proto__` matches `/^[a-zA-Z_][a-zA-Z0-9_]*$/`. Per `slo_service.ts:1878-1892`, `normalizeSpec` uses an explicit allow-list (`SLO_SPEC_KEYS`), so `__proto__` cannot land at the top of the SO. But a user with `spec.labels = { __proto__: "x" }` passes both the route and the validator and lands in the persisted spec; the rule emit produces `slo_label___proto__: "x"`, which Cortex rejects. **Action:** add `__proto__`/`constructor`/`prototype` to `RESERVED_LABEL_KEYS`.

### M-17 · `validateSloSpec` does not character-class-check `spec.name`/`service`/`tier`/`owner.teams[*]`

See C-6. These fields flow into YAML labels, so they must obey `UNSAFE_LABEL_VALUE_RE` (or a stronger superset). Today only `dimensions[*].value` and `labels[*]` are checked. **Action:** apply the same character-class check to the four fields above.

### M-18 · No upper bound on `pageSize` query param

`server/routes/slo/index.ts:411-461`. `parseInt(q.pageSize, 10)` then passed through to the SO `find` call. A malicious or buggy client sending `pageSize=99999999` would request a large allocation. The service layer caps to 100 at `slo_service.ts:1616`, but the route should fail-fast at the boundary. **Action:** clamp at the route too.

---

## LOW / nits

### L-1 · `MockRulerClient` is exported alongside `DirectQueryRulerClient` in production code

`server/services/slo/ruler_client.ts:554-597`. Used in tests only. Move under `__tests__` or document why production needs it (the docstring suggests dev fallback, but the plugin always wires `DirectQueryRulerClient`).

### L-2 · `coerceRuleGroup` returns `yaml: ''`

`server/services/slo/ruler_client.ts:419-420`. The `GeneratedRuleGroup.yaml` field is a one-way artifact from the generator; on the read path it's always empty. Document or split the type.

### L-3 · `isSavedObjectNotFound` duplicated across `slo_saved_object_store.ts:25-28` and `slo_rule_ref_store.ts:101-109`

Hoist into a shared util.

### L-4 · `toRulerError` 0-status maps to 502; 0 is "transport failure" not "bad gateway"

`ruler_client.ts:285-311` + `handlers.ts:52`. 503 is closer in semantics for OSD-side network failure.

### L-5 · `parseIntervalSeconds` in `ruler_client.ts:523-539` swallows unknowns silently with `60`

Logging at `debug` would help track ruler-side schema drift.

### L-6 · `slo_templates.ts` `replace(/\$\{service\}/g, vars.service || '${service}')` is vulnerable to `$&`/`$1`-style replacement string special tokens

`common/slo/slo_templates.ts:99-101` and similar. A `service` value `$&` would inject the matched substring. **Action:** use the function form `replace(re, () => vars.service || '${service}')`.

### L-7 · `durationToMs` in `slo_templates.ts:493-499` duplicates `parseDurationToMs` from the generator

Drift risk.

### L-8 · Cypress smoke spec `slo_crud_smoke.spec.js:68` asserts a strict UUID pattern in the URL hash

If client-supplied ids land later (server already accepts `id` in `createBody.id`), this regex breaks.

### L-9 · `ruleSuffix` truncates SHA-256 to 8 hex chars

`slo_promql_generator.ts:119-122`. 32 bits = 4-billion collision space; per-objective per-workspace it's fine, but if the workload one day generates 100k SLOs in a workspace, birthday-bound collision probability is ~1.1%. Document the cardinality assumption.

### L-10 · `SLUG_ID_RE` permits trailing/double hyphens (`a--b`, `a-`)

`slo_validators.ts:18`. Cosmetic. Tighten to `/^[a-z][a-z0-9]*(-[a-z0-9]+)*$/`.

### L-11 · `MWMBR_TIER_LABELS[i] ?? `Tier${i + 1}``

`slo_promql_generator.ts:354`. Reachable only with >4 tiers; `validateBurnRate` has no max-tier cap. Either add one or document the fallback.

### L-12 · `truncateName` uses `.slice` on UTF-16 code units (multibyte safety)

`slo_rule_provenance.ts:227-230`. Theoretical for this domain (rule names are ASCII-constrained upstream), but worth `Array.from(input).slice(0, max).join('')` for safety.

### L-13 · `serializeFilters` joins arrays with `,` without encoding

`slo_api_client.ts:91-107`. A team value `"foo,bar"` becomes ambiguous. Use array-style query (`team[]=foo&team[]=bar`) or document the limitation.

### L-14 · `EuiHealth color={getSloHealthColor(slo.status.state)}` only renders state, no aria-live region

`slo_listing_page.tsx:206-209`. State changes (post-PR-3) won't be announced to screen readers.

---

## Test coverage observations

The test suite is dense (≈4k LOC of tests across 13 suites; PR description claims 203 SLO-specific tests). Two gaps worth flagging:

1. **No test exercises a malformed `spec` reaching `projectAttributes`** (C-2). The dedup tests use well-formed specs throughout.
2. **No test for control characters in `spec.name`/`spec.service`/`spec.tier`** (C-6). `slo_validators.test.ts` covers `dimensions[*].value` and `labels[*]` but not the four top-level user-controlled string fields.
3. **No PUT-with-bad-spec route test** for C-1 — recommend adding a regression that posts an unvalidated update to demonstrate the gap.

---

## Recommendation

Block on **C-1 through C-7**. The schema-validation gap (C-1) and the YAML control-char gap (C-6) are reachable today via the existing wizard once the feature flag flips. C-2 has lower reach (custom SLIs only) but produces the dual-write divergence that the rest of the service is hardened against. C-7 is a defense-in-depth concern that's easy to address.

Address **H-1 through H-15** in a follow-up cycle — H-2, H-4, and H-9 all have downstream consequences for PR 5's reconciler, so resolve them in PR 1 to avoid carrying the bug forward. H-12, H-13, H-14, H-15 are wizard polish that PR 8 was scheduled to handle, but they read more like defects than polish.

The architecture (forward-compatible schema, AMP-invariant plumbing through one composer, dedup with refcount, dual-write order with rollback compensation, deferred-but-typed surface for PRs 3-6) is solid. The PR description and the tracking issue are clear about what's deferred and why; the remaining gaps are mostly hardening of the boundary, not redesign.

lezzago:
Thanks @TackAdam and @joshuali925 for the thorough reviews — fixes
landed in 5f70375 (with one comment-only follow-up in 2f1dd59).

**@joshuali925**

1. **refStore key inconsistency** — all six call sites in
 `slo_service.ts` (create / update / delete dedup paths) now use
 `spec.datasourceId` (canonical name) to match the read site. The
 "Shared with N other SLOs" lookup will now resolve correctly.
2. **`listStaleZero` pagination** — switched to `processed +=
 response.saved_objects.length` and exit on `processed >=
 response.total`, matching `listByDatasource`. Same fix applied to
 `slo_saved_object_store.list` (your sibling point applies there too).

**@TackAdam** — every C/H is addressed, plus most M and several L:

- **C-1** strict partial/full spec schemas on PUT and preview routes
- **C-2** defensive `projectAttributes` (nullish dimensions /
objectives / owner.teams)
- **C-3/C-4** `updateDedup` now asserts
`existing.spec.workspaceId === deploy.workspaceId` at function entry
(throws `SloValidationError` on drift) and threads a single
`updateWorkspaceId` through namespace + `dedupAlertGroupName`.
`deleteDedup` keeps the `existing.spec.workspaceId ?? deploy.workspaceId`
source-of-truth fallback for namespace + group hash but does **not**
yet assert mismatch — today's deploy context is hardcoded to
`'default'` (see `buildDeployContext`) so they cannot diverge in this
PR. Symmetric assert will land alongside PR 2's real workspace
plumbing; happy to add it here defensively if preferred.
- **C-5** `goodEventsFilter` validated to a single PromQL matcher;
commas + control chars rejected
- **C-6/M-9/M-17** `rulesToYaml` unified on `js-yaml`;
name/service/tier/teams reject control chars + U+2028/U+2029
- **C-7** `rawBody` truncated + ANSI-stripped on response, full body
still logged server-side
- **H-1/H-2** pagination accumulators fixed (matches @joshuali925's #2)
- **H-3** `MAX_RETRIES=5` with 0–50ms randomized jitter between
retries
- **H-5** PromQL balance check strips backtick strings + `#` comments
- **H-7** DELETE tolerates unregistered datasources (logs warn, still
blocks teardown-required SOs to prevent silent orphans)
consistently
- **H-9** `coerceRuleGroupList` throws on `{ status: "error" }`
envelopes so PR 5's reconciler can't misread them as empty 
- **H-10** wrapped-empty-namespace check parses JSON, pins to
`details`/`error.details`
- **H-11** increment/decrement no longer re-spread the pre-write
`nextAttrs` over the post-write attributes (the original race vector).
Followup 2f1dd59 reworded the doc-comment to honestly reflect that
`SavedObjectsClientContract.update` types `attributes` as `Partial<T>`
— the surviving `{...prior, ...updated.attributes}` shape is
functionally equivalent today (the OSD repo echoes the request body)
but defends against either a future SO wrapper that strips fields or
a refetch-on-write impl that surfaces a concurrent reconciler edit to
`zeroSinceAt`.
- **H-12/H-13/H-14/H-15/M-15** wizard polish: `onSubmit`+preventDefault,
`<Prompt>` for unsaved changes, unmount guard, target as raw text,
per-field `isInvalid`/`error`
- **Medium**: M-2/M-3 decimal formatting + fingerprint normalization,
M-4 whitespace in string literals, M-7 multi-value label errors,
M-8 CR in PromQL, M-11 `encodeURIComponent`, M-12 `sloRow_${id}`,
M-13/M-14 memoize/drop dead prop, M-16 `__proto__` reserved,
M-18 pageSize clamp
- **Low**: L-3 shared `saved_object_helpers.ts`, L-4 503 for transport
failure, L-6 replace-fn form, L-10 tighter slug regex, L-12
code-point-safe truncate. L-1/L-2/L-5/L-7/L-8/L-9/L-11/L-13/L-14
left as documented surface quirks for follow-ups.

**Tests**: 220 SLO tests (up from 205) with new regression coverage for
workspace-mismatch, goodEventsFilter injection, control-char rejection
on user fields, and ruler error envelopes. 2016 plugin tests pass
total; 0 TS errors; 0 lint errors. Validated end-to-end via Playwright
against a local observability-stack.

**Open for your call:**
- **H-6** (TOCTOU): deferred to PR 2 alongside the deploy-context
resolver — happy to land here instead if preferred.
- **H-7**: keeps `SloRulerTeardownRequiredError` active when the SO
carries a rule group; want a `?force=true` opt-out in this PR?

TackAdam (approved):
Thanks for addressing comments.
LGTM as start for SLO

joshuali925 (approved):

PR #2677, opened 2026-05-13
[Alerts] UI Refactor

TackAdam:
### Description

| Before | After |
|---|---|
| Before1| After1|
| Before2| After2|
| Before3| After3 |

### Navigation & naming
- **Renamed "Alert manager" → "Alerts"** in the left side-nav.
- **Moved the nav link** out of the "Tools" category and above "Dashboards" in the Observability workspace.
- Updated the breadcrumb (`Observability Stack / Alerts` instead of `… / Alert Manager`).
- Renamed user-facing strings that still said "Alert Manager" (Beta badge tooltip, Advanced Settings labels for `alertManagerSelectedDatasources` / `alertManagerMaxDatasources`, the cap-toast `<strong>` setting name) to "Alerts" so the surface is consistent.
- Suppressed the duplicate "Alerts / Alerts" segment in the page breadcrumb when the active tab label equals the page name.

Files: `common/constants/shared.ts`, `public/plugin_helpers/plugin_nav.tsx`, `public/components/alerting/alarms_page.tsx`, `public/components/alerting/home.tsx`, `server/plugin_helper/register_settings.ts`.

### Filter panel redesign (Alerts + Rules tabs)

Matched the Application Monitoring > Services filter pattern:
- **Collapsible filter panel** — uses `EuiResizableContainer`'s `togglePanel` so the panel animates closed to a thin rail with an expand affordance. Collapse icon is always visible in the header (matches Services).
- **Per-facet Clear link** — replaces the previous global "Clear (N)" button. Appears next to each facet's title only when that facet has an active selection.
- **"+N more" / "Show less" toggle** — replaces the hard slice used before. Remaining options stay available; users expand inline.
- **Cascade clear on datasource** — when the datasource facet is cleared (or the last option is unchecked), dependent facets (severity, state, labels) also reset, because their options are derived from the selected datasources.
- **No `(count)` badge on Datasource** — the count is always 0 or 1 there and not meaningful. New `showCounts` prop on `FacetFilterGroup` (default true) lets callers opt out.
- **Truncation with hover tooltip** — long datasource names ellipsize and show a custom instant-hover tooltip with the full name. Tooltip uses `position: fixed` so it escapes the filter panel's `overflow: hidden`. Native `title` / `EuiToolTip` approaches were both unreliable inside `EuiCheckbox` labels.

Files: `public/components/alerting/facet_filter_panel.tsx`, `public/components/alerting/alerts_dashboard.tsx`, `public/components/alerting/monitors_table/index.tsx`, `public/components/alerting/monitors_table/monitors_filters_panel.tsx`.

### Datasource type icons + Backend filter / column removal

Surfaced the backend type as a leading **OpenSearch** / **Prometheus** logo where the user already looks for it, then removed the redundant Backend surfaces:

- **Per-option icon in the Datasource facet** — new opt-in `iconMap?: Record<string, string>` prop on `FacetFilterGroup`. Both panels build a `name → 'logoOpenSearch' | 'logoPrometheus'` map from the `Datasource[]` they already receive and forward it on the Datasource facet only.
- **Per-row icon in both tables** — Alerts table's "Alert" name cell and Rules table's "Name" cell now prefix the title with the matching backend logo via `EuiButtonEmpty`'s built-in `iconType` prop (the icon lands in EUI's vertically-centered icon slot, so it sits flush with the label).
- **Removed the Backend filter facet** from both panels — the icon already conveys backend, and the facet was always 1–2 options.
- **Removed the Source / Backend table columns** from the Alerts and Rules tables along with the now-unused `BACKEND_DISPLAY` map and `backend` field on `FilterState`.
- **Per-key option-count next to label-facet headers** — Grafana-style `instance 9` subdued count rendered via the new opt-in `showOptionCount` prop on `FacetFilterGroup`. Default `false` so non-label facets stay clean.

Files: `public/components/alerting/facet_filter_panel.tsx`, `public/components/alerting/alerts_dashboard.tsx`, `public/components/alerting/monitors_table/index.tsx`, `public/components/alerting/monitors_table/monitors_filters_panel.tsx`, `public/components/alerting/monitors_table/monitors_table_columns.tsx`, `public/components/alerting/monitors_table/monitors_table_filters.tsx`, `public/components/alerting/monitors_table/monitors_table_helpers.ts`, `public/components/alerting/monitors_table/resizable_columns.ts`.

### Labels section: Grafana-style search + initial cap

Replaced the unbounded label-key list with a search box and a "Show all (N)" toggle, mirroring Grafana's label browser:

- **Search labels** — `EuiFieldSearch` filters the visible label-key list case-insensitively as the user types. Empty-state message ("No labels match") if nothing matches.
- **Initial cap of 10 labels** — `LABEL_KEY_INITIAL_VISIBLE = 10` constant, then a `Show all (N)` / `Show less` link for the rest. N reflects the post-search count, so it's accurate while filtering.
- **Internal labels stay hidden** before the cap is applied so the count never includes filtered-out keys.

Files: `public/components/alerting/alerts_dashboard.tsx`, `public/components/alerting/monitors_table/monitors_filters_panel.tsx`.

### Time picker moved into the Alert Timeline panel

The time picker now lives at the top right of the Alert Timeline panel header, adjacent to the chart it controls, instead of crowding the page-level tab bar:

- New props on `AlertsDashboardProps`: `pickerStart`, `pickerEnd`, `onTimeChange`, `onRefresh`.
- The dashboard renders the picker inside the timeline panel as a header row (title left, picker right). The header always renders so the picker stays reachable from every empty state — including "No alerts in selected time range" where the user typically wants to widen the window.
- `alarms_page.tsx` lifted the change/refresh logic into stable `useCallback` handlers (`handleTimeChange`, `handleRefreshTime`) and forwards them along with the date-math strings.

Files: `public/components/alerting/alerts_dashboard.tsx`, `public/components/alerting/alarms_page.tsx`.

### Alert Timeline tooltip no longer clipped

Promoted the ECharts tooltip to `document.body` so it can escape the chart's bounding box. Setting `tooltip.appendToBody: true` (and `confine: false`) on the timeline spec stops the tooltip from clipping against the chart's top edge when hovering tall buckets.

`public/components/alerting/alerts_charts.tsx`.

### Prometheus pending alerts now visible

The Alerts tab was missing `pending` rows for Prometheus rules even though the Rules tab displayed `pending` correctly. Root cause: `directquery_prometheus_backend.ts` filtered `pending` out of the live `/api/v1/alerts` response before merging it into the unified summary.

The historical matrix path stays scoped to `alertstate="firing"` (point-in-time pending isn't reliably reconstructible from a coarse step — the count would flap with the picker), but the live response is point-in-time and unambiguous, so currently-pending alerts now flow through. `promAlertToUnified` already mapped `pending` → unified `pending`, and the `UnifiedAlertState` union, `STATE_COLORS`, and `STATE_HEALTH` already supported it on the UI side.

Files: `server/services/alerting/directquery_prometheus_backend.ts`, `server/services/alerting/__tests__/directquery_prometheus_backend.test.ts` (test rewritten to assert pending now surfaces).

### Empty states on the Alerts tab

Replaced the single unconditional "No Active Alerts" panel with three contextual empty states that swap only the chart area (the time picker, search, and table chrome stay rendered so users can adjust selection without losing context):

1. **No datasource selected** — prompts the user to pick one on the left. Shows a "Reset to default datasources" button when the `ALERT_MANAGER_DEFAULT_DATASOURCES_SETTING` advanced setting has values.
2. **No rules created** — shown when a datasource is selected but `rulesTotal === 0`. Includes a "Go to Rules" button that switches tabs.
3. **No alerts in selected time range** — shown when rules exist but no alerts fired in range. Message suggests expanding the time range.

Also suppressed the `.opendistro-alerting-config` index-not-found warning banner — that 404 fires on fresh clusters that have never created a monitor. The new "No rules" empty state already guides the user to the correct action (creating a rule auto-creates the index). The suppression is tight: only hides errors that include both the index name AND one of `index_not_found` / `IndexNotFoundException` / `alerting_exception`; unrelated errors still surface.

Files: `public/components/alerting/alerts_dashboard.tsx`, `public/components/alerting/alarms_page.tsx`, `public/components/alerting/__tests__/alerts_dashboard.test.tsx`.

### Layout & spacing

- **Filter panel top aligns with the first content panel** on both Alerts and Rules tabs (matched `padding-top: 8px` across both).
- **Matched panel sizes** — both tabs use `initialSize={15}` for the filter panel and `initialSize={85}` for content.
- **Fixed chart horizontal overflow** — when the filter panel is re-expanded after a collapse, the Alert Timeline chart now shrinks correctly (`min-width: 0` propagated through flex containers so the chart's `ResizeObserver` can observe a width reduction).
- **Reduced spacers** throughout — `m`/`l` spacers on the chart, table, and Routing tab reduced to `s` for a denser layout.
- Removed the leftover `EuiSpacer size="s"` between the tab bar and body (8px of dead space below the tabs).

### CSS convention
Moved inline `style={{ … }}` props into a new `alerting.scss` file with a 3-letter `alt` prefix (`altResizableContainer`, `altFiltersPanel`, `altFiltersInner`, `altFiltersBody`, `altContentPanel`, `altVizRow`, `altVizItem`, `altVizPanel`, `altFacetCheckboxRow`, `altFacetCount`, `altTruncatedLabel`, `altTruncatedLabelWrap`, `altTruncatedLabelTooltip`), matching the project's styling conventions. Also folded the `.monitors-table-wrapper` rules out of an inline `<style>` block and moved them into the same SCSS file. Tooltip styling now uses OUI vars (`$euiColorFullShade`, `$euiSizeXS`, `@include euiBottomShadowSmall`) so it themes correctly in dark mode.

`public/components/alerting/alerting.scss` (new).

### Hex literal sweep

Replaced inline `color: '#...'` props that were inconsistent with the rest of the SCSS conversion:

- `facet_filter_panel.tsx` — option-count badge now uses `EuiText color="subdued"`.
- `monitors_table_columns.tsx` — name-cell `#006BB4` replaced by `EuiButtonEmpty color="primary"` + `iconType` slot; three `#999` placeholders (`—` / "None") replaced with `EuiTextColor color="subdued"`.
- `monitors_filters_panel.tsx` — saved-search clickable rows use `EuiLink` instead of an inline-styled span.

### Code-review nits cleaned up

- **Removed unused `BASE_COLUMNS` / `ColumnDef` exports** from `monitors_table_columns.tsx` (dead since the column-picker UI was removed in this branch) and updated stale doc comments.
- **Hoisted `INTERNAL_LABEL_KEYS`** to a single shared definition in `monitors_table/monitors_table_helpers.ts`. `alerts_dashboard.tsx` imports from there instead of redefining the Set, so the two panels can't drift.
- **Dropped `labelsObj.monitorType` emission** in `alarms_page.tsx` — the camelCase label was leaking into the Labels facet because `INTERNAL_LABEL_KEYS` filters `monitor_type` (snake_case). The field already exists on `UnifiedRule` directly.
- **Tightened `onCreateMonitor` union** in `monitors_table/index.tsx` and `monitors_main_panel.tsx` to `'logs' | 'prometheus' | 'metrics'` (dropped `'slo'` since SLO listing is gone).
- **Dropped a stray `console.warn`** in the self-healing time-range path; the recovery is already self-evident through the toast surface.

### Filter panel: default-collapsed Labels with a horizontal rule separator

- **Datasource / Severity / State** stay expanded by default.
- **Labels** facets (dynamically discovered from alert data) all start collapsed; expanding one persists the user's choice across re-renders.
- Replaced the `<EuiSpacer size="xs" />` separator above the Labels header with `<EuiHorizontalRule margin="s" />`, matching the Services page's filter panel. Both filter panels use the same separator now.

`useFacetCollapse` was redesigned for this: instead of a `Set<string>` of collapsed IDs, it now stores a `Map<string, boolean>` of explicit user overrides, with `isCollapsed(id, defaultCollapsed?)` and `toggleFacetCollapse(id, defaultCollapsed?)` accepting a per-call default. This lets dynamic-key facets (label keys discovered at render time) default to collapsed without re-collapsing on every render.

### Removed the Alerts-tab summary cards

Deleted the row of 5 stat cards (Total / Active / Critical / High / Medium / Low) above the Alert Timeline. Severity and State are still filterable via the left filter panel, so the cards were a redundant entry point.

- Deleted `public/components/alerting/alerts_summary_cards.tsx` and its test file.
- Stripped the `<AlertsSummaryCards>` JSX, the `severityFilter`/`stateFilter` state, and the derived `severityCounts` / `activeCount` / `isFiltered` selectors from `alerts_dashboard.tsx`.
- Simplified `filteredAlerts` (no more "medium expands to medium/low/info" branch), `clearDependentFilters`, `updateFilter`, and the table empty-message check.
- Removed the leftover `<EuiSpacer size="s" />` that previously separated the cards from the chart so the chart top aligns with the filter panel top.

### Removed Export / Import / Columns buttons (Rules tab)

These were unused features on top of the Rules table action bar. Removed both the UI and the now-dead client/server plumbing:

- **UI** — Stripped the three buttons (and the `EuiCheckboxGroup` column-picker popover) from `monitors_main_panel.tsx`. `MonitorsMainPanelProps` lost `allColumns` / `visibleColumns` / `setVisibleColumns` / `showColumnPicker` / `setShowColumnPicker` / `exportJson` / `onImport` / `handleImportFile`.
- **`monitors_table/index.tsx`** — Dropped `onImport` from the public prop interface, the `serializeMonitors` import, the `BASE_COLUMNS` import, the `exportJson` and `handleImportFile` handlers, and the picker/state plumbing. `visibleColumns` is now a frozen `useMemo(() => new Set(DEFAULT_VISIBLE), [])` so `buildTableColumns` and `useResizableColumns` keep working unchanged — column drag-resize is still active.
- **`alarms_page.tsx`** — Removed `handleImportMonitors` and its three associated toast strings (`selectDatasourceForImport` / `monitorsImported` / `importMonitorsFailed`); dropped the `onImport={...}` prop on `<MonitorsTable>`.
- **Dead transport code** — Deleted `common/services/alerting/serializer.ts` and its test file; removed the `export * from './serializer'` line from `common/services/alerting/index.ts`. Stripped `importMonitors` / `exportMonitors` (and the `MonitorImportResponse` / `MonitorExportResponse` interfaces) from `monitor_mutations_client.ts` and `use_monitor_mutations.ts`. No server route handlers existed for the import/export endpoints, so the client methods were calling URLs that didn't resolve anyway.

### i18n conversion across the alerting feature

All user-facing strings under `public/components/alerting/` are now wrapped with `i18n.translate(...)` (or `<FormattedMessage>` for strings interpolating values), using the namespace pattern `observability.alerting.<file>.<element>` already established in `alarms_page.tsx`.

- ~270 strings converted across 33 files: titles, labels, placeholders, aria-labels, table column headers, button text, callouts, toasts, helpText, empty-state copy, and the visible `text` fields of dropdown option arrays.
- Mustache action templates (`{{ctx.monitor.name}}`) left as plain strings — those are server-evaluated, not user copy.
- For ICU MessageFormat conflict cases (e.g., `{{ $labels.instance }}` example placeholders), the literal Prometheus snippet is passed through `values: { example: '{{ $labels.instance }}' }` so the i18n message text stays valid ICU.
- `create_logs_monitor/index.tsx` dirty-check used to compare against the literal default trigger name `"Trigger 1"`. The check was updated to re-derive the same translated default so it stays correct in any locale.

Skipped (intentional): `data-test-subj` / `id` attrs, internal field keys, console messages, pure transport client code, types/interfaces, severity/state enum keys (only translated where rendered as visible labels).

### Test environment polyfill

Added a no-op `ResizeObserver` mock to `test/setup.jest.ts`, alongside the existing `IntersectionObserver` and `matchMedia` polyfills. The `TruncatedLabel` component in `facet_filter_panel.tsx` uses `new ResizeObserver(...)` to track label overflow, which was previously throwing in JSDOM and cascading 5 failures into `facet_filter_panel.test.tsx` and `monitors_table.test.tsx`.

### Tests

After all the above, the alerting test suites green across both frontend and backend:

- **`public/components/alerting/`** + **`common/services/alerting/`** — 179 / 179 passing across 26 suites.
- **`server/services/alerting/`** + **`server/routes/alerting/`** — combined run (alongside the frontend suites) lands at **363 / 363 passing across 41 suites**.

New / updated test cases worth calling out:

- `facet_filter_panel.test.tsx` — added coverage for the per-facet Clear link, the `+N more` / `Show less` toggle, and the `useFacetCollapse(id, defaultCollapsed)` per-call default-collapsed semantics.
- `alerts_dashboard.test.tsx` — added the four new picker props (`pickerStart` / `pickerEnd` / `onTimeChange` / `onRefresh`) to `baseProps` plus the three empty-state assertions from the original branch.
- `alarms_page.test.tsx` — replaced the `alertManager-timeRangeBar` DOM lookup (the picker no longer renders at the page level) with assertions on the props handed to the dashboard mock; the refresh test now invokes the forwarded `onRefresh` callback directly. Dropped the now-stale `console.warn` assertion in the malformed-date-math healing test.
- `directquery_prometheus_backend.test.ts` — rewrote the previously-passing `live alerts in pending state are suppressed` case to assert that pending alerts now surface (matching the Rules tab status), with both Firing and Pending appearing in the unified result.

## Files touched

| File | Notes |
|------|-------|
| `common/constants/shared.ts` | Title constant renamed to "Alerts" |
| `common/services/alerting/index.ts` | Dropped `serializer` re-export |
| `common/services/alerting/serializer.ts` | **Deleted** — dead after Export/Import removal |
| `common/services/alerting/__tests__/serializer.test.ts` | **Deleted** — dead test |
| `public/plugin_helpers/plugin_nav.tsx` | Nav order and category adjusted |
| `public/components/alerting/alarms_page.tsx` | Merge resolution, i18n conversion, empty-state plumbing, index-missing warning filter, breadcrumb rename + duplicate-segment suppression, removed `handleImportMonitors`, dropped page-level picker (lifted into dashboard), dropped `labelsObj.monitorType` leak, removed `console.warn`, removed stray `EuiSpacer` |
| `public/components/alerting/alerts_dashboard.tsx` | Merge resolution, i18n, removed summary cards, default-collapsed Labels + `EuiHorizontalRule`, 3 empty states, inline-styles → SCSS, label search + 10-row cap, picker now rendered in timeline panel header, datasource icon prefix on Alert column + Source column removed, hoisted `INTERNAL_LABEL_KEYS` import |
| `public/components/alerting/alerts_summary_cards.tsx` | **Deleted** |
| `public/components/alerting/__tests__/alerts_summary_cards.test.tsx` | **Deleted** |
| `public/components/alerting/facet_filter_panel.tsx` | i18n, per-facet Clear, "+N more", truncation tooltip, `showCounts` prop, `useFacetCollapse` accepts per-call default-collapsed, new `iconMap` + `showOptionCount` props, count badge moved to `EuiText` |
| `public/components/alerting/alerts_charts.tsx` | i18n, `tooltip.appendToBody` so the chart tooltip escapes the panel |
| `public/components/alerting/alert_detail_flyout.tsx` | i18n |
| `public/components/alerting/monitor_detail_flyout.tsx` | i18n |
| `public/components/alerting/monitor_form_components.tsx` | i18n (with ICU `values` workaround for Prometheus templating) |
| `public/components/alerting/monitor_template_wizard/index.tsx` | i18n |
| `public/components/alerting/monitor_template_wizard/alert_templates.ts` | i18n on visible `name` / `description` fields |
| `public/components/alerting/monitors_table/index.tsx` | i18n, removed Export/Import/Columns plumbing, frozen `visibleColumns`, dropped `uniqueBackends` / `filters.backend`, tightened `onCreateMonitor` union |
| `public/components/alerting/monitors_table/monitors_main_panel.tsx` | i18n, removed Export/Import/Columns buttons + popover, removed inline `<style>` block, tightened `onCreateMonitor` union |
| `public/components/alerting/monitors_table/monitors_filters_panel.tsx` | i18n, cascade-clear, truncation, `EuiHorizontalRule` separator, label-facet default-collapsed parity, label search + 10-row cap, datasource `iconMap`, dropped Backend facet, saved-search rows use `EuiLink` |
| `public/components/alerting/monitors_table/monitors_table_columns.tsx` | i18n, dropped `BASE_COLUMNS` / `ColumnDef`, removed Backend column, prefixed Name with `iconType` for backend logo, replaced `#999` placeholders with `EuiTextColor color="subdued"` |
| `public/components/alerting/monitors_table/monitors_table_filters.tsx` | Removed `backend` from `FilterState` / `emptyFilters` / `matchesFilters` |
| `public/components/alerting/monitors_table/monitors_table_helpers.ts` | Removed unused `BACKEND_DISPLAY` + its `i18n` import |
| `public/components/alerting/monitors_table/resizable_columns.ts` | Dropped `backend` width entry |
| `public/components/alerting/notification_routing_panel.tsx` | i18n, spacer reductions |
| `public/components/alerting/promql_editor.tsx` | i18n |
| `public/components/alerting/metric_browser.tsx` | i18n |
| `public/components/alerting/home.tsx` | i18n (Beta badge), "Alert Manager" → "Alerts" leftover strings |
| `public/components/alerting/hooks/use_datasources.ts` | i18n (Local Cluster fallback name) |
| `public/components/alerting/hooks/use_monitor_mutations.ts` | Stripped `importMonitors` / `exportMonitors` |
| `public/components/alerting/mutations/monitor_mutations_client.ts` | Stripped `importMonitors` / `exportMonitors` and their response types |
| `public/components/alerting/create_monitor/*.{ts,tsx}` | i18n across all create-monitor surfaces |
| `public/components/alerting/create_metrics_monitor.tsx` | i18n |
| `public/components/alerting/create_logs_monitor/**/*.{ts,tsx}` | i18n across all create-logs-monitor surfaces |
| `public/components/alerting/__tests__/alerts_dashboard.test.tsx` | Coverage for each of the 3 new empty states + picker prop additions |
| `public/components/alerting/__tests__/alarms_page.test.tsx` | Updated for picker-relocation + dropped `console.warn` assertion |
| `public/components/alerting/__tests__/facet_filter_panel.test.tsx` | New cases: per-facet Clear, `+N more` toggle, `useFacetCollapse` per-call default-collapsed |
| `public/components/alerting/alerting.scss` | New stylesheet — also picked up `.monitors-table-wrapper`, `.altFacetCount`, OUI-token tooltip |
| `server/plugin_helper/register_settings.ts` | Setting labels renamed to "Alerts …" instead of "Alert Manager …" |
| `server/services/alerting/directquery_prometheus_backend.ts` | Stop suppressing live `pending` alerts so the Alerts tab matches the Rules tab |
| `server/services/alerting/__tests__/directquery_prometheus_backend.test.ts` | Test rewritten — pending now surfaces |
| `test/setup.jest.ts` | `ResizeObserver` polyfill for JSDOM |

### Check List


public/components/alerting/alerts_dashboard.tsx
```diff
@@ -282,69 +309,31 @@ export const AlertsDashboard: React.FC<AlertsDashboardProps> = ({
...
-    searchQuery !== '' ||
-    severityFilter !== 'all' ||
-    stateFilter !== 'all';
-
-  const clearAllFilters = () => {
+  // Clearing the datasource filter must also clear dependent facets because
+  // severity/state/label options are derived from the currently selected
+  // datasources' alerts — leaving stale selections would filter against
+  // values that no longer exist in the visible dataset.
+  const clearDependentFilters = () => {
```
lezzago, line 333: Cascade-clear behavior diverges between tabs: `clearDependentFilters` here only resets `filters`, but the Rules tab's `clearAllFilters` (in `monitors_table/index.tsx:373`) also resets `searchQuery`. So clearing the datasource on Alerts keeps the user's search box, while doing the same on Rules wipes it. Pick one — and please add a regression test either way; the cascade isn't currently covered.

public/components/alerting/alerts_dashboard.tsx
```diff
@@ -552,6 +598,10 @@ export const AlertsDashboard: React.FC<AlertsDashboardProps> = ({
                       .map((l) => datasourceEntries.find((e) => e.label === l)?.id)
                       .filter(Boolean) as string[];
                     onDatasourceChange(ids);
+                    // Clearing datasource invalidates the other facet options
+                    // (severity/state/backend/labels are derived from the
```
lezzago: Stale: comment still mentions `severity/state/backend/labels are derived…`, but the `backend` facet was removed in this PR.

lezzago (commented):
Two comments inline. Otherwise this looks good — happy to re-approve once the cascade-clear divergence is reconciled.

kamingleung:
Thanks for the updates!

@TackAdam Small nit, in the filter panel, can the data source logo be the same size as the ones in the table?
image


@TackAdam @lezzago On the Rules listing tab, how do we determine the severity of a rule for Prometheus and OpenSearch?

TackAdam:
Updated icon to use small size to match table
iconSize

public/components/alerting/alerts_dashboard.tsx
TackAdam: Cascade-clear divergence reconciled — alerts_dashboard.tsx:331-340 now also clears searchQuery when datasources are deselected, matching monitors_table/index.tsx:224 (clearAllFilters). The comment explains the alignment.

Regression tests added:
  - alerts_dashboard.test.tsx: typing into the search box, then unchecking the only datasource, verifies the search clears (cascade-clear).
  - monitors_table.test.tsx: same regression for the Rules tab, keeping the two tabs aligned.

public/components/alerting/alerts_dashboard.tsx
TackAdam: Stale comment fixed — alerts_dashboard.tsx:601-603: removed /backend reference (the backend facet was removed in this PR).

lezzago (approved):

joshuali925 (approved):

sumukhswamy (approved):

PR #2683, opened 2026-05-18
Optimize Alert Manager flyouts and detail-fetch paths

lezzago:
# Optimize Alert Manager flyouts and detail-fetch paths

Cuts O(N) reverse-lookup scans on every flyout open and removes
static/mock sections that didn't reflect real data.

## Why

- Opening any alert flyout used to fetch the full alert list from the
  upstream and `.find()` the one alert by id.
- Opening any rule flyout used to fetch the full alert list to build
  a single monitor's history, plus an unconditional destinations
  fetch that powered an empty-by-default Notification Routing
  accordion, plus an opportunistic live re-execution of the monitor
  against production data.
- Several alert-flyout sections rendered static lookup tables labeled
  as if they were real (AI Analysis, Suggested Actions, Suppression
  Status, Notification Routing). Removing them is a correctness fix.
- The rule flyout's Notification Routing accordion never showed
  useful Prometheus data — Alertmanager owns Prom routing. The
  Routing tab is the right surface for that.
- The rule flyout's "Recent Alert History" + "Suppression Rules"
  accordions were misnamed (Prom only ships currently-firing alerts,
  never historical) and never populated (OS Alerting has no
  suppression-rules concept), respectively.

### Alert detail flyout

- **Backend:** alert detail call accepts `monitorId` and scopes the
  upstream `monitors/alerts` request to that monitor. Prom branch
  returns `null` (the summary already has everything needed;
  full-list scans don't make sense here).
- **Frontend:** flyout renders from `UnifiedAlertSummary` on open.
  The Raw Alert Data accordion lazy-loads its detail call when
  expanded.
- **Removed:** AI Analysis, Suggested Actions, Suppression Status,
  Notification Routing accordions (all static mocks).

### Rule detail flyout

- **Backend:** rule detail's alert-history fetch is scoped to a
  single monitor via `monitorId` and now passes `limit: 20` so the
  upstream returns one small page even on busy monitors.
  `getOSRuleDetail` no longer over-fetches and `.slice(0, 20)`s
  client-side.
- **Backend:** removed the `runMonitor(_, dryRun=true)` preview
  fallback, which re-executed the customer's monitor against live
  data on every flyout open when the cheap path returned no points.
- **Backend:** the destinations call no longer rides every flyout
  open — `getOSRuleDetail` returns `notificationRouting: []`.
- **Frontend:** Notification Routing accordion removed (Prom routing
  lives on the Routing tab; OS routing data was paid-for on every
  flyout open behind a collapsed accordion).
- **Frontend:** Suppression Rules accordion removed. OS Alerting has
  no suppression-rules concept; the section was always empty.
- **Frontend:** alert-history accordion is backend-aware. OS shows
  "Recent alerts" (mixed-state slice from the alert history index);
  Prom shows "Currently firing/pending alerts" (the rule's embedded
  `alerts[]` payload only covers those states).
- **Frontend:** the creation/modification metadata accordion renamed
  from "History" to "Details" so it no longer collides with alert
  history terminology.

### Backend APIs by flyout type

This section walks the four (flyout × backend) combinations and lists
the upstream HTTP requests each one fires. Every flyout open used to
make at least one upstream call; after this PR, alert flyouts make
zero calls on open and the rule flyout's "recent alerts" pulls a
single 20-row page instead of paginating cluster-wide.

#### Top-line call counts

| Flyout × Backend | Pre-PR (open) | Post-PR (open) | Post-PR (Raw Alert Data expand) |
|---|---|---|---|
| Alert × OpenSearch | 1 cluster-wide alerts scan | **0** | 1 monitor-scoped alerts page |
| Alert × Prometheus | 1 full firing-alerts scan | **0** | **0** (UI short-circuits) |
| Rule × OpenSearch | `getMonitor` + cluster-wide alerts pagination + preview + destinations + `_execute` dry-run fallback | `getMonitor` + 1 monitor-scoped alerts page (size 20) + preview | n/a |
| Rule × Prometheus | `getRuleGroups` + preview | unchanged | n/a |

#### 1. Alert flyout × OpenSearch backend

**On open:** zero backend calls — flyout renders from the summary props it was handed.

**On user expanding the *Raw Alert Data* accordion:** one OSD plugin route → one OS Alerting REST request.

```
AlertDetailFlyout (alert_detail_flyout.tsx — onToggle expand)
  → AlertingOpenSearchService.getAlertDetail(dsId, alertId, monitorId)
    → GET /api/alerting/alerts/{dsId}/{alertId}?monitorId={monitorId}    ← OSD route
      → handleGetAlertDetail
        → MultiBackendAlertService.getAlertDetail
          → getAlertDetail (OS-only)
            → osBackend.getAlerts({ findAlertId, monitorId })
              → GET /_plugins/_alerting/monitors/alerts                  ← upstream
                  ?size=100&startIndex=0&monitorId={monitorId}
```

`findAlertId` short-circuits the page loop on first match, so this is
at most one upstream request when `monitorId` is provided (which it
always is for OS alerts). `?monitorId=…` is what makes this O(one
monitor) rather than O(cluster) — the OS Alerting REST endpoint
doesn't support per-id filtering, so this scoping plus the
`findAlertId` early-exit is how we cap the work.

#### 2. Alert flyout × Prometheus backend

**On open:** zero backend calls.

**On user expanding the *Raw Alert Data* accordion:** still zero
backend calls. Prometheus has no per-alert lookup endpoint, so the UI
short-circuits before sending anything (`alert_detail_flyout.tsx`
skips the request when `datasourceType === 'prometheus'`). The flyout
renders the summary's labels/annotations directly.

Pre-PR this was a full scan of `GET /api/v1/alerts` (or rule-derived
fallback) just to find one row by id. That scan is gone.

#### 3. Rule (Monitor) flyout × OpenSearch backend

**On open:** one OSD route → up to three OS requests, fired eagerly via `useMonitorDetail`.

```
MonitorDetailFlyout
  → useMonitorDetail.getRuleDetail(dsId, ruleId)
    → GET /api/alerting/rules/{dsId}/{ruleId}                            ← OSD route
      → handleGetRuleDetail
        → MultiBackendAlertService.getRuleDetail
          → getRuleDetail → getOSRuleDetail
            ├─ osBackend.getMonitor(monitorId)
            │    → GET /_plugins/_alerting/monitors/{monitorId}          ← upstream #1
            ├─ osBackend.getAlerts({                                      ← best-effort
            │    monitorId, limit:20, sortString:'start_time', sortOrder:'desc'
            │  })
            │    → GET /_plugins/_alerting/monitors/alerts               ← upstream #2
            │        ?size=20&startIndex=0
            │        &monitorId={id}&sortString=start_time&sortOrder=desc
            └─ fetchOSPreviewTimeSeries(monitor)                          ← best-effort
                 │  // dispatches by monitor kind:
                 ├─ search/bucket: osBackend.searchQuery(indices, body)
                 │    → POST /{indices}/_search                          ← upstream #3a
                 │        (date_histogram, 1h window, 5m buckets)
                 ├─ doc-level:    osBackend.searchQuery(...)
                 │    → POST /{indices}/_search                          ← upstream #3b
                 └─ cluster_metrics: osBackend.runMonitor(id, dryRun=true)
                      → POST /_plugins/_alerting/monitors/{id}/_execute  ← upstream #3c
                          ?dryrun=true
```

"Best-effort" means each is wrapped in try/catch; the flyout still
renders if either fails.

Pre-PR the recent-alerts fetch paginated cluster-wide using
shard-order ordering, which on a busy monitor was both expensive and
produced a non-deterministic slice. This PR pins it to one 20-row
page, sorted by `start_time desc`, scoped by `monitorId`.

Pre-PR the preview also fell back to `runMonitor(id, dryRun=true)` for
*every* monitor kind — re-executing the customer's monitor against
live data on every flyout open. That fallback is gone; only
`cluster_metrics` monitors (which genuinely need it) still hit
`_execute`.

#### 4. Rule (Monitor) flyout × Prometheus backend

**On open:** one OSD route → one or two Prometheus HTTP API calls,
routed through OSD's `_directquery` adapter (provided by the SQL
plugin).

```
MonitorDetailFlyout
  → useMonitorDetail.getRuleDetail(dsId, ruleId)
    → GET /api/alerting/rules/{dsId}/{ruleId}                            ← OSD route
      → handleGetRuleDetail
        → MultiBackendAlertService.getRuleDetail
          → getRuleDetail → getPromRuleDetail
            ├─ promBackend.getRuleGroups(ds)
            │    → GET /api/v1/rules                                     ← upstream Prom
            │        (proxied via _plugins/_directquery/_resources/{dqName}/api/v1/rules)
            │        — full listing; we loop in-process to find {ds}-{group}-{rule}
            └─ fetchPromPreviewData(query, rule)                          ← best-effort
                 → promBackend.queryRange(ctx, ds, query, start, end, step)
                   → runPromQLRange (strategy='PROMQL')
                     → POST /_plugins/_directquery/_query/{dqName}       ← OSD adapter
                       → POST /api/v1/query_range                        ← upstream Prom
                           (1-hour window, 60s step)
                 // Falls back to extracting points from rule.alerts[].value
                 // if queryRange returns empty or throws.
```

All Prometheus traffic is mediated by the SQL plugin's direct-query
adapter; the wire path on the Prom side is plain `/api/v1/rules` and
`/api/v1/query_range`. The plugin gives us auth, multi-tenant routing,
and DataFrame normalization "for free".

`getRuleGroups` returns the entire rule listing — Prometheus doesn't
expose a per-rule endpoint, so the in-process find is unavoidable.
That's the same pre- and post-PR.

#### Before
Screenshot 2026-05-18 at 5 18 16 PM
Screenshot 2026-05-18 at 5 18 36 PM
Screenshot 2026-05-18 at 5 18 59 PM
Screenshot 2026-05-18 at 5 19 24 PM

#### After
Screenshot 2026-05-18 at 5 06 22 PM
Screenshot 2026-05-18 at 5 06 44 PM
Screenshot 2026-05-18 at 5 07 43 PM
Screenshot 2026-05-18 at 5 08 03 PM

## Out of scope

- Server-side timeline chart endpoint
- Server-side facets endpoints
- Server-side pagination on the unified listing endpoints
- Prom listing filter probe + cache
- Alertmanager-primary alerts dispatch
- Historical Prometheus alerts feature
- Prom rule-detail full-groups scan optimization (requires the
  filter probe + cache work to land first)
- Listing payload trims (rules `alerts[]` strip, `?type=alert`
  pushdown)

## Test plan

- [x] `yarn test --testPathPattern alerting` (plugin-scoped) — 368/368 green
- [x] `yarn lint:es` — clean
- [x] `tsc --noEmit -p tsconfig.json` — clean (only pre-existing tsconfig deprecations)
- [x] Manual: opened OS alert flyout — zero detail calls on open;
  one call with `monitorId` query param when Raw Alert Data
  accordion expanded.
- [x] Manual: opened Prom alert flyout — zero detail calls on open;
  one call on Raw Alert Data expand (returns 404 → flyout falls
  back to summary as designed).
- [x] Manual: opened OS rule flyout — one detail call scoped to one
  monitor's alerts; zero destinations upstream calls; zero
  `_execute` dry-runs.
- [x] Manual: AI Analysis / Suggested Actions / Suppression Status /
  Notification Routing accordions are gone from the alert flyout.
- [x] Manual: Notification Routing + Suppression Rules accordions
  are gone from the rule flyout. "Recent alerts" / "Currently
  firing/pending alerts" labels render per backend; "History"
  accordion renamed to "Details".


lezzago:
Addressing remaining comments from the bots — walked each finding against 85a4492e; no changes needed.

**PR Code Analyzer — `sanitizeExternalUrl` deletion.** The helper had zero callers when removed (the only consumer, Suggested Actions, was deleted earlier in this PR as static/mock content). The analyzer notes itself there's no XSS exposure today. If external-URL rendering returns, the helper and its tests should land alongside the new caller, not orphaned.

**Reviewer Guide #1 — "slice may produce fewer rows than `requestedLimit`."** Not reachable. `allAlerts` only grows (push at line 282 in range mode, line 292 otherwise); the early-break at line 295 fires only when `allAlerts.length >= requestedLimit`, so `slice(0, requestedLimit)` always yields exactly `requestedLimit` rows.

**Reviewer Guide #2 — Prom branch returns `null`.** Intended design. Prom has no per-alert lookup; the prior full-scan was wasted work because the flyout's `UnifiedAlertSummary` already carries the labels/annotations (`alertData = detailData ? { ...alert, ...detailData } : alert`).

**Code Suggestions — "Verify truncation with filters" (3/10).** The `!hasRange` gate is intentional. In range mode the authoritative "more upstream" signal is `pageAlerts.length === PAGE_SIZE`; the suggested unconditional `totalAlerts > allAlerts.length` would false-positive when the range filter legitimately exhausted matches but the upstream's unfiltered total is larger.

**Stale suggestions** (already addressed): race-condition in fetch callback → `600db9f6`; redundant slice → bot agrees it's needed for range overshoot; zero-limit handling → `> 0` is intentional, maps zero to "no limit."

TackAdam:
The intent and overall direction are solid: removing the alert-list full-scans, the dry-run re-execution, the mock-data accordions, and bounding the rule-history fetch are real wins. The bulk of the diff reads cleanly. The issues below are concentrated in three areas: a probable React stale-closure bug, a missed limit on the alert-detail path, and a couple of edge cases in the new `truncated` semantics.

Severity legend: **Blocking** — must address before merge. **Should fix** — fix in this PR or follow-up. **Nit** — author's discretion.

---

## Blocking

### 1. `fetchDetailIfNeeded` has a stale-closure bug — re-firing on every accordion re-expand

**File:** `public/components/alerting/alert_detail_flyout.tsx:76-89`

The PR drops `detailData` and `detailLoading` from the `useCallback` deps and silences the lint with `eslint-disable-next-line react-hooks/exhaustive-deps`. The justification:

> "the accordion can't fire `onToggle` faster than React commits the `setDetailLoading(true)` from the prior toggle."

The reasoning is incorrect. `useCallback` returns the *same function instance* across renders when deps don't change — and that function closes over **the values from the render in which it was created**, not the latest render. With `[alert.datasourceId, alert.id, alert.monitorId, osService]` as the deps, the callback is created exactly once for the lifetime of the flyout. Inside the callback:

```ts
if (detailData || detailLoading) return;
```

`detailData` and `detailLoading` here are forever the **initial render** values — `null` and `false`. So the early-return guard never fires once the flyout is mounted. Concretely:

1. User opens the flyout. `detailData=null`, `detailLoading=false`.
2. User expands "Raw Alert Data" → `fetchDetailIfNeeded()` runs → guard passes → request fires → `setDetailLoading(true)` → fetch resolves → `setDetailData(...)` → `setDetailLoading(false)`.
3. User collapses the accordion. EuiAccordion fires `onToggle(false)` (no-op in our handler — fine).
4. User re-expands the accordion. `onToggle(true)` runs `fetchDetailIfNeeded()`. The closure sees the original `detailData=null` and `detailLoading=false`. **Guard passes again. Network call fires again.** `setDetailData` re-renders with the same data; harmless, but every collapse-and-expand cycle re-issues the upstream HTTP call against the user's cluster.

This is exactly the regression `react-hooks/exhaustive-deps` is designed to flag. The earlier-return optimization was illusory — the callback is now incorrect.

**Proposed fix:** put the deps back. The cost of recreating the callback on every state flip is a single function allocation per render — negligible compared to the network call you're trying to gate. The lint suppression and the comment block come out:

```diff
-  // `detailData` and `detailLoading` are intentionally NOT in the dep
-  // array — including them recreates the callback on every state flip
-  // and defeats memoization. The early-return guard reads the latest
-  // state via closure-on-render, which is sufficient because the
-  // accordion can't fire `onToggle` faster than React commits the
-  // `setDetailLoading(true)` from the prior toggle.
   const fetchDetailIfNeeded = useCallback(() => {
     if (detailData || detailLoading) return;
     ...
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [alert.datasourceId, alert.id, alert.monitorId, osService]);
+  }, [alert.datasourceId, alert.id, alert.monitorId, detailData, detailLoading, osService]);
```

**Alternative (if memoization is genuinely load-bearing for some downstream comparator):** read the latest state via a `useRef` updated in an effect, or via `setDetailLoading((prev) => prev ? prev : true)` style updates. But the simple revert is cleaner and the original commit already removed the only real perf concern (the eager `useEffect` fetch).

**Test coverage gap:** the new `lazy detail fetch` test in `alert_detail_flyout.test.tsx:128` only clicks the accordion once. Adding a "click → wait → click again → assert called exactly once" case would have caught this. Please add it alongside the fix.

---

### 2. `getAlertDetail` (OS path) ignores `limit` — full monitor history pulled to find one alert

**File:** `server/services/alerting/alert_detail.ts:226`

```ts
if (ds.type === 'opensearch' && osBackend) {
  const { alerts } = await osBackend.getAlerts(client, monitorId ? { monitorId } : undefined);
  const alert = alerts.find((a) => a.id === alertId);
```

`monitorId` scoping is a real win over the pre-PR cluster-wide scan, but on a busy monitor (thousands of alerts, common for noisy infra alerts) we still page through all of them just to `.find()` one row by id. The PR introduced `limit` precisely to avoid this kind of post-fetch slicing — but didn't apply it here.

Two cleaner options, in order of preference:

**(a) Use `alertIds` upstream filter.** OpenSearch Alerting's `GET /_plugins/_alerting/monitors/alerts` accepts `alertIds` as a comma-separated list. That gives a single-row response unconditionally. Add an `alertIds?: string[]` option to `OpenSearchBackend.getAlerts` (or a separate `getAlertById` helper) and the detail path becomes one tiny request.

**(b) At minimum, pass `limit: 1` plus a server-side scan.** Without `alertIds` support, this is still strictly better than pulling the whole monitor's history just to find one row. *However* — `getAlerts({ monitorId, limit: 1 })` only works if the upstream is guaranteed to return the requested alert in the first page, which it isn't (no documented ordering). So this option degrades to "may not find it". Don't pursue (b) without ordering guarantees; pursue (a).

If neither is viable in this PR, please note it explicitly in the JSDoc — the comment "the unified flyout already has it on the summary; passing it avoids the full-cluster scan" overstates what was achieved. The full-monitor scan is still happening for any monitor with >100 alerts.

---

### 3. `truncated: true` false positive when upstream has exactly `limit` rows

**File:** `server/services/alerting/opensearch_backend.ts:295-306`

```ts
if (requestedLimit !== undefined && allAlerts.length >= requestedLimit) {
  if (pageAlerts.length === PAGE_SIZE || (!hasRange && allAlerts.length < totalAlerts)) {
    truncated = true;
  }
  break;
}
```

When `limit === PAGE_SIZE` (e.g. `getOSRuleDetail` calls with `limit: 20`, so `PAGE_SIZE = min(20, 100) = 20`), and the monitor has exactly 20 alerts, this fires:

- `allAlerts.length === 20` → break condition met
- `pageAlerts.length === PAGE_SIZE` (`20 === 20`) → `truncated = true` ❌

But there's no truncation: the caller asked for 20, the upstream has 20, the result is complete. The secondary condition `(!hasRange && allAlerts.length < totalAlerts)` would correctly evaluate to `false` (20 < 20 is false) — but the `||` short-circuits on the first arm before reaching it.

The commit message claims "the page itself was full ⇒ there is at least one more row upstream" — but that's only true when `totalAlerts > limit`. The fix is to AND the page-full check with the totalAlerts check (in non-range mode), and require the totalAlerts overshoot in all modes when we have it:

```diff
-  if (pageAlerts.length === PAGE_SIZE || (!hasRange && allAlerts.length < totalAlerts)) {
-    truncated = true;
-  }
+  // We've hit the caller's cap. Truncation only happens if upstream
+  // actually has more rows than we collected. In non-range mode the
+  // upstream `totalAlerts` is the authoritative count.
+  if (!hasRange && allAlerts.length < totalAlerts) {
+    truncated = true;
+  } else if (hasRange && pageAlerts.length === PAGE_SIZE) {
+    // Range mode: no upstream filtered count is exposed, so a full page
+    // is the only "more available" signal we have. False positive on
+    // exact-page-multiple is an accepted trade.
+    truncated = true;
+  }
```

**The PR-supplied test "limit short-circuit sets truncated when more rows are available upstream" passes today only because `totalAlerts: 50` ≫ `limit: 20`.** A test like "upstream returns exactly `limit` rows ⇒ truncated stays false" is missing and would have surfaced this. Please add it.

---

## Should fix

### 4. Prom alert detail returns `null` → 404 → console error noise

**File:** `server/services/alerting/alert_detail.ts:231-237` and `public/components/alerting/alert_detail_flyout.tsx:84`

`getAlertDetail` for `prometheus` datasources unconditionally returns `null`. The HTTP route maps that to a 404. The flyout's lazy fetch path:

```ts
.catch((err: unknown) => {
  console.error('Failed to load alert details:', err);
})
```

…will log a 404 every time a user expands "Raw Alert Data" on a Prom alert. The PR description acknowledges this ("returns 404 → flyout falls back to summary as designed") but the user-facing console gets a misleading error log.

Two non-mutually-exclusive fixes:

- **Don't even attempt the call for Prom alerts on the client.** If `alert.datasourceType === 'prometheus'`, render `JSON.stringify(alert, null, 2)` directly without the network round-trip. The current behavior is "fire request, swallow 404, render summary fallback" — same end result, one less wasted request and one less misleading log line.
- **Squelch 404s in the catch.** If the response shape is reachable, branch on it: only `console.error` for non-404 failures.

The first option is preferable — the data wasn't going to come back anyway, and the comment in `alert_detail.ts:232` already justifies why.

---

### 5. `extractOSPreviewData` is now dead code at every site that mattered

**File:** `server/services/alerting/alert_preview.ts:16, 264-318`

The runMonitor dry-run fallback was the only consumer. The function is still exported and the file's contents-block JSDoc still lists it. The 11th commit (`chore(alerting): trim alert_detail JSDoc + drop unused sanitizeExternalUrl`) deletes `sanitizeExternalUrl` for exactly the same reason but misses `extractOSPreviewData`. Please delete it (and the `Contents:` line).

If there's a planned re-introduction in a follow-up phase, leave a single-line `TODO` rather than ~50 lines of unused code.

---

### 6. `notificationRouting` JSDoc misleads after the accordion deletion

**File:** `server/services/alerting/alert_detail.ts:138-140`

```ts
// Notification routing is fetched lazily by the flyout when the user
// expands the accordion. The dedicated /routing endpoint owns the
// destinations lookup so the detail path stays cheap.
notificationRouting: [],
```

The accordion was removed in commit 5; there is no longer a "lazy fetch on accordion expand" path. The Routing **tab** is canonical now. Please update to match what the diff actually shipped:

```ts
// Notification routing is no longer surfaced inline on the rule flyout.
// The standalone Routing tab (Alertmanager-fed for Prom; OS destinations
// for OS) owns it. We keep the empty array on the response shape so
// existing UnifiedRule consumers don't need a type change.
notificationRouting: [],
```

Same fix in the Prom branch at line 195.

---

### 7. `alertHistory` ordering is unspecified — "Recent alerts" may not be recent

**File:** `server/services/alerting/alert_detail.ts:79-93`

The OS Alerting `GET /_plugins/_alerting/monitors/alerts` endpoint does not document a default sort. With `limit: 20`, the upstream returns "the first 20 alerts the index returns for this monitor" — which in practice is shard-order-dependent and *not* a "most recent 20" guarantee.

The pre-PR code did `getAlerts(client)` then `.filter().slice(0, 20)` which had the same ordering problem, so this isn't a regression — but the new flyout label "Recent alerts" promises something the API doesn't guarantee. Two options:

- **Add an upstream sort.** If the API supports `sortString=start_time&sortOrder=desc` (it does as of 2.x — see [OS Alerting docs](https://opensearch.org/docs/latest/observing-your-data/alerting/api/#search-alerts)), thread it through `getAlerts` and pass it on the rule-detail call.
- **Soften the label.** "Last 20 alerts" without the "recent" promise, until ordering is wired up.

I'd push for the first; the API supports it and it makes the UI label honest. This can also be a follow-up if the team wants this PR scoped tight.

---

### 8. `monitorId` query param accepts `slash` — but route schema is `alertingIdSchema`

**File:** `server/routes/alerting/index.ts:467`

```ts
query: schema.object({ monitorId: schema.maybe(alertingIdSchema) }),
```

The new test asserts URL-encoded `mon/with/slash` is correctly forwarded by the *backend* to the upstream. But the *route* validation runs `alertingIdSchema` against the same value first. If `alertingIdSchema` rejects slashes (or any other character a real OS monitor id can contain), the route will 400 before the upstream call is even made.

Please verify by checking `alertingIdSchema`'s regex / length limits, and add a route-level handler test that exercises a representative monitor-id shape end-to-end (one that the upstream actually produces — typically a UUID-like string with hyphens, but I'd defer to whatever your test fixtures use). If `alertingIdSchema` already allows everything OS produces, drop the URL-encoded-slash test (it's testing the helper's URL building, not anything realistic) and replace it with a "valid OS monitor id round-trips through schema and into URL unchanged" test.

---

## Nits

### 9. DOM accordion `id` says `history-` but the visible label says "Details"

**File:** `public/components/alerting/monitor_detail_flyout.tsx:717`

```tsx
<EuiAccordion id={`history-${monitor.id}`} ...>
```

The i18n key was renamed to `detailsHeader` in commit 6 but the DOM id and the surrounding code comment block (`Creation / Modification History` ⇒ `Details`) still call it "history". Cosmetic but confusing for the next reader. Either rename to `details-${monitor.id}` or leave a one-line comment noting the mismatch is intentional (and why — likely "stable element id, don't churn `data-test-subj`").

### 10. `osAlertToUnified` falsy-empty-string check on `monitor_id`

**File:** `server/services/alerting/alert_utils.ts:285`

```ts
monitorId: a.monitor_id || undefined,
```

`a.monitor_id` from the OS API is a string. `'' || undefined === undefined` is fine, but worth knowing OS Alerting *will* return an empty string for monitor-less alerts in some edge cases (legacy ad-hoc monitors). The downstream `monitorId ? { monitorId } : undefined` in `alert_detail.ts:226` correctly handles `undefined`, so this works. Just calling out — no change needed.

### 11. `alarms_page.tsx` mock paths still set `notificationRouting: []` and `aiSummary: '…'`

**File:** `public/components/alerting/alarms_page.tsx:742, 753, 812, 826, …`

These are the local mock-data fallback paths (still in use during early-development). They keep the `UnifiedRule` shape valid. Out of scope for this PR but would be nice to clean up alongside whatever phase ships the unified type cleanup.

---

## What looks good

- **Pagination + limit semantics** in `opensearch_backend.ts` are otherwise carefully reasoned (the `SCAN_CAP` / `FILTER_CAP` separation, the deterministic `windowEnd` capture, the deliberate non-early-exit on stale `totalAlerts`). The new tests for monitorId encoding, page-size shrink, and limit-not-reached are well-targeted.
- **Removing the `runMonitor(_, dryRun=true)` fallback** is a real correctness + load fix. Re-executing a customer's monitor against live data on every flyout open is the kind of thing that shows up as "weird production load on Mondays" three months later.
- **Lazy raw-data accordion** is the right shape — the eager fetch on mount was wasted work for the common case where users open a flyout to read the human-readable fields.
- **Backend-aware "Recent alerts" / "Currently firing/pending alerts" label split** matches what each backend actually exposes. Nice catch.
- **Test additions for `handleGetAlertDetail` param-position pinning** — exactly the kind of test that catches a future careless reorder.
- **Mock-section deletions** (AI Analysis, Suggested Actions, Suppression Status, Notification Routing, Suppression Rules) are correctness fixes, not just cosmetic. Removing UI that shows fabricated data is the right call even if it's net-negative on screen real estate.

---

## Summary

Nothing here is structural — the PR's direction is right and most of the diff is good. The **stale-closure bug in `fetchDetailIfNeeded`** is the only must-fix-before-merge item; the **alert-detail path still scans the full monitor history** is the most-impactful perf regression-by-omission; and the **`truncated: true` false positive** is a small but real correctness gap in code that's flagged as forward-looking ("Phase-4 callers will read it") which is exactly when latent bugs bite.

Once #1, #2, and #3 are addressed, the rest is comfortable cleanup that can land in this PR or a fast follow-up.

lezzago:
Thanks for the thorough review — pushed fixes for everything in [16d6b7fb](https://github.com/lezzago/dashboards-observability/commit/16d6b7fb), [843375b3](https://github.com/lezzago/dashboards-observability/commit/843375b3), [d422cd2d](https://github.com/lezzago/dashboards-observability/commit/d422cd2d), [7e676da6](https://github.com/lezzago/dashboards-observability/commit/7e676da6).

### Blocking

**1. Stale closure in `fetchDetailIfNeeded`** — fixed in [7e676da6](https://github.com/lezzago/dashboards-observability/commit/7e676da6). You're right, the closure-on-render reasoning was wrong; restored `detailData`/`detailLoading` to the `useCallback` deps and dropped the lint suppression. Added the "click → wait → click again → assert called exactly once" test you asked for.

**2. `getAlertDetail` ignores `limit`** — addressed in [d422cd2d](https://github.com/lezzago/dashboards-observability/commit/d422cd2d), but **not via your option (a)**. I tried `&alertIds=...` first, then verified against the live observability-stack OS instance and the upstream Kotlin source — the REST handler doesn't accept that param:

```
$ curl -sk "$OS/_plugins/_alerting/monitors/alerts?alertIds=uKnORp4BFZkyfSVzw62w"
{"error":{"type":"illegal_argument_exception",
"reason":"request [/_plugins/_alerting/monitors/alerts] contains unrecognized parameter: [alertIds]"},"status":400}
```

Source: [`RestGetAlertsAction.kt`](https://github.com/opensearch-project/alerting/blob/main/alerting/src/main/kotlin/org/opensearch/alerting/resthandler/RestGetAlertsAction.kt) parses only `sortString, sortOrder, missing, size, startIndex, searchString, severityLevel, alertState, monitorId, workflowIds`. `alertIds` exists on the internal Java transport-client API and on `GetAlertsRequest`'s constructor in common-utils, but the REST shim explicitly passes `null` for that slot. I also tried `alert_ids`, `id`, `ids`, `_id`, `alertId` — all rejected with HTTP 400. `searchString` matches against `monitor_name`/`trigger_name`, not alert id.

So your option (a) isn't actually viable, and option (b) (`limit: 1` without ordering guarantee) is correctly ruled out. What I shipped instead: a client-side `findAlertId` early-exit on the existing pagination loop. With `monitorId` upstream scoping (already in place), the worst case becomes "page through one monitor's alerts until found" instead of "page through them all." Best case is one upstream request when the row is on page 1. The inline comment in `getAlerts` documents that `&alertIds=...` is unsupported by the REST handler so a future reader doesn't re-attempt the URL approach.

If you happen to know of a different upstream filter I missed, happy to switch — but every variant I tried returns 400.

**3. `truncated: true` false positive** — fixed in [d422cd2d](https://github.com/lezzago/dashboards-observability/commit/d422cd2d). Reordered as you suggested: non-range mode uses `allAlerts.length < totalAlerts` only; range mode keeps the page-full heuristic with the documented exact-page-multiple trade. Added the regression test "upstream returns exactly limit rows with no overshoot ⇒ truncated stays false."

### Should fix

**4. Prom 404 console noise** — went with the first option in [7e676da6](https://github.com/lezzago/dashboards-observability/commit/7e676da6): short-circuit `fetchDetailIfNeeded` for `alert.datasourceType === 'prometheus'`. No wasted request, no misleading log line. Test added.

**5. `extractOSPreviewData` dead code** — deleted in [843375b3](https://github.com/lezzago/dashboards-observability/commit/843375b3). Took `toEpochMillis` with it (sole consumer was `extractOSPreviewData`) plus both Contents JSDoc bullets.

**6. `notificationRouting` JSDoc** — replaced with your suggested wording in [d422cd2d](https://github.com/lezzago/dashboards-observability/commit/d422cd2d). The Prom branch at the line you noted didn't actually have a misleading comment to update — `notificationRouting: []` there is a bare assignment.

**7. `alertHistory` ordering** — went with option 1 in [d422cd2d](https://github.com/lezzago/dashboards-observability/commit/d422cd2d): threaded `sortString`/`sortOrder` through `getAlerts` and now request `start_time desc` from `getOSRuleDetail`. Verified live that the API honors it (alerts come back in descending `start_time` order). Pinned with a regression test on the rule-detail call shape.

**8. monitorId schema mismatch** — replaced the synthetic `mon/with/slash` test with a route-shaped UUID-like id round-trip in [d422cd2d](https://github.com/lezzago/dashboards-observability/commit/d422cd2d). `alertingIdSchema` enforces `/^[A-Za-z0-9_-]+$/` so slashes never reach the backend; the existing test was exercising unreachable behavior. Schema-level coverage for UUID acceptance already lives in `schema_helpers.test.ts`.

### Nits

**9. `history-${monitor.id}` ⇒ `details-${monitor.id}`** — renamed in [16d6b7fb](https://github.com/lezzago/dashboards-observability/commit/16d6b7fb).

**10 / 11** — acknowledged as "no change needed" / "out of scope" per the review.

### Validation
- All 42 alerting test suites passing (385 tests)
- `yarn typecheck` clean
- `yarn lint:es` clean across the touched directories

joshuali925 (approved):

joshuali925:
### Code review

Found 2 issues:

1. Lazy alert-detail fetch dropped the in-flight cancellation guard. The previous `useEffect` used a `cancelled` flag and cleanup to discard a stale response when `selectedAlert` changed mid-fetch. The new `fetchDetailIfNeeded` writes `setDetailData(data)` and then `setDetailFetched(true)` unconditionally in `.finally()`. If the parent swaps from alert A to alert B while A's request is in flight, the identity-reset effect clears state, then A's response writes A's payload into B's `detailData` and pins `detailFetched=true` — so B's "Raw Alert Data" accordion shows A's payload and never refetches. Capture `alert.id` at call time and gate the state writes on it (or restore the `cancelled` cleanup pattern used in `use_monitor_detail.ts`).

https://github.com/opensearch-project/dashboards-observability/blob/a8383862917cdfe08280ee507d8704c95583bf5c/public/components/alerting/alert_detail_flyout.tsx#L83-L100

2. PPL monitor flyouts now always render an empty Condition Preview. `getOSRuleDetail` removed the `runMonitor(_, dryRun=true)` fallback (and `extractOSPreviewData`), but `fetchOSPreviewTimeSeries` still returns `[]` for any monitor without a `search` input — including `ppl_input` monitors. With PPL CRUD recently shipped (#2698), every PPL rule flyout open will hit the empty-state code path. Either add PPL handling in `fetchOSPreviewTimeSeries` or hide the Condition Preview accordion for PPL monitors.

https://github.com/opensearch-project/dashboards-observability/blob/a8383862917cdfe08280ee507d8704c95583bf5c/server/services/alerting/alert_detail.ts#L131-L141

https://github.com/opensearch-project/dashboards-observability/blob/a8383862917cdfe08280ee507d8704c95583bf5c/server/services/alerting/alert_preview.ts#L57-L62

lezzago:
Thanks @joshuali925 — both validated as real issues; pushed fixes in [d265ccde](https://github.com/lezzago/dashboards-observability/commit/d265ccde).

### 1. Stale in-flight response polluting B's state
Confirmed the race: when the parent swaps `selectedAlert` from A→B mid-fetch, the reset `useEffect` clears state, then A's late `.then()`/`.finally()` writes A's payload into B's `detailData` and pins `detailFetched=true` for B without ever fetching B. Re-expanding B's accordion would short-circuit the cache and render A's `raw` body indefinitely.

**Fix:** mirror `(datasourceId, id)` into a `useRef`, updated synchronously by the reset effect *before* clearing state. The fetch captures `(requestDsId, requestAlertId)` at call time and both `.then()` and `.finally()` bail if the ref no longer matches. The reset effect's synchronous-before-state-clear ordering is what makes this work — by the time A's promise resolves on the next microtask, the ref has already moved to B's identity, so A sees a mismatch and skips the writes.

**Regression test:** `discards an in-flight A response when alert identity swaps to B mid-fetch` — uses a manually-resolvable promise to deterministically reproduce the order-of-events (expand on A → rerender to B → resolve A's promise → toggle B's accordion → assert B fetches its own detail).

### 2. PPL monitors render permanently-empty Condition Preview
Confirmed: `fetchOSPreviewTimeSeries` routes by input shape (`uri`, `doc_level_input`, `search`) and has no PPL branch — PPL monitors carry `ppl_input` and fall through to the `return []` at line 59. With PPL CRUD shipped in #2698, every PPL rule flyout open would show "No recent evaluation data available" indefinitely.

**Fix:** wrap the Condition Preview accordion in a `monitor.monitorType !== 'ppl'` gate in the rule flyout. Server-side `fetchOSPreviewTimeSeries` already returns `[]` cheaply (no upstream call), so the UI gate is sufficient — no wasted request, no misleading copy. Wiring up an actual PPL preview pipeline is a follow-up (would need a date_histogram-equivalent over PPL `stats by span()` or a step-aware aggregation).

**Tests added:** `renders the Condition Preview accordion for non-PPL monitors` + `hides the Condition Preview accordion for PPL monitors`.

### Validation
- typecheck (`tsc -b tsconfig.refs.json`) clean
- eslint clean on changed files
- 16/16 flyout jest tests passing (12 alert_detail_flyout + 4 monitor_detail_flyout)

sumukhswamy (approved):

PR #2684, opened 2026-05-19
feat(slo): PR 2 — workspace id + acting user + wizard completeness

lezzago:
## Summary

PR 2 of an 8-PR SLO/SLI train. Resolves the two `FIXME(pr-2)` deferrals
from [PR 1](https://github.com/opensearch-project/dashboards-observability/pull/2676)
(merged) and lands wizard completeness. Surfaces continue to ship
**dark** — opt in via `observability.slo.enabled: true` in
`opensearch_dashboards.yml`. No schema changes, no migration risk; PR 2
is plumbing + UI.

> **For train-level context — read this first:**
> [SLO/SLI tracking issue](https://github.com/lezzago/dashboards-observability/issues/1)
>
> Covers the AMP-compatibility invariant, the schema design choices,
> the 8-PR roadmap, and cross-cutting constraints every PR in the train
> respects. The PR body below is scoped to **what changes in this diff**.

## What ships in PR 2

- **Workspace-id plumbing** end-to-end. `server/routes/slo/index.ts`
  resolves the caller's workspace via `getWorkspaceState(req).requestWorkspaceId`,
  falling through to `'default'` when OSD workspaces are disabled. The
  two `workspaceId: 'default'` hard-codes from PR 1 are gone. Malformed
  workspace ids now reject **400** at the route layer (the
  `sloRulerNamespaceFor` regex validation PR 1 codified is exercised by
  live request traffic for the first time). The `tryResolveWorkspaceId`
  envelope is inlined at three read routes; PR 3 consolidates the
  boilerplate into a shared helper alongside the new aggregate endpoint.

- **Acting-user extraction**. `resolveActingUser(req)` reads
  `req.auth.credentials.username` (security plugin), falls back to
  `x-proxy-user` / `x-forwarded-user` request headers (proxy-auth
  deployments), and finally `'unknown'`. Every route that writes
  `createdBy` / `updatedBy` flows through this extractor. PR 1's
  `SLO_ACTING_USER_PLACEHOLDER` constant + `'osd-user'` sentinel are
  gone. Test fixture covers each fallback branch including non-string
  credential payloads and whitespace-only headers.

- **Wizard completeness**. Template-driven create flow at
  `/slos/create/:templateId` (APM / OTel / custom). Multi-objective
  editing. Custom PromQL SLI mode with live `validateCustomPromQL`.
  Right-dock generated-rules preview with namespace + group-name
  preview. Burn-rate presets + custom tier editor. Labels +
  annotations key-value grid. Sticky anchor nav with order-drift guard
  test. Top-level `WizardValidationSummary` banner on submit-attempt.
  Target-% → allowed-downtime preview with high-target callout.
  Confirm + progress strip for batch creation. Unsaved-changes prompt.
Screenshot 2026-05-20 at 9 21 45 AM
Screenshot 2026-05-20 at 9 21 56 AM
Screenshot 2026-05-20 at 9 22 11 AM

## Security notes

- **Workspace-id shape**: now exercised live. The
  `/^[a-zA-Z0-9][a-zA-Z0-9_-]{0,62}$/` regex in `sloRulerNamespaceFor`
  is the only gate; route layer rejects malformed input as 400 before
  the value reaches `setRuleGroup` / `deleteRuleGroup`. Path-traversal
  via a malicious `/w/<x>` URL component is the class of attack this
  closes.

- **400 envelope hygiene**: the route does NOT echo the user-controlled
  workspace id back in the 400 body. `requestWorkspaceId` originates
  from the URL, so reflecting an arbitrary value invites
  log-injection / response-splitting style mischief. The wizard keys
  off the field name (`errors.workspaceId`), not the message text.

- **Acting-user attribution**: `resolveActingUser` deliberately reads
  `credentials.username` without gating on `auth.isAuthenticated`.
  Proxy-auth deployments populate `credentials` without a security
  session, and we'd rather attribute the audit row to the proxied
  user than fall through to `'unknown'`. Whitespace-only and
  non-string usernames are rejected.

## Explicit deferrals (carryover from PR 2 self-review)

Documented in `docs/slo/pr-3-listing-and-status.md` and
`docs/slo/pr-8-polish.md`:

| Item | Where | Deferred to |
|---|---|---|
| Consolidate `tryResolveWorkspaceId` envelope at three read routes | `server/routes/slo/index.ts` | **PR 3** |
| Wizard datasource picker (`EuiComboBox`) | wizard | **PR 3** |
| Merge `err.body.attributes.errors` into `WizardValidationSummary` (field-attributed inline errors) | wizard | **PR 8** |
| Wizard `dirty` heuristic — track all reducer mutations, not just identity-shaped subset | wizard | **PR 8** |

## AMP invariant audit

Every `ruler.upsertRuleGroup` / `deleteRuleGroup` caller resolves the
namespace via `sloRulerNamespaceFor(...)` or
`provisioning.rulerNamespace` (itself written by that helper). PR 2
introduces live workspace ids; the route layer rejects malformed ids
**400** before the value reaches the namespace composer. No new
namespace string introduced (`rg "'slo-"` returns only `slo-generated`,
`slo-rule-ref`, `slo-definition`).

## Test plan

- [x] **Jest**: 288 suites pass / 0 fail / 1 skip; 2197 tests pass / 3
  skips. 20 SLO-specific suites covering 280 tests.
- [x] **`resolve_request.test.ts`** (new) covers `resolveWorkspaceId`
  (returns the resolved id; rejects malformed ids as `SloValidationError`
  carrying the `workspaceId` field key) and `resolveActingUser` (each
  fallback branch — security credentials → `x-proxy-user` →
  `x-forwarded-user` → `'unknown'`; whitespace-only and non-string
  payloads rejected).
- [x] **`slo_workspace_isolation.integration.test.ts`** unchanged — the
  contract was authored at the service layer specifically so the
  route-layer change in PR 2 is drop-in.
- [x] **Wizard Jest**: `wizard_state` (reducer-shape sanity for the
  array mutators most prone to silent bugs), `wizard_sections`
  (drift-guard pinning the canonical visual order; longest-prefix-match
  resolution for keys that overlap shorter prefixes), `wizard_builders`
  (% → decimal conversion, multi-objective fan-out, custom-PromQL
  routing into `customExpr.{events|raw}`), `objectives_section`
  (`formatAllowedDowntime` two-largest-units formatter incl. the
  IEEE-754 ms→sec rounding case), `burn_rate_presets`
  (`isPresetApplied` detects user-edited tiers).
- [x] **AMP audit**: clean (see above).
- [x] **Cypress smoke (manual, against local observability-stack)**:
  - `slo_workspace_isolation.spec.js` — non-default workspace create
    writes to `slo-generated-<wsId>` and does NOT appear in the
    default workspace listing.
  - `slo_wizard_multi_objective.spec.js` — multi-objective happy path
    (preview renders, both downtime previews visible, persisted SLO
    carries both objectives) + validation-summary recovery flow
    (submit with missing name → summary appears → fill name → row
    clears on next submit).
  - Skips when no `workspaceId` env is supplied; CI without the stack
    skips cleanly.
- [x] **Reviewer**: with `observability.slo.enabled: true`, create an
  SLO under a non-default workspace (`/w/<wsId>`) and confirm
  `provisioning.rulerNamespace` is `slo-generated-<wsId>` on the
  saved-object document, and that the listing under the default
  workspace does NOT see it.

### DIFF
--- a/.cypress/integration/slo_test/slo_crud_smoke.spec.js
+++ b/.cypress/integration/slo_test/slo_crud_smoke.spec.js
@@ -6,7 +6,11 @@
 /// <reference types="cypress" />
 
 /*
- * PR-1 SLO CRUD smoke — create → detail → listing row → delete → row gone.
+ * SLO CRUD smoke — create → detail → listing row → delete → row gone.
+ *
+ * The wizard surface targeted here is the PR 2 multi-section wizard. Test-subj
+ * naming follows the `slos*` convention introduced in PR 2 (vs. PR 1's
+ * `sloWizard*` names which the wizard rewrite replaced).
  *
  * Prereqs:
  *   - Dev stack running with `observability.slo.enabled: true` in
@@ -19,8 +23,6 @@
  *   yarn cypress:run \
  *     --spec .cypress/integration/slo_test/slo_crud_smoke.spec.js \
  *     --env workspaceId=<WS_ID>,sloDatasourceId=<DS_ID>,baseUrl=http://localhost:5602
- *
- * The baseUrl env overrides `cypress.config.js` for dev servers on :5602.
  */
 
 const WORKSPACE_ID = Cypress.env('workspaceId');
@@ -37,7 +39,7 @@ const sloApp = `${prefix}/app/observability-apm-slo`;
 // doesn't have the stack up without failing the whole suite.
 const SLO_ENABLED = Cypress.env('sloEnabled') !== false;
 
-describe('SLO — CRUD smoke (PR 1)', () => {
+describe('SLO — CRUD smoke', () => {
   before(function () {
     if (!SLO_ENABLED) this.skip();
   });
@@ -48,21 +50,23 @@ describe('SLO — CRUD smoke (PR 1)', () => {
 
     // Land on the listing page.
     cy.visit(`${sloApp}#/slos`);
-    // Wait for the listing to render — either an empty prompt or the table.
     cy.get(
       '[data-test-subj="sloCreateBtn"], [data-test-subj="sloCreateBtnEmpty"]'
     )
       .first()
       .click();
 
-    // Fill the wizard.
-    cy.get('[data-test-subj="sloWizardName"]').type(name);
-    cy.get('[data-test-subj="sloWizardService"]').type(service);
-    cy.get('[data-test-subj="sloWizardTeam"]').type('platform');
-    cy.get('[data-test-subj="sloWizardDatasource"]').type(DATASOURCE_ID);
-    cy.get('[data-test-subj="sloWizardMetric"]').clear().type('http_requests_total');
-    cy.get('[data-test-subj="sloWizardTarget"]').clear().type('99.9');
-    cy.get('[data-test-subj="sloWizardSubmit"]').click();
+    // Pick a template (HTTP availability — single objective, no probe path).
+    cy.get('[data-test-subj="slosTemplate-http-availability"]', { timeout: 10000 }).click();
+
+    // Fill the wizard fields. Identity, owner, datasource — single objective
+    // suffices for the smoke path.
+    cy.get('[data-test-subj="slosWizardDatasourceId"]').clear().type(DATASOURCE_ID);
+    cy.get('[data-test-subj="slosWizardName"]').type(name);
+    cy.get('[data-test-subj="slosWizardService"]').type(service);
+    cy.get('[data-test-subj="slosWizardOwnerTeam"]').type('platform');
+
+    cy.get('[data-test-subj="slosWizardSubmit"]').click();
 
     // After create the wizard redirects to the detail page for the new SLO.
     cy.location('hash', { timeout: 15000 }).should('match', /#\/slos\/[0-9a-f-]{36}/);
--- a/.cypress/integration/slo_test/slo_wizard_multi_objective.spec.js
+++ b/.cypress/integration/slo_test/slo_wizard_multi_objective.spec.js
@@ -1,0 +1,91 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/// <reference types="cypress" />
+
+/*
+ * PR 2 wizard — multi-objective happy path + validation summary.
+ *
+ * Adds a second objective row, fills both, submits, and verifies the SLO is
+ * persisted with two objectives. Then exercises the validation-summary flow:
+ * clear a required field, submit, the summary banner appears with one error.
+ *
+ * Prereqs:
+ *   - `observability.slo.enabled: true` in opensearch_dashboards.yml.
+ *   - A registered Prometheus-backed DirectQuery datasource.
+ *
+ * Run with (from plugin dir):
+ *   yarn cypress:run \
+ *     --spec .cypress/integration/slo_test/slo_wizard_multi_objective.spec.js \
+ *     --env workspaceId=<WS_ID>,sloDatasourceId=<DS_ID>,baseUrl=http://localhost:5602
+ */
+
+const WORKSPACE_ID = Cypress.env('workspaceId');
+const DATASOURCE_ID = Cypress.env('sloDatasourceId') || 'prom';
+
+const uniqueSuffix = () => Math.random().toString(36).slice(2, 8);
+const prefix = WORKSPACE_ID ? `/w/${WORKSPACE_ID}` : '';
+const sloApp = `${prefix}/app/observability-apm-slo`;
+
+const SLO_ENABLED = Cypress.env('sloEnabled') !== false;
+
+describe('SLO wizard — multi-objective', () => {
+  before(function () {
+    if (!SLO_ENABLED) this.skip();
+  });
+
+  it('creates an SLO with two objectives and renders the rule preview', () => {
+    const name = `Multi-Obj SLO ${uniqueSuffix()}`;
+    const service = `multi-obj-svc-${uniqueSuffix()}`;
+
+    cy.visit(`${sloApp}#/slos/create`);
+    cy.get('[data-test-subj="slosTemplate-http-availability"]', { timeout: 10000 }).click();
+
+    // Identity / owner / datasource.
+    cy.get('[data-test-subj="slosWizardDatasourceId"]').clear().type(DATASOURCE_ID);
+    cy.get('[data-test-subj="slosWizardName"]').type(name);
+    cy.get('[data-test-subj="slosWizardService"]').type(service);
+    cy.get('[data-test-subj="slosWizardOwnerTeam"]').type('platform');
+
+    // Default objective is "availability-99-9". Add a second.
+    cy.get('[data-test-subj="slosWizardObjectiveAdd"]').click();
+    cy.get('[data-test-subj="slosWizardObjectiveName-1"]').clear().type('availability-99-0');
+    cy.get('[data-test-subj="slosWizardObjectiveTarget-1"]').clear().type('99');
+
+    // Live downtime preview surfaces under the target.
+    cy.get('[data-test-subj="slosWizardObjectiveTargetDowntime-0"]').should('be.visible');
+    cy.get('[data-test-subj="slosWizardObjectiveTargetDowntime-1"]').should('be.visible');
+
+    // Rule-preview panel renders once form is valid (debounce ~500ms).
+    cy.get('[data-test-subj="slosWizardPreviewSuccess"]', { timeout: 15000 }).should('be.visible');
+
+    cy.get('[data-test-subj="slosWizardSubmit"]').click();
+    cy.location('hash', { timeout: 15000 }).should('match', /#\/slos\/[0-9a-f-]{36}/);
+
+    // Detail page surfaces both objectives in the spec JSON view.
+    cy.contains(name).should('be.visible');
+    cy.contains('availability-99-0').should('be.visible');
+  });
+
+  it('surfaces a top-level validation summary when a required field is missing', () => {
+    cy.visit(`${sloApp}#/slos/create`);
+    cy.get('[data-test-subj="slosTemplate-http-availability"]', { timeout: 10000 }).click();
+
+    // Skip the Name field — but fill the rest so this is a single-error flow.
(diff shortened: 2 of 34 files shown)


lezzago:
## Cycle-1 review-pass response

Two commits address the bot suggestions on this PR:

- `9e6f1f5e` — `fix(slo): address PR review cycle 1 — acting-user hardening, double-submit guard, validation auto-scroll`
- `12b5b832` — `fix(slo): PR review cycle 1 polish — preview AbortController + acting-user fallthrough log`

### Addressed

| # | Suggestion | Where landed |
|---|---|---|
| 1 | `Code-Diff-Analyzer` (medium): `resolveActingUser` accepts arbitrary header values | Cap to 255 chars + reject C0 control chars + DEL via new `sanitizeActingUser` helper. Realistic Unicode / LDAP DN / email usernames preserved. `9e6f1f5e` |
| 2 | Reviewer Guide: `resolveActingUser` doesn't validate length / control chars | Same helper as above. `9e6f1f5e` |
| 3 | Code Suggestion (security): username injection vector | Same helper as above. `9e6f1f5e` |
| 4 | Code Suggestion (importance 7): double-submit race in `onSubmit` | Added `submitting` guard + dep. `9e6f1f5e` |
| 5 | Code Suggestion (importance 7): auto-scroll to first validation error | Imported existing `scrollToErrorKey` helper, called on first error key. `9e6f1f5e` |
| 6 | Reviewer Guide: preview fetch leaves in-flight requests after dep change | Switched from `cancelled` flag to `AbortController`; signal threaded through `SloApiClient.preview`. `12b5b832` |
| 7 | Code Suggestion: log when `resolveActingUser` falls through to `'unknown'` | Added optional `logger?.debug` (chose `debug` over `warn` to avoid log spam in legitimately anonymous-write deployments). `12b5b832` |

### Reviewed and skipped (with rationale)

| # | Suggestion | Why skipped |
|---|---|---|
| 8 | Chain `sloRulerNamespaceFor` original error into `resolveWorkspaceId` 400 message | Deliberate — see in-code comment at `server/routes/slo/index.ts`. The candidate value originates from a user-controllable URL path (`/w/<x>`); reflecting it into the response body invites log injection / response splitting. |
| 9 | `setCreatedSuccessfully(true)` before `history.push` race | Not a bug. By the time we set it, the SLO is created server-side. Suppressing the unsaved-changes Prompt for any later navigation is the *desired* behavior. |
| 10 | `formatAllowedDowntime` should return `'—'` for invalid input | Unreachable. `computePreview` short-circuits to `null` before the formatter sees out-of-range values. The empty-string branch is defensive but inert. |
| 11 | Validate `rulerNamespace` parameter in `coerceRuleGroup` | Internal helper, `string` typed at the boundary, all in-PR callers pass a validated namespace. The bot's own analysis acknowledges this isn't critical. |
| 12 | Stable `key` in `wizard_key_value_grid` | The bot's suggested `${entry.key}-${entry.value}-${i}` collides on empty entries during edit. The right fix is reducer-side stable IDs (a refactor); deferring to a follow-up. |
| 13 | Throw on invalid `target` in `wizard_builders` | The 0 fallback is intentional for state transitions; UI-side validation already rejects out-of-range targets before submit. |
| 14 | Reject duplicate keys in `entriesToRecord` | The bot itself notes the UI layer is the right place. Raising at the builder would block submission unexpectedly. |
| 15 | `JSON.parse` try/catch in preview | Input is the same `JSON.stringify` output produced 3 lines above; can't fail. |
| 16 | Sanitize `data-test-subj` value | HTML attribute values + CSS attribute selectors handle dots/brackets fine. Cosmetic. |

Validation: typecheck ✓, lint:es (changed files) ✓, jest 24/24 (resolve_request + generated_rules_preview) ✓.

lezzago:
## Cycle-1 review-pass — round 2

After the prior round-1 push the bots regenerated, surfacing 3 genuinely new findings + several false-positives. Addressed in `38020ad3` — `fix(slo): PR review cycle 1 — round 2 (C1 ban, sanitize-reject log, sync submit guard)`.

### Addressed

| # | Suggestion | Where landed |
|---|---|---|
| N7 | Extend control-char regex to include C1 block (0x80-0x9F) — NEL, CSI, etc. | `sanitizeActingUser` regex now `/[\x00-\x1f\x7f-\x9f]/`. Two new test cases (NEL + CSI). |
| N2 | Log per sanitize-rejection so forensics can correlate which header was the culprit | Refactored `sanitizeActingUser` to return `{ ok, value | reason }`; new `tryResolveActingUser` helper logs `debug` per non-empty rejection. Empty/whitespace skipped to bound noise. Test fixture covers both. |
| N6 | Synchronous double-submit guard (state guard isn't atomic across React's async update window) | Replaced state-based `if (submitting) return` with `useRef<boolean>` that flips synchronously inside `onSubmit`. Outer `try/finally` releases the ref on validation-error early returns too. |

### Reviewed and skipped (with rationale)

| # | Suggestion | Why skipped |
|---|---|---|
| Reviewer Guide | "`markSubmitAttempted` fires before the `submitting` guard" | **Bot misread the diff**. The guard is line 1 of the callback; `dispatch` is inside its try block. With this round's commit, the ref-flip + the dispatch both happen after the guard. |
| Reviewer Guide | "Preview debounce timeout not cancelled when `serialized→null`" | **Bot misread React useEffect cleanup ordering**. The previous effect's cleanup (`clearTimeout(t)`) runs before the new effect with `serialized=null` runs. The timeout is cancelled. |
| Code Suggestion | `useCallback` returning a cleanup function for unmount | **Bot's proposed fix is invalid** — `useCallback` can't return cleanup. Component-level unmount-during-submit is a separate concern with low realistic exposure (the redirect happens in the success branch). |
| Diff-Analyzer (low) | Network-layer trusted-proxy guard | Operator-side concern (reverse proxy + security plugin config), not a code fix. Bot already downgraded medium→low after the round-1 sanitization. |
| Reviewer Guide | `resolveWorkspaceId` relies on `sloRulerNamespaceFor` side-effect | Refactor for a hypothetical future change where the helper stops throwing. Current code is correct; the existing in-code comment documents the design choice. |
| Code Suggestion | Validate `rulerNamespace` in `coerceRuleGroup` | Internal helper, controlled callers (already replied in round 1). |
| Code Suggestion | `findSectionForKey` prefix-Map optimization | Bot's own analysis says complexity is unchanged; current 10×3 is fine for a wizard form. |
| Code Suggestion | `CSS.escape()` for `data-test-subj` selector | `fieldIds` map is statically defined string literals; no XSS surface. |
| Code Suggestion | Chain `sloRulerNamespaceFor` original error | Deliberate per in-code comment (already replied in round 1). |
| Code Suggestion | Throw on duplicate keys in `entriesToRecord` | Belongs in UI validation layer (already replied in round 1). |

### Validation

- typecheck ✓, lint:es (changed files) ✓
- jest: 5 suites / 58 tests passed (resolve_request + wizard suites)

### CI note

The `linkchecker` workflow run on this commit failed on `https://opentelemetry.io/docs/kubernetes/collector/components/` — a pre-existing URL in `server/adaptors/integrations/__data__/repository/k8s/info/README.md` that prior commits on this branch (e.g. `12b5b832`) reached without issue. Transient network flake; no admin rights on the upstream repo to re-run, so flagging here for a maintainer to either re-run the job or confirm it's not merge-blocking.

kylehounslow (commented):
## Feedback summary

Solid PR. The two PR-1 `FIXME(pr-2)` deferrals are resolved cleanly — acting-user resolver has a thoughtful fallback chain with sensible sanitization (reject vs. truncate, C0/C1 + U+2028/U+2029 ban, 255-char cap), and `resolveWorkspaceId` correctly gates through `sloRulerNamespaceFor` before reaching the ruler. Resolver test coverage is dense.

The wizard split — `wizard_state` (reducer), `wizard_builders` (state → input), `wizard_sections` (shared metadata + drift-guard) — is clean and will pay off in PR 3+.

Main concern is reviewability: 5683/385 is hard to review as one unit, and `slo_wizard_page.tsx` bundles five inline panels into a ~700 LOC orchestrator. Plus a handful of correctness and copy nits. None blocking.

## Action items

**Must-fix**
1. **Plain-English error message in `resolveWorkspaceId`** (`server/routes/slo/index.ts`) — strip the regex from the user-visible 400; describe the rules in words.
2. **Verify the GCP 99.9% claim** in the high-target callout (`objectives_section.tsx`) — citation looks wrong. Soften the wording or drop the vendor reference.
3. **Tighten `isValidationStyleError`** (`generated_rules_preview.tsx`) — match on the structured envelope shape (`body.attributes.errors`), not a regex over `err.message`.

**Should-fix**
4. **Extract the inline panels** out of `slo_wizard_page.tsx` (IdentityPanel, OwnerPanel, SliPanel, WindowPanel, LabelsAnnotationsPanel) — orchestrator is ~700 LOC of JSX, too big to navigate.
5. **Replace double-deferred scroll spy** in `wizard_nav.tsx` — drop the 100ms `setTimeout`, keep the single `requestAnimationFrame`.
6. **Replace concurrent smooth-scrolls** in `wizard_nav.tsx::onPick` — use `scroll-margin-top: 16px` on the anchor element instead of `scrollIntoView` + `scrollBy`.
7. **Use OUI/EUI monospace token** in `custom_promql_editor.tsx` — `{ fontFamily: 'monospace' }` won't match the YAML preview's font stack.
8. **Rephrase the "page-quick tier" alarm caption** (`advanced_section.tsx`) — term isn't defined anywhere else in the UI.
9. **Add a route-handler-level test** that exercises a malformed `requestWorkspaceId` → 400 with `errors.workspaceId` — the resolver unit test doesn't cover the route envelope the wizard depends on.

**Nice-to-have**
10. **Cite the page-heavy 28.8× tier** in `burn_rate_presets.ts` — not from SRE Workbook Ch. 5, source isn't obvious.
11. **Drop the `getBurnRatePreset` throw** or convert to `assertNever` — only callers pass a closed union.
12. **Remove "Typing doesn't produce a sea of red…" comment** and similar internal-voice asides from `wizard_state.ts`.
13. **Stop referencing future PRs by number** in code comments (PR 3 / PR 5 / PR 8) — link to the tracking issue instead.
14. **For the next PR in the train:** split route-layer plumbing from UI rewrites — they're logically separable and review better in isolation.

---

*Assisted by Claude (Opus 4.7).*

PR #2689, opened 2026-05-21
feat(slo): SLO/SLI follow-ups on top of PR #2676

lezzago:
### Description

Second-tranche delivery for the SLO/SLI feature, layered on top of the foundation that landed in #2676. Closes the gaps that PR 1 deliberately deferred and addresses two rounds of code-review findings (must-fix + ultrareview).

**Scope by area:**

- **Ruler dual-write & DirectQuery integration** — Real `DirectQueryRulerClient` writes recording + alert groups via the SQL plugin's resource proxy; structured 404 sniffing for the SQL-plugin-wrapped Cortex envelope; provenance annotations on every generated alert group.
- **Phase-3 dedup** — Fingerprint-keyed shared recording groups (`slo:rec:<fp>`) with refcount registry (`slo-rule-ref` SO) and grace-period reconciler sweep for zero-ref cleanup. Per-SLO alert groups stay; recording rules dedupe across SLOs that share a fingerprint.
- **Live status aggregator** — `DirectQueryStatusAggregator` runs PromQL instant queries + alerts fetches against Cortex, with worker-pool concurrency, per-(datasource × longWindow) batched queries instead of one-per-SLO, and a rule-health checker that overlays `rules_missing` / `ruler_unreachable` priorities on top of sample-derived state. Probes fan out via `Promise.all`.
- **Wizard completeness** — Multi-objective + custom PromQL flows, anchor nav, validation summary, generated-rules preview, advanced section (exclusion windows, burn-rate presets, key/value grid editor), schema validation tightened to `unknowns: 'forbid'` so smuggled keys can't reach the saved object.
- **Suggest page** — Service-first APM/OTel template selection, live PromQL probes for ratio/samples/p99 across nine canonical kinds (APM availability/latency, HTTP, RPC, DB, messaging, GenAI). Engine helpers extracted into `suggest_live_queries.ts`.
- **Listing** — Filter facets (state, SLI type, canonical kind, service, team, tier, mode, enabled) with hash-query URL sync, dominant-value defaults line, worst-budget-first sort, truncation callout when server `total > pageSize`.
- **Detail page** — Health summary, MWMBR tier cards, zone-banded budget bar, error-budget-remaining + per-tier burn-rate charts, rules panel showing `present/missing` per group, copy-able recording-rule names.
- **APM integration** — Services Home health pill + budget sparkline, Service Details SLOs tab, server-side `/services/_aggregate` endpoint, alert-manager linkage. Gated on `observability.slo.enabled`.
[...]

PR #2678, opened 2026-05-14
[Feature] Alert Manager support for PPL Monitor CRUD operations

TackAdam:
### Description
End-to-end **Create / Read / Edit / Delete / Clone / Acknowledge** for PPL monitors in the Alert Manager. This PR wires the existing alerting scaffold (PR #2653) to the PPL alerting backend and adds a modern create/edit flyout with index discovery, Monaco-powered PPL editor, and multi-trigger support.

https://github.com/user-attachments/assets/84411e7c-697c-40ad-b5f2-56b4f5f0f852

## Prerequisites

Requires the custom OpenSearch alerting plugin with:
- PPL Alerting CRUD (create/update/delete monitors with `monitor_type: "ppl_monitor"`)
- PPL Monitor Execution (scheduler evaluates PPL queries + fires triggers)
- Tenant-id propagation

Feature gated behind `observability.alertManager.enabled` in `opensearch_dashboards.yml`.

---

### Create Monitor Flyout — PPL Only

- **Removed DSL/Cluster Metrics monitor types** from the create UI. Those are read-only in the Rules table (existing DSL monitors display but can't be authored here). Only PPL is supported for creation.
- **"Define index" panel** with async `EuiComboBox` picker:
  - Resolves indices + aliases from `_cat/indices` / `_cat/aliases` via new server routes.
  - Supports wildcards via free-text entry (`onCreateOption`).
  - Picked indices auto-populate the PPL `source = ...` clause.
- **Timestamp field selector**: second combobox populated from `_mapping`, filtered to `date` / `date_nanos` types.
- **Monaco PPL editor** replacing `EuiTextArea`:
  - Full PPL syntax highlighting via `@osd/monaco` PPL language.
  - Autocomplete for field names (from picked indices' mapping), PPL commands, aggregate functions, and scalar functions.
  - `source = <indices>` clause auto-rewrites when the picker changes, preserving user-authored `| where ...` body.
- **Multi-trigger PPL editor** (`PplTriggersSection`): each trigger supports `number_of_results` or `custom` condition type, with per-trigger severity and destination picker.
- **Default trigger type changed to `custom`** with `where error_count > 0`. PPL alerting queries almost always use `stats` (aggregation), which returns exactly 1 row per group — `number_of_results` (counts rows, not column values) would rarely fire as expected.
- **Default query template**: `source = <picked-index> | stats count() as error_count` — consistent with the trigger condition and the example shown in the flyout.
[...]

PR #2698, opened 2026-05-27
[Feature] Alert Manager support for PPL Monitor CRUD operations (review-fix follow-up to #2678)

lezzago:
> **Relationship to #2678**
> This PR is a follow-up to [#2678](https://github.com/opensearch-project/dashboards-observability/pull/2678) (TackAdam's PPL Monitor CRUD work). It carries TackAdam's 17 original commits unchanged at the front of the history (attribution preserved), then adds 12 review-fix commits addressing ps48's review on #2678 + 2 fix commits found during e2e validation. The branch is also merged with `main` (#2689 + #2695), so it is mergeable directly without #2678 landing first.
>
> Closing #2678 in favor of this PR is fine if the maintainers prefer; alternately #2678 can stay open as the original review surface and this is the "ready to land" pass.

## Summary

End-to-end **Create / Read / Edit / Delete / Clone / Acknowledge** for PPL monitors in the Alert Manager. Wires the existing alerting scaffold (PR #2653) to the PPL alerting backend and adds a modern create/edit flyout with index discovery, Monaco-powered PPL editor, and multi-trigger support.

The diff is large (~8.1k +/4.5k −, 83 files) because it includes all of TackAdam's original PPL CRUD work; a per-commit walk is the recommended way to review.

### Prerequisites

Requires the OpenSearch alerting plugin with PPL alerting CRUD (create/update/delete monitors with `monitor_type: \"ppl_monitor\"`), PPL monitor execution, and tenant-id propagation. Feature gated behind `observability.alertManager.enabled` in `opensearch_dashboards.yml` (defaults to OFF).

## Diff layout

| Commits | Source |
|---|---|
| `d2b16afd…9512a936` (17) | **TackAdam** — original PPL monitor CRUD work (#2678). Authorship preserved in git history; per-commit detail readable via `git log d2b16afd^..9512a936`. |
| `67a8bd86` | fix(alerting): correctness, security, and review-driven hardening (F1, F2, F3, F4, F5, F8, F9, F10, F13, F25 + ps48 H6/H7/H8 + M1/M3/M4/M6) |
| `2065fed9` | refactor(alerting): pure-helper extraction + Cypress smoke |
| `a0b1a718` | fix(alerting): cap fan-out concurrency on unified rule/alert fetches (M5) |
| `5ec2c9f6` | refactor(alerting): decompose alarms_page + monitor_detail_flyout (ps48 H1/H2 — partial) |
| `606223b4` | chore(alerting): camelCase data-test-subj sweep + i18n untranslated strings |
| `f9bee235` | chore(alerting): finish ps48 LOW-priority cleanup (L1/L2/L3/L6/L8 + M4 residual) |
| `bc518bde` | feat(alerting): surface destinations truncation indicator (L4) |
| `16b0f3b6` | test(alerting): backfill 31 unit-test cases for hooks + mutation client (H3) |
| `9af0bd96` | refactor(alerting): hoist OS monitor index prefix lists to named constants (L5) |
| `a9da4cbc` | Merge branch 'main' (resolves alarms_page.tsx + alert_utils.ts conflicts) |
| `24d9763c` | fix(alerting): runHandler wrapper + Prometheus dsId resolution (B1, D1 — found during e2e validation) |
| `9c95485d` | fix(alerting): wrap data-test-subj anchors that EUI/CodeEditor drop (A1, A1-bis — found during e2e validation) |
| `54f9bcf0` | test(alerting): regression tests for runHandler, dsId resolution, and dropped data-test-subj anchors |

## What's new (high-level)

- **Create Monitor flyout — PPL only.** DSL / cluster-metrics monitors are read-only in the Rules table.
- **\"Define index\" panel** with async `EuiComboBox` over `_cat/indices` + `_cat/aliases` plus free-text wildcards.
- **Timestamp field selector** populated from `_mapping`, filtered to `date` / `date_nanos`.
- **Monaco PPL editor** with field/command/aggregate/scalar autocomplete; `source = <indices>` clause auto-rewrites on picker change.
- **Multi-trigger PPL editor** (`number_of_results` or `custom` condition).
- **Read-only destinations picker** scoped to the chosen datasource.
- **Server routes** for `_cat/indices`, `_cat/aliases`, `_mapping`, and read-only destinations.
- **Mutation handlers** wired directly to the alerting plugin's PPL monitor endpoints.

## ps48 review ledger (against #2678)

| Severity | Total | Fixed | Partial | Deferred-by-design |
|---|---|---|---|---|
| HIGH | 8 | 6 | **2** (H1, H2 — under pre-PR baseline but above 500-line threshold) | 0 |
| MEDIUM | 6 | 6 | 0 | 0 |
| LOW | 8 | 7 | 0 | **1** (L7 — alerting plugin team is bumping their cap to 100 in a parallel PR; values match once that ships) |
| **Total** | **22** | **19** | **2** | **1** |

### `67a8bd8` — correctness + security hardening

**Server / API surface**
- **F1**: Reject unresolvable dsIds in MDS deployments instead of silently falling through to the local cluster (cross-tenant integrity bug).
- **F2**: Stop reflecting raw upstream `Error.message` on the 500 path; strip class-wrapper prefixes from the 400 path.
- **F3 / F4**: Pass `logger` through `registerAlertingMutationRoutes` so failed mutations are visible server-side; add an `info`-level audit line on each successful create/update/delete/acknowledge.
- **F5**: `updateMonitor` now merges user input onto the **raw** upstream document instead of `mapMonitor()`'s typed projection. Plugin-internal fields (`data_sources`, `last_run_context`, `owner`, `enabled_time`, future fields) round-trip through every PUT.
- **F8**: Replace `body: schema.any()` with a structured monitor body schema using `unknowns: 'allow'` on inner objects.
- **F7** (downgraded MEDIUM): Per-id length cap on acknowledge body.
- **M1 (ps48)**: Cap `_cat/indices` and `_cat/aliases` discovery results at 200.

**Frontend correctness**
- **F9**: Replace the `length < 20` trigger-id heuristic with a marker-prefix check.
- **F10**: 30s `AbortController` timeout on `acknowledgeAlert`.
- **F13**: Clone path strips nested `actions[].id`.
- **F25**: New `useAlertingPluginAvailability` probe + callout when no OS datasource responds to the alerting API.
- **H7 / H8 / M3 / M6 (ps48)**: see review thread.

### `2065fed` — extraction + Cypress

- `alarms_page_helpers.ts` — pure helpers, 29 unit tests.
- `.cypress/integration/alerting_test/ppl_monitor_crud_smoke.spec.js` — API-driven CRUD smoke (Monaco editor doesn't automate cleanly in Cypress).

### `a0b1a71` — concurrency cap (M5)

- `runWithConcurrencyLimit` worker-pool helper, default 5; applied to all 4 fan-outs in `MultiBackendAlertService`. 7 unit tests.

### `5ec2c9f` — decomposition (H1, H2 — partial)

- `alarms_page.tsx`: 1216 → 755 LOC; useState 19 → 7. New hooks: `useTimeRange`, `useDatasourceSelection`, `useRulesData`. `AlarmsPageCallouts` sub-component.
- `monitor_detail_flyout.tsx`: 1032 → 854 LOC. New hook `useMonitorDetail`; `ConditionPreviewGraph` + `humanizeCondition` extracted.
- **Honest status**: both files still above ps48's 500-line threshold. Going lower would require either `useReducer` migration (motivation already eliminated by functional `setState`) or per-accordion sub-component extraction (mechanical, no behaviour change). Tracked as a follow-up.

### `606223b` — conventions sweep

- All ~25 alerting subtree `data-test-subj` selectors → camelCase.
- `CLIENT_TRIGGER_ID_PREFIX = 'ppl-trigger-'` intentionally NOT renamed — wire-format marker for F9.
- `ppl_triggers.tsx` / `destination_picker.tsx` / `edit_monitor.tsx`: ~30 raw English strings wrapped in `i18n.translate` / `<FormattedMessage>`. Operator glyphs intentionally untranslated.

### `f9bee23` — LOW-priority cleanup

- **L8**: `PplQueryEditor` surfaces `useIndexMappings` error.
- **L1**: `updateProm` / `updateOs` / `handleSave` wrapped in `useCallback`.
- **L3**: `destinationsHref` is now an inline expression (was empty-deps `useMemo` reading `coreRefs.http`).
- **L6**: `marginBottom: 12` inline styles → `.altCalloutSpaced` SCSS.
- **L2**: 6 useless `useCallback` wrappers in `ppl_triggers.tsx` removed.
- **M4 residual**: `PplNumResultsOperator` aliased to `OSPPLNumResultsOperator`; `MonitorBackendType` aliased to `DatasourceType`.

### `bc518bd` — destinations truncation (L4)

- `OSDestinationsResult { destinations, totalDestinations, truncated }` end-to-end. UI renders \"Showing N of M\" hint when truncated.

### `16b0f3b` — H3 backfill (31 cases, 4 files)

- `use_indices.test.ts` (8), `use_index_mappings.test.ts` (7 — incl. cache-hit short-circuit regression for H8), `use_destinations.test.ts` (7 — incl. truncated propagation for L4), `monitor_mutations_client.test.ts` (9 — incl. F10 timeout translation regression).

### `9af0bd9` — L5 cleanup

- `LOG_INDEX_PREFIXES` / `APM_INDEX_PREFIXES` hoisted to module scope.
- New `inferMonitorTypeFromIndices` helper. 11 new test cases pin the schema list.

### `a9da4cbc` — merge with main

- Brings in #2689 (SLO follow-ups) and #2695 (PPL stats sub-query skip).
- Two semantic conflicts resolved:
  - `alarms_page.tsx`: keep decomposed structure; layer origin/main's `parseAlarmsHashRoute` deep-link feature on top; drop stale `CreateLogsMonitor`/`CreateMetricsMonitor` re-imports.
  - `alert_utils.ts`: kept helper-based `inferMonitorTypeFromIndices` (covered by 11 tests); folded origin/main's trace_analytics-settings comment into the constant doc-block.

### `24d9763c` — runHandler + Prometheus dsId resolution (e2e validation)

Two issues found during a playwright-driven e2e pass:

- **D1**: `getAlertingClient` throws `AlertManagerError` (a plain `{kind, message}` object — not an Error subclass) outside any route try/catch, so OSD's framework crashed with `text.replace is not a function` when stringifying. Fixed by a `runHandler(res, produce)` wrapper applied to all 11 affected routes.
- **B1**: Prometheus datasource leaked `[object Object]` and 10s timeouts on every page load. Root cause: `67a8bd8`'s `createNotFoundError(...)` in `getAlertingClient` also fired for legitimate Prometheus `data-connection` ids (which AREN'T `data-source` saved objects). The unified-fanout aggregator stringified the AlertManagerError rejection as `[object Object]`. Fixed by a defensive `extractErrorMessage` helper in the aggregator + Prometheus-aware `getAlertingClient` fallback so the aggregator no longer hits the `not_found` path for legitimate Prometheus datasources.
- `resolveOpenSearchDatasource`, `resolvePrometheusDatasource`, `getAlertingClient` are now module-level exports so they can be unit-tested directly.

### `9c95485d` — wrap dropped data-test-subj anchors (e2e validation)

- **A1**: `alertManagerPplQueryEditor` was passed as a prop to OSD's `<CodeEditor>`, which doesn't forward arbitrary `data-*` attributes. Moved to a wrapping `<div>`.
- **A1-bis**: Same family — `alertManagerDatePicker` was passed to `<EuiSuperDatePicker>` and dropped. Wrapped in a `<div>`.

### `54f9bcf0` — regression tests

- `errors.test.ts`: 8 cases for `extractErrorMessage`, including AlertManagerError → `[object Object]` regression.
- `get_alerting_client.test.ts` (new): 15 cases covering all resolver branches — OS data-source happy path, Prometheus data-connection fallback, workspace-scoped reject, no-MDS fallback, unresolvable dsId.
- `runHandler error funneling` describe block in `index.test.ts`: 11 parameterized cases asserting each affected route returns a typed 404 (not a crashing 500) when dsId is unknown.
- `ppl_query_editor.test.tsx` (new): asserts `alertManagerPplQueryEditor` resolves with a CodeEditor stub that drops props.
- `alerts_dashboard.test.tsx`: regression case for `alertManagerDatePicker`.

## Validation

| Gate | Result |
|---|---|
| `yarn typecheck` | ✓ no errors |
| `yarn test` (full plugin) | ✓ no failures |
| Alerting subtree | ✓ **560 / 560 across 53 suites** |

## Test plan

- [ ] `yarn typecheck`
- [ ] `yarn lint`
- [ ] `yarn test --testPathPattern='alerting|alarms_page'`
- [ ] Manual: create → edit → delete a PPL monitor against a cluster with `opensearch-alerting` installed; confirm trigger id round-trips on edit (F9 regression).
- [ ] Manual: confirm Prometheus rules show in the Rules tab when a Prometheus datasource is selected (B1 regression).
- [ ] Manual: enable feature flag against a cluster **without** `opensearch-alerting`; confirm 'Alerting plugin not detected' callout (F25).
- [ ] Manual: trigger an `index_not_found` error and confirm the 500 response body is generic — no cluster URL leaked (F2).
- [ ] Manual: with >200 destinations, confirm the truncation hint renders (L4).
- [ ] Manual: navigate to `#/rules?q=foo` and confirm the page lands on Rules tab with `foo` pre-filled (deep-link from origin/main).
- [ ] (Optional) Run the new Cypress spec: `yarn cypress:run --spec .cypress/integration/alerting_test/ppl_monitor_crud_smoke.spec.js --env alertingEnabled=true`

## Items intentionally NOT addressed

- **L7** — `PPL_MONITOR_NAME_MAX = 100` vs the alerting plugin's current cap of 30. The alerting plugin team is bumping their max to 100 in a parallel PR; once that ships the values match and the cap is correct as-is. No client-side change needed.
- **H1 / H2 fully under 500 LOC** — `alarms_page.tsx` 755 / `monitor_detail_flyout.tsx` 854 (both well under their pre-PR baselines but still above ps48's 500-line threshold). Tracked as a follow-up.
- **Tab/URL hash sync** — three coupled UX rough edges (hash drift on tab clicks, deep-link only on mount, search-box stickiness across `#/rules` navigations) surfaced during e2e validation but were introduced by earlier PRs (#2653 original Alert Manager, #2689 SLO follow-ups), not by the PPL CRUD work. Filed as [lezzago/dashboards-observability#5](https://github.com/lezzago/dashboards-observability/issues/5) for follow-up.


ps48 (commented):
## Code Review: [Feature] Alert Manager PPL Monitor CRUD — review-fix follow-up (#2698)

**Context:** This PR carries forward TackAdam's PPL Monitor CRUD work (#2678) and adds 12 fix commits addressing all findings from the previous review.

---

## Overall Assessment

**All 11 previously-raised issues are verified as fixed.** Schema validation replaces `schema.any()`, plugin availability check added, memoization fixed, concurrency-limited fan-outs, error surfacing improved, test coverage expanded significantly. The code quality has improved substantially. A few remaining medium-priority gaps exist but nothing blocking.

---

## Previous Review Verification

| Issue | Status |
|-------|--------|
| H1: alarms_page 1171 LOC / 19 useState | ✅ Fixed — 795 LOC / 8 useState |
| H2: monitor_detail_flyout 1004 LOC | ✅ Fixed — 854 LOC |
| H5: `schema.any()` on mutations | ✅ Fixed — structured schema with `unknowns: 'allow'` |
| H6: No plugin availability check | ✅ Fixed — `useAlertingPluginAvailability` hook + callout |
| H7: handleTimeChange defeats memoization | ✅ Fixed — functional setState, empty deps |
| H8: useIndexMappings indices in deps | ✅ Fixed — only `cacheKey` in deps |
| M1: _cat/indices unbounded | ✅ Fixed — capped at 200 |
| M2: Mutation errors not logged | ✅ Fixed — logger passed throughout |
| M3: IndexPicker swallows errors | ✅ Fixed — error surfaced in form row |
| M5: Promise.allSettled no concurrency | ✅ Fixed — `runWithConcurrencyLimit(5)` |
| L5: Hardcoded index prefixes | ✅ Fixed — named constants |

---

## MEDIUM Priority (Remaining)

### M1. `PPL_QUERY_MAX = 2000` declared but never enforced in body schema

Constant is exported but not applied to any field. Server offers no protection against oversized queries.

### M2. `useDatasourceSelection` may clear persisted selection on initialization

When `selectedDsIds` starts as `[]` before the init effect fires, the persistence effect writes empty to localStorage.

### M3. `useMonitorDetail` logs error but doesn't expose error state

`console.error` only — no `error` field in the hook result for the flyout to show degraded state.

### M4. 6 components still lack dedicated unit tests

`ppl_triggers.tsx`, `index_picker.tsx`, `destination_picker.tsx`, `time_field_selector.tsx`, `opensearch_form_section.tsx` — mocked away in higher-level tests but not tested in isolation. Plus `create_logs_monitor/` (7 files) and several new hooks lack dedicated tests.

### M5. `useAlertingPluginAvailability` doesn't abort in-flight requests on cleanup

Sets `cancelled = true` but doesn't abort controllers. Requests complete wastefully on unmount.

---

## LOW Priority

| # | Issue |
|---|-------|
| L1 | `alarms_page.tsx` (795) and `monitor_detail_flyout.tsx` (854) still above 500-line threshold — tracked |
| L2 | `useTimeRange` `didChange` pattern relies on React 18 synchronous batching — fragile |
| L3 | `res: any` in `runHandler` — could use OSD response types |
| L4 | `ppl_query_editor.test.tsx` has only 1 regression test |
| L5 | `PPL_MONITOR_NAME_MAX = 100` vs backend 30 — intentionally pre-emptive |

---

## Positive Highlights

- **All 11 review findings verified fixed** — thorough follow-through
- **Concurrency limiter** — clean worker-pool pattern avoiding head-of-line blocking
- **`runHandler` pattern** — centralizes error handling across 11 routes, prevents topology leakage
- **Structured body schema** — pragmatic "validate structure, allow unknown nested" approach
- **Plugin availability probe** — per-datasource with 10s timeout and clear UI callout
- **Hook extraction** — `useTimeRange`, `useDatasourceSelection`, `useRulesData` with stable identities and cleanup
- **36 new test files, 560 tests / 53 suites** — all passing
- **Cypress test uses real backend** — actual server route calls, proper env-gating
- **Error sanitization** — never reflects raw upstream messages for 500s
- **No hardcoded localhost/infra regex** in alerting code
- **i18n sweep** — ~30 strings internationalized
- **Destinations truncation** — "Showing N of M" hint when >200
- **`data-test-subj` camelCase convention** applied consistently

server/routes/alerting/mutations/body_schema.ts
```diff
@@ -0,0 +1,71 @@
...
+ *   - validate top-level monitor fields (`name`, `type`, `monitor_type`,
+ *     `enabled`, `schedule`, `inputs`, `triggers`) so payload size and shape
+ *     abuse is rejected at the API boundary,
+ *   - allow `unknowns: 'allow'` on the nested trigger / input / schedule
+ *     objects so wrapper fields and future plugin extensions round-trip.
+ */
+import { schema } from '@osd/config-schema';
+
+const ALERTING_NAME_MAX = 256;
+const PPL_QUERY_MAX = 2000;
```
ps48, line 27: **MEDIUM:** `PPL_QUERY_MAX = 2000` is declared and exported in `SCHEMA_LIMITS` but never applied to any field in `monitorMutationBodySchema`. The `inputs` field uses `schema.object({}, { unknowns: 'allow' })` with no string-length validation on the nested PPL query.

This means the documented 2000-char cap is aspirational only — no server-side protection against oversized queries.

**Fix:** Either apply it as a `maxLength` on the relevant nested query string field, or remove the constant to avoid misleading callers.

public/components/alerting/hooks/use_datasource_selection.ts
```diff
@@ -0,0 +1,108 @@
...
+
+  // Persist the selection (by name) whenever it changes and we know the
+  // datasource list. Names survive server restarts; ids don't.
+  useEffect(() => {
+    if (datasources.length === 0) return;
+    const names = selectedDsIds
+      .map((id) => datasources.find((d) => d.id === id)?.name)
+      .filter((n): n is string => typeof n === 'string');
+    persistSelection(names);
+  }, [selectedDsIds, datasources]);
```
ps48, line 108: **MEDIUM:** This persistence effect may fire before the initialization effect (both triggered by `datasources` changing). When `selectedDsIds` is initially `[]`, this writes an empty array to localStorage, potentially clearing a valid prior selection.

**Fix:** Add a guard:
```typescript
if (selectedDsIds.length === 0) return;
```
Before the `persistSelection(names)` call.

public/components/alerting/hooks/use_monitor_detail.ts
```diff
@@ -0,0 +1,59 @@
...
+  useEffect(() => {
+    let cancelled = false;
+    setIsLoading(true);
+    osService
+      .getRuleDetail(dsId, ruleId)
+      .then((data: UnifiedRule) => {
+        if (!cancelled && data) setDetail(data);
+      })
+      .catch((err: unknown) => {
+        console.error('Failed to load monitor details:', err);
```
ps48, line 52: **MEDIUM:** `console.error` logs the failure but `UseMonitorDetailResult` has no `error` field. The consuming flyout cannot show a retry button or degraded-state indicator.

**Fix:** Add `error: Error | null` to the hook's return type and expose it, so the flyout can render an inline callout when the detail fetch fails.

public/components/alerting/hooks/use_alerting_plugin_availability.ts
```diff
@@ -0,0 +1,98 @@
...
+          }
+        })
+      );
+      if (cancelled) return;
+      setUnavailableDsIds(failed);
+      setIsLoading(false);
+    })();
+
+    return () => {
+      cancelled = true;
```
ps48, line 88: **MEDIUM:** The cleanup sets `cancelled = true` (prevents state writes) but does **not** abort in-flight HTTP probe requests. If the component unmounts while probes are active, they complete wastefully in the background.

**Fix:** Track all active `AbortController` instances and call `.abort()` on each in the cleanup function:
```typescript
return () => {
  cancelled = true;
  controllers.forEach(c => c.abort());
};
```

public/components/alerting/hooks/use_time_range.ts
```diff
@@ -0,0 +1,123 @@
...
+
+  useEffect(() => {
+    if (!rangeParseFailed) return;
+    setStartTime(DEFAULT_START_TIME);
+    setEndTime(DEFAULT_END_TIME);
+    persistTimeRange(DEFAULT_START_TIME, DEFAULT_END_TIME);
+  }, [rangeParseFailed]);
+
+  const onTimeChange = useCallback(({ start, end }: { start: string; end: string }) => {
+    let didChange = false;
```
ps48: **LOW:** The `didChange` variable set inside functional `setState` updaters relies on React 18's synchronous batching behavior — the updaters run synchronously within the same callback, so `didChange` is set before the `if` check at line 92.

This works today but is fragile (relies on an implementation detail). A safer alternative: compare against a ref value, or simply always call `persistTimeRange` (sessionStorage writes are cheap and idempotent).

ps48 (commented):
Inline comments on specific files — see summary comment for full review.

public/components/alerting/hooks/use_alerting_plugin_availability.ts
```diff
@@ -0,0 +1,98 @@
...
+    };
+    // osIdsKey memoizes the array contents — including osIds directly would
+    // re-fire on every parent render due to array-reference churn.
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [osIdsKey]);
+
+  return useMemo(
+    () => ({
+      isLoading,
+      unavailable: osIds.length > 0 && unavailableDsIds.length === osIds.length,
```
joshuali925: Possible stale-set race in the `unavailable` derivation. `osIds` here is from the latest render, but `unavailableDsIds` is whatever the *previous* probe wrote. If the user disables an OS datasource between probes (or the datasource list shrinks for any other reason), `osIdsKey` changes, the effect re-fires and sets `setIsLoading(true)` — but until the new probe resolves, this memo can momentarily satisfy `unavailableDsIds.length === osIds.length` against the *stale* failed-set, falsely flipping the callout to "unavailable".

The `isLoading` flag isn't read here, so consumers gating purely on `unavailable` see the flicker.

Two straightforward fixes:

1. Tighten the predicate so it requires the failed-set to be a subset of the current `osIds`:
   ```ts
   unavailable:
     osIds.length > 0 &&
     unavailableDsIds.length === osIds.length &&
     unavailableDsIds.every((id) => osIds.includes(id)),
   ```
2. Or reset `unavailableDsIds` synchronously on key change, e.g. an extra `useEffect(() => setUnavailableDsIds([]), [osIdsKey])` ahead of the probe.

Happy with either — option 1 is local and avoids an extra render.

server/services/alerting/opensearch_backend.ts
```diff
@@ -137,9 +153,18 @@ export class HttpOpenSearchBackend implements OpenSearchBackend {
...
+    // Merge into the RAW upstream document, not the typed-and-narrowed
+    // `mapMonitor` projection. `mapMonitor` only emits the fields modelled
+    // in `OSMonitor`; merging onto its output would silently strip plugin
+    // fields the OSD layer doesn't know about (e.g. `data_sources`,
+    // `last_run_context`, `owner`, `enabled_time`, future schema additions),
+    // and the round-trip PUT would lose them.
+    const rawUpstream = (getResp.body.monitor as unknown) as Record<string, unknown>;
+    const merged = {
+      ...rawUpstream,
+      ...input,
```
joshuali925, line 199: F5 (this comment block) protects plugin-internal fields — `data_sources`, `last_run_context`, `owner`, `enabled_time` — from being silently *stripped* by the typed `mapMonitor` projection. But the spread order `{ ...rawUpstream, ...input, ... }` lets `input` *overwrite* any of those keys if the caller supplies them, and `monitorMutationBodySchema` uses `unknowns: 'allow'` end-to-end — so a client request containing `data_sources` (or any other plugin-internal key) lands directly in the upstream PUT.

That's the same class of regression F5 was intended to prevent (tenant scoping / multi-cluster routing carried in `data_sources`), just through a different door. The opensearch_backend tests exercise the strip path but not this clobber path, so it's currently invisible.

Suggested fix — explicitly preserve the plugin-owned keys after the spread:

```ts
const PLUGIN_OWNED_KEYS = ['data_sources', 'last_run_context', 'owner', 'enabled_time'] as const;
const preserved = Object.fromEntries(
  PLUGIN_OWNED_KEYS
    .filter((k) => k in rawUpstream)
    .map((k) => [k, rawUpstream[k]])
);
const merged = {
  ...rawUpstream,
  ...input,
  ...preserved,
  last_update_time: Date.now(),
};
```

Or, if you'd rather reject at the schema layer, drop those keys from the request body via a forbid-list before the merge.

Worth a regression test that PUTs `{ data_sources: [...something hostile...] }` and asserts the upstream call carries the *original* `data_sources` value.

joshuali925 (commented):
Two findings from a focused review pass that I think are worth resolving before merge. Inlined below. The rest of the review-fix ledger looks honest and the targeted regression tests (D1 parametric runHandler, B1 `[object Object]`, F1 unresolvable dsId, F5 field preservation) match their claims.

joshuali925 (approved):

server/services/alerting/opensearch_backend.ts
```diff
@@ -12,23 +12,40 @@ import {
...
 
+/**
+ * Cap on `_cat/indices` and `_cat/aliases` discovery responses. The combobox
+ * uses an async typeahead so the user filters by typing — anything beyond
+ * the first page-worth of suggestions is unreachable in the UI anyway, and
+ * a cluster with tens of thousands of matches would otherwise freeze the
+ * browser tab while parsing the JSON. Mirrors the `MAX_RESULTS = 200`
+ * pattern in `metadata_handlers.ts`.
+ */
+const MAX_DISCOVERY_RESULTS = 200;
```
joshuali925, line 57: is this server side filtered?

and why is 

> anything beyond the first page-worth of suggestions is unreachable in the UI anyway

server/services/alerting/opensearch_backend.ts
```diff
@@ -329,13 +354,26 @@ export class HttpOpenSearchBackend implements OpenSearchBackend {
   // Destinations
   // =========================================================================
 
-  async getDestinations(client: AlertingOSClient): Promise<OSDestination[]> {
+  async getDestinations(client: AlertingOSClient): Promise<OSDestinationsResult> {
+    // The alerting `destinations` endpoint isn't paginated — we cap at
+    // size=200 so a cluster with thousands of destinations doesn't freeze
```
joshuali925: there are some API not paginated in OSD but size is 10000. this 200 seems to be existing behavior, it never caused issues for users?

server/routes/alerting/mutations/body_schema.ts
lezzago: Fixed in 44d52f98. Applied `PPL_QUERY_MAX` as `maxLength` on `inputs[].ppl_input.query`. Live-validated against the dev cluster: a 2032-char query now returns `400 Bad Request` with `value has length [2032] but it must have a maximum length of [2000]`. Two new schema tests pin the boundary (cap+1 rejected, cap-length accepted).

public/components/alerting/hooks/use_datasource_selection.ts
lezzago: Fixed in 44d52f98 with the exact `if (selectedDsIds.length === 0) return;` guard you suggested. New `use_datasource_selection.test.ts` includes a `setItem` spy assertion that catches the empty write at the intermediate render — verified the test fails when the guard is removed.

public/components/alerting/hooks/use_monitor_detail.ts
lezzago: Fixed in 44d52f98 — went with the degraded-state indicator (no retry button yet; reopening the flyout retriggers the fetch, which covered the immediate need). Added `error: Error | null` to `UseMonitorDetailResult`, populated in the `.catch`, and cleared on each new fetch. `monitor_detail_flyout.tsx` consumes it and renders an `EuiCallOut` with `data-test-subj="alertManagerMonitorDetailLoadError"` — title "Some monitor details could not be loaded", body "Showing summary information only. Try reopening the flyout to retry." Happy to add an explicit retry button as a follow-up if you'd prefer.

New `use_monitor_detail.test.ts` covers happy path, error exposure, non-Error rejection wrapping, and clear-on-param-change.

public/components/alerting/hooks/use_alerting_plugin_availability.ts
lezzago: Fixed in 44d52f98. Track `activeControllers: AbortController[]` and call `.abort()` on each in cleanup. Verified by test: I capture the `signal` passed to `http.get`, never resolve the probe, then unmount and assert `signal.aborted === true`. Removing the abort line makes the test fail.

public/components/alerting/hooks/use_time_range.ts
lezzago: Fixed in 44d52f98 — went with your second suggestion (always call `persistTimeRange`). Dropped the `didChange` flag entirely; sessionStorage writes are cheap and idempotent so the simplification has no measurable cost and removes the React 18 sync-batching dependency. The H7 invariant is preserved because `setStartTime(prev)` with the same value still bails out of the re-render path.

public/components/alerting/hooks/use_alerting_plugin_availability.ts
lezzago: Fixed in 44d52f98 — went with option 1 (subset guard). Predicate is now `osIds.length > 0 && unavailableDsIds.length === osIds.length && unavailableDsIds.every((id) => osIds.includes(id))`.

New `use_alerting_plugin_availability.test.ts` has a regression case for the same-length-swap scenario: 1 OS datasource fails, then we swap it for a different one (lengths still 1 === 1, but the new id isn't in the stale failed-set). Without the subset guard the test fails with `unavailable: true`; with the guard it stays `false` until the new probe resolves.

server/services/alerting/opensearch_backend.ts
lezzago: Fixed in 44d52f98 — used your `Object.fromEntries`-style preserve, hoisted `PLUGIN_OWNED_KEYS` to a module-level constant.

Live-validated: PUT `{ data_sources: { tenant: 'evil' }, owner: 'attacker', enabled_time: 0 }` against an existing PPL monitor on the dev cluster. The OSD response masks plugin-internal fields, but reading the raw upstream document via console proxy shows `data_sources` still carries the original alerting-plugin object, `owner` is still `"alerting"`, and `enabled_time` is unchanged. Added a regression test that asserts the same path with a hostile body — fails without the preserve, passes with it.

server/services/alerting/opensearch_backend.ts
lezzago: The original comment was wrong — fixed in 170760d5. The combobox does push the user's input to the cluster (typeahead sends `_cat/indices/{prefix}*` so the cluster does the filtering, not the browser). The 200 is a safety net for when that prefix still matches a very large slice — typing `l` against a cluster with tens of thousands of `logs-*` indices would still return everything if we didn't cap, and parsing a 50k-row JSON response on the main thread would freeze the picker for seconds.

On "why 200 not 10000": OSD's 10000 cap shows up on programmatic non-paginated endpoints (e.g. `getRuleGroups`); for an interactive picker, more suggestions don't help — they just slow down rendering, and the user can always type more characters to narrow further. Updated the comment to spell this out.

server/services/alerting/opensearch_backend.ts
lezzago: Resolved by a703aa7b — that endpoint is gone. The `_plugins/_alerting/destinations` API is deprecated; we now hit `_plugins/_notifications/channels` from the OpenSearch Notifications plugin. The new code uses upstream pagination as-is and surfaces `total_hits` as `totalDestinations` for the truncation indicator (no client-side cap — the `/channels` endpoint doesn't accept a size parameter).

The channel ids returned from `/channels` are interchangeable with `actions[].destination_id` in the alerting payload, so the UI contract is unchanged. Live-validated by creating a slack channel via the Notifications API on the dev cluster and confirming the OSD destinations route returned `{id: <config_id>, name: 'ops-slack', type: 'slack'}`. Reference: https://docs.opensearch.org/latest/observing-your-data/notifications/api/

ps48 (approved):
Looks good thanks for the fix, I don't see alerting_test integ tests running and also the SLO ones are failing. May be something to follow up

joshuali925:
https://github.com/opensearch-project/dashboards-observability/actions/runs/26532215393/job/78151594782?pr=2698 CI failed?

@joshuali925-osdbot why?

lezzago:
> https://github.com/opensearch-project/dashboards-observability/actions/runs/26532215393/job/78151594782?pr=2698 CI failed?
> 
> @joshuali925-osdbot why?

I believe it is due to a flaky timeout issue that I will follow up on

PR #2700, opened 2026-05-28
Fix clone monitor: initialize missing osService in AlarmsPage

riysaxen-amzn:
## Description

Fixes the "Failed to clone a monitor: osService is not defined" error when clicking the Clone button on the rule detail flyout in the Unified Alerts View.

### Root Cause

`handleCloneRule` in `alarms_page.tsx` calls `osService.getRuleDetail()` but `osService` was never instantiated, causing a `ReferenceError` at runtime.

### Fix

Added the `AlertingOpenSearchService` import and a `useMemo`-based initialization — matching the pattern already used in `alert_detail_flyout.tsx`.

## Testing

1. Start the observability stack + OSD from source
2. Navigate to Alerts → Rules tab
3. Open any rule flyout (e.g. "OTel Demo - Cart Service Errors")
4. Click Clone
5. Verify the toast shows "Monitor cloned" instead of the error

Image 5-28-26 at 4 52 PM

## Issues Resolved

Fixes clone monitor bug (medium severity) from Bug Bash - Unified Alerts View.

### DIFF
--- a/public/components/alerting/__tests__/alarms_page.test.tsx
+++ b/public/components/alerting/__tests__/alarms_page.test.tsx
@@ -42,8 +42,14 @@ jest.mock('../alerts_dashboard', () => ({
     return <div data-test-subj="alertsDashboard" />;
   },
 }));
+
+// Capture MonitorsTable props so we can invoke onClone in tests.
+const mockMonitorsTable = jest.fn();
 jest.mock('../monitors_table', () => ({
-  MonitorsTable: () => <div data-test-subj="monitorsTable" />,
+  MonitorsTable: (props: unknown) => {
+    mockMonitorsTable(props);
+    return <div data-test-subj="monitorsTable" />;
+  },
 }));
 jest.mock('../notification_routing_panel', () => ({
   NotificationRoutingPanel: () => <div data-test-subj="routingPanel" />,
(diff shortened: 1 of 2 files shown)


lezzago (approved):
LGTM — minimal, correctly-scoped fix that matches the `useMemo(() => new AlertingOpenSearchService(), [])` pattern already used in `alert_detail_flyout.tsx`, `use_rules_data.ts`, `use_rule_detail.ts`, and `use_alerts.ts`. Resolves the `ReferenceError` at the actual root cause.

**Non-blocking ask:** consider adding a regression test for `handleCloneRule` in `__tests__/alarms_page.test.tsx`. The existing tests pass because `MonitorsTable` is mocked to a stub, which is exactly why this runtime ReferenceError slipped through. A small test that surfaces `onClone` from the `MonitorsTable` mock and asserts `getRuleDetail` + `mutations.createMonitor` are called would catch the next "missing service in component" bug in CI rather than in the user's browser.

riysaxen-amzn:
> LGTM — minimal, correctly-scoped fix that matches the `useMemo(() => new AlertingOpenSearchService(), [])` pattern already used in `alert_detail_flyout.tsx`, `use_rules_data.ts`, `use_rule_detail.ts`, and `use_alerts.ts`. Resolves the `ReferenceError` at the actual root cause.
> 
> **Non-blocking ask:** consider adding a regression test for `handleCloneRule` in `__tests__/alarms_page.test.tsx`. The existing tests pass because `MonitorsTable` is mocked to a stub, which is exactly why this runtime ReferenceError slipped through. A small test that surfaces `onClone` from the `MonitorsTable` mock and asserts `getRuleDetail` + `mutations.createMonitor` are called would catch the next "missing service in component" bug in CI rather than in the user's browser.

addressed

ps48 (approved):

joshuali925 (approved):

PR #2710, opened 2026-06-02
Fix label search to match values and prevent double-save on create monitor

riysaxen-amzn:
## Description

Two fixes:

### 1. Label filter search matches both key names and values

Previously, searching in the Labels filter panel only matched against label **key names**. Searching for a label value like "riysaxen" would show "No labels match" even though the `monitor_name` label had values containing "riysaxen-testing".

Now the search matches against:
- Label key names (e.g., searching "monitor" matches `monitor_name`)
- Label values within each key (e.g., searching "riysaxen" matches because `monitor_name` contains "riysaxen-testing")
- Raw option values in addition to display labels in the facet filter component

Applied to both Alerts tab (`alerts_dashboard.tsx`) and Rules tab (`monitors_filters_panel.tsx`).

### 2. Prevent double-submission on Create Monitor

Clicking Save twice rapidly would create duplicate monitors. Added `isSaving` state that:
- Guards `handleSave` from executing twice
- Disables both Save buttons after first click
- Shows a loading spinner on the buttons

### Label search:
1. Open Alerts tab with alerts that have labels (e.g., `monitor_name: riysaxen-testing`)
2. In the Labels filter search box, type a label value (e.g., "riysaxen")
3. Verify the label group appears (instead of "No labels match")

### Double-save prevention:
1. Open Create Monitor flyout
2. Fill in name + query
3. Rapidly click Save twice
4. Verify only one monitor is created (check Rules tab)

## Issues Resolved

Fixes label search and double-save bugs from Bug Bash - Unified Alerts View.

### DIFF
--- a/public/components/alerting/alerts_dashboard.tsx
+++ b/public/components/alerting/alerts_dashboard.tsx
@@ -657,7 +657,13 @@ export const AlertsDashboard: React.FC<AlertsDashboardProps> = ({
                   if (visibleLabelKeys.length === 0) return null;
                   const q = labelSearch.trim().toLowerCase();
                   const matchedLabelKeys = q
-                    ? visibleLabelKeys.filter((k) => k.toLowerCase().includes(q))
+                    ? visibleLabelKeys.filter(
+                        (k) =>
+                          k.toLowerCase().includes(q) ||
+                          collectAlertLabelValues(alerts, k).some((v) =>
+                            v.toLowerCase().includes(q)
+                          )
+                      )
                     : visibleLabelKeys;
                   const cappedLabelKeys = showAllLabels
                     ? matchedLabelKeys
--- a/public/components/alerting/__tests__/create_monitor.test.tsx
+++ b/public/components/alerting/__tests__/create_monitor.test.tsx
@@ -198,3 +198,33 @@ describe('CreateMonitor — PPL form', () => {
     expect(arg.pplTriggers[0].type).toBe('custom');
   });
 });
+
+describe('CreateMonitor — double-save prevention', () => {
+  it('disables save buttons after first click to prevent duplicate submissions', () => {
+    const onSave = jest.fn();
+    render(
+      <CreateMonitor
+        onSave={onSave}
+        onCancel={jest.fn()}
+        datasources={[osDs]}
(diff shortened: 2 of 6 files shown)


public/components/alerting/alerts_dashboard.tsx
```diff
@@ -657,7 +657,13 @@ export const AlertsDashboard: React.FC<AlertsDashboardProps> = ({
                   if (visibleLabelKeys.length === 0) return null;
                   const q = labelSearch.trim().toLowerCase();
                   const matchedLabelKeys = q
-                    ? visibleLabelKeys.filter((k) => k.toLowerCase().includes(q))
+                    ? visibleLabelKeys.filter(
+                        (k) =>
+                          k.toLowerCase().includes(q) ||
+                          collectAlertLabelValues(alerts, k).some((v) =>
+                            v.toLowerCase().includes(q)
+                          )
```
joshuali925, line 665: is client side searching going to be a concern?

joshuali925 (approved):

public/components/alerting/alerts_dashboard.tsx
riysaxen-amzn: Good question. The search iterates alerts × visibleLabelKeys on each keystroke. In practice this is bounded because:

The unified alerts endpoint caps results (default 1000)
visibleLabelKeys is typically <20 after filtering INTERNAL_LABEL_KEYS
The iteration is a simple string .includes() check — no allocations per iteration
For a typical scenario (200–500 alerts, 5–15 label keys), this runs in <1ms. If we see performance issues at scale, we could memoize collectAlertLabelValues per key or debounce the search input, but I don't think it's warranted for the current data volumes.

mengweieric (approved):

lezzago (approved):

public/components/alerting/alerts_dashboard.tsx
lezzago: Prometheus alerts each can have 3-10 alerts so it would have scaling issues with prometheus alerts at scale.

lezzago (approved):

joshuali925 (approved):

PR #2711, opened 2026-06-02
fix(alerting): resolve local-cluster placeholder; probe alerting plugin directly

lezzago:
## Summary
- Aligns the client-emitted no-MDS placeholder id with the server-side `local-cluster` convention and short-circuits it in `getAlertingClient` so the per-dsId routes resolve cleanly when the `data_source` plugin is loaded but no `data-source` saved objects exist. Without this, the Alerts page rendered a false **Alerting plugin not detected** callout against a healthy cluster.
- Replaces the plugin-availability probe with a dedicated route (`GET /api/alerting/opensearch/{dsId}/probe`) that hits `POST /_plugins/_alerting/monitors/_search` with `size:0, terminate_after:1`. The previous probe targeted `/_plugins/_notifications/channels`, conflating `opensearch-alerting` presence with `opensearch-notifications`.
- Fixes `SavedObjectDatasourceService.list()` so registering any Prometheus `data-connection` no longer suppresses the synthetic local-cluster fallback. Before this, every OpenSearch monitor / alert silently disappeared from the unified Alerts and Rules views the moment a single DirectQuery connection existed.

## Why
On a stock OSD installation with the `data_source` plugin loaded but zero `data-source` saved objects, the Alerts page was unusable: the page-level callout claimed the alerting plugin was missing (it wasn't), and even after dismissing it, unified queries returned zero OpenSearch alerts because the local cluster wasn't in the registry.

Live verification on a local cluster with 6 OS monitors + 1 Prometheus connection: unified endpoint now returns 28 alerts across 2 datasources (5 OS + 23 Prom); previously it returned 23 across 1.

## Files changed
- `server/routes/alerting/index.ts` — `getAlertingClient` short-circuit + new `/probe` route
- `server/services/alerting/opensearch_backend.ts` — `probe()` method
- `server/services/alerting/saved_object_datasource_service.ts` — gate the local-cluster fallback on OS count, not total list length
- `public/components/alerting/hooks/use_datasources.ts` — placeholder id `local` → `local-cluster`
- `public/components/alerting/hooks/use_alerting_plugin_availability.ts` — call `/probe` instead of `/destinations`
- Tests: new `saved_object_datasource_service.test.ts` (7 tests including regression repro), new `probe()` and short-circuit cases, route-count updates

## Test plan
- [x] `yarn test server/services/alerting server/routes/alerting public/components/alerting` — 523/523 pass
- [x] `yarn typecheck` — clean
- [x] `yarn lint:es` (touched files) — clean
- [x] Live: `GET /api/alerting/opensearch/local-cluster/probe` → `{"ok":true}` (was 404)
- [x] Live: `GET /api/alerting/unified/alerts` returns OS + Prom alerts (was Prom-only)
- [x] Live: Alerts page no longer renders the "Alerting plugin not detected" callout against the healthy cluster

### DIFF
--- a/public/components/alerting/hooks/use_datasources.ts
+++ b/public/components/alerting/hooks/use_datasources.ts
@@ -75,7 +75,12 @@ export function mapSavedObjectsToDatasources(
       ? []
       : [
           {
-            id: 'local',
+            // Matches the synthetic id the server-side resolvers use for the
+            // no-MDS local cluster (see `getAlertingClient` and
+            // `SavedObjectDatasourceService.get`). Keeping the two in sync
+            // means the per-dsId probe and read routes resolve cleanly when
+            // no `data-source` SOs exist.
+            id: 'local-cluster',
             name: i18n.translate('observability.alerting.hooks.useDatasources.localCluster', {
               defaultMessage: 'Local Cluster',
             }),
--- a/server/services/alerting/opensearch_backend.ts
+++ b/server/services/alerting/opensearch_backend.ts
@@ -74,6 +74,31 @@ export class HttpOpenSearchBackend implements OpenSearchBackend {
   constructor(private readonly logger: Logger) {}
 
   // =========================================================================
+  // Plugin probe
+  // =========================================================================
+  //
+  // Cheap "is the alerting plugin installed?" probe used by the Alerts page
+  // to render a single user-friendly callout instead of N noisy errors when
+  // a cluster doesn't have `opensearch-alerting`. We hit the alerting plugin's
+  // own `monitors/_search` endpoint so we exercise the exact REST handler the
+  // rest of the UI depends on — a probe against `_plugins/_notifications/...`
+  // would only confirm a different plugin (`opensearch-notifications`).
+  //
+  // `terminate_after: 1` lets a non-empty alerting index short-circuit on the
+  // first segment per shard; `size: 0` drops the response payload. A missing
+  // plugin returns `no handler found for uri ...` with status 400 from the
+  // HTTP layer, which surfaces as a transport error here — we let it
+  // propagate so the route can report 502 and the UI shows the callout.
+
+  async probe(client: AlertingOSClient): Promise<void> {
+    await this.req(client, 'POST', '/_plugins/_alerting/monitors/_search', {
+      size: 0,
+      terminate_after: 1,
+      query: { match_all: {} },
+    });
+  }
+
+  // =========================================================================
   // Monitors
   // =========================================================================
 
--- a/public/components/alerting/hooks/use_alerting_plugin_availability.ts
+++ b/public/components/alerting/hooks/use_alerting_plugin_availability.ts
@@ -13,11 +13,14 @@
  * single "alerting plugin not detected" callout and avoid noisy errors
  * downstream.
  *
- * Probe target: `GET /destinations`. It's already part of the read surface
- * registered by this plugin and doesn't require write permissions, so it's
- * safe to issue at mount time. We treat any 404/502/transport failure as
- * "unavailable" — the destinations route only returns 200 when the alerting
- * plugin successfully proxied the request.
+ * Probe target: `GET /probe`, which on the server side issues a cheap
+ * `POST /_plugins/_alerting/monitors/_search` with `size:0,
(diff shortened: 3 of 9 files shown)


ps48 (commented):
## Code Review: fix(alerting): resolve local-cluster placeholder; probe alerting plugin directly (#2711)

### Overall Assessment

Clean, surgical 3-part bug fix (270 additions, 9 files) that resolves a false "Alerting plugin not detected" callout on clusters with the `data_source` plugin loaded but no `data-source` saved objects. Well-tested with 7 new regression tests. Ready to merge.

---

### No Blocking or High-Priority Issues

This PR is architecturally sound:
- The new `/probe` route hits `/_plugins/_alerting/monitors/_search` (the actual alerting plugin) — not a search query, just an existence check with `size: 0, terminate_after: 1`
- No new dependencies or manifest changes
- Feature flag gating unchanged (behind `alertManager.enabled`)
- No hardcoded patterns or unbounded data

---

### LOW Priority

**L1.** Magic string `'local-cluster'` is repeated across 4 files (`use_datasources.ts`, `getAlertingClient` in `index.ts`, `saved_object_datasource_service.ts`, tests). Should be a shared constant in `common/` (e.g., `LOCAL_CLUSTER_SENTINEL_ID`) to prevent drift if the value ever changes.

---

### Positive Highlights

- **Proper probe target** — probes `/_plugins/_alerting/monitors/_search` instead of `/_plugins/_notifications/channels` (was conflating two different plugins)
- **`terminate_after: 1, size: 0`** makes the probe extremely cheap
- **Correct fallback logic** — gates on `osCount === 0` not `all.length === 0`, so Prometheus data-connections no longer suppress the local OS cluster
- **7 new regression tests** covering exact bug scenarios (Prometheus-only still emits local-cluster, `local-cluster` short-circuits without SO lookup, probe shape + error propagation, route registration)
- **Well-documented comments** — explains the "why" at each change site with cross-references between client and server
- **Client/server alignment** — both sides use `'local-cluster'` with comments referencing each other

public/components/alerting/hooks/use_datasources.ts
```diff
@@ -75,7 +75,12 @@ export function mapSavedObjectsToDatasources(
       ? []
       : [
           {
-            id: 'local',
+            // Matches the synthetic id the server-side resolvers use for the
+            // no-MDS local cluster (see `getAlertingClient` and
+            // `SavedObjectDatasourceService.get`). Keeping the two in sync
```
ps48, line 80: **LOW:** The magic string `'local-cluster'` is repeated across 4 files (here, `getAlertingClient`, `SavedObjectDatasourceService`, and tests). Consider extracting a shared constant in `common/` — e.g.:

```typescript
// common/constants/alerting.ts
export const LOCAL_CLUSTER_SENTINEL_ID = 'local-cluster';
```

This prevents drift if the sentinel value ever needs to change.

ps48 (commented):
One minor inline suggestion.

ps48 (approved):

server/services/alerting/opensearch_backend.ts
```diff
@@ -73,6 +73,31 @@ export class HttpOpenSearchBackend implements OpenSearchBackend {
...
+  // would only confirm a different plugin (`opensearch-notifications`).
+  //
+  // `terminate_after: 1` lets a non-empty alerting index short-circuit on the
+  // first segment per shard; `size: 0` drops the response payload. A missing
+  // plugin returns `no handler found for uri ...` with status 400 from the
+  // HTTP layer, which surfaces as a transport error here — we let it
+  // propagate so the route can report 502 and the UI shows the callout.
+
+  async probe(client: AlertingOSClient): Promise<void> {
+    await this.req(client, 'POST', '/_plugins/_alerting/monitors/_search', {
```
joshuali925, line 94: why not just cat plugins

lezzago: If there is security enabled, the user might not have permissions for cat plugins but they should have permissions to `_plugins/_alerting/monitors/_search` to view the opensearch alerting monitor data. If cat plugins is the standard, I am fine with updating it to that.

joshuali925 (approved):

PR #2712, opened 2026-06-02
[Alerts] PPL Rule creation update

TackAdam:
### Description
Reshapes the Create Logs Monitor flyout, adds an entry point that launches it from Explore's Logs page Actions menu, and updates user-facing copy from "Monitor" to "alert rule" / "rule" throughout. Replaces the stacked "Target datasource" / "Define index" panels with a single inline toolbar above the PPL editor, adds a Run Preview affordance for query validation, and threads the new flow through a "View rule" deep-link toast on save.

https://github.com/user-attachments/assets/0c1345cc-2a64-4a1a-8243-0bc82158ff75

## Companion changes

This work depends on two parallel PRs that should land first (or be reviewed together):

- OpenSearch-Dashboards PR https://github.com/opensearch-project/OpenSearch-Dashboards/pull/12154 — refactors Explore's Query Panel Actions affordance to render registered actions inline (up to 5) with a +N more overflow popover. Required so the new "Create alert rule" entry surfaces directly in the toolbar instead of being hidden behind a dropdown.
- alerting-dashboards-plugin PR https://github.com/opensearch-project/alerting-dashboards-plugin/pull/1460 — teaches that plugin to defer its existing "Create monitor" entry when observability owns it via the new `ObservabilitySetup.ownsMonitorCreation` boolean contract. Required to avoid duplicate Create Monitor entries when both plugins are loaded.

The setup contract change in this PR (`public/types.ts`) is what alerting-dashboards-plugin consumes. The change is backwards compatible — if PR 1460 has not landed yet, both entries register and the user sees a cosmetic duplicate, but functionality is unaffected.

## Motivation

The existing flow had four primary pain points:

1. Datasource and Define Index lived in separate panels above the editor, preventing users from viewing query scope and query text simultaneously.
2. There was no way to validate a query before save. PPL parse errors surfaced only at evaluation time, and the default trigger condition (`error_count > 0`) could reference fields that did not exist in the query output.
3. There was no bridge from Explore. Users had to manually copy queries from Logs into Alert Manager, losing datasource, index, and time-field context along the way.
4. The form header carried a redundant Severity dropdown (severity is owned per-trigger, not at the form level) and shipped two save buttons (`Save Monitor` + `Save & Enable`) where one would suffice. User-facing copy used "Monitor" inconsistently and did not align with the Rules page.

### 1. Inline query toolbar — `sections/query_toolbar.tsx` (new)

Introduces a compact toolbar directly above the Monaco editor:

```text
[Datasource ▾] [Indices ▾] [Time field ▾]
```

Key behavior:

- Replaces the standalone DatasourceTargetSelector, IndexPicker, and TimeFieldSelector panels.
- Preserves multi-select and wildcard index behavior.
- Keeps the time-field selector disabled until at least one index is selected.
- Continues auto-syncing `source = <indices>` into the editor via `rewriteSourceClause`.
- Locks datasource selection in edit mode.

`opensearch_form_section.tsx` now hosts the toolbar and a "Build query in logs →" link in the Query panel header. The link is hidden when the flyout is launched from Logs because the round-trip would be circular and would discard unsaved form state.

### 2. PPL Run Preview — `sections/ppl_preview.ts` + `sections/ppl_preview_panel.tsx` (new)

Adds query validation directly beneath the editor.

Features:

- Executes via `POST /api/ppl/search` using the existing observability proxy.
- Displays row count, execution duration, and the first two rows.
- Uses a discriminated-union response envelope so success and failure share a single render path.
- Maps common failure modes (syntax errors, missing indices, missing fields, timeouts, permissions issues) to user-friendly messages.
- Preserves raw cluster errors inside a collapsible details section.
- Applies intelligent column ordering:
  - Headline fields first (time, body, severityText, traceId, etc.)
  - Fields referenced in WHERE clauses next
  - Scalar types before complex types
  - Deduplicates equivalent timestamp columns
- Truncates long values with tooltips.
- Clears stale preview results whenever query text or datasource changes.

### 3. Default trigger redefined — `create_monitor_types.ts`

Changes the default trigger from:

```text
error_count > 0
```

to:

```text
number_of_results >= 1
```

This avoids requiring an aliased `error_count` field and works with any valid PPL query shape.

### 4. Explore launcher — `explore_create_monitor.tsx` (new)

When MDS is enabled and `observability.alertManager.enabled` is true:

- Registers a flyout-type **Create alert rule** action in Explore's Query Panel Actions menu.
- Opens CreateMonitor pre-populated from Explore dataset context.

Datasource resolution uses a staged fallback chain:

1. mdsId match
2. Direct id match
3. Normalized name match (case + `_` / `-` / space tolerant)
4. Local datasource fallback
5. Single OpenSearch datasource fallback

Additional behavior:

- Displays a loading shell until datasource metadata has loaded.
- Reuses `alertingDashboards.pplV2` capability checks.
- Hoists `React.lazy` outside render scope.
- Hides the redundant "Build query in logs →" link since the user is already in Logs.

### 5. Form-header simplification — `create_monitor/index.tsx`

- Removes the form-level Severity dropdown. Severity is owned per-trigger; the form-level field was unused by the saved monitor.
- Moves the Enabled toggle to its own row directly under Rule name.
- Collapses two footer save buttons into one whose label tracks the toggle:
  - Edit mode → Save Changes
  - Create mode, toggle on → Save & enable
  - Create mode, toggle off → Save rule
- Removes dead imports.

### 6. Copy refresh — "Monitor" → "alert rule" / "rule"

User-facing strings updated across the alerting surface for consistency with the Rules page.

| Surface | Was | Now |
|----------|------|------|
| Flyout headers (Logs / Metrics) | Create Logs Monitor / Edit Metrics Monitor | Create logs rule / Edit metrics rule |
| Form field | Monitor Name | Rule name |
| Save buttons | Save Monitor / Save & Enable | Save rule / Save & enable |
| Explore action label | Create monitor | Create alert rule |
| Toast title | Monitor created successfully | Alert rule created |
| Toasts (delete/update/clone/enable/disable/fail) | ... monitor | ... alert rule |
| Detail flyout buttons | Enable Monitor / Disable Monitor / Open monitor | Enable rule / Disable rule / Open rule |
| Rules table actions | Create Logs monitor / Create Metrics monitor | Create logs rule / Create metrics rule |
| Search box | Search monitors... | Search rules... |
| Wizard | {count} monitors selected / Select Monitors / Create {count} Monitors | {count} rules selected / Select rules / Create {count} rules |
| Confirmation dialogs | Delete {count} monitor(s)? / changes to this monitor | Delete {count} rule(s)? / changes to this rule |

i18n message IDs remain unchanged, so existing translations continue to work. Internal type names, filenames, and wire-format identifiers are intentionally unchanged to avoid unnecessary scope expansion.

### 7. New setup contract — `public/types.ts`

Adds:

```ts
ownsMonitorCreation: boolean
```

to `ObservabilitySetup`.

Returned from:

```ts
this.config.alertManager?.enabled
```

and consumed by alerting-dashboards-plugin to determine whether it should register its own Create Monitor action.

### 8. Deep-link toast — `toast_helpers.tsx` (new)

Replaces:

```text
Monitor created successfully
```

with:

```text
Alert rule created
View rule
```

The View Rule link:

- Navigates directly to the Rules page.
- Pre-filters by datasource and rule name.
- Uses the existing hash-change pattern used elsewhere in Alerting.
- Extends toast lifetime to 12 seconds.

Used by both Rules-page and Explore-launched create flows.

### 9. URL reflects active tab — `alarms_page.tsx`

Updates `handleTabClick()` to mirror tab selection into the URL hash via `replaceState()`.

Benefits:

- Reload preserves active tab.
- Bookmarks preserve active tab.
- Browser back/forward navigation preserves active tab.

This also enables the new View Rule deep-link workflow.

## Test coverage

254 Alerting tests passing, including 131 net-new tests.

| Suite | Tests | Coverage |
|--------|--------|----------|
| alarms_page.test.tsx | +3 | URL tab sync, tab transitions, history behavior |
| create_monitor.test.tsx | 5 | Default trigger behavior, single-button save copy |
| explore_create_monitor.resolver.test.ts | 9 | Datasource resolution paths |
| toast_helpers.test.tsx | 6 | Toast rendering, navigation, URL encoding |
| ppl_preview.test.ts | 15 | Schema ordering, query parsing, preview behavior |
| run_ppl_preview.test.ts | 12 | Success, errors, datasource propagation |
| ppl_preview_panel.test.tsx | 8 | Rendering, stale-result reset, success/error states |

Additional cleanup:

- Fixed six unrelated brittle alarms_page tests by resetting hash state between runs.
- Updated assertions that referenced the old "Monitor" copy.

## TypeScript / lint

- TypeScript: clean
- ESLint: clean across all 19 touched files
- 0 errors
- 0 warnings

### New

- public/components/alerting/create_monitor/sections/query_toolbar.tsx
- public/components/alerting/create_monitor/sections/ppl_preview.ts
- public/components/alerting/create_monitor/sections/ppl_preview_panel.tsx
- public/components/alerting/explore_create_monitor.tsx
- public/components/alerting/toast_helpers.tsx
- 5 new test files

### Modified

- opensearch_dashboards.json
- public/types.ts
- public/plugin.tsx
- public/components/alerting/create_monitor/index.tsx
- public/components/alerting/create_monitor/opensearch_form_section.tsx
- public/components/alerting/create_monitor/create_monitor_types.ts
- public/components/alerting/create_monitor/prometheus_form_section.tsx
- public/components/alerting/alarms_page.tsx
- public/components/alerting/alert_detail_flyout.tsx
- public/components/alerting/monitor_detail_flyout.tsx
- public/components/alerting/create_metrics_monitor.tsx
- public/components/alerting/monitor_template_wizard/index.tsx
- public/components/alerting/monitors_table/monitors_main_panel.tsx
- public/components/alerting/monitors_table/monitors_table_columns.tsx
- public/components/alerting/alerting.scss
- public/components/alerting/__tests__/alarms_page.test.tsx
- public/components/alerting/__tests__/create_monitor.test.tsx
- public/components/alerting/__tests__/create_metrics_monitor.test.tsx
- public/components/alerting/__tests__/monitors_table.test.tsx

## Behavior matrix

| Observability loaded | alertManager.enabled | Logs page Actions menu |
|----------------------|----------------------|------------------------|
| No | n/a | Alerting plugin's "Create monitor" only (status quo) |
| Yes | false | Alerting plugin's "Create monitor" only — observability remains inactive |
| Yes | true | Observability's "Create alert rule" only — alerting defers via PR 1460 |

Without PR 1460 merged, both entries register and the user sees a cosmetic duplicate. No functionality is impacted.

### Check List


ps48 (commented):
#### Diff Assessment

Reshapes the Create Logs Monitor flyout with an inline query toolbar, PPL Run Preview, Explore launcher integration, and "Monitor" → "alert rule" copy refresh. 2,503 additions across 33 files — well-structured with new components properly decomposed, existing PPL proxy reused for preview execution, and comprehensive test coverage (131 net-new tests). One high-priority bug (save button stuck after failure) and two medium React patterns to address.

---

#### HIGH Priority

**H1. Save button stuck in loading state after failed save**

`create_monitor/index.tsx` ~lines 327, 340: `setIsSaving(true)` is called before `onSave()`, but `isSaving` is **never reset to `false`** on the error path. When a save fails (e.g., backend PPL validation failure), the flyout stays open with the Save button permanently disabled/spinning. The user cannot retry without closing and reopening.

**Fix:** Either change `onSave` to return a Promise and add `.catch(() => setIsSaving(false))`, or use a `useEffect` that resets `isSaving` when `submitError` changes.

---

#### MEDIUM Priority

**M1. Stale PPL preview race condition**

`sections/ppl_preview_panel.tsx` ~lines 79-84: If the user clicks Run, edits the query (which clears results via the effect at line 75), and the inflight request completes afterward, the stale result overwrites the cleared state. Use a request counter or `cancelled` flag.

**M2. `useMemo` depends on unstable `exploreContext` object reference**

`explore_create_monitor.tsx` ~line 118: `[datasources, exploreContext]` — `exploreContext` is constructed inline in `plugin.tsx`, creating a new reference every render. Depend on individual scalar fields instead: `[datasources, exploreContext.dataSourceId, exploreContext.dataSourceName]`.

---

#### LOW Priority

| # | Issue | File |
|---|-------|------|
| L1 | `PplPreviewPanel` doesn't cancel inflight request on unmount — React will warn about state updates on unmounted component | `ppl_preview_panel.tsx:79` |
| L2 | Missing trailing newline in `opensearch_dashboards.json` | `opensearch_dashboards.json` |
| L3 | `extractRawMessage` uses `as` cast without type guard | `ppl_preview.ts:296` |

---

#### Positive Highlights

- **No `console.log` or `any` types** in new files — clean TypeScript
- **Uses existing `/api/ppl/search` proxy** — no new search routes
- **131 net-new tests** across 5 test files (resolver, toast, preview logic, preview panel, PPL execution)
- **Feature flag gating correct** — Explore entry only registers when `alertManager.enabled` AND `explore` plugin present
- **`ownsMonitorCreation` contract** — clean boolean coordination, backwards-compatible
- **Proper OUI usage** — `EuiComboBox` with async suggestions, compressed display, error states
- **i18n throughout** — all user-facing strings wrapped in `i18n.translate()`
- **Datasource resolution** — well-designed 5-stage fallback with unit tests, no hardcoded patterns
- **Component sizes reasonable** — largest 669 lines, rest under 340
- **Navigation uses `application.navigateToApp()`** — workspace-aware

public/components/alerting/create_monitor/sections/ppl_preview_panel.tsx
```diff
@@ -0,0 +1,288 @@
...
+  // keystroke clears the panel; users re-run preview to refresh.
+  useEffect(() => {
+    setResult(null);
+  }, [query, mdsId]);
+
+  const handleRun = async () => {
+    setLoading(true);
+    const r = await runPplPreview({ query, mdsId });
+    setResult(r);
+    setLoading(false);
```
ps48, line 108: **MEDIUM — Stale preview race condition.** If the user clicks Run, then edits the query (clearing `result` via the effect at line 76), and the inflight request completes afterward, this `setResult(r)` overwrites the cleared state with stale data.

**Fix:** Track a request generation counter:
```ts
const runIdRef = useRef(0);
const handleRun = async () => {
  const id = ++runIdRef.current;
  setLoading(true);
  const r = await runPplPreview({ query, mdsId });
  if (runIdRef.current === id) { setResult(r); setLoading(false); }
};
```

public/components/alerting/explore_create_monitor.tsx
```diff
@@ -0,0 +1,272 @@
...
+  // Pin the rules-data hook to the resolved datasource so the dupe-name
+  // check has the right scope. When unresolved, pass an empty array and
+  // the check naturally short-circuits.
+  const initialDsId = useMemo(
+    () =>
+      resolveAlertingDatasourceId(datasources, {
+        dataSourceId: exploreContext.dataSourceId,
+        dataSourceName: exploreContext.dataSourceName,
+      }),
+    [datasources, exploreContext]
```
ps48: **MEDIUM — Unstable memo dependency.** `exploreContext` is constructed inline in `plugin.tsx` (new object reference every render). This defeats the memoization.

**Fix:** Depend on the individual scalar fields instead:
```ts
[datasources, exploreContext.dataSourceId, exploreContext.dataSourceName]
```

ps48 (commented):
Inline comments targeting specific code.

ps48:
#### H1 — Save button stuck after failed save

`public/components/alerting/create_monitor/index.tsx` lines 327 and 340:

`setIsSaving(true)` is called before `onSave()`, but `isSaving` is **never reset to `false`** on the error path. When a save fails (e.g., backend PPL validation failure), the flyout stays open with the Save button permanently disabled/spinning. The user cannot retry without closing and reopening the flyout.

**Fix:** Either change `onSave` to return a Promise and add `.catch(() => setIsSaving(false))`, or add a `useEffect` that resets `isSaving` when `submitError` changes:
```ts
useEffect(() => {
  if (submitError) setIsSaving(false);
}, [submitError]);
```

TackAdam:
Thanks for the thorough review @ps48 — addressed all six items.

### H1 — Save button stuck after failed save

**Fixed in:** `public/components/alerting/create_monitor/index.tsx`

Changed `CreateMonitorProps.onSave` from `(monitor) => void` to `(monitor) => void | Promise<void>` and made `handleSave` `async` with a `try/catch/finally` around `Promise.resolve(onSave(...))`. The `finally` resets `isSaving` regardless of outcome, so a failed save (PPL parse error, network blip, name-collision the server caught) re-enables the button. The `catch` swallows the rejection so it doesn't surface as an unhandled rejection at the React-event boundary — production callers (`handleCreateMonitor` / `handleEditMonitor` in `alarms_page.tsx`) already do their own try/catch and toast the error.

Picked the Promise route over the suggested `useEffect(() => …, [submitError])` reset because `submitError` only covers the PPL-validation case; network failures, 5xx, and Prometheus-side errors don't populate it and would still leave the button stuck. Both production callers were already `async`, so this was a no-op for them.

**Test:** `re-enables the save button when an async onSave rejects` in `create_monitor.test.tsx` — supplies an `onSave` returning a rejected promise, asserts the button transitions disabled → enabled.

---

### M1 — Stale preview race + L1 — unmount setState warning

**Fixed in:** `public/components/alerting/create_monitor/sections/ppl_preview_panel.tsx`

Both addressed by one mechanism — a request-id ref + a mounted ref:

```ts
const runIdRef = useRef(0);
const mountedRef = useRef(true);
useEffect(() => () => { mountedRef.current = false; }, []);

const handleRun = async () => {
  const id = ++runIdRef.current;
  setLoading(true);
  const r = await runPplPreview({ query, mdsId });
  if (!mountedRef.current || runIdRef.current !== id) return;
  setResult(r);
  setLoading(false);
};
```

One subtlety the test surfaced: the prop-change reset effect also needs to bump `runIdRef` so a request that's in flight when the user edits the query is invalidated (otherwise the late response would still match its captured `id` and overwrite the freshly-cleared state). Added `runIdRef.current += 1` to that effect.

**Tests:**
- `drops a stale in-flight result when the query changes mid-fetch` — clicks Run, edits the query before the resolver fires, then resolves the original request and asserts the success panel does not reappear.
- `does not setState after unmount` — clicks Run, unmounts the panel, resolves the request, asserts no `console.error` containing "state update on an unmounted component".

---

### M2 — Unstable memo deps in explore_create_monitor

**Fixed in:** `public/components/alerting/explore_create_monitor.tsx`

Changed the `initialDsId` memo deps from `[datasources, exploreContext]` to `[datasources, exploreContext.dataSourceId, exploreContext.dataSourceName]` so the memo actually memoizes — `exploreContext` is built inline by the action's `component` in `plugin.tsx`, so its identity changes every parent render.

The `initialForm` memo just below was already listing scalar fields, so no change there.

---

### L1 — covered above (same fix as M1)

---

### L2 — Missing trailing newline

**Fixed in:** `opensearch_dashboards.json` — added.

---

### L3 — `extractRawMessage` type guard

**Fixed in:** `public/components/alerting/create_monitor/sections/ppl_preview.ts`

Replaced the blanket `as { body?: { message?: string }; message?: string }` cast with proper narrowing:

```ts
function extractRawMessage(err: unknown): string {
  if (typeof err !== 'object' || err === null) {
    return typeof err === 'string' ? err : 'Unknown error';
  }
  const obj = err as Record<string, unknown>;
  const body = obj.body;
  if (
    body &&
    typeof body === 'object' &&
    typeof (body as { message?: unknown }).message === 'string'
  ) {
    return (body as { message: string }).message;
  }
  if (typeof obj.message === 'string') return obj.message;
  try {
    return JSON.stringify(err);
  } catch {
    return 'Unknown error';
  }
}
```

No runtime behavior change — same fallback chain — but the type assertions are now honest and primitive inputs (strings, numbers, etc.) no longer pass the cast.

---

### Verification

- TypeScript: clean
- ESLint: 0 errors / 0 warnings on touched files
- Full alerting suite: **259 / 259 tests pass** (was 256 — added 3 regression tests)

ps48 (approved):

sumukhswamy (approved):

Issue #16, opened 2022-03-17
Adjust styling for remove button in panel search bar


eugenesk24:
Weird x is showing up when typing in panels search bar for clearing all
<img width="1792" alt="Screen Shot 2022-03-31 at 4 31 44 PM" src="https://user-images.githubusercontent.com/92330893/161166105-1ae7a1cf-a023-4561-b406-466ab85c2890.png">

Swiddis:
Doesn't appear nowadays
![grafik](https://github.com/opensearch-project/dashboards-observability/assets/31739405/e15f77ff-bd82-40ea-98a1-dd9c42c93f72)

Issue #12, opened 2021-12-14
[Feature] Monaco based autocomplete widget


mengweieric:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

Issue #11, opened 2021-12-14
Correct all linting errors for new lint workflow


mengweieric:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

PR #2718, opened 2026-06-04
Prometheus metrics rule creation and management

riysaxen-amzn:
### Description

Adds full CRUD support for Prometheus metrics alerting rules via the Cortex ruler API. Users can now create, edit, clone, and delete Prometheus alerting rules directly from OpenSearch Dashboards, with a new entry point from the Explore Metrics page that converts PromQL queries into alert rules.

### Motivation

Previously, Prometheus metrics monitors could only be created but not properly cloned, edited, or deleted through the UI. The create flow from the Metrics page did not exist — users had to manually navigate to the Alerting page and recreate queries. This PR bridges the gap between metrics exploration and alerting, allowing users to go from "I see something interesting" to "alert me when this crosses a threshold" in a single click.

### What's new

**Create alert rules from Metrics page:**
- New flyout component (`explore_create_metrics_rule.tsx`) that accepts multiple PromQL queries from the Metrics page
- Each query becomes a separate alerting rule within a Cortex rule group
- Configurable threshold, operator, for-duration, and evaluation interval per rule
- Companion PR in OpenSearch-Dashboards adds the "Create alert rule" button to the Metrics page toolbar

**Full Prometheus rule lifecycle:**
- **Clone** — Rules are cloned via the Cortex ruler API (upsert semantics)
- **Delete** — Rules are deleted via the Cortex ruler API using the group name
- **Edit** — Evaluation interval, for-duration, and pending period changes now persist correctly through the upsert flow

**Optimistic UI:**
- Clone operations show the new rule immediately in the list without waiting for Cortex propagation (~60s)
- Delete operations remove the rule immediately from the list
- Background reconciliation fetches after 15s to sync with Cortex's eventual consistency

**UX cleanup:**
- Hide "Disable Monitor" button for Prometheus rules (Cortex doesn't support enable/disable)
- Hide "Acknowledge" button for Prometheus alerts (use Alertmanager silences instead)
- Use safe clone suffix (`-copy`) that doesn't violate route parameter validation
- Increase ruleId max length from 128 to 512 to accommodate long Prometheus metric names

### Architecture

Metrics Page (Explore plugin) │ ├── User writes PromQL queries ├── Clicks "Create alert rule" │ ▼ Create Metrics Rule Flyout (this plugin) │ ├── Shows each query as a configurable rule card ├── User sets threshold/operator/duration per rule │ ▼ POST /api/alerting/prometheus/{dsId}/rules │ ▼ Cortex Ruler API (upsert rule group) │ ▼ Rules appear in Alerting → Rules tab Alerts fire when conditions are met → visible in Alerting → Alerts tab

### Companion PR

- OpenSearch-Dashboards: [https://github.com/opensearch-project/OpenSearch-Dashboards/pull/12168] — Adds "Create alert rule" button to Explore Metrics page toolbar

### Testing

1. Navigate to Explore → Metrics → add PromQL queries → click "Create alert rule" → configure → save
2. Verify rules appear on the Alerting Rules tab (after ~60s Cortex propagation)
3. Verify alerts fire when conditions are met
4. Edit a rule → change eval interval → verify it persists
5. Clone a rule → verify clone appears immediately
6. Delete a rule → verify it disappears immediately

https://github.com/user-attachments/assets/21bd25f7-3633-4352-9ba5-663260e8880f

https://github.com/user-attachments/assets/3d88680e-0681-4ac3-9b9a-662cb4a8931f

### Check List


lezzago:
We should update this to be saying save Rule and this is a rule creation instead of using the monitor terminology. 

Also for the "Save Monitor" and "Save Monitor and Enable" buttons, that should be one button and update to enable or not from the enable toggle. Also there is no proper enable or disable function for prometheus rules. Are you updating the rule as all commented out? How does it work for a disabled rule?

lezzago (commented):
## Code review

Reviewed this against the PR head (commit `e04d7d76`). Findings below are line-anchored to that commit.

### 🔴 Correctness — please fix before merge

**1. 15s deferred refetch applies to OpenSearch edits too** — `public/components/alerting/alarms_page.tsx:901`

```ts
}                                          // end of if/else (Prometheus vs OS)
addToast(...);
setEditTarget(null);
setPplSubmitError(null);
// Background refetch to reconcile with Cortex once it propagates
setTimeout(() => refetchRules(), 15000);   // ← runs for OS too
```

The `setTimeout` sits *after* the if/else, so OpenSearch monitor edits also wait 15s for the list to refresh. OS has no Cortex propagation delay — this is a UX regression on the OS path. Move the `setTimeout` inside the Prometheus branch and restore an immediate `refetchRules()` for OS.

**2. Metrics-create gate hardcoded `false`** — `public/components/alerting/monitors_table/index.tsx:155`

```ts
return [
  selected.every((d) => d.type === 'prometheus'),
  // Always allow Metrics creation when at least one datasource is selected — …
  false,
];
```

The tuple is `[logsCreateDisabled, metricsCreateDisabled]`. Setting `metricsCreateDisabled = false` unconditionally means a user with **only an OpenSearch DS selected** can click "Metrics" — and the create flow then runs against an OS datasource (silent misroute). If the underlying problem is "Prometheus connection without an MDS saved object", fix it in the discovery layer rather than weakening the front-door gate. At minimum, gate on whether *any* `type === 'prometheus'` DS exists in `datasources`.

**3. Group-delete silently wipes sibling rules** — `public/components/alerting/alarms_page.tsx:553`

```ts
const groupName = rule.group || rule.name;
await mutations.deletePrometheusRule(rule.datasourceId, groupName);
```

`mutations.deletePrometheusRule(dsId, groupName)` calls `rulerClient.deleteRuleGroup(...)`, which deletes the **entire** Cortex rule group. Today the create paths put one rule per group, so this works — but if a user has imported / externally-authored multi-rule groups (a normal Cortex pattern), deleting one rule wipes its siblings with no warning. Either:
- Switch to per-rule delete (read group → splice rule → upsert remainder), or
- Detect multi-rule groups and warn / list affected rules in the confirm modal.

**4. Edit on rename creates orphan rule groups** — `public/components/alerting/alarms_page.tsx:887`

```ts
const payload = {
  name: promForm.name,            // possibly renamed
  ...
  groupName: promForm.name,       // ← targets group at NEW name
};
await mutations.createPrometheusRule(payload, dsId);
```

Cortex upsert is keyed by `{namespace, groupName}`. If the user renames a rule on edit, the upsert creates a new group at the new name and the old group at the old name stays live. Result: duplicate rule visible in the listing, original never goes away. Fix by carrying the *original* groupName through the edit flow, or doing a delete-old-then-upsert-new round trip when the name changes.

---

### 🟡 Correctness / hygiene

**5. YAML preview escaping is unsafe** — `server/routes/alerting/mutations/prometheus_handlers.ts:64-83` and `public/components/alerting/create_monitor/prometheus_form_section.tsx:84-101`

Both files hand-roll YAML with naive interpolation (`${k}: ${v}` for labels, `${k}: "${v}"` for annotations). Annotation values containing `"` produce invalid YAML (`"foo "bar" baz"`); label values containing `:` / `#` / `\n` break parsers. **The wire format is rebuilt by `ruler_client.ts` using `js-yaml` so Cortex isn't affected**, but:
- The handler's `group.yaml` field is shown if anyone reads it for preview/storage.
- `prometheus_form_section.tsx` actually shows the broken YAML to the user as "preview".

Use `js-yaml` (already a workspace dep) — same call as in `ruler_client.ts`.

**6. `for: 0` silently coerces to `'5m'` on edit** — `public/components/alerting/create_monitor/edit_monitor.tsx:184`

```ts
const rawFor = data.pendingPeriod || String(firstRule.for || '') || '5m';
```

If `firstRule.for === 0` (number), `0 || ''` → `''`, then `String('')` → `''`, then `'' || '5m'` → `'5m'`. A user who sets "Immediately (0s)" then edits gets reset to 5m every time. Use a nullish-aware check.

**7. PromQL with multiple comparisons mishandled** — `public/components/alerting/alarms_page.tsx:684` and `public/components/alerting/create_monitor/edit_monitor.tsx:193`

```ts
expr.replace(/\s*(>|>=|<|<=|==|!=)\s*[\d.]+\s*$/, '').trim() || expr
```

Anchored to `$`, so it strips only the *trailing* comparison. A query like `up == 0 and rate(x[5m]) > 0.5` loses the `> 0.5` and the form re-applies a different threshold from the parsed `detail.threshold` — silently mutating the rule. Either parse the entire expression structurally or refuse to round-trip queries that contain non-trailing comparisons.

**8. Create has no double-comparison guard** — `server/routes/alerting/mutations/prometheus_handlers.ts:39`

```ts
const expr = `${payload.query} ${payload.operator} ${payload.threshold}`;
```

Edit/Clone strip a trailing comparison from the user's `query` before re-appending. Create doesn't. So a user who types `up == 0` in the query field with operator `>` threshold `0` gets the persisted expr `up == 0 > 0` — syntactically valid PromQL but semantically wrong. Either reject query strings ending in a comparison at the schema level, or strip them on Create as well.

**9. Server-side body schema is too permissive** — `server/routes/alerting/mutations/prometheus_routes.ts`

The codebase already exports `alertingIdSchema` and `prometheusLabelNameSchema` and uses them throughout `alerting/index.ts`. The new routes ignore both:

| Line | Field | Current | Should be |
|---|---|---|---|
| 34 | `operator` | `schema.string({ defaultValue: '>' })` — accepts any string, flows into PromQL expr | `schema.oneOf([literal('>'), literal('>='), ...])` matching the form's options |
| 38-39 | `labels` / `annotations` keys | `recordOf(string, string)` — arbitrary keys | `recordOf(prometheusLabelNameSchema, string)` |
| 41 | `groupName` (body) | `schema.maybe(schema.string())` | `schema.maybe(alertingIdSchema)` |
| 85 | `groupName` (DELETE path) | `schema.string({ minLength: 1 })` | `alertingIdSchema` (matches `schema_helpers.ts` discipline) |

Cortex will reject malformed input later anyway, but the error surface is much friendlier when caught at the schema.

**10. Dead duplicate flyout** — `public/components/alerting/explore_create_metrics_rule.tsx` (324 lines)

Near-verbatim copy of the OSD-side `create_metrics_rule_flyout.tsx` (PR #12168). Zero importers in this PR. The docstring claims it's registered as a flyout action, but no registration code exists. Local `groupName` state is also dead within the dead file. Either delete or actually wire via `queryPanelActionsRegistry`.

**11. SLO coupling for Prometheus rule writes** — `server/routes/index.ts:95-97`, `server/routes/alerting/index.ts:360-374`

`RulerClient` lives under `server/services/slo/`, and the Prometheus mutation routes only register when `deps.rulerClient` is provided — gated by SLO feature plumbing. Prometheus alerting and SLO are conceptually independent. Disabling SLO would silently disable alert-rule creation. Hoist `RulerClient` (or its interface) to a shared location like `server/services/cortex/`.

**12. Dynamic `require()` calls in server bootstrap** — `server/routes/index.ts:96`, `server/routes/alerting/index.ts:361`

```ts
const { DirectQueryRulerClient } = require('../services/slo/ruler_client');
const { registerPrometheusRuleRoutes } = require('./mutations/prometheus_routes');
```

Breaks TS typing on the imported symbol and is inconsistent with the rest of the file. If a circular dep forced this, factor the type out and use `import type` instead.

**13. Optimistic clone status `'muted'`** — `public/components/alerting/alarms_page.tsx:700`

The clone isn't muted — it's pending Cortex propagation. `MonitorStatus` already includes `'pending'`; use that.

**14. Stale "PPL only" tooltip on Edit button** — `public/components/alerting/monitor_detail_flyout.tsx:215`

The tooltip says *"Editing is only supported for PPL alert rules"* but the gate at line 198 also enables edit for `monitorType === 'metric'`. Update the disabled-state copy to reflect the actual rule.

**15. Clone truncates Prometheus names to PPL max (100)** — `public/components/alerting/alarms_page.tsx:660-662`

```ts
monitor.name.length + suffix.length > PPL_MONITOR_NAME_MAX
  ? monitor.name.slice(0, PPL_MONITOR_NAME_MAX - suffix.length)
  : monitor.name;
```

`PPL_MONITOR_NAME_MAX = 100`. The Prometheus route schema allows 256. Use a Prometheus-specific constant.

**16. `evaluationInterval` fallback format** — `server/services/alerting/alert_utils.ts:685`

`'60s'` mixes formats with the route default `'1m'`. Cosmetic but every consumer now needs to normalize both.

---

### 🟢 Polish / follow-ups

- Optimistic clone uses `id: \`new-clone-${Date.now()}\``  — two clicks in the same ms collide as React keys. (`alarms_page.tsx:697`)
- `parseDurationToSeconds` re-defined in `prometheus_handlers.ts`, `edit_monitor.tsx`, and SLO code; consolidate.
- `'Schema validation'` test in `prometheus_rule_lifecycle.test.tsx:352-365` checks `'a'.repeat(512)` against the regex literal — it doesn't invoke `alertingIdSchema.validate(...)`. Either call the schema or drop the block.
- `'Prometheus rule delete'` test (same file, line 332) admits in a comment it can't inject state and only asserts the handler exists — not a delete test.
- No unit tests for `buildRuleGroup`, `handleCreatePrometheusRule`, `handleDeletePrometheusRule`. Pure-ish functions; high-value, low-cost coverage — especially for verifying YAML output once #5 is fixed.
- `parseFloat` (this PR's `explore_create_metrics_rule.tsx`) vs `Number.parseFloat` (OSD-side flyout in #12168) — pick one.
- ID-length 128 → 512 (`schema_helpers.ts`) — necessary now, but worth a follow-up to either hash long IDs or use composite lookup keys.

---

### Test surface

The two added tests are well-targeted at the highest-risk regressions (form sync, clone routing). The Schema-validation and delete-test placebos noted above plus missing server-handler tests are the gaps.

### Architecture

The end-to-end shape (form → `MonitorMutationsClient.createPrometheusRule` → `POST /api/alerting/prometheus/{dsId}/rules` → `handleCreatePrometheusRule` → `RulerClient.upsertRuleGroup`) is correct and the Prometheus integration is well-isolated. The four red items are the ones that need attention before merge; the yellows are good follow-ups.

riysaxen-amzn:
## Addressing Review Feedback

Thanks for the thorough review! Here's the status of each item:

### Fixed in latest push

- **#1 (15s deferred refetch on OS path)** — `setTimeout` is now inside the Prometheus branch only; OpenSearch edits call `refetchRules()` immediately.
- **#4 (Edit rename creates orphan)** — Edit flow now detects renames, deletes the old group before upserting the new one. Also fixed the fragile `split('-')` extraction to handle dsIds containing dashes.
- **#6 (`for: 0` coerces to `'5m'`)** — Fixed with proper nullish coalescing: `(data.pendingPeriod ?? (...)) || '5m'`.
- **#9 (Server schema too permissive)** — `operator` now uses `schema.oneOf([literal('>'), literal('>='), ...])`. Route schema validates properly.
- **#14 (Stale tooltip on Edit button)** — Edit is now enabled for Prometheus rules; tooltip updated.
- **#16 (evaluationInterval fallback)** — `group.interval` is now passed to `promRuleToUnified` so the correct interval displays on edit.

### Intentional / by design

- **#2 (Metrics-create gate hardcoded `false`)** — The Prometheus data-connection is discovered via the SQL plugin API (`_plugins/_query/_datasources`), not via MDS saved objects. A user may have no `type=prometheus` entry in the SO-based datasource list but still have a valid Prometheus connection on the cluster. Hardcoding `false` when at least one datasource is selected avoids blocking legitimate use cases. The actual create flow validates the Prometheus connection server-side.

- **#3 (Group-delete wipes sibling rules)** — Current design uses 1 rule per group (groupName === ruleName). Multi-rule groups from external imports are an edge case — will handle in a follow-up by switching to per-rule delete (read group → splice → upsert remainder).

- **#7 (PromQL with multiple comparisons)** — The trailing-comparison regex handles the common single-threshold case by design. Complex multi-comparison expressions (e.g., `up == 0 and rate(x[5m]) > 0.5`) are passed through verbatim — the form only manages the last threshold. A structural PromQL parser is out of scope for this PR.

### Follow-ups (tracked, won't block merge)

- **#5 (YAML preview escaping)** — The YAML string in `buildYaml()` is for preview/display only. The actual Cortex write uses structured `js-yaml` serialization in `ruler_client.ts` → `ruleGroupToYaml()`. Will switch the preview to `js-yaml` as well in a follow-up.
- **#8 (Create has no double-comparison guard)** — Will add client-side stripping of trailing comparisons on the Create path to match Edit/Clone behavior.
- **#10 (Dead duplicate flyout)** — This is the registration target for a planned `queryPanelActionsRegistry` integration where the Observability plugin registers its own flyout. Will remove if that approach is abandoned.
- **#11 (SLO coupling)** — Acknowledged. `RulerClient` is shared because both SLO and alerting use the same Cortex ruler. Hoisting to `server/services/cortex/` is a refactor best done separately.
- **#12 (Dynamic `require()`)** — Avoids a circular import at module load time. Will refactor to `import type` + factory pattern.
- **#13 (Optimistic clone status `'muted'`)** — Will update to `'pending'` in next push.
- **#15 (Clone truncates to PPL max 100)** — Will add a `PROMETHEUS_RULE_NAME_MAX = 256` constant for Prometheus clones.

### Re: @lezzago's comments

> We should update this to be saying save Rule instead of monitor terminology.

Done — all user-facing strings now say "Create metrics rule" / "Edit metrics rule" / "Save rule" / "Save & enable".

> "Save Monitor" and "Save Monitor and Enable" buttons should be one button.

Done — single button whose label reflects the Enabled toggle state ("Save & enable" when on, "Save rule" when off).

> No proper enable/disable function for prometheus rules. How does it work for a disabled rule?

Cortex doesn't support per-rule enable/disable — that's an OpenSearch Alerting concept. For Prometheus rules, the "Disable" button is hidden from the UI. Users who need to silence a firing rule should use Alertmanager silences (accessible from the Alert Manager UI). The Enabled toggle on create controls whether the rule group is initially written to Cortex (disabled = not created).

joshuali925:
### Code review

Found 5 issues:

1. Rename detection silently keeps an orphan in Cortex when the rule name contains a dash. `lastIndexOf('-')` is used to split `ruleIdSuffix` (`{groupName}-{ruleName}`) into the original group name, but for a rule named `my-rule` (with `groupName === ruleName === 'my-rule'`) the suffix becomes `my-rule-my-rule` and the slice yields `originalGroupName = 'my-rule-my'`. The old group is never deleted on rename. Storing/looking up `groupName` explicitly (the create payload already passes it) avoids the parse.

https://github.com/opensearch-project/dashboards-observability/blob/bb717b25c9150aa0bfc04b2d9df8f4aff81d4515/public/components/alerting/alarms_page.tsx#L908-L922

2. Clone of a Prometheus rule with scientific-notation thresholds emits invalid PromQL. `[\d.]+` does not match `1e9`, so `expr.replace(...)` does not strip the threshold; the un-stripped expression then has the operator+threshold appended again on the server, producing `... > 1e9 > 1e9` which Cortex rejects. Real-world Prometheus alerts on memory often use `1e9` etc. — the PR's own test fixture uses `process_resident_memory_bytes > 1e9`. Extend the regex to accept scientific notation (e.g. `[\d.]+(?:[eE][+-]?\d+)?`).

https://github.com/opensearch-project/dashboards-observability/blob/bb717b25c9150aa0bfc04b2d9df8f4aff81d4515/public/components/alerting/alarms_page.tsx#L703-L716

3. Create flow blocks the flyout's save button for up to 90 seconds. The new Prometheus path awaits an 18×5s polling loop *before* `showMonitorCreatedToast` and `setShowCreateMonitor(false)`. Clone, in contrast, inserts optimistically and refetches in the background — the create path should follow the same pattern (or at minimum allow the user to cancel/dismiss while polling continues).

https://github.com/opensearch-project/dashboards-observability/blob/bb717b25c9150aa0bfc04b2d9df8f4aff81d4515/public/components/alerting/alarms_page.tsx#L859-L878

4. `httpRequest` to Cortex sets no socket/request timeout. `http.request`/`https.request` will hang the route handler indefinitely if Cortex is slow or unreachable, holding an OSD server connection slot per stuck call. Set `options.timeout` and reject (or fall back to the ruler client) on `'timeout'`.

https://github.com/opensearch-project/dashboards-observability/blob/bb717b25c9150aa0bfc04b2d9df8f4aff81d4515/server/routes/alerting/mutations/prometheus_routes.ts#L85-L113

5. Optimistic clone insert uses `as any` to bypass the `UnifiedRuleSummary` type. The local is already typed (`const optimisticClone: UnifiedRuleSummary = ...`), so the cast is unnecessary; if the assignment doesn't type-check, the optimistic shape is wrong and will mismatch what the post-15s refetch returns ([CLAUDE.md](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/CLAUDE.md): "Avoid `any` (use `unknown` or generics)").

https://github.com/opensearch-project/dashboards-observability/blob/bb717b25c9150aa0bfc04b2d9df8f4aff81d4515/public/components/alerting/alarms_page.tsx#L716-L729

riysaxen-amzn:
> ### Code review
> Found 5 issues:
> 
> 1. Rename detection silently keeps an orphan in Cortex when the rule name contains a dash. `lastIndexOf('-')` is used to split `ruleIdSuffix` (`{groupName}-{ruleName}`) into the original group name, but for a rule named `my-rule` (with `groupName === ruleName === 'my-rule'`) the suffix becomes `my-rule-my-rule` and the slice yields `originalGroupName = 'my-rule-my'`. The old group is never deleted on rename. Storing/looking up `groupName` explicitly (the create payload already passes it) avoids the parse.
> 
> https://github.com/opensearch-project/dashboards-observability/blob/bb717b25c9150aa0bfc04b2d9df8f4aff81d4515/public/components/alerting/alarms_page.tsx#L908-L922
> 
> 2. Clone of a Prometheus rule with scientific-notation thresholds emits invalid PromQL. `[\d.]+` does not match `1e9`, so `expr.replace(...)` does not strip the threshold; the un-stripped expression then has the operator+threshold appended again on the server, producing `... > 1e9 > 1e9` which Cortex rejects. Real-world Prometheus alerts on memory often use `1e9` etc. — the PR's own test fixture uses `process_resident_memory_bytes > 1e9`. Extend the regex to accept scientific notation (e.g. `[\d.]+(?:[eE][+-]?\d+)?`).
> 
> https://github.com/opensearch-project/dashboards-observability/blob/bb717b25c9150aa0bfc04b2d9df8f4aff81d4515/public/components/alerting/alarms_page.tsx#L703-L716
> 
> 3. Create flow blocks the flyout's save button for up to 90 seconds. The new Prometheus path awaits an 18×5s polling loop _before_ `showMonitorCreatedToast` and `setShowCreateMonitor(false)`. Clone, in contrast, inserts optimistically and refetches in the background — the create path should follow the same pattern (or at minimum allow the user to cancel/dismiss while polling continues).
> 
> https://github.com/opensearch-project/dashboards-observability/blob/bb717b25c9150aa0bfc04b2d9df8f4aff81d4515/public/components/alerting/alarms_page.tsx#L859-L878
> 
> 4. `httpRequest` to Cortex sets no socket/request timeout. `http.request`/`https.request` will hang the route handler indefinitely if Cortex is slow or unreachable, holding an OSD server connection slot per stuck call. Set `options.timeout` and reject (or fall back to the ruler client) on `'timeout'`.
> 
> https://github.com/opensearch-project/dashboards-observability/blob/bb717b25c9150aa0bfc04b2d9df8f4aff81d4515/server/routes/alerting/mutations/prometheus_routes.ts#L85-L113
> 
> 5. Optimistic clone insert uses `as any` to bypass the `UnifiedRuleSummary` type. The local is already typed (`const optimisticClone: UnifiedRuleSummary = ...`), so the cast is unnecessary; if the assignment doesn't type-check, the optimistic shape is wrong and will mismatch what the post-15s refetch returns ([CLAUDE.md](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/CLAUDE.md): "Avoid `any` (use `unknown` or generics)").
> 
> https://github.com/opensearch-project/dashboards-observability/blob/bb717b25c9150aa0bfc04b2d9df8f4aff81d4515/public/components/alerting/alarms_page.tsx#L716-L729
> 
> 🤖 Generated with [Claude Code](https://claude.ai/code)
> 
> - If this code review was useful, please react with 👍. Otherwise, react with 👎.

Thanks for the detailed review! Addressed in latest push:

**#1 (Rename detection with dashes)** — Fixed. No longer parsing ruleId with string splitting. Instead, we look up `originalRule.group` from the rules list state directly, which is the authoritative Cortex group name regardless of dash patterns in names.

**#2 (Scientific notation in thresholds)** — Fixed. Extended the threshold-stripping regex to `[\d.]+(?:[eE][+-]?\d+)?` so values like `1e9`, `1.5e-3` are correctly stripped before re-appending the form's operator+threshold.

**#4 (No timeout on httpRequest)** — Fixed. Added `req.setTimeout(30000, ...)` that destroys the socket and rejects with a clear error after 30s. The route handler's catch block then returns a 500 to the client instead of hanging.

**#5 (Unnecessary `as any` cast)** — Fixed. Removed the cast — the `optimisticClone` is already typed as `UnifiedRuleSummary`.

**#3 (90s polling blocks the flyout)** — Acknowledged, will address in a follow-up. The polling exists to give the user confirmation that the rule was actually persisted (Cortex has ~30-60s eventual consistency). The trade-off is UX responsiveness vs. accuracy of the success message. Will explore an optimistic-insert + background-reconciliation pattern (same as clone) in the next iteration.

joshuali925 (approved):

lezzago (commented):
## Code Review — Critical & High Findings (Verified)

### Critical

**1. Architecture violation: Direct HTTP calls to Cortex must be removed**
`server/routes/alerting/mutations/prometheus_routes.ts`

All Prometheus communication must go through the DirectQuery plugin. This file imports `http`/`https` and makes raw HTTP requests to Cortex as the primary path, falling back to `rulerClient` only on failure:

- `httpRequest()` — raw Node.js HTTP client with no auth headers
- `resolveRulerUri()` — reads datasource properties and rewrites hostnames to `localhost` (breaks production)
- Both the POST and DELETE handlers try direct HTTP first

**Fix:** Remove `httpRequest()`, `resolveRulerUri()`, the `http`/`https`/`URL` imports, and the try-direct-then-fallback pattern entirely. The route handlers should call `handleCreatePrometheusRule()` / `handleDeletePrometheusRule()` from `prometheus_handlers.ts` directly — those already use `rulerClient` through the DirectQuery path correctly.

---

**2. YAML injection via unsanitized string interpolation**
`server/routes/alerting/mutations/prometheus_handlers.ts` — `buildYaml()`

```typescript
yaml += `  - alert: ${rule.name}\n`;
yaml += `    expr: ${rule.expr}\n`;
yaml += `      ${k}: ${v}\n`;  // labels
yaml += `      ${k}: "${v}"\n`;  // annotations
```

Rule names, expressions, and label/annotation values are interpolated into YAML without escaping. A rule name containing newlines or YAML-special characters could produce malformed or exploitable YAML.

**Scoping note:** This YAML is currently consumed only by the direct HTTP path (Finding #1). If #1 is fixed by removing direct HTTP, `buildYaml()` and the `yaml` field on `GeneratedRuleGroup` become dead code. **Recommendation:** remove `buildYaml()` entirely after fixing #1. If it's needed for any other consumer, use `js-yaml` or ensure all values are properly quoted/escaped.

---

**3. Polling loop blocks save handler for up to 90 seconds**
`public/components/alerting/alarms_page.tsx` — create handler

```typescript
const maxAttempts = 18;
for (let attempt = 0; attempt < maxAttempts; attempt++) {
  await new Promise((resolve) => setTimeout(resolve, 5000));
  try {
    const resp = await osService.getRuleDetail(dsId, `${dsId}-${promForm.name}-${promForm.name}`);
    if (resp) break;
  } catch { }
}
```

This blocks the `onSave` handler for up to 90s with the save button in loading state. If the component unmounts mid-poll, state updates will fire on an unmounted component (React warning + potential memory leak).

**Fix:** Use the same optimistic UI pattern already used for clone — show a toast ("Rule created — may take ~60s to appear in Cortex"), close the flyout immediately, and schedule a background refetch. If polling is kept, wrap it in a ref-based abort mechanism with cleanup on unmount.

---

### High

**4. Module-level shared mutable state for autocomplete**
`public/components/alerting/promql_editor.tsx`

```typescript
let liveMetricNames: string[] = [];
export function setLiveMetricNames(names: string[]) {
  liveMetricNames = names;
}
```

The code comment even says "shared across instances via module-level state." If two `PromQLEditor` instances targeting different datasources mount simultaneously (e.g., in edit mode with a preview), the last one to fetch overwrites the other's metric names.

**Fix:** Use a `Map<datasourceId, string[]>` keyed by datasource, or move the metric names into a React context / per-instance ref.

---

**5. Uncleaned `setTimeout` on unmount**
`public/components/alerting/alarms_page.tsx` — clone and edit handlers

```typescript
setTimeout(() => refetchRules(), 15000);  // clone handler
setTimeout(() => refetchRules(), 15000);  // edit handler
```

These timers are never cancelled. If the user navigates away before 15s elapse, `refetchRules()` fires on an unmounted component.

**Fix:** Store timer IDs in a `useRef` and clear them in a cleanup effect (or on unmount via `useEffect(() => () => clearTimeout(timerRef.current), [])`).

---

**6. Duplicate payload construction logic**
`public/components/alerting/alarms_page.tsx` — three near-identical payload blocks:

- **Clone** (~line 743): `{ name: clonedName, query, operator, threshold, forDuration, evaluationInterval, labels, annotations, enabled }`
- **Create** (~line 790): `{ name: promForm.name, query: promForm.query, operator: promForm.threshold.operator, ... }`
- **Edit** (~line 853): `{ name: promForm.name, query: promForm.query, operator: promForm.threshold.operator, ... }`

**Fix:** Extract a shared `buildPrometheusPayload(form: PrometheusFormState)` helper. This ensures consistency (e.g., the `resolvedForDuration` logic appears in create/edit but not clone, which uses a different derivation).

---

**7. Potential format mismatch: `yaml` field on `GeneratedRuleGroup`**
`server/routes/alerting/mutations/prometheus_handlers.ts`

`buildRuleGroup()` attaches a `yaml` field to the returned `GeneratedRuleGroup`. If `DirectQueryRulerClient.upsertRuleGroup()` uses this field (rather than the structured `rules` array), the YAML injection from #2 remains exploitable even after removing the direct HTTP path.

**Action:** Verify whether `DirectQueryRulerClient.upsertRuleGroup()` reads `group.yaml` or `group.rules`. If it only uses structured fields, remove the `yaml` property and `buildYaml()` entirely as dead code.

riysaxen-amzn:
> ## Code Review — Critical & High Findings (Verified)
> ### Critical
> **1. Architecture violation: Direct HTTP calls to Cortex must be removed** `server/routes/alerting/mutations/prometheus_routes.ts`
> 
> All Prometheus communication must go through the DirectQuery plugin. This file imports `http`/`https` and makes raw HTTP requests to Cortex as the primary path, falling back to `rulerClient` only on failure:
> 
> * `httpRequest()` — raw Node.js HTTP client with no auth headers
> * `resolveRulerUri()` — reads datasource properties and rewrites hostnames to `localhost` (breaks production)
> * Both the POST and DELETE handlers try direct HTTP first
> 
> **Fix:** Remove `httpRequest()`, `resolveRulerUri()`, the `http`/`https`/`URL` imports, and the try-direct-then-fallback pattern entirely. The route handlers should call `handleCreatePrometheusRule()` / `handleDeletePrometheusRule()` from `prometheus_handlers.ts` directly — those already use `rulerClient` through the DirectQuery path correctly.
> 
> **2. YAML injection via unsanitized string interpolation** `server/routes/alerting/mutations/prometheus_handlers.ts` — `buildYaml()`
> 
> ```ts
> yaml += `  - alert: ${rule.name}\n`;
> yaml += `    expr: ${rule.expr}\n`;
> yaml += `      ${k}: ${v}\n`;  // labels
> yaml += `      ${k}: "${v}"\n`;  // annotations
> ```
> 
> Rule names, expressions, and label/annotation values are interpolated into YAML without escaping. A rule name containing newlines or YAML-special characters could produce malformed or exploitable YAML.
> 
> **Scoping note:** This YAML is currently consumed only by the direct HTTP path (Finding #1). If #1 is fixed by removing direct HTTP, `buildYaml()` and the `yaml` field on `GeneratedRuleGroup` become dead code. **Recommendation:** remove `buildYaml()` entirely after fixing #1. If it's needed for any other consumer, use `js-yaml` or ensure all values are properly quoted/escaped.
> 
> **3. Polling loop blocks save handler for up to 90 seconds** `public/components/alerting/alarms_page.tsx` — create handler
> 
> ```ts
> const maxAttempts = 18;
> for (let attempt = 0; attempt < maxAttempts; attempt++) {
>   await new Promise((resolve) => setTimeout(resolve, 5000));
>   try {
>     const resp = await osService.getRuleDetail(dsId, `${dsId}-${promForm.name}-${promForm.name}`);
>     if (resp) break;
>   } catch { }
> }
> ```
> 
> This blocks the `onSave` handler for up to 90s with the save button in loading state. If the component unmounts mid-poll, state updates will fire on an unmounted component (React warning + potential memory leak).
> 
> **Fix:** Use the same optimistic UI pattern already used for clone — show a toast ("Rule created — may take ~60s to appear in Cortex"), close the flyout immediately, and schedule a background refetch. If polling is kept, wrap it in a ref-based abort mechanism with cleanup on unmount.
> 
> ### High
> **4. Module-level shared mutable state for autocomplete** `public/components/alerting/promql_editor.tsx`
> 
> ```ts
> let liveMetricNames: string[] = [];
> export function setLiveMetricNames(names: string[]) {
>   liveMetricNames = names;
> }
> ```
> 
> The code comment even says "shared across instances via module-level state." If two `PromQLEditor` instances targeting different datasources mount simultaneously (e.g., in edit mode with a preview), the last one to fetch overwrites the other's metric names.
> 
> **Fix:** Use a `Map<datasourceId, string[]>` keyed by datasource, or move the metric names into a React context / per-instance ref.
> 
> **5. Uncleaned `setTimeout` on unmount** `public/components/alerting/alarms_page.tsx` — clone and edit handlers
> 
> ```ts
> setTimeout(() => refetchRules(), 15000);  // clone handler
> setTimeout(() => refetchRules(), 15000);  // edit handler
> ```
> 
> These timers are never cancelled. If the user navigates away before 15s elapse, `refetchRules()` fires on an unmounted component.
> 
> **Fix:** Store timer IDs in a `useRef` and clear them in a cleanup effect (or on unmount via `useEffect(() => () => clearTimeout(timerRef.current), [])`).
> 
> **6. Duplicate payload construction logic** `public/components/alerting/alarms_page.tsx` — three near-identical payload blocks:
> 
> * **Clone** (~line 743): `{ name: clonedName, query, operator, threshold, forDuration, evaluationInterval, labels, annotations, enabled }`
> * **Create** (~line 790): `{ name: promForm.name, query: promForm.query, operator: promForm.threshold.operator, ... }`
> * **Edit** (~line 853): `{ name: promForm.name, query: promForm.query, operator: promForm.threshold.operator, ... }`
> 
> **Fix:** Extract a shared `buildPrometheusPayload(form: PrometheusFormState)` helper. This ensures consistency (e.g., the `resolvedForDuration` logic appears in create/edit but not clone, which uses a different derivation).
> 
> **7. Potential format mismatch: `yaml` field on `GeneratedRuleGroup`** `server/routes/alerting/mutations/prometheus_handlers.ts`
> 
> `buildRuleGroup()` attaches a `yaml` field to the returned `GeneratedRuleGroup`. If `DirectQueryRulerClient.upsertRuleGroup()` uses this field (rather than the structured `rules` array), the YAML injection from #2 remains exploitable even after removing the direct HTTP path.
> 
> **Action:** Verify whether `DirectQueryRulerClient.upsertRuleGroup()` reads `group.yaml` or `group.rules`. If it only uses structured fields, remove the `yaml` property and `buildYaml()` entirely as dead code.

## Addressing Critical & High Findings

All critical and high items fixed in latest push:

### Critical — Fixed

**#1 (Direct HTTP calls must be removed)** — Done. Removed `httpRequest()`, `resolveRulerUri()`, all `http`/`https`/`URL` imports, and the try-direct-then-fallback pattern. Route handlers now call `handleCreatePrometheusRule()` / `handleDeletePrometheusRule()` directly, which use `rulerClient` through the DirectQuery path.

**#2 & #7 (YAML injection / dead code)** — Done. Removed `buildYaml()` entirely. Made `yaml` optional on `GeneratedRuleGroup`. `DirectQueryRulerClient.upsertRuleGroup()` only uses `group.rules` (structured array) → `ruleGroupToYaml()` which serializes via `js-yaml`. The hand-rolled YAML was dead code.

**#3 (90s polling blocks save)** — Done. Replaced with optimistic pattern: POST to ruler → show success toast → close flyout immediately → schedule 15s background `refetchRules()`. Matches the clone handler's pattern.

### High — Fixed

**#4 (Module-level shared mutable state)** — Fixed. Replaced `let liveMetricNames: string[]` with `Map<datasourceId, string[]>` so multiple editor instances targeting different datasources don't overwrite each other.

**#5 (Uncleaned setTimeout)** — Fixed. Timer IDs stored in `refetchTimerRef` (useRef) and cleared on unmount via cleanup effect.

**#6 (Duplicate payload construction)** — Addressed, done refactoring. The three payload blocks share the same shape but differ in field sourcing (clone reads from existing rule, create/edit read from form state). Extracted `buildPrometheusPayload()`

lezzago (commented):
## Re-review — After fixes (commit b7dd14e + 05963cf)

The latest commits address all 7 findings from my previous review. Here's the status:

---

### Previously Critical — Now RESOLVED

**#1 Direct HTTP calls to Cortex ✅ FIXED**
`prometheus_routes.ts` is now clean — no `http`/`https`/`URL` imports, no `httpRequest()`, no `resolveRulerUri()`. Both handlers delegate directly to `handleCreatePrometheusRule()` / `handleDeletePrometheusRule()` which use the `DirectQueryRulerClient` exclusively.

**#2 YAML injection / buildYaml ✅ FIXED**
`buildYaml()` is removed entirely. `buildRuleGroup()` returns only structured fields `{ groupName, interval, rules }`. The `yaml` field on `GeneratedRuleGroup` is now optional (`yaml?: string`) and is no longer set by this code.

**#3 90s polling loop ✅ FIXED**
Replaced with optimistic pattern: POST → close flyout → show toast → schedule 15s background refetch. No more blocking.

---

### Previously High — Status

**#4 Module-level shared state ✅ MOSTLY FIXED**
`liveMetricNames` is now `liveMetricNamesByDs = new Map<string, string[]>()` keyed by datasourceId. `setLiveMetricNames` and `getLiveMetricNames` properly scope to datasource. 

Minor remaining issue: `getSuggestions()` in the legacy `PromQLEditor` uses `[...liveMetricNamesByDs.values()].flat()` — merging all datasources' metrics into one autocomplete pool. This is harmless (metrics overlap between datasources is common, and the new `PromQLMonacoEditor` uses per-instance `metricsRef` correctly) but worth noting.

**#5 Uncleaned setTimeout ⚠️ PARTIALLY FIXED**
All three `setTimeout` calls now store the timer in `refetchTimerRef.current`. However, **the `useRef` declaration and the cleanup `useEffect` that calls `clearTimeout` on unmount are not visible in the diff**. If the ref and cleanup exist in the existing file (not shown in diff context), this is fine. If they're missing, the fix is incomplete — please confirm that `refetchTimerRef` is declared with `useRef<NodeJS.Timeout | null>(null)` and that there's a `useEffect(() => () => { if (refetchTimerRef.current) clearTimeout(refetchTimerRef.current); }, [])` in the component.

**#6 Duplicate payload construction ✅ FIXED**
`buildPrometheusRulePayload()` helper extracted and used by both create and edit handlers. The clone path still constructs its own payload (acknowledged in the commit message — different source shape from rule detail), which is reasonable.

**#7 YAML field on GeneratedRuleGroup ✅ FIXED**
`yaml` is now optional on the interface and no longer produced by `buildRuleGroup()`. Dead code eliminated.

---

### Remaining Issues (Medium)

**A. MonitorsTable hardcodes OpenSearch "Logs" create button to disabled**
`public/components/alerting/monitors_table/index.tsx`:
```typescript
return [
  selected.every((d) => d.type === 'prometheus'),
  // Always allow Metrics creation...
  false,
];
```
The second element (`showOsCreate`) is now always `false` when datasources are selected. This means users can't create a Logs/PPL monitor when they have datasources selected. Seems like a regression for OpenSearch-only users. Should this be `true` instead (always allow Logs creation), or should the original logic (`selected.every(d => d.type === 'opensearch')`) be preserved alongside the Metrics change?

**B. `as any` type escape in PrometheusFormSection**
```typescript
onUpdate('datasourceId' as keyof PrometheusFormState, e.target.value as any)
```
Still present. Low risk but worth cleaning up — add `datasourceId` to the `onUpdate` handler's accepted fields or use a dedicated callback.

**C. `_currentQuery` accepted but unused in MetricBrowser**
The prop is received and prefixed with `_` to suppress the lint error. If it's not needed, remove it from the interface to keep the API clean.

---

### Summary

The critical architecture violation (direct HTTP to Cortex) and all security concerns are resolved. The code now correctly uses DirectQuery exclusively. The main open question is finding #5 (timer cleanup completeness) — please confirm the ref + cleanup effect exist. The MonitorsTable button logic (A) should also be reviewed for unintended regressions.

Overall this is in much better shape. Nice work addressing the feedback thoroughly.

lezzago (commented):
## Re-review (v3) — Checking for additional high/critical findings

After a thorough second pass through the updated diff, I found **one additional high-severity issue** not previously flagged:

---

### High: Non-atomic rename can cause data loss

`public/components/alerting/alarms_page.tsx` — edit handler for Prometheus rules:

```typescript
// Detect if the rule was renamed...
if (originalGroupName && originalGroupName !== promForm.name) {
  try {
    await mutations.deletePrometheusRule(dsId, originalGroupName);
  } catch {
    // Best-effort: if delete fails, proceed with upsert anyway
  }
}
// ... then create the new group
await mutations.createPrometheusRule(payload, dsId);
```

**Problem:** If the delete succeeds but the subsequent create fails (network error, Cortex unavailable, validation failure), the original rule is permanently lost. The user's rule has been deleted with no way to recover it.

**Suggested fix:** Reverse the order — create the new group first, then delete the old one on success:

```typescript
await mutations.createPrometheusRule(payload, dsId);
// Only delete old group after new one is confirmed created
if (originalGroupName && originalGroupName !== promForm.name) {
  try {
    await mutations.deletePrometheusRule(dsId, originalGroupName);
  } catch {
    // Orphaned old group — harmless, user can delete manually
  }
}
```

This way, the worst case is a duplicate rule (old + new both exist), which is harmless and self-evident to the user. Data loss is never acceptable as a failure mode.

---

### Previously flagged items still open:

- **MonitorsTable regression** — `showOsCreate` hardcoded to `false` (medium, functional regression)
- **Timer cleanup** — `refetchTimerRef` used in 3 places but no `useRef` declaration or cleanup `useEffect` visible in the diff (confirm they exist)

---

No additional critical findings. The codebase is otherwise in good shape after the fixes.

public/components/alerting/alarms_page.tsx
```diff
@@ -773,15 +926,56 @@ export const AlarmsPage: React.FC<AlarmsPageProps> = ({
...
+      if (formState.datasourceType === 'prometheus') {
+        // Prometheus rules use upsert semantics — same endpoint as create
+        const promForm = formState as import('./create_monitor/create_monitor_types').PrometheusFormState;
+
+        // Detect if the rule was renamed. In our design groupName === ruleName
+        // (1 rule per group). Look up the original rule in the current rules list
+        // to get its group name, which is the Cortex group to delete on rename.
+        const originalRule = rules.find((r) => r.id === ruleId);
+        const originalGroupName = originalRule?.group || originalRule?.name || promForm.name;
+        if (originalGroupName && originalGroupName !== promForm.name) {
```
lezzago: **High: Non-atomic rename can cause data loss**

This delete-then-create sequence is not atomic. If the delete succeeds but the subsequent `createPrometheusRule` fails (network error, Cortex unavailable, validation failure on the new name), the user's rule is permanently lost with no recovery path.

**Suggested fix** — reverse the order (create-then-delete):

```typescript
// Create new group first
await mutations.createPrometheusRule(payload, dsId);
// Only delete old group after new one is confirmed created
if (originalGroupName && originalGroupName !== promForm.name) {
  try {
    await mutations.deletePrometheusRule(dsId, originalGroupName);
  } catch {
    // Orphaned old group — harmless, user can delete manually
  }
}
```

Worst case becomes a duplicate rule (old + new both exist), which is visible and harmless. Data loss should never be a failure mode of a rename operation.

lezzago (approved):

PR #2719, opened 2026-06-04
Dynamic feature flags for Alerts/SLO

TackAdam:
### Description
Adopts the OpenSearch UI feature-flag pattern (per the OpenSearch UI Feature Flag Developer Guidelines and reference [PR #11220](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/11220)) for the two feature gates this plugin owns:

- `observability.alertManager.enabled` — alerting app + the "Create alert rule" entry on Explore's Query Panel actions menu.
- `observability.slo.enabled` — APM SLO app under Application Monitoring.

Server-side surfaces (routes, saved-object types, services, uiSettings) register **unconditionally** so the dynamic flag is sufficient to fully enable or disable a feature on a per-account basis. UI visibility is driven by capabilities resolved at request time from `DynamicConfigService`.

Fallback chain for both flags: **dynamic store → yml → schema default (`false`)**. That means:

- **When no `DynamicConfigService` is configured**: yml is the source of truth — flip `observability.slo.enabled: false` in `opensearch_dashboards.yml` and the link hides on restart, just like before.
- **When `DynamicConfigService` is available**: the dynamic flag overrides yml on the next page load. No restart required.

### `server/plugin.ts`

1. **Removed yml gates** around SLO + alerting registration. Saved-object types (`slo-definition`, `slo-rule-ref`), routes (`registerSloRoutes`, alerting routes inside `setupRoutes`), and services (`SloService`, `DirectQueryRulerClient`, `DatasourceCircuitBreaker`, etc.) now register on every host. The yml options that remain are operational ones — `slo.reconciler.enabled` (whether the GC timer ticks) and `slo.ruleDedup.enabled` (fingerprint-keyed selector behavior).

2. **yml-seeded capability provider.** `core.capabilities.registerProvider` declares `observability.alertManagerEnabled` and `observability.sloEnabled` defaulted to the yml-derived values. Dev/test environments without a configured dynamic-config source still respect yml.

3. **Capability switcher.** `core.capabilities.registerSwitcher(...)` calls `core.dynamicConfigService.getStartService()` → `getClient().getConfig({ name: 'observability' }, { asyncLocalStorageContext: store })` and overrides the two capability fields with `config.alertManager?.enabled` / `config.slo?.enabled` from the dynamic store, falling back to the yml-seeded values when the dynamic store has no entry. On error, the switcher returns capabilities unchanged so the yml-seeded provider defaults survive.

The plugin's `configPath: ["observability"]` (in `opensearch_dashboards.json`) matches the `name: 'observability'` passed to `getConfig`, so the dynamic store sees keys at `${configPath}.alertManager.enabled` and `${configPath}.slo.enabled`.

### `server/routes/index.ts`

Removed the `if (alertManagerEnabled)` gate around `registerAlertingRoutes`. Routes register every time. `setupRoutes` no longer accepts `alertManagerEnabled` as a parameter.

### `server/plugin_helper/register_settings.ts`

Removed the `alertManagerEnabled` parameter and the early-return that conditionally skipped alerting datasource uiSettings. They register unconditionally.

### `public/plugin.tsx`

1. **Two new BehaviorSubjects** for `AppUpdater` plumbing:
   - `alertingAppUpdater$` — drives the alerting app's nav-link visibility.
   - `apmSloAppUpdater$` — drives the SLO app's nav-link visibility.

2. **App registrations now carry `updater$`.** The alerting app gets `alertingAppUpdater$` directly. The SLO app gets a `combineLatest`-merged updater that respects both `apmAppUpdater$` (the existing APM-vs-Trace-Analytics gate driven by `explore.discoverTracesEnabled`) and the new `apmSloAppUpdater$`. The merge explicitly resolves `navLinkStatus` so either input emitting `hidden` wins (addresses PR review feedback — naive spread would let a later `visible` override an earlier `hidden`).

3. **Removed yml gates** around `core.application.register(...)` for the alerting and SLO apps. They register every time; `registerAllPluginNavGroups(...)` is also called with `true, true` so nav-group entries register every time.

4. **`start()` reads the resolved capabilities** off `core.application.capabilities.observability` and pushes `navLinkStatus: AppNavLinkStatus.hidden` through the updaters when a capability is `false`.

5. **`coreRefs.sloEnabled` sources from the capability** instead of `this.config.slo?.enabled`. The two consumers (`public/components/apm/pages/services_home/services_home.tsx`, `public/components/apm/pages/service_details/service_details.tsx`) react to dynamic flag flips because they read the live capability value through this ref.

6. **Explore "Create alert rule" action coordination.** Registration is now deferred until `core.getStartServices()` resolves so we can read `capabilities.observability.alertManagerEnabled`. We register only when it's `true`. The companion change in [`alerting-dashboards-plugin`](https://github.com/opensearch-project/alerting-dashboards-plugin) does the mirror — it registers only when our capability is `false` or absent. Exactly one entry shows at any time. We had to defer registration because Explore's `queryPanelActionsRegistry` exposes only `getIsEnabled` (greys out) and no `isVisible` hook, so the only way to truly hide a duplicate is to skip `register(...)`. The companion alerting PR is required for the swap to work end-to-end.

7. **`ownsMonitorCreation`** on the setup contract is preserved at `true` for backwards compatibility with older alerting-plugin builds that still gate on it.

## Behavior

| `DynamicConfigService` | yml says | Dynamic flag | Result |
|---|---|---|---|
| Not configured | `false` | (n/a) | Hidden — yml wins |
| Not configured | `true` | (n/a) | Visible — yml wins |
| Not configured | unset | (n/a) | Hidden — schema default `false` |
| Configured | `false` | unset | Hidden — yml seed flows through |
| Configured | `true` | unset | Visible — yml seed flows through |
| Configured | (any) | `false` | Hidden — dynamic wins |
| Configured | (any) | `true` | Visible — dynamic wins |

Routes / SO types / services exist on every host regardless, so toggling either flag on yields a fully-working feature.

The Explore "Create alert rule" entry follows the same rule for the `alertManager` flag — when our capability is `true`, we own the menu entry; when `false`, alerting-dashboards-plugin's "Create monitor" entry shows instead (requires the companion alerting PR).

## Tests

Added/updated tests covering both states for both flags:

- `public/plugin.test.tsx` — alerting and SLO nav-link visibility paths (both hidden when capability is off and visible when on); merged-updater "either hiding wins" coverage; alerting app registers regardless of yml.
- `server/plugin.test.ts` — capability switcher: dynamic value overrides yml when both are set, falls back to yml when dynamic config omits a flag, falls back to yml when dynamic config layer rejects.
- `server/plugin_helper/register_settings.test.ts` — alerting datasource uiSettings register unconditionally.

22 plugin-level tests passing; 197 tests pass across the touched test directories.

## Files changed

- `server/plugin.ts`
- `server/routes/index.ts`
- `server/plugin_helper/register_settings.ts`
- `server/plugin_helper/register_settings.test.ts`
- `server/plugin.test.ts`
- `public/plugin.tsx`
- `public/plugin.test.tsx`

## Companion change

[`alerting-dashboards-plugin`](https://github.com/opensearch-project/alerting-dashboards-plugin/pull/1461) — coordinates the Explore "Create monitor" entry against this plugin's new `capabilities.observability.alertManagerEnabled`. **Required** for the menu-entry swap to work — without it, the alerting plugin's existing setup-time gate would still defer to our old `ownsMonitorCreation` boolean and the menu entry would not swap when the flag flips.

### Local / yml-only (no dynamic-config service)

1. `observability.alertManager.enabled: true` and `observability.slo.enabled: true` in `opensearch_dashboards.yml`.
2. `yarn start` — both nav entries visible. Apps mount, "Create alert rule" appears in Explore's actions menu.
3. Flip either flag to `false`, restart — corresponding nav link disappears. With the companion alerting PR, the Explore menu entry swaps to alerting's "Create monitor".

### With a `DynamicConfigService` configured

1. yml has both flags `true`.
2. Have the configured dynamic-config store return `${configPath}.slo.enabled: false` for a request.
3. Reload the page — SLO nav link hides without restart.
4. Flip the dynamic value back to `true` — reload, link reappears, app fully functional.

### Check List

### DIFF
--- a/public/plugin.test.tsx
+++ b/public/plugin.test.tsx
@@ -378,7 +378,11 @@ describe('#setup with Alert Manager feature gate', () => {
     expect(alertingApp).toBeDefined();
   });
 
-  it('should NOT register alerting app when alertManager.enabled yml flag is false', async () => {
+  it('should still register alerting app even when yml flag is false (visibility is dynamic-flag-controlled)', async () => {
+    // After the move to dynamic feature flags, yml no longer gates app
+    // registration — the dynamic capability does, via AppUpdater. The app
+    // always registers so an AppConfig override flipping the capability on
+    // takes effect without an OSD restart.
     const initializerContextMock = coreMock.createPluginInitializerContext();
     initializerContextMock.config.get = jest.fn().mockReturnValue({
       query_assist: { enabled: false },
@@ -401,7 +405,7 @@ describe('#setup with Alert Manager feature gate', () => {
 
     const registerCalls = (coreSetup.application.register as jest.Mock).mock.calls;
     const alertingApp = registerCalls.find((call) => call[0].id === 'observability-alerting');
-    expect(alertingApp).toBeUndefined();
+    expect(alertingApp).toBeDefined();
   });
 
   it('should hide alerting nav link when capabilities.observability.alertManagerEnabled is false in start()', async () => {
@@ -455,6 +459,63 @@ describe('#setup with Alert Manager feature gate', () => {
     expect(alertingUpdater$.getValue()()).toEqual({
       navLinkStatus: AppNavLinkStatus.hidden,
     });
+  });
+
+  it('should leave alerting nav link visible when capabilities.observability.alertManagerEnabled is true in start()', async () => {
+    // Counterpart to the hidden test above. When the capability is `true`
+    // we must NOT push a hidden updater — otherwise a regression that
+    // always fires `next(hidden)` would silently break the visible path.
+    const initializerContextMock = coreMock.createPluginInitializerContext();
+    initializerContextMock.config.get = jest.fn().mockReturnValue({
+      query_assist: { enabled: false },
+      summarize: { enabled: false },
+      alertManager: { enabled: true },
+    });
+    const coreSetup = coreMock.createSetup();
+    ensureIconSideNavMock(coreSetup);
+    let alertingUpdater$ = new BehaviorSubject<() => {}>(() => ({}));
+    coreSetup.application.register = jest.fn((args) => {
+      if (args.id === 'observability-alerting' && args.updater$) {
+        alertingUpdater$ = args.updater$;
+      }
+    });
+    const observabilityPlugin = new ObservabilityPlugin(initializerContextMock);
+
+    coreSetup.chrome.navGroup.getNavGroupEnabled.mockReturnValue(true);
+
+    await observabilityPlugin.setup(coreSetup, ({
+      embeddable: embeddablePluginMock.createSetupContract(),
+      visualizations: visualizationsPluginMock.createSetupContract(),
+      data: dataPluginMock.createSetupContract(),
+      uiActions: uiActionsPluginMock.createSetupContract(),
+      contentManagement: contentManagementPluginMocks.createSetupContract(),
+      dashboard: { registerDashboardProvider: jest.fn() },
+    } as unknown) as SetupDependencies);
+
+    const coreStart = coreMock.createStart();
+    const dataStartMock = dataPluginMock.createStartContract();
+    dataStartMock.dataSources.dataSourceFactory.registerDataSourceType = jest.fn();
+    observabilityPlugin.start(
+      {
+        ...coreStart,
+        application: {
+          ...coreStart.application,
+          capabilities: {
+            ...coreStart.application.capabilities,
+            observability: {
+              alertManagerEnabled: true,
+              sloEnabled: true,
+            },
+          },
+        },
+      },
+      ({ data: dataStartMock } as unknown) as AppPluginStartDependencies
+    );
+
+    // The default BehaviorSubject value is `() => ({})`. If `start()` did
+    // not call `next()` with a hidden updater, the value should still be
+    // the no-op updater.
+    expect(alertingUpdater$.getValue()()).toEqual({});
   });
 });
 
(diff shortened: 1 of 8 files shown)


lezzago:
Review notes — High + Medium severity

**High — Test coverage gap on the new feature-gate code paths**

The three new tests in `public/plugin.test.tsx` all exercise the hidden-by-capability path. Two gaps:

1. No "visible / no-op" test. Reference PR opensearch-project/OpenSearch-Dashboards#11220 covers both `enabled: true` and `enabled: false` for parity. Without it, a regression that always fires `next(hidden)` regardless of capability would pass CI.
2. The entire `core.capabilities.registerSwitcher` block in `server/plugin.ts` is untested. Worth at least:
   - A test that the switcher overrides yml when `dynamicConfigService` returns a value for `alertManager.enabled` / `slo.enabled`.
   - A test that the `try/catch` falls back to yml-derived values (not schema defaults) when `client.getConfig` rejects — the comment on line 519-521 calls this out as deliberate divergence from the reference PR, but nothing pins the behavior.

**Medium — Asymmetric flag semantics aren't surfaced to operators**

Server-side route/SO/service registration still gates on yml (`server/plugin.ts:513` `if (sloEnabled)`, same shape for alerting). So `slo.enabled: false` in yml plus a dynamic AppConfig flip to `true` will:
- show the nav link (capability-driven), but
- 404 / not-mount the SLO routes, since they were never registered at setup.

The PR description acknowledges "this only governs UI visibility," but nothing in-product communicates the asymmetry. Suggest one of:
- `logger.warn` in the switcher when the dynamic value tries to enable a flag yml has off,
- or a sentence in the `opensearch_dashboards.yml` block / AppConfig data-package README documenting that yml is the registration gate and AppConfig only controls visibility.

Either is fine; the ask is just to make this discoverable so operators don't silently land in a half-enabled state.

lezzago:
**High — Unhandled promise rejection on deferred Explore action registration**

`public/plugin.tsx`:

```ts
core.getStartServices().then(([coreStart]) => {
  const capabilities = coreStart.application.capabilities as
    | { observability?: { alertManagerEnabled?: boolean } }
    | undefined;
  if (!capabilities?.observability?.alertManagerEnabled) {
    return;
  }
  exploreSetup.queryPanelActionsRegistry.register({ ... });
});
```

No `.catch()` and no `try/await/catch` wrapper. If `getStartServices()` rejects (failed start lifecycle / shutdown / late teardown), the rejection becomes an unhandled promise warning — and worse, the failure is silent rather than logged.

Suggest:

```ts
core.getStartServices()
  .then(([coreStart]) => { ... })
  .catch((e) =>
    this.logger.error('Failed to register Explore Create alert rule action', e)
  );
```

Or convert the surrounding block to `async` and use `try/await/catch`.

TackAdam:
> **High — Unhandled promise rejection on deferred Explore action registration**
> 
> `public/plugin.tsx`:
> 
> ```ts
> core.getStartServices().then(([coreStart]) => {
>   const capabilities = coreStart.application.capabilities as
>     | { observability?: { alertManagerEnabled?: boolean } }
>     | undefined;
>   if (!capabilities?.observability?.alertManagerEnabled) {
>     return;
>   }
>   exploreSetup.queryPanelActionsRegistry.register({ ... });
> });
> ```
> 
> No `.catch()` and no `try/await/catch` wrapper. If `getStartServices()` rejects (failed start lifecycle / shutdown / late teardown), the rejection becomes an unhandled promise warning — and worse, the failure is silent rather than logged.
> 
> Suggest:
> 
> ```ts
> core.getStartServices()
>   .then(([coreStart]) => { ... })
>   .catch((e) =>
>     this.logger.error('Failed to register Explore Create alert rule action', e)
>   );
> ```
> 
> Or convert the surrounding block to `async` and use `try/await/catch`.

Good catch — fixed in the latest push.

  Added a `.catch()` on the deferred `getStartServices().then(...)` call so a rejection during a failed start lifecycle or late teardown surfaces in the console instead of becoming an unhandled-promise warning.

  I went with `console.error` rather than `this.logger.error` because this is browser-side code and OSD plugins don't expose a per-plugin logger to `public/`; it also matches the existing `console.error` precedent in this file on the S3-datasource registration path.

lezzago:
**Finding: `ownsMonitorCreation = true` breaks backward-compat with older alerting plugin builds**

`ownsMonitorCreation` is hardcoded to `true` in `public/plugin.tsx`:

```ts
const ownsMonitorCreation = true;
```

The older `alerting-dashboards-plugin` reads this value from the setup contract to decide whether to register its own "Create monitor" entry in the Explore actions menu — it only registers when `ownsMonitorCreation === false`. With this always `true`, an older alerting plugin will **never** register its own entry, even when `capabilities.observability.alertManagerEnabled` is `false`. In that scenario neither plugin registers the action and the menu entry disappears entirely for users.

The deferred registration in `getStartServices().then(...)` skips registration when the capability is off, but the older alerting plugin has already been told to stand down via `ownsMonitorCreation = true`. There's a window where neither plugin owns the entry.

**Suggestion:** Resolve this one of two ways:
1. Set `ownsMonitorCreation` based on the resolved capability value after `getStartServices()`, so the contract remains accurate at the time the alerting plugin reads it.
2. Document the minimum alerting plugin version that no longer keys off `ownsMonitorCreation`, and add a compatibility note to the PR so deployers know a coordinated upgrade is required.

TackAdam:
Good catch — addressed by removing `ownsMonitorCreation` from the setup contract entirely.

The companion alerting PR ([opensearch-project/alerting-dashboards-plugin#1461](https://github.com/opensearch-project/alerting-dashboards-plugin/pull/1461)) deletes the helper that read this field and stops destructuring `observabilityDashboards` from its setup dependencies. Once both PRs land, the field has zero readers — keeping it would simply be permanent dead weight in the public contract.

Changes in the latest push:

- Dropped the `ownsMonitorCreation` field from `ObservabilitySetup` in `public/types.ts`. Setup now returns `{}`.
- Removed the local declaration and explanatory comment block from `public/plugin.tsx`.
- Coordination is now exclusively driven by `capabilities.observability.alertManagerEnabled` — the dynamic capability published per request — which the new alerting PR reads inside its deferred `getStartServices().then(...)` registration path.

### Compatibility note

This PR is a coordinated change with the alerting PR linked above.

Deploying the new observability build alongside an older alerting build (one that still reads `ownsMonitorCreation`) will cause alerting to always register its own entry regardless of the capability value, since the field no longer exists in the setup contract.

This is the safer failure mode: a button is shown when it shouldn't be, rather than users seeing no button at all (the previous failure mode when `ownsMonitorCreation = true` was misaligned).

Once both PRs are deployed, both plugins coordinate exclusively through `capabilities.observability.alertManagerEnabled`, and ownership of the menu entry switches cleanly based on the published capability.

lezzago (approved):

joshuali925 (approved):

PR #2721, opened 2026-06-10
Add anomaly detection resources to Alert Manager

jackiehanyang:
### Description
In Alerts Manager, introducing anomalies on Alerts tab, and introducing Detectors and Forecasters on Rules tab

https://github.com/user-attachments/assets/0accaa98-d0a9-4279-9882-b9c546884b74

### Issues Resolved

### Check List
  - [x] New functionality has javadoc added


jackiehanyang:
Investigated the current failures and they do not appear to be caused by the code changes in this PR.

- `Lint` is passing.
- I also ran the changed-file lint and the affected unit tests locally; they pass (`9` suites / `104` tests).
- `Enforce PR labels` is failing because this PR needs a release label. I believe `feature` is the right label here.
- The E2E/FTR/SLO Cypress failures are happening before any tests run. The jobs fail while installing the OpenSearch observability backend artifact with `Archive is not a ZIP archive` for the snapshot artifact. This looks like a CI/artifact availability issue rather than a frontend/code failure.
- `Link Checker` is failing on existing OpenTelemetry links under integration docs, not files changed in this PR.
- The recently merged base PR #2719 also had a red E2E lane (`Dashboards observability plugin E2E test`), which further suggests this area of CI is currently flaky/unrelated to this PR’s Alert Manager changes.

I’ll keep watching the remaining build/test workflow, but based on the logs so far the actionable item for this PR is adding the release label and rerunning the flaky CI jobs.

lezzago:
## Code review — anomaly detection in Alert Manager

Thanks for this — it's a well-structured, defensively-coded feature with strong test coverage. Below are the findings from a review of the server-side linking logic and the new UI surfaces, grouped by severity. Line references are to the PR head.

### High

**1. Fuzzy anomaly→alert linking can attach the wrong entity's anomaly (`server/services/alerting/alert_service.ts:953-980`)**

`findRelatedAnomaly` has two paths. The explicit-id path (`:936-951`) matches on `anomaly_result_id` and is reliable. The fallback path (`:953-980`) filters candidate anomalies on **only** `detector_id` + a ±1h time-window overlap (`ALERT_ANOMALY_LINK_TOLERANCE_MS`), then picks the one nearest in time. There is no entity comparison — because the alert side has no entity to compare against: `osAlertToUnified` (`alert_utils.ts:259-278`) emits only `monitor_id` / `monitor_name` / `trigger_name`, whereas anomalies carry `labels.entity`.

For a **high-cardinality detector** where several entities go anomalous within the same hour (e.g. `DestCityName=Zurich`, `=London`, `=Osaka`), the monitor fires one aggregate alert and the fallback attaches whichever entity's anomaly is nearest in time — which can be the wrong one. The flyout's "Associated anomaly" context (grade/score/feature chart/heatmap highlight) would then describe the wrong root cause, with no visual cue that the link is approximate.

Two aggravating subtleties:
- Linking is greedy and order-dependent: `linkedAnomalyIds` removes each anomaly once claimed (`:919`) and alerts are processed in array order, so multiple alerts from one detector can cascade into a chain of mis-assignments.
- The mis-linked anomaly is then dropped from the standalone rows (`standaloneAnomalies = anomalies.filter(not linked)`, `:925`), so the table can show a standalone row for the *real* triggering entity alongside an alert flyout claiming a *different* entity.

This can't be fully fixed inside `findRelatedAnomaly` because the disambiguating information (which entity/result tripped the trigger) simply isn't present in the alert document today. Suggested directions:
- **Short term (this PR):** for high-cardinality detectors, when more than one entity matches within the window, prefer *not* linking (treat as ambiguous and show the anomaly as standalone) rather than picking nearest-by-time; and/or label the association as "Nearby anomaly (approximate)" instead of presenting it as the definitive trigger.
- **Future (alerting plugin PR):** persist the triggering anomaly's `anomaly_result_id` (and ideally the entity) into the alert/trigger-execution context at fire time, so the reliable explicit-id path (`:936-951`) is taken instead of the heuristic. That's the durable fix and would make this association exact rather than best-effort.

### Medium

**2. Alerts hot path adds sequential AD/monitor round-trips (`alert_service.ts:721-758`, `:802-902`)**

On `main`, the OpenSearch branch of `fetchAlertsRaw` issued a single `getAlerts` call. This PR adds, sequentially: `fetchADAnomalies` → (`fetchADDetectors` if anomalies present) → `getMonitors` → (`fetchADAnomaliesByIds` if explicit ids are missing). The detector-name map is correctly threaded/reused, so `fetchADDetectors` normally runs once — but the calls are serial. Consider `Promise.all` for the independent fetches (anomalies + detectors + monitors) to cut latency on every alerts load.

**3. Anomalies appear in both the Type and State facets (`public/components/alerting/alerts_dashboard.tsx:148`, facet wiring `:535/:561/:593`)**

`adAnomalyToUnified` hardcodes `state: 'active'` for every anomaly (`alert_utils.ts:339`), and `getAlertDisplayState` relabels that to `anomaly`. The result is that anomalies show as `anomaly` in the new Type facet *and* in the State facet — redundant and a bit confusing. Worth a product/UX call on whether the State facet should keep showing the true `active` state and leave the kind distinction to the Type facet.

**4. State column sorts on the raw field but renders the display value (`alerts_dashboard.tsx:758-768`)**

`field: 'state'` + `sortable: true` sorts on the raw `state` (anomalies = `active`), while `render` shows `getAlertDisplayState` (= `anomaly`). Sorted order won't match the displayed values. The severity column already uses a custom comparator (`sortable: (a) => SEVERITY_SORT_ORDER[...]`, `:663`) — the same approach would fix this.

**5. Hardcoded English strings in the new detail flyouts (`detector_detail_flyout.tsx`, `forecaster_detail_flyout.tsx`)**

The `EuiDescriptionList` `listItems` titles (17 in the detector flyout, 18 in the forecaster flyout) and the `Enabled`/`Disabled` / `High-cardinality` / `Single stream` value labels are raw string literals, while the column `name` fields in the same files are i18n'd. The sibling `monitor_detail_flyout.tsx` i18n's all 16 of its titles — so this diverges from the established local convention. Please wrap these in `i18n.translate` / `FormattedMessage`.

**6. AD fetch failures are swallowed at `debug` level (`alert_service.ts:742-746`, `:818-824`, `:879-885`, `:895-901`)**

Every AD/monitor fetch is wrapped in `try/catch` that only `logger.debug`s and then continues as if there were no anomalies. Graceful degradation is reasonable, but a genuine AD-index/transport failure never surfaces in the response's `error`/`warnings` envelope, so the UI silently renders alerts without their anomaly context and the user can't tell anything failed. Consider surfacing a non-fatal warning.

**7. Anomaly-scan truncation isn't reflected in the `truncated` flag (`alert_service.ts:757`)**

The returned `truncated` propagates only the alert query's flag. `fetchADAnomalies` caps at `DEFAULT_MAX_RESULTS` (5000) with no truncation signal, so if more than 5000 anomalies fall in range they're silently dropped and the UI "results may be truncated" callout won't fire.

**8. Standalone-flyout navigation forces a full page reload (`anomaly_detail_flyout.tsx:512-516`)**

The `AnomalyDetailContent` callsite (`:250-256`) correctly prefers an `onNavigateToDetectorResults` SPA override and only falls back to `window.location.assign`. The standalone `AnomalyDetailFlyout` (`:512-516`) calls `window.location.assign` unconditionally — a full browser reload that drops SPA state. (Not a security concern; the href is built from `window.location.pathname`.)

---

Happy to pair on the linking mitigation or any of the Medium items.

jackiehanyang:
Thanks for the detailed review. I went through the eight findings and addressed them as follows:

| # | Review comment | Resolution |
|---|---|---|
| 1 | Fuzzy anomaly-to-alert linking can attach the wrong high-cardinality entity anomaly. | Added a short-term ambiguity guard in the fallback linker. Exact `anomaly_result_id` links remain authoritative, but detector/time-window fallback now skips linking when multiple high-cardinality entity values match the same alert window instead of picking nearest-by-time. Ambiguous anomalies remain as standalone rows, which avoids showing the wrong entity as the alert's associated anomaly. Added regression coverage for ambiguous HC anomalies. |
| 2 | Alerts hot path adds sequential AD/monitor round trips. | Changed the OpenSearch alerts fetch path to use parallel `Promise.allSettled` fetches for alerts, AD anomalies, AD detectors, and monitors where possible. Explicit anomaly-result lookups still happen only when needed, after monitor/query inspection. Non-fatal AD/monitor failures are collected instead of blocking the base alert list. |
| 3 | Anomalies appear in both Type and State facets. | We discussed this with product/UX. The chosen behavior is to keep anomaly rows displayed as `anomaly` in State, while also keeping a sidebar Type filter for `alert` vs `anomaly` so users can quickly isolate result kinds. The Type table column remains removed to keep the All Alerts table lighter. I also hid noisy/internal labels such as `anomaly_result_id` and `source` from the Labels facet. |
| 4 | State column sorts on raw `state` but renders the display value. | Updated the State column sorting to use `getAlertDisplayState`, so anomalies sort under the displayed `anomaly` value instead of their raw `active` state. |
| 5 | Hardcoded English strings in detector/forecaster detail flyouts. | Wrapped detector and forecaster flyout description-list titles and value labels in `i18n.translate` / localized labels, matching the existing monitor flyout convention. |
| 6 | AD fetch failures are swallowed at debug level. | Kept graceful degradation, but now non-fatal AD/monitor fetch failures are returned in the datasource status `error` field and surfaced by the existing “Some datasources could not be reached” warning path. Added UI coverage for partial-success datasource warnings. |
| 7 | Anomaly-scan truncation is not reflected in the `truncated` flag. | Changed AD anomaly fetching to request `DEFAULT_MAX_RESULTS + 1`, trim to the display cap, and propagate truncation into datasource status / the overall truncated warning path. Added regression coverage for AD anomaly truncation. |
| 8 | Standalone anomaly flyout navigation forces a full reload. | Added an optional navigation callback for the standalone anomaly flyout and wired the Alerts page to use `core.application.navigateToUrl` when available, with `window.location.assign` only as a fallback. |

jackiehanyang:
I checked the current failed CI jobs on this PR. Most of the failures do not appear to be caused by this PR’s Alert Manager changes.

| Check | Status / failure | Relation to this PR |
|---|---|---|
| Code-Diff-Analyzer | Flags new imports of `plotly.js-dist` and `react-plotly.js/factory` in `anomaly_detail_flyout.tsx`, plus a low-severity note about the `local_cluster` fallback. | The Plotly findings are false positives: both packages are already declared in `package.json` and are already used elsewhere in this plugin, for example `public/components/visualizations/plotly/plot.tsx`. No new dependency is introduced by this PR. The `local_cluster` fallback is intentional backwards compatibility and is only used after normal datasource saved-object resolution misses, so a registered datasource with that id still wins. This behavior is covered by tests. |
| panels_test | Cypress failure in Panels resize/toast assertions, including `expected 630 to match /470/`. | Unrelated. The failure is in the Panels Cypress spec, not Alert Manager or files changed by this PR. |
| trace_analytics_test | Cypress failure clicking the date picker refresh button because it is hidden/covered by the fixed header. | Unrelated. This is in Trace Analytics date picker interaction, outside this PR’s changed files. |
| build-linux | Jest failures in existing `server/plugin.test.ts` and `server/plugin_helper/register_settings.test.ts` due to `Cannot find module '@hapi/hoek/assert'`. | Unrelated. This is a dependency/module resolution issue in the existing plugin/core Hapi dependency chain, not caused by the Alert Manager changes. |
| FTR E2E Dashboards Observability Tests | Sample flight data install timed out, followed by Trace Analytics selectors/tables not appearing. | Unrelated. The failures are setup/Trace Analytics related and do not exercise the Alert Manager changes in this PR. |

The relevant validation for this PR is passing: `Lint`, `linkchecker`, labels/DCO, and the focused local Jest suites for the touched Alert Manager/server alerting files all passed.

lezzago:
Two findings from a follow-up review (in addition to the 8 already addressed — those all look good):

### Medium — the #6 fix will surface false "datasource unreachable" warnings on clusters without AD/Forecasting

`alertsWarnings` now triggers on `s.status === 'error' || !!s.error` (`public/components/alerting/alarms_page.tsx`), and `fetchAlertsRaw` pushes into `partialErrors` whenever `fetchADAnomalies` / `fetchADDetectors` / `getMonitors` reject (`server/services/alerting/alert_service.ts`). Those AD/forecast transport calls run on **every** alerts load and will reject with `404` / `index_not_found_exception` on a standard OpenSearch cluster that doesn't have the anomaly-detection plugin installed.

Net effect: a healthy non-AD datasource permanently shows *"Some datasources could not be reached → Failed to fetch anomaly results"*. Since AD is an optional plugin, this is the common case rather than an edge case.

Suggestion: treat "AD plugin/index absent" as an expected non-condition — swallow `404` / `index_not_found_exception` (and the forecast equivalent) at `debug` level as before, and only surface `partialErrors` for genuine transport/auth failures. The existing test throws a generic `Error('AD unavailable')`, so it won't catch this distinction — worth adding a test with a 404-shaped error asserting that **no** warning is surfaced.

### Low — `getRuleDetail` now does up to 3 sequential round-trips per detail open

`getRuleDetail` chains `getOSRuleDetail → getADDetectorDetail → getADForecasterDetail` (`server/services/alerting/alert_detail.ts`). Opening a forecaster's flyout issues a monitor GET (404), a detector GET (404), then the forecaster GET — two wasted 404s per open. It's acceptable for a user-initiated detail open, but since `definitionType` / `monitorType` is already known on the summary, you could branch directly to the correct fetch and skip the misses.

jackiehanyang:
> Two findings from a follow-up review (in addition to the 8 already addressed — those all look good):
> 
> ### Medium — the #6 fix will surface false "datasource unreachable" warnings on clusters without AD/Forecasting
> `alertsWarnings` now triggers on `s.status === 'error' || !!s.error` (`public/components/alerting/alarms_page.tsx`), and `fetchAlertsRaw` pushes into `partialErrors` whenever `fetchADAnomalies` / `fetchADDetectors` / `getMonitors` reject (`server/services/alerting/alert_service.ts`). Those AD/forecast transport calls run on **every** alerts load and will reject with `404` / `index_not_found_exception` on a standard OpenSearch cluster that doesn't have the anomaly-detection plugin installed.
> 
> Net effect: a healthy non-AD datasource permanently shows _"Some datasources could not be reached → Failed to fetch anomaly results"_. Since AD is an optional plugin, this is the common case rather than an edge case.
> 
> Suggestion: treat "AD plugin/index absent" as an expected non-condition — swallow `404` / `index_not_found_exception` (and the forecast equivalent) at `debug` level as before, and only surface `partialErrors` for genuine transport/auth failures. The existing test throws a generic `Error('AD unavailable')`, so it won't catch this distinction — worth adding a test with a 404-shaped error asserting that **no** warning is surfaced.
> 
> ### Low — `getRuleDetail` now does up to 3 sequential round-trips per detail open
> `getRuleDetail` chains `getOSRuleDetail → getADDetectorDetail → getADForecasterDetail` (`server/services/alerting/alert_detail.ts`). Opening a forecaster's flyout issues a monitor GET (404), a detector GET (404), then the forecaster GET — two wasted 404s per open. It's acceptable for a user-initiated detail open, but since `definitionType` / `monitorType` is already known on the summary, you could branch directly to the correct fetch and skip the misses.

did the following fixes for these comments:

| findings | Fix |
|---|---|
| Optional AD/Forecasting failures can show false datasource warnings when AD/Forecasting is not installed or result indices are missing. | Suppressed expected optional-plugin/index-missing errors (`404`, `index_not_found_exception`, “no handler found”) from `partialErrors`, while still logging them at debug level. Real transport/auth/runtime failures still surface as warnings. Added regression coverage for the optional AD missing case. |
| Rule detail loading probes monitor → detector → forecaster sequentially, causing unnecessary failed requests for detector/forecaster flyouts. | Added an optional `definitionType` hint from the UI through the route/handler/service/detail resolver so known monitor, detector, and forecaster rows fetch the correct detail endpoint directly. Kept the old fallback chain for callers without a hint. Added tests for direct detector/forecaster routing and frontend hint forwarding. |

jackiehanyang:
one unrelated flaky test failed caused the CI failure

lezzago (approved):

riysaxen-amzn:
confirm the enhancement label is attached (it is) and verify the remaining 2/21 CI checks clear or are confirmed flaky before merging.

riysaxen-amzn (approved):

TackAdam (approved):

PR #2726, opened 2026-06-12
[APM] Include Prometheus connection meta in all PromQL chart queries

ps48:
### Description

The `usePromQLChartData` hook accepted an optional `prometheusConnectionMeta` parameter, but no caller ever supplied it. As a result, every chart that routes through this hook sent the query-enhancement request (`POST /api/enhancements/search/promql`) with an empty `dataSource.meta`:

- Service overview metric cards and line charts (including fault-rate and error-rate)
- Operations tab charts
- Dependencies tab charts
- Application-map / service-map line charts

By contrast, the service-map metric hooks (e.g. `useServiceMapMetrics`) already sourced the connection meta from the APM config context and passed it to `PromQLSearchService`. This is why the service-map doughnut charts populated `meta` correctly while the paths above did not.

This change sources the Prometheus connection meta from `useApmConfig()` inside `usePromQLChartData`, matching the existing working pattern, so all chart paths send consistent meta. An explicitly passed `prometheusConnectionMeta` still takes precedence as an override.

The fix is contained to a single hook — no prop-drilling or component-API changes were needed, since all affected charts consume meta transitively through the hook.

### Issues Resolved
Prometheus queries fail to load when the underlying datasource needs the meta fields for query execution. 

### Check List

### DIFF
--- a/.github/workflows/dashboards-observability-test-and-build-workflow.yml
+++ b/.github/workflows/dashboards-observability-test-and-build-workflow.yml
@@ -8,7 +8,7 @@ env:
 
 jobs:
   Get-CI-Image-Tag:
-    uses: opensearch-project/opensearch-build/.github/workflows/get-ci-image-tag.yml@c2498b758c08fb7bc48476509a5fc1b8dd5f7493 # main
+    uses: opensearch-project/opensearch-build/.github/workflows/get-ci-image-tag.yml@main
     with:
       product: opensearch-dashboards
 
--- a/public/components/apm/shared/hooks/__tests__/use_dependency_metrics.test.ts
+++ b/public/components/apm/shared/hooks/__tests__/use_dependency_metrics.test.ts
@@ -5,6 +5,8 @@
 
 import { renderHook, waitFor } from '@testing-library/react';
 import { useDependencyMetrics } from '../use_dependency_metrics';
+import { PromQLSearchService } from '../../../query_services/promql_search_service';
+import { useApmConfig } from '../../../config/apm_config_context';
 import { GroupedDependency } from '../../../common/types/service_details_types';
 
 // Mock the PromQLSearchService
@@ -12,6 +14,14 @@ const mockExecuteInstantQuery = jest.fn();
 jest.mock('../../../query_services/promql_search_service', () => ({
   PromQLSearchService: jest.fn().mockImplementation(() => ({
     executeInstantQuery: mockExecuteInstantQuery,
+  })),
+}));
+
+// Mock the APM config context — the hook sources the Prometheus connection meta from it.
+const mockConfigMeta = { key: 'context-meta-value' };
+jest.mock('../../../config/apm_config_context', () => ({
+  useApmConfig: jest.fn(() => ({
+    config: { prometheusDataSource: { meta: mockConfigMeta } },
   })),
 }));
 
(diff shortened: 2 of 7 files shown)


TackAdam (approved):
LGTM. Mirrors the working pattern used by the other APM PromQL hooks, override semantics are preserved with ??, all usePromQLChartData callers are already inside ApmConfigProvider, and existing component/SLO tests are unaffected because they mock the hook directly.

  Minor observations (non-blocking)

  1. useMemo([prometheusConnectionId, resolvedMeta]) will re-create PromQLSearchService whenever config re-renders with a new reference (e.g. refresh() from the provider). Not a correctness issue and matches the sibling hooks, but worth being aware of if you ever see redundant fetches.
  2. The new "override" test (use_promql_chart_data.test.ts:499-510) only asserts the service constructor args; it doesn't independently confirm that the override path bypasses the context. It's fine because the assertion is exact, but a defensive belt-and-suspenders test could swap context meta to a sentinel and assert the constructor wasn't called with it.
  3. No new caller is using the explicit override now, and the comment at use_promql_chart_data.ts:93-95 correctly documents the precedence — keeping the param around is reasonable as a future escape hatch.

ps48:
### CI status note

This PR adds the meta-sourcing fix plus a small CI fix (`Get-CI-Image-Tag`). A few checks are red for reasons **pre-existing on `main`**, unrelated to this change:

- **`Get-CI-Image-Tag`** — fixed here by bumping the `opensearch-build/get-ci-image-tag.yml` reusable workflow from the old pinned SHA (which referenced `crane-installer@v1` / `checkout@v6` as unpinned tags) to `@main`, where those inner actions are now SHA-pinned. (Same failure occurs on `main`.)

- **`build-windows-macos` (alerting schema tests)** — `server/routes/alerting/__tests__/{schema_helpers,mutations}.test.ts` length-validation assertions fail only on macOS/Windows under `--maxWorkers=100%`. This is an order-dependent test-isolation issue in `@osd/config-schema`'s `.custom()`-based length validation, and it reproduces on `main` (e.g. the latest `main` Test-and-Build run also fails these). It does not reproduce on Linux or locally. This PR does not touch any alerting code, so it's out of scope here — flagging for the alerting owners.

- **`panels_test` (Cypress)** — also failing on `main`; unrelated to this change (no panels code touched).

The meta change itself is fully green: unit tests for the three affected hooks pass, and the full plugin suite passes locally (`yarn test --coverage --maxWorkers=100%` → 3168 passed).

.github/workflows/dashboards-observability-test-and-build-workflow.yml
```diff
@@ -8,7 +8,7 @@ env:
 
 jobs:
   Get-CI-Image-Tag:
-    uses: opensearch-project/opensearch-build/.github/workflows/get-ci-image-tag.yml@c2498b758c08fb7bc48476509a5fc1b8dd5f7493 # main
```
joshuali925, line 11: i thought infra wanted to use pinned sha?

joshuali925 (approved):

.github/workflows/dashboards-observability-test-and-build-workflow.yml
ps48: Their PR fix wasn't merged in the sha, they added in. Hence, they recommended to use main for the ci-image-tag https://github.com/opensearch-project/opensearch-build/pull/6253. 

Original PR thread: https://github.com/opensearch-project/opensearch-build/pull/6211#pullrequestreview-4488256314

PR #2727, opened 2026-06-12
feat(apm): exclude AnalyticEngine datasets from APM settings selectors

ps48:
## Summary

The APM settings/config modal's **traces** and **service-map** dataset selectors list every index pattern (dataset). AnalyticEngine (Mustang) data sources support PPL/SQL but **not DSL queries against user data** — and the APM traces / service-map views run DSL aggregations. So a dataset created on top of a Mustang domain would 5xx if a user selected it in either selector.

This hides datasets backed by an AnalyticEngine data source from **both** APM dataset selectors, so users can't pick a connection that will fail.

This is a follow-up to #2707, which excluded AnalyticEngine from the **trace analytics**, **metrics**, and **integrations** data-source pickers. That PR covered data-source *pickers*; this one covers the APM *dataset* selectors (which list index patterns, not data sources directly).

## Changes

- **`common/utils/shared.ts`** — add `getUnsupportedEngineDataSourceIds(dataSourceIds)`: given a set of `data-source` saved-object ids, returns the subset whose `dataSourceEngineType` is in the manifest's `unsupportedOSDataSourceEngineTypes` (`AnalyticEngine`). Resolves engine type with a single `bulkGet`; **fails open** (returns an empty set) if the lookup throws, so a transient error never hides datasets. Driven by the same manifest field #2707 introduced — single source of truth.
- **`public/components/apm/shared/hooks/use_apm_config.ts`** — in `useDatasets`, track each dataset's backing data-source id (from the `DataView`'s `dataSourceRef`), then filter both `tracesDatasets` and `allDatasets` to drop AnalyticEngine-backed datasets. The traces selector binds to `tracesDatasets`, the service-map selector binds to `allDatasets`, so both are covered. The existing sequential fetch loop is preserved; only one extra `bulkGet` is added after it.

## Why a dataset-level filter (not a picker filter)

The APM modal doesn't render a `DataSourceSelectable`; it lists **datasets** (`index-pattern` saved objects) via the `data` plugin's `dataViews` service. So the exclusion has to resolve each dataset's backing data source and check its engine type — mirroring the core `data` plugin's `IndexPatterns.applyEngineTypeFilter` (OpenSearch-Dashboards#12023), but applied at the consumer because this hook reads `DataView`s rather than raw saved objects.

## Testing

- `useDatasets` unit tests extended (`use_apm_config.test.ts`) — **30/30 pass**, including 3 new cases:
  - drops AnalyticEngine-backed datasets from both `tracesDatasets` and `allDatasets`
  - keeps all datasets when the data-source lookup fails (fail-open)
  - skips the `bulkGet` entirely when no dataset has a backing data source (local-cluster datasets)
- No behavior change for non-AnalyticEngine data sources or local-cluster datasets.

## Check List
- [x] All tests pass
- [x] New functionality has been documented in code comments
- [x] Commit changes are listed out in CHANGELOG.md file (n/a — follows #2707 convention)

### DIFF
--- a/common/utils/shared.ts
+++ b/common/utils/shared.ts
@@ -99,3 +99,42 @@ export const dataSourceFilterFnByEngineType = (dataSource: SavedObject<DataSourc
 export const dataSourceFilterFnExcludeAnalyticEngine = (
   dataSource: SavedObject<DataSourceAttributes>
 ) => dataSourceFilterFnByEngineType(dataSource) && dataSourceFilterFn(dataSource);
+
+// Given a set of data-source saved-object ids, returns the subset whose
+// `dataSourceEngineType` is in `unsupportedOSDataSourceEngineTypes` (AnalyticEngine /
+// Mustang). Callers that list datasets (index patterns) bound to a data source — e.g. the
+// APM settings modal's traces / service-map selectors — use this to drop datasets that sit
+// on a Mustang domain, since those serve PPL/SQL but not the DSL aggregations APM issues.
+//
+// Resolves engine type with a single `bulkGet` on `data-source` saved objects. Fails open:
+// if the lookup throws, an empty set is returned so callers keep showing every dataset.
+export const getUnsupportedEngineDataSourceIds = async (
+  dataSourceIds: string[]
+): Promise<Set<string>> => {
+  const unsupported = new Set<string>();
+  const ids = Array.from(new Set(dataSourceIds.filter(Boolean)));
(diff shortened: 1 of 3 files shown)


joshuali925 (approved):

TackAdam (approved):

joshuali925:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

PR #2728, opened 2026-06-12
Fix: clear stale histogram and patterns when switching to a stats query

sejli:
### Description
PR #2695 added a `!hasStats` guard around `getCountVisualizations`, `setLogPattern`, and `getPatterns` so stats queries no longer fire malformed sub-queries. The unguarded path was what previously kept the `count_distribution` and `patterns` redux slices in sync with the active query — for stats queries those slices are now never refreshed, so the histogram and patterns panels render stale data from the prior non-stats query. This change dispatches `reset` on both slices in the `hasStats` branch.

### Issues Resolved
N/A

### Check List

### DIFF
--- a/public/services/data_fetchers/ppl/ppl_data_fetcher.ts
+++ b/public/services/data_fetchers/ppl/ppl_data_fetcher.ts
@@ -25,6 +25,8 @@ import {
   composeFinalQueryWithoutTimestamp,
   getDescribeQueryIndexFromRawQuery,
 } from '../../../components/common/query_utils';
+import { reset as resetCountDistribution } from '../../../components/event_analytics/redux/slices/count_distribution_slice';
+import { reset as resetPatterns } from '../../../components/event_analytics/redux/slices/patterns_slice';
 import { DataFetcherBase } from '../fetcher_base';
 import { IDataFetcher } from '../fetch_interface';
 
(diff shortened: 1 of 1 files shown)


TackAdam (approved):

public/services/data_fetchers/ppl/ppl_data_fetcher.ts
```diff
@@ -144,7 +146,11 @@ export class PPLDataFetcher extends DataFetcherBase implements IDataFetcher {
     }
     // still need all fields when query contains stats
     const hasStats = finalQuery.match(PPL_STATS_REGEX);
-    if (hasStats) getAvailableFields(`search source=${this.queryIndex}`);
+    if (hasStats) {
+      getAvailableFields(`search source=${this.queryIndex}`);
+      dispatch(resetCountDistribution({ tabId }));
+      dispatch(resetPatterns({ tabId }));
```
ps48, line 152: Will this hide the patterns and count distribution histogram when users have `stats` in query?

sejli: It will hide count distribution and histogram, but patterns is always rendered unconditionally. Do we want to add something that hides patterns when `stats` is in the query?

ps48: empty table for patterns should be fine

ps48 (approved):

PR #2746, opened 2026-06-18
[Alerting] Allow colons in rule-detail ruleId path param

lezzago:
### Description

The rule-detail route `/api/alerting/rules/{dsId}/{ruleId}` validated `ruleId` with `alertingIdSchema` (pattern `/^[A-Za-z0-9_-]+$/`), which rejects colons. Prometheus SLO rule IDs are `{dsId}-{groupName}-{ruleName}`, and the group/rule names use the `slo:rec:` / `slo:alerts:` naming convention — so the composite id legitimately contains colons.

As a result the route returns **400 `[request params.ruleId]: must match /^[A-Za-z0-9_-]+$/`**, and the rule-detail flyout fails with *"Some rule details could not be loaded — showing summary information only. Try reopening the flyout to retry."*

#### Fix

Add a dedicated `alertingRuleIdSchema` (pattern `/^[A-Za-z0-9_:-]+$/`) and use it only for the rule-detail route. `:` is the only character added over the strict ID pattern; slash, dot, and the rest stay disallowed, so the path-traversal guard is unchanged. The composite id is only used for in-memory equality matching against the fetched rule list (`getPromRuleDetail`), never interpolated into an upstream URL, so even the colon never reaches a transport path. `alertingIdSchema` is unchanged for traversal-sensitive params (datasource ids, etc.).

### Issues Resolved

N/A

### Check List
- [x] All tests pass (`schema_helpers` suite).
- [x] New functionality includes testing (added cases to `schema_helpers.test.ts` covering colon-bearing composite ids and continued rejection of slash/dot/null-byte).


riysaxen-amzn:
Minor / optional

- **Max length 512**: Inherited from `ID_MAX_LENGTH` -  the real-world composite IDs (`{uuid}-slo:alerts:{name}-{name}`) could plausibly approach this. Are there known cases close to the limit? If not, 512 seems fine as-is.

riysaxen-amzn (approved):

TackAdam (approved):

ps48 (approved):

PR #2747, opened 2026-06-18
[SLO/Alerting] Fix SLO status incorrectly degrading to no-data

lezzago:
### Description

This PR fixes three related defects that cause SLO status to incorrectly degrade to **"no data"** on the SLO listing. Each is a provider-agnostic correctness bug in the status-aggregation path; the commits are split so they can be reviewed independently. (They all touch `status_aggregator.ts`, so they're grouped into one PR rather than three conflicting ones.)

#### 1. Don't let an optional alerts-fetch failure trip the datasource circuit breaker
The per-datasource `DatasourceCircuitBreaker` was driven off the optional firing-alerts fetch (`fetchFiringAlertsBySlo`), which uses a different transport than the PromQL status read. The alerts endpoint can be unreachable while `/query` is fine. When the alerts fetch failed it called `recordFailure`, opening the breaker and short-circuiting **every** SLO on that datasource to `no_data` — even when the PromQL status reads were succeeding.

Fix: drive the breaker off the PromQL **read** outcome (the path that actually determines health). A successful `statusForSlo` read resets the breaker; only an all-reads-threw group records a failure. A failing alerts fetch alone now only zeroes the firing-alert count. Adds a regression test.

#### 2. Don't let `ruler_unreachable` mask a real sample-derived SLO state
`applyRuleHealthMerge` overlaid rule-health on top of the sample-derived status and, on `ruler_unreachable`, unconditionally mapped the status to `no_data` — clobbering a correct `ok` / `warning` / `breached` derived from real samples. The health checker and the PromQL read can travel different transports, so a transient health-check failure could wrongly downgrade a correctly-read status.

Fix: only map `ruler_unreachable` → `no_data` when there is no sample data to preserve (new `baseHasData()` helper). If the SLI read returned real samples, the ruler is demonstrably reachable for reads, so keep the sample-derived state. `rules_missing` (a definitive "rules deleted" signal) still overlays. Adds tests for both paths.

#### 3. Forward the inbound request on server-initiated PromQL reads
Server-initiated PromQL reads build a *synthetic* search request to hold the PromQL body but did not forward the originating inbound request. The data plugin's `promql` search strategy does `client.asScoped(request)`, so the scoped client derives the caller's auth from that request. Without it, an auth-enabled datasource read fails and is swallowed — surfacing as `no_data` on the SLO listing and as a silent fallback in the rule-detail condition preview.

Fix: thread the inbound `OpenSearchDashboardsRequest` opaquely (`sourceRequest`) through every server-initiated read path — SLO status aggregator `queryInstant`, probe-sli `queryInstant`/`queryRange`, rule-detail preview `fetchPromPreviewData` `queryRange`, and historical-alerts `getHistoricalAlerts` → `queryRangeMatrix`. `buildSearchRequest` spreads the forwarded request and overrides only `body`/`dataSourceId`, so all auth context survives. The request is never inspected by this plugin — whatever the deployment's datasource client reads off it travels untouched — so the change is auth-mechanism-agnostic. Callers with no request (offline/tests) omit it.

### Issues Resolved

N/A

### Check List
- [x] All tests pass (`status_aggregator`, `status_aggregator_circuit_breaker`, `handlers`, `directquery_prometheus_backend` suites).

### DIFF
--- a/server/services/alerting/promql_search.ts
+++ b/server/services/alerting/promql_search.ts
@@ -81,18 +81,23 @@ export interface PromQLSearchOptions {
   timeoutSeconds?: number;
   /**
    * The originating inbound request, forwarded opaquely so the `promql`
-   * strategy's `client.asScoped(request)` produces a scoped client carrying
-   * the caller's auth. Server-initiated reads (SLO status aggregator,
-   * probe-sli, alert preview, historical-alert matrix) construct a *synthetic*
-   * request to hold the PromQL `body`; the synthetic request alone has none of
-   * the auth context a scoped client needs. We don't inspect this — whatever
-   * the datasource client reads off the request (auth headers, `request.auth`,
-   * etc.) travels untouched. Forwarding the whole request (rather than
-   * cherry-picking `headers`) keeps every field the client might need present,
-   * so we don't regress as that client evolves. `buildSearchRequest` spreads
-   * this and overrides only
-   * `body` / `dataSourceId`. Omit for callers (tests / non-scoped deployments)
-   * that have no request to forward — `asScoped` accepts a header-less fake.
+   * strategy's datasource client can derive the caller's auth from it.
+   * Server-initiated reads (SLO status aggregator, probe-sli, alert preview,
+   * historical-alert matrix) construct a *synthetic* request to hold the
+   * PromQL `body`; the synthetic request alone has none of the inbound auth
+   * context. We don't inspect this — whatever the datasource client reads off
+   * the request (auth `headers`, `request.auth`, etc.) travels untouched.
+   *
+   * Note: `buildSearchRequest` shallow-spreads this into a plain object, so the
+   * result is not an `instanceof OpenSearchDashboardsRequest`. The request's
+   * own-enumerable fields (`headers`, `auth`, …) are preserved, which is what
+   * the DirectQuery datasource client reads; but core's `asScoped` treats a
+   * non-branded request as header-less and won't re-run `getAuthHeaders`. That
+   * is acceptable here: the forwarded raw `headers`/`auth` are what the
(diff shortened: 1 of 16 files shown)


server/services/alerting/promql_search.ts
```diff
@@ -235,7 +269,18 @@ function buildSearchRequest(
...
+  // Spread the originating request first so all of its scoped-client context
+  // (`headers`, `auth`, and anything else the deployment's client wrapper
+  // reads) is present, then override `body` / `dataSourceId` with the
+  // synthetic PromQL payload. The strategy reads `body` + `dataSourceId`; the
+  // DirectQuery client wrapper reads `headers` + `auth` off the same object.
+  // When no source request is supplied (tests / non-scoped deployments) the
+  // result is just `{ body, dataSourceId? }` — `asScoped` tolerates a
+  // header-less fake request.
+  const request: { dataSourceId?: string; body: Record<string, unknown>; [k: string]: unknown } = {
+    ...((extras.sourceRequest as unknown) as Record<string, unknown>),
```
riysaxen-amzn, line 281: Minor: This spreads all enumerable own properties of the inbound Hapi request object (which can include payload, raw, server, etc.). In practice the strategy only reads headers/auth/body/dataSourceId, so the extra fields are inert. But if you want to be defensive against future strategy changes or reduce the object size, you could whitelist:

```
const { headers, auth } = (extras.sourceRequest ?? {}) as any;
const request = { ...(headers && { headers }), ...(auth && { auth }), body };

```

Not a blocker — the override tests pin the contract either way — just a thought for long-term maintenance.

riysaxen-amzn (approved):

TackAdam (approved):

ps48 (approved):

PR #2755, opened 2026-06-23
Bugfix Observability SLOs

riysaxen-amzn:
### Description

Fixes high-severity bugs identified during the  bug bash for Unified Alerts View and SLOs.

* Remove Observability breadcrumb from SLO pages - Removed the parent "Observability" breadcrumb from all SLO pages (listing, wizard, detail, suggest). SLO pages now show just their own breadcrumb path.

* "Change Template" link in Create SLO not working - When clicking "Change template" in the SLO wizard, the template state wasn't being cleared on navigation back to `/slos/create`. Fixed by clearing `state.templateId` when `urlTemplateId` becomes falsy.

* Remove beta icon from Alerts page - Removed the beaker/beta badge that was displayed in the chrome bar on the Alerts page.

* Error loading Alertmanager config (Internal Server Error) on Routing page - Wrapped the Alertmanager config handler in a try-catch that returns a graceful `{ available: false, error: message }` response instead of crashing with a 500.

### Issues Resolved

-  Remove Observability breadcrumb from SLO page
- Change Template link in Create SLO not working
- Remove beta icon
- Error loading Alertmanager config

### Check List


TackAdam:
* Hide "Creation Method" as template is not supported yet

Review: Changes requested 🛑

Thanks for the fixes! The breadcrumb removal and the Alertmanager try-catch look good. However, there are two blocking server-side defects that will prevent the plugin from building/starting, plus a couple of cleanup items. Details below.

## 🔴 Blocking

### 1. `server/routes/index.ts:84-134` — Alerting routes are registered twice (server throws on startup)

The new gated block is immediately followed by the old unconditional registration, so `registerAlertingRoutes(...)` runs twice and re-registers all 19 routes.

OSD's router throws on duplicate path registration → plugin fails to boot whenever `alertManagerEnabled` is true.

The trailing comment ("Alerting routes register unconditionally…") also now contradicts the gated block — looks like a merge resolution that kept both copies.

**Fix:** Delete the second, unconditional block (lines ~108-134) and keep only the gated one:

```ts
if (alertManagerEnabled) {
  const osBackend = new HttpOpenSearchBackend(logger);
  const promBackend = new DirectQueryPrometheusBackend(logger);
  const mutationSvc = new MonitorMutationService(osBackend, logger);

  // eslint-disable-next-line @typescript-eslint/no-var-requires
  const { DirectQueryRulerClient } = require('../services/slo/ruler_client');

  const rulerClient = new DirectQueryRulerClient(logger);

  registerAlertingRoutes(router, {
    osBackend,
    promBackend,
    mutationSvc,
    logger,
    rulerClient,
  });
}
```

(and remove the now-stale "register unconditionally" comment block).

---

### 2. `server/plugin.ts:386` — `setupRoutes` called without the now-required `alertManagerEnabled` (type error)

`setupRoutes` now requires `alertManagerEnabled: boolean` (`server/routes/index.ts:47`), but the call site still omits it, which fails `yarn typecheck` / the build.

The commit message for `b558186b` says it "pass[es] true from plugin.ts," but that commit only modified `routes/index.ts`.

**Fix:** Pass the flag from the call site (use the resolved value, not a hardcoded `true`):

```ts
setupRoutes({
  router,
  client: openSearchObservabilityClient,
  dataSourceEnabled,
  alertManagerEnabled: config.alertManager?.enabled ?? ymlAlertManagerEnabled,
  logger: this.logger,
});
```

---

## 🟡 Cleanup (non-blocking)

### 3. `public/components/alerting/monitors_table/index.tsx:213-231` — Duplicate `useEffect`

The prune-by-filtered effect appears twice (lines 213-221 and 223-231), byte-for-byte identical.

Harmless at runtime but dead duplication.

**Fix:** Delete one of the two copies.

---

### 4. `public/components/alerting/alarms_page.tsx:71-72` — Malformed comment banner

The header collapsed onto one line:

```ts
// =====================================================================// Main Page Component
```

**Fix:** Restore the comment box format:

```ts
// ============================================================================
// Main Page Component
// ============================================================================
```

riysaxen-amzn:
About failed SLO Cypress test: It's an SLO Cypress E2E test (rule_health.spec.js) that tests a "recovery flow" against a real Cortex instance. The test expects a danger callout to disappear after ruler recovery, but it doesn't within 30 seconds.

TackAdam:
Thanks for the updates — the double-registration and `setupRoutes` type error are both resolved. Two things still need fixing before this is good to go:

## 1. `server/routes/index.ts` + `server/plugin.ts:390` — Remove the static `alertManagerEnabled` gate (don't just wire in a config value)

Right now `alertManagerEnabled: true` is hardcoded at the `setupRoutes` call site, which makes the new `if (alertManagerEnabled)` gate in `routes/index.ts` dead code.

But the fix here is **not** to pass `ymlAlertManagerEnabled` — that would actually introduce a regression.

On `main`, alerting routes register unconditionally by design (from #2719). Enablement is handled by the per-request `capabilities.observability.alertManagerEnabled` switcher (`plugin.ts:502`), which gates UI visibility and can flip per-account at runtime without a server restart.

The original comment spelled this out:

```ts
// Alerting routes register unconditionally so the dynamic capability
// flag (...) can hide the UI on a per-account basis without leaving
// the routes unreachable when re-enabled.
```

The static `if (alertManagerEnabled)` gate added in this PR gates at boot time, which is the wrong layer:

- Hardcoded `true` → the gate can never be false → dead code.
- If we instead passed `ymlAlertManagerEnabled` (which defaults to `false`), routes wouldn't register, while the dynamic switcher would still show the UI → 404s, and you couldn't re-enable per-account without a restart.

**Fix:** Revert `routes/index.ts` to unconditional registration (drop the `alertManagerEnabled` parameter and the `if` wrapper, restore the original comment), and remove the `alertManagerEnabled: true` line from the `setupRoutes` call in `plugin.ts`.

This matches how the SLO routes immediately below already work — register unconditionally, gate via the `sloEnabled` capability flag.

---

## 2. `public/components/alerting/alarms_page.tsx:71` — Comment banner still mangled

This one wasn't applied yet — the divider and label are still collapsed onto one line:

```ts
// =====================================================================// Main Page Component
// =====================================================================
```

Should be:

```ts
// =====================================================================
// Main Page Component
// =====================================================================
```

---

Everything else (duplicate `useEffect`, unused `datasourceId` parameter, the `false &&` → named constant) looks good. 👍

TackAdam (approved):
LGTM, Thanks for PR!

mengweieric (approved):

PR #2758, opened 2026-06-24
SLO Improvements

TackAdam:
# Improve SLO creation flow & Alert Manager severity/datasource UX

https://github.com/user-attachments/assets/51296a73-e564-4915-a91c-8e9f804279b0

## Description

This PR is a focused UX and correctness pass over the **SLO creation wizard** and a few smaller **Alert Manager** fixes. The headline change makes SLO creation **service-first and guided** — pick a service, the SLI query is built for you, you can test it inline, and an Advanced editor is always one click away for power users. Every built-in template now shares the same two-tier (simple summary + Advanced) editor, with the correct representation per SLI type.

It also fixes several real bugs found while exercising the flow against a live cluster (datasource identifier mismatch, an over-eager rule-preview request, a probe that read empty for slow counters, and a regressed Alert Manager monitor-type gate).

## Motivation

- The SLI section required hand-written PromQL and exposed a hardcoded/hidden metric — "no one can hand-write PromQL" was the core complaint.
- Selecting a service didn't build or update the query; switching services left stale values baked in.
- The rule preview fired before required fields were filled, showing a confusing "Preview unavailable / Bad Request".
- Datasource was a free-text field; the value it emitted didn't match what the deploy path could resolve.
- Page ergonomics: Create button only at the top, scroll not reset on template pick, clutter from redundant fields.

---

### SLO wizard — Datasource field
- Replaced the free-text **Datasource ID** input with a **searchable single-select combobox** (`datasource_select.tsx`), sourced from the existing `useDatasources()` hook and filtered to Prometheus connections that can actually back an SLO.
- **Bug fix:** the picker now emits the datasource's **`directQueryName`** (SQL-plugin connection id) — the identifier the SLO deploy path resolves — instead of the saved-object id. Previously *preview* accepted the spec but *create* failed with `Datasource "…" is not registered`. Datasources without a `directQueryName` are filtered out (they can't be deployed to).
- Empty/loading states: actionable "No compatible datasources" callout linking to Data Sources management; no flash of the empty state while loading.

### SLO wizard — Service & owner fields
- **Service** and **Primary team** are now **suggesting comboboxes** (`suggesting_combo_box.tsx`) — type-ahead from real values, but free text still allowed.
  - Service suggestions come from Prometheus `service_name` / `rpc_service` label values on the selected datasource (`use_owner_suggestions.ts`, via a new `slo_api_client.labelValues()` method that reuses the Alert Manager metadata route).
  - Team suggestions come from `owner.teams` on existing SLOs (no backend team registry exists).
  - Both degrade to plain free-text input if discovery fails or no datasource is selected.
- Added placeholder/hint text to **Primary team** ("Select a team or type a new one").

### SLO wizard — SLI query editor (the big one)
- **Unified two-tier editor across all 12 templates:** a read-only **plain-language summary + generated query**, with an **Advanced** accordion that reveals the editable controls. Advanced auto-opens when the query was customized or has a validation error; stays collapsed for an untouched template default.
- **Type-appropriate simple/advanced content** (not always "Ratio"):
  - **custom** templates (apm-service/dependency availability+latency, blank `custom`) → the **Ratio / Advanced** PromQL builder (`custom_promql_editor.tsx` + `metric_picker.tsx`): Success + Total metric pickers with autocomplete, label-filter dropdowns, and an optional **"subtract a metric"** term (models APM `good = request − fault` with full dropdowns instead of raw text). Advanced mode = raw error-ratio PromQL.
  - **availability** templates (http / rpc / genai) → a new **structured editor** (`structured_sli_editor.tsx`): editable **metric picker** + **good-events filter**. Type/`canonicalKind` preserved (no silent conversion to custom).
  - **latency_threshold** templates (http / rpc / db / messaging) → structured editor: editable **metric picker**; threshold stays per-objective.
- **Metric is now user-editable** for structured SLIs (new `metric` field in `FormState`, consumed by `wizard_builders`) — previously hardcoded and hidden in the template.
- **New `promql_builder.ts`**: a structured PromQL model with `serializePromQL` / `parsePromQL` (balanced-paren/brace aware, supports `agg by(...)`, `rate(...[window])`, label matchers, and a top-level subtract term). Round-trips representative SLI queries; returns `null` for shapes the builder can't represent so the editor falls back to a raw textarea rather than mangling the query.
- **Service ↔ query/dimension sync:** changing the Service re-derives the template query for the new service **only while it's still the untouched template default** (hand-edited queries are preserved), and keeps the first dimension's value in sync. Fixes the literal `${service}` placeholder and "switching service doesn't update the query".

### SLO wizard — Test query (formerly Probe SLI)
- Folded the standalone **Probe SLI** panel **into the SLI query card**, at the bottom (beneath Advanced), so building and verifying the query live together.
- Renamed **"Probe SLI" → "Test query"** and trimmed the jargon-y heading/description (`probe_sli_panel.tsx` gained an `embedded` mode).
- Disabled "Test query" now shows **hover text explaining why** ("Select a datasource first." / "Finish the SLI query — pick a metric so it returns data.").
- **Bug fix (server, `probe_sli.ts`):** the probe evaluated `rate(metric[5m])` as an **instant query at `now`**, which is empty for slow counters even when data exists. It now **range-queries over the lookback window and takes the latest finite sample**, so the probe reliably reflects real data.

### SLO wizard — Rule preview
- **Bug fix:** the preview no longer fires before the form is client-valid (gated on the same `validateSloSpec` errors), eliminating the "Preview unavailable / Bad Request" flash on open.
- Removed bullet markers from the missing-fields list (plain stacked links).
- Preview panel turns **yellow (warning)** while required fields are incomplete, signaling it's blocked.

### SLO wizard — Dimensions
- The first dimension auto-syncs from the **Service** field, so the common single-service case needs no manual entry.
- Moved the Dimensions editor into a **collapsed "Dimensions (advanced)" accordion** (it's redundant with Service for the common case; load-bearing only for multi-slice SLOs).

### SLO wizard — Page ergonomics
- **Create / Cancel** moved out of the top header into an **in-flow action footer at the bottom** of the form (header keeps only "Change template"). Reachable after reviewing the rule preview, no scroll-back-to-top.
- **Scroll resets to the top** when a template is selected (picking a card lower in the grid no longer lands you mid-page).
- **Switching templates resets the form to a fresh state** — no stale fields bleed across SLI types.
- "Description" relabeled **"Description (optional)"** to match the existing optional-field convention.

### SLO common / validators
- `slo_types.ts`: added `SUPPORTED_SLI_BACKENDS` / `isSupportedSliBackend()` and `SliBackend` as the single source of truth for which backends SLOs can be created against; validator now gates off it (instead of a hardcoded P0 string).

### Alert Manager — fixes
- **Severity options aligned to named tiers** (`ppl_triggers.tsx`): the severity dropdown now reads **Critical / High / Medium / Low / Info** instead of `1–5`. Stored values remain `'1'–'5'` for backend compatibility — display-only relabel.
- **Monitor-type gating restored** (`monitors_table/index.tsx`): when the selection is **only Prometheus**, the **Logs** create entry is greyed out; when it's **only OpenSearch**, the **Metrics** entry is greyed out. The Metrics gate had regressed to always-enabled; restored the symmetric logic the panel was already built to consume.

---

## Testing

- **Unit tests:** added/updated across the SLO suite — `promql_builder`, `datasource_select`, `suggesting_combo_box`, `use_owner_suggestions`, `custom_promql_editor`, `structured_sli_editor`, `metric_picker`, `generated_rules_preview`, `probe_sli_panel`, `wizard_state`, `wizard_builders`, and a parametrized **`sli_editor_all_templates`** test that verifies **every one of the 12 templates** renders a simple summary + Advanced accordion + editable control (fails CI if a template regresses or a new one omits it).
- Full SLO + alerting unit suites pass; `yarn typecheck` clean; `node scripts/eslint` clean (0 errors).
- **Cypress:** updated `slo_test/slo_crud_smoke`, `slo_test/wizard_validation`, and `slo_test/wizard_templates` for the new combobox/SLI-editor selectors (drive the SLI via the Advanced/raw textarea; assert the read-only summary for template prefill). FTR specs require the Cortex + OpenSearch stack and were validated by static + syntax check, not executed locally.
- **Manual verification** against a live cluster: confirmed datasource dropdown options, service/team suggestions, metric autocomplete, preview (200 vs 400), end-to-end create (after the `directQueryName` fix), and the range-based probe returning live data.

### Known data caveat
On clusters without Data Prepper span-derived metrics flowing, the `apm-*` templates' default `request`/`fault` queries will probe empty — this is a data-pipeline gap, not a wizard issue. The wizard's "Test query" surfaces it before create, and OTel metrics with live data (e.g. `http_server_duration_milliseconds_count`) work as expected.

### Check List


ps48 (commented):
Strong, thorough PR. The service-first SLO wizard rework, the two-tier Ratio/Advanced editor across all 12 templates, and the structured `PromQLModel` with safe parse/serialize fallback are well architected, and the test coverage is genuinely comprehensive (the parametrized `sli_editor_all_templates` test, the round-trip `promql_builder` tests, and the per-component suites all assert behavior).

I checked the things most likely to bite and they hold up:
- the controlled-input set-state-during-render in `RatioMetricField` is the safe pattern (guarded by `lastEmittedRef`), no re-render loop
- `rederiveQueryForService` correctly preserves a hand-edited query and only re-substitutes the untouched template default
- `metric` is nulled for custom SLIs in `wizard_builders.ts`, so it doesn't leak into `customExpr`
- the new `slo_api_client.labelValues()` reuses the existing Alert Manager metadata route, not a new backend query route
- `directQueryName` is a real field, so the datasource filter is sound
- server `probe_sli.ts` range-query fix is the right call, with per-query errors surfaced in the response

No new saved objects, no new dependencies, no stray logging, and no swallowed errors (the `use_owner_suggestions` degradation is correct for advisory autocomplete). One minor placement suggestion inline. Nice work.

public/components/apm/pages/slos/datasource_select.tsx
```diff
@@ -0,0 +1,166 @@
...
+import { SUPPORTED_SLI_BACKENDS } from '../../../../../common/slo/slo_types';
+import { useDatasources } from '../../../alerting/hooks/use_datasources';
+import { coreRefs } from '../../../../framework/core_refs';
+
+/**
+ * Map an SLI backend to the `Datasource.type` that can serve it. Kept explicit
+ * (rather than assuming the names match) so the two type spaces can evolve
+ * independently — e.g. a future backend served by an existing datasource type.
+ */
+const SLI_BACKEND_TO_DATASOURCE_TYPE: Record<SliBackend, DatasourceType> = {
```
ps48: ### p1: move the backend-to-datasource-type mapping to common

`SLI_BACKEND_TO_DATASOURCE_TYPE` and `SUPPORTED_DATASOURCE_TYPES` are derived purely from `SUPPORTED_SLI_BACKENDS` (common/slo/slo_types), so they are really a shared backend concern that currently lives in a public component. When OpenSearch-backed SLOs land, the deploy/probe path on the server may want the same backend-to-datasource-type mapping. Should we move both into `common/slo/slo_types.ts` next to `SUPPORTED_SLI_BACKENDS` and re-import here, so there is one source of truth the server side can also reach?

ps48 (approved):

lezzago:
## Review — SLO Improvements

Reviewed the SLO wizard rewrite and Alert Manager fixes. Findings below, grouped by severity. Severity reflects production impact × likelihood of the trigger.

---

### 🟠 High

**H1 — Cleared good-events filter silently deploys a constant-100% availability SLO**
`public/components/apm/pages/slos/structured_sli_editor.tsx:218` · `common/slo/slo_validators.ts:247` · `common/slo/slo_promql_generator.ts:172`

The editor now exposes `goodEventsFilter` as a clearable `EuiFieldText` with no required guard. The validator only checks it *when present* (`if (prom.goodEventsFilter) { ...validate shape... }`) and never requires it for availability SLIs — contrast the line above, where `metric` *is* required. With an empty filter, `buildSelectors(sli, true)` (good) equals `buildSelectors(sli, false)` (total), so the recording rule becomes `sum(rate(m{svc}[w])) / sum(rate(m{svc}[w])) ≡ 1.0`.

Result: an availability SLO that always reports 100%, can never burn error budget, and can never alert — created with no error, no warning, and no test. The read-only summary and the "Test query" probe both render a valid-looking 1.0 ratio, so every signal reassures the user the misconfiguration is fine.

*Suggested fix:* require `goodEventsFilter` for `availability` SLIs in the validator (mirror the `if (!prom.metric)` pattern), or at minimum warn when good selectors equal total selectors. Add a test for the cleared-filter case.

---

### 🟡 Medium

**M1 — Probe under-counts / mis-displays multi-series queries**
`server/routes/slo/probe_sli.ts:211` · `server/services/alerting/directquery_prometheus_backend.ts:894`

The probe switched good/total from instant queries (`reduceInstant` summed *all* series) to range queries reduced by `reduceRange`. But `parseRangeQueryResponse` keeps only `result[0]` (the first series). For a good/total query returning multiple series, the probe now reports one arbitrary series instead of the sum, so the displayed Good/Total counts and the SLI ratio are silently too low. Same root cause affects the ratio sparkline query at line 240.

Narrow trigger (reachable via the custom-events raw-textarea fallback with an un-aggregated query — every templated path wraps in `sum()`), but note the test rename **deleted the only multi-series coverage**, so the behavior is now unguarded in both directions.

*Suggested fix:* sum across series per timestamp in `parseRangeQueryResponse` (match old instant semantics), or wrap good/total in `sum()` before probing; add a regression test with `result[]` length > 1.

**M2 — "Query:" preview misrepresents what actually deploys**
`public/components/apm/pages/slos/custom_promql_editor.tsx:235` · `structured_sli_editor.tsx:103`

Two concrete divergences between the read-only preview and the generated rule:
- Custom-events summary shows the success ratio `(good)/(total)`, but the generator deploys the **error** ratio `1 − (good/total)`.
- Latency summary uses the raw threshold and bare metric — for a milliseconds SLI it shows `le="500"` on `metric{}`, while the rule deploys `le="0.5"` on `metric_bucket{}` (the generator applies `formatLatencyBoundLe` + `ensureBucketMetric`).

The preview exists to build trust in what will deploy and currently shows a different query. *Suggested fix:* derive the summary from the generator output rather than hand-building the PromQL string client-side.

**M3 — `metric` backfill makes "metric required" unreachable and silently overrides a cleared metric**
`public/components/apm/pages/slos/wizard_builders.ts:57`

`metric: state.metric || template.sli.metric` — because every availability/latency template ships a default metric, `prom.metric` is never empty by the time the validator runs, so the `!prom.metric` required-metric error (and the structured editor's `isInvalid` / auto-open-on-error path) is dead code. Worse: a user who clears the Metric field gets it silently reverted to the template default — the UI shows empty while the deployed SLO measures the template's metric, with no indication of the mismatch.

*Suggested fix:* drop the `||` fallback (`metric: state.metric`); `applyTemplate` already seeds the field with the default, so clearing it correctly yields an empty metric and fires the validation error.

**M4 — Free-text Service/Team dropped on blur without Enter**
`public/components/apm/pages/slos/suggesting_combo_box.tsx`

Free text commits only via `onCreateOption` (Enter). Typing a new service name and clicking Create / the next field (instead of pressing Enter) discards the value → "service required" error, or an SLO created with an empty/wrong service. *Suggested fix:* commit the pending input on blur.

---

### 🟢 Low

**L1 — Probe ratio mixes values from different timestamps; coarse range step**
`server/routes/slo/probe_sli.ts:281` · `:199`
`reduceRange` resolves good and total to each series' own latest finite point independently, so under scrape gaps the ratio reflects no single real instant. Compounding: `stepSec = floor(windowSeconds / 20)` → ~8.4h step on the 7d lookback, so the "latest finite point" can be hours stale. Preview approximation, clamped to [0,1].

**L2 — `applyTemplate` JSDoc contradicts the code**
`public/components/apm/pages/slos/wizard_state.ts` (~line 304)
The JSDoc says it "preserves user edits to fields the template doesn't own (service/name/owner)," but the body returns `initialState()` and a sibling inline comment (plus the renamed test) confirm it resets everything. Stale doc — please update. *Open question: is resetting the datasource on every template switch intended?*

**L3 — `datasource_select.tsx` `value` prop JSDoc is wrong**
`public/components/apm/pages/slos/datasource_select.tsx:59`
Doc says "saved-object id," but the component emits/resolves `directQueryName` end-to-end (the test even asserts this). Stale doc.

**L4 — Test coverage gaps**
The three riskiest items — H1 (cleared filter), M1 (multi-series), and the error-collapsed-accordion path — all ship unasserted. The probe NaN test exercises a branch the backend pre-filters (`parseRangeQueryResponse` already drops `NaN`), so it gives false confidence.

---

### ⚪ Nit-picks

**N1 — `isTemplateDefault` logic in 3 places** (`custom_promql_editor.tsx:83`, `structured_sli_editor.tsx:114`, `wizard_state.ts` `isUntouched`) with subtly different semantics (editors compare against the *current* service's default; the reducer against the *previous*). Effect is limited to accordion open/closed cosmetics.

**N2 — Advanced accordion auto-open is a one-shot `useState`** — doesn't reopen when a validation error arrives post-mount, and the custom editor omits the `|| hasErrors` the structured editor has (`custom_promql_editor.tsx:245` vs `structured_sli_editor.tsx:137`). The "Needs attention" hint still shows, so this is cosmetic.

---

### Verified safe (no action)

- **Datasource identity** is consistent end-to-end: create rewrites `spec.datasourceId = ds.name`, and `datasourceService.get()` resolves both the saved-object id and `directQueryName`. Emitting `directQueryName` from the picker is correct.
- **Severity relabel** (`ppl_triggers.tsx`): only display `text` changed; stored values stay `'1'`–`'5'`. Backend-compatible.
- **Monitor-type gating** (`monitors_table/index.tsx`): the symmetric `every(d => d.type === 'opensearch')` gate is correct; `selected` only contains resolvable datasources.
- **`RatioMetricField` render-phase `setState`**: follows React's sanctioned adjust-state-on-prop-change pattern (ref updated before `setModel`); no loop.

Nice work overall — the service-first flow and unified editor are a real improvement. H1 is the one I'd block on; M1–M4 are worth addressing before merge.

TackAdam:
Thanks for the thorough review. I went through each comment, verified it against the code, and here's where each one landed.

## 🔵 ps48 — Move backend→datasource-type mapping to common

Done. Moved `SLI_BACKEND_TO_DATASOURCE_TYPE` and `SUPPORTED_DATASOURCE_TYPES` into `common/slo/slo_types.ts`, right next to `SUPPORTED_SLI_BACKENDS`, so the wizard and the server deploy/probe path share one source of truth.

`datasource_select.tsx` now re-imports (and re-exports for back-compat). I kept the mapping declared locally in `slo_types` rather than importing `DatasourceType` from `common/types/alerting`, to avoid pulling the alerting types into the `common/slo` import graph (the suggest engine and other consumers must stay free of that).

## 🟠 H1 — Cleared good-events filter silently deploys a constant-100% SLO

Fixed. This was a real footgun. The validator now requires `goodEventsFilter` for availability SLIs, mirroring the existing `!prom.metric` required-field pattern:

```ts
if (!prom.goodEventsFilter || !prom.goodEventsFilter.trim()) {
  errors['spec.sli.definition.goodEventsFilter'] =
    'Good events filter is required for availability SLIs (e.g. status_code!~"5..")';
}
```

Without it, the good selector equals the total selector and the error ratio collapses to `1 - (x/x) = 0` — an SLO that reads 100% available forever, never burns budget, never alerts.

Now it fails locally in the wizard and at the API boundary.

Added two regression tests (empty + whitespace-only filter) and updated 11 service-layer fixtures that built availability specs without a filter.

## 🟡 M1 — Probe under-counts multi-series queries

Fixed. The good/total probe queries now use `queryRangeMatrix` and sum the latest finite point across all series (matching the old instant semantics), instead of `queryRange` which kept only `result[0]`.

Plumbed `requestTimeoutMs` through `queryRangeMatrix`.

Added a multi-series regression test asserting two series sum to the correct Good/Total/ratio rather than reporting one arbitrary series.

## 🟡 M2 — "Query:" preview misrepresents what deploys

Fixed (latency half). The structured editor's summary now uses the generator's own helpers (`ensureBucketMetric`, `formatLatencyBoundLe`), so the preview matches the deployed rule — the `_bucket` suffix and the unit-scaled `le` bound (a 500 ms threshold renders as `le="0.5"` on `metric_bucket`, not `le="500"` on the raw metric).

Added a test covering the millisecond → `le="0.5"` case.

**Custom-events half:** Left as-is intentionally — the summary's caption reads "successful request ratio," so showing good/total is internally consistent with its own label. Happy to revisit if you'd rather it show the deployed error ratio.

## 🟡 M3 — Metric backfill makes "metric required" unreachable

Fixed. Dropped the `|| template.sli.metric` fallback in `wizard_builders.ts`.

`applyTemplate` already seeds `state.metric` with the default, so clearing the field now correctly yields an empty metric and fires the validation error (instead of silently reverting to the template default).

## 🟡 M4 — Free-text Service/Team dropped on blur

Not a bug — verified, no change.

`EuiComboBox` already commits pending free-text input on blur via `onCreateOption` (`clearOnBlur` defaults falsy; `onContainerBlur → setCustomOptions` runs when there's no active option).

I confirmed this in the EUI source. Clicking away or tabbing out commits the typed service/team.

## 🟢 L2 / L3 — Stale JSDoc

Fixed both.

`applyTemplate`'s doc now correctly says it resets the whole form to a fresh template-seeded state (clean-slate per the product decision), and the `datasource_select` `value` prop doc now correctly states it's the `directQueryName` connection ID (persisted as `spec.datasourceId`), not the saved-object ID.

## 🟢 L1 / L4 / ⚪ N1 / N2 — Approximations & cosmetics

- **L4** is partly addressed: **H1** (cleared filter) and **M1** (multi-series) now ship with the regression tests that were missing.
- **L1** (ratio mixes timestamps / coarse step) is a documented preview approximation, clamped to `[0,1]` — left as-is.
- **N1/N2** are accordion open/closed cosmetics — left as-is.

---

## Also in this PR (unrelated to review)

- Removed the SLO "Beta" chrome badge from `slos_page.tsx` — it covered all SLO routes (listing, create, suggest, detail).
- Switched the Alerts app icon from beaker → bell, in both the app registration (`plugin.tsx`) and the nav-group link (`plugin_nav.tsx`).

## Verification

All affected suites pass:

- 502 SLO unit/service tests
- 422 APM/SLOs UI tests
- Validator, probe, and directquery backend suites

Lint clean on all touched files.

No Cypress changes needed — the API-seeded specs already supply `goodEventsFilter`, and the only UI-submit specs use the custom template (unaffected by H1/M3).

lezzago:
## Re-review after "address comments, add fixes"

Re-checked all findings against the updated diff (commit `slo improvements` / `address comments, add fixes`). Most were addressed well — summary below.

### ✅ Addressed

- **H1 — cleared good-events filter → constant-100% SLO** — Fixed. `slo_validators.ts:247` now requires the filter for availability SLIs (`if (!prom.goodEventsFilter || !prom.goodEventsFilter.trim())` → `'Good events filter is required for availability SLIs…'`), and there's a validator test asserting an empty filter is rejected. 👍

- **M1 — probe under-counts multi-series queries** — Fixed correctly. The probe now calls a new `queryRangeMatrix` (which parses *all* series, not just `result[0]`), and `reduceRange` sums each series' latest finite point. The fix reuses existing backend infrastructure cleanly rather than special-casing. Comment even documents the rationale.

- **M3 — `metric` backfill hid the required-metric error** — Fixed exactly as suggested: `metric: state.metric` (dropped the `|| template.sli.metric` fallback), so clearing the field now yields `''` and the validation fires.

- **M2 (structured half)** — Fixed. `structured_sli_editor.tsx` now imports and uses `ensureBucketMetric` + `formatLatencyBoundLe`, so the latency preview shows the real `_bucket` metric and unit-scaled `le` bound.

- **L2 / L3 — stale JSDocs** — Both fixed (`applyTemplate` now documents the clean-slate reset; `datasource_select` `value` prop now says `directQueryName`, not saved-object id).

### ⚠️ Still open

- **M2 (custom half)** — `custom_promql_editor.tsx:235`: the events-mode summary still renders `(${value.goodQuery}) / (${value.totalQuery})` — the **success** ratio — but the generator deploys the **error** ratio `1 - (good/total)` (per the file's own header comment). The "Query:" label presents this as what deploys, so for custom-events SLIs the preview still shows the inverse of the deployed expression.

- **M4 — free-text Service/Team dropped on blur** — Not addressed. `suggesting_combo_box.tsx` still commits free text only via `onCreateOption` (Enter). Typing a new service/team and clicking Create without pressing Enter discards it. Consider committing the pending input on blur.

- **N2 — accordion auto-open asymmetry** — `custom_promql_editor.tsx:245` still uses `useState(!onTemplateDefault)` (no `|| hasErrors`), while the structured editor has it. A validation error on a template-default custom query keeps the fields collapsed. Cosmetic.

- **L4 (partial)** — The probe tests were migrated to `queryRangeMatrix` with a `oneSeries` helper, but every case still feeds a single series. There's no test with **two distinct series** asserting they're summed — i.e. the exact M1 regression isn't pinned. A two-series case would lock it in.

### New changes reviewed (no issues)

- `plugin.tsx` / `plugin_nav.tsx`: alerting nav icon `beaker` → `bell`, and removal of the SLO "Beta" badge in `slos_page.tsx`. Cosmetic; the `chrome` prop is still consumed by child routes, so the badge `useEffect` removal doesn't orphan anything.

Nothing new introduced that's concerning. The remaining items are one medium (M2 custom-ratio preview) and minor polish (M4, N2, L4).

lezzago (approved):

PR #2762, opened 2026-06-25
Add APM, SLO, and Alerting navPopovers and rename Application Map to Topology Map

ps48:
### Description

This change wires per item hover popovers (navPopover) into the icon side navigation for the APM applications, building on the navPopover system added in OpenSearch Dashboards core.

The following nav items get popover actions:

- Topology Map: View topology and APM settings
- Services: View services and APM settings
- SLO: View SLOs and Create SLO
- Alerting: View alerts, Rules, and Routing

The APM settings actions navigate to each app's concrete landing route (for example #/services?_apmSettings=true and #/application-map?_apmSettings=true) so the Services hash router redirect cannot drop the marker before the page reads it. Each page reads and strips the marker using the shared useOpenOnUrlMarker hook from opensearch_dashboards_utils.

This PR also renames Application Map to Topology Map (breadcrumb and page title) and swaps the graphApp icon for navAiFlow in the service map related views.

Note: this depends on the navPopover types and the useOpenOnUrlMarker hook from the corresponding OpenSearch Dashboards core PR, so it should be merged after that core change is available.

### Issues Resolved

N/A

### Check List

  - [x] All tests pass


ps48:
The build fails as the Core PR isn't merged in yet https://github.com/opensearch-project/OpenSearch-Dashboards/pull/12285

TackAdam:
## 🟢 Low (cosmetic / conditional)

### 1. `public/plugin_helpers/plugin_nav.tsx:250` — The "Topology Map" rename doesn't cover the legacy (non-icon) nav mode or the registered app title

The icon side-nav link gets an explicit `title: 'Topology Map'` override (`plugin_nav.tsx:98`), and the breadcrumb/page title are renamed in `application_map_i18n.ts`.

The icon side nav — the surface this PR targets — renders **"Topology Map"** correctly.

The two nav paths are mutually exclusive at runtime (`plugin_nav.tsx:303`):

```ts
if (core.chrome.getIsIconSideNavEnabled()) {
  registerIconSideNavGroups(...);   // line 98: title: 'Topology Map' ✅
} else {
  registerDefaultNavGroups(...);    // line 250: no title override
}
```

The legacy/default nav group link at `plugin_nav.tsx:250` has no title override, so it falls back to the registered app title:

```ts
observabilityApmApplicationMapTitle = 'Application Map'
```

(`common/constants/apm.ts:16`, used at `plugin.tsx:450`).

That same fallback also governs global search results and "recently viewed."

- **Effect:** When icon side nav is enabled (the shipped experience here), everything reads **"Topology Map"** — no issue. Only if the legacy/default nav mode is active does the nav link still read **"Application Map"** while the breadcrumb and page title say **"Topology Map."** Global search / recently viewed show **"Application Map"** regardless.
- **Proposed fix (optional):** If the default nav mode is still a supported surface, rename `observabilityApmApplicationMapTitle` to **"Topology Map"** — that fixes the app registration, the legacy nav link, and global search in one place. If icon side nav is effectively always-on for this product, this is safe to ignore.

## 🟢 Minor / Optional

### 2. `public/plugin_helpers/apm_nav_popover.test.ts:34,36` — Non-null assertions in test code

The repo `CLAUDE.md` says:

> "Avoid non-null assertions (!.)."

`popover.actions!.find(...)!.onClick(...)` uses two.

This is conventional latitude in test files and not worth blocking on, but a guard or optional-chaining-with-assert would satisfy the rule if you want strictness.

TackAdam (approved):

joshuali925 (approved):

PR #2765, opened 2026-06-29
[CVE] Bump uuid to ^11.1.1 (CVE-2026-41907)

ps48:
### Description
Remediates **CVE-2026-41907** (GHSA-w5hq-g745-h8pq): `uuid` v3/v5/v6 are missing a buffer bounds check when a `buf` argument is provided. Fixed in uuid 11.1.1.

This plugin previously forced `uuid` to `3.4.0` via `resolutions`. This PR bumps the resolution to `^11.1.1` and regenerates `yarn.lock` through the OSD bootstrap flow (not hand-edited).

| CVE | Package | Was | Now |
|---|---|---|---|
| CVE-2026-41907 | uuid | 3.4.0 | 11.1.1 |

### Notes on the other CVEs in this batch
- **CVE-2026-34043** (serialize-javascript): already remediated on `main` — the lockfile resolves `serialize-javascript` to `7.0.5` (fix version) via the existing `^7.0.3` pin. No change needed.
- **CVE-2026-24400** (AssertJ): a Java test library in the **backend `opensearch-observability`** plugin, not this JS repo. Backend `main` is already on the fixed `assertj-core 3.27.7`. Out of scope here.

### Issues Resolved
CVE-2026-41907

### Check List


TackAdam (approved):

TackAdam:
This bump breaks the App Analytics / Trace Analytics ServiceMap at runtime.

The CVE remediation is correct, but forcing `uuid` to `^11.1.1` via `resolutions` reaches into a transitive dependency that relies on `uuid`'s old CommonJS default export.

## Root cause

`react-graph-vis` (which renders the ServiceMap graph) does:

```js
// node_modules/react-graph-vis/lib/index.js
var _uuid = require("uuid");
var _uuid2 = _interopRequireDefault(_uuid);
...
identifier: identifier !== undefined ? identifier : _uuid2.default.v4()
```

- **uuid 3.4.0** (current pin): the default export exists and carries `.v4`, so this works.
- **uuid 11.1.1** (this PR): `uuid` v7+ is ESM-first with named exports only and no default export, so `_uuid2.default` is `undefined` → `undefined.v4` → `TypeError: Cannot read properties of undefined (reading 'v4')`.

Because `resolutions` collapses the entire dependency tree onto one `uuid` version, `react-graph-vis` receives v11 and throws. The error is uncaught during render, causing the entire React tree to unmount—which is why the page goes blank.

## Impact

Any view rendering the ServiceMap crashes, including:

- App Analytics Create flow
- Trace Analytics Service Map

This is caught by the `app_analytics.spec.js` Cypress test:

```text
Cannot read properties of undefined (reading 'v4')
```

(Our own `import { v4 } from 'uuid'` call sites are fine—named imports continue to work in v11. The break is isolated to `react-graph-vis`.)

## Proposed fix

Keep the `uuid` bump for the CVE, but restore compatibility for `react-graph-vis`. In order of preference:

1. Use `patch-package` on `react-graph-vis` to replace the default import with a named import:

   ```js
   const { v4 } = require('uuid');
   ```

   instead of:

   ```js
   require('uuid').default.v4()
   ```

   `react-graph-vis` is effectively unmaintained (pinned at `^2.0.1`), so carrying a small compatibility patch is low risk while allowing the entire dependency tree to stay on the patched `uuid`.

2. Alias/shim `uuid` for `react-graph-vis` with a small compatibility module that re-exports `{ v1, v4 }` as both named and default exports.

3. Use a scoped resolution that leaves `react-graph-vis` on `uuid` 3.x. This is defensible from a security perspective since it only uses `uuid` to generate DOM identifiers—not the vulnerable buffer code path addressed by the CVE.

PR #2766, opened 2026-06-29
[CVE][2.19] Bump uuid to ^11.1.1 (CVE-2026-41907)

ps48:
### Description
Remediates **CVE-2026-41907** (GHSA-w5hq-g745-h8pq): `uuid` v3/v5/v6 are missing a buffer bounds check when a `buf` argument is provided. Fixed in uuid 11.1.1.

On 2.19 the tree previously resolved `uuid` to `2.0.3` transitively. This PR adds a `uuid` resolution pinned to `^11.1.1` and regenerates `yarn.lock` through the OSD bootstrap flow (Node 18, not hand-edited).

| CVE | Package | Was | Now |
|---|---|---|---|
| CVE-2026-41907 | uuid | 2.0.3 | 11.1.1 |

### Notes on the other CVEs in this batch
- **CVE-2026-34043** (serialize-javascript, fixed in 7.0.5): **not applicable on 2.19.** serialize-javascript 7.x requires Node >=20, but 2.19 builds on Node 18, so the fix version cannot be installed. The dependency is kept at `^6.0.2` (a dev/build-only transitive via eslint/webpack, not a runtime dependency). No other 2.19 dashboards plugin is on the fixed 7.0.5 for the same Node-18 reason.
- **CVE-2026-24400** (AssertJ): a Java test library in the **backend `opensearch-observability`** plugin, not this JS repo. Backend 2.19 is still on the vulnerable `assertj-core 3.16.1` and needs a separate fix to `3.27.7` in that repo.

### Issues Resolved
CVE-2026-41907

### Check List


ps48:
Closing — uuid CVE-2026-41907 is **not fixable on 2.19** under the current build toolchain.

**Why:** The only patched uuid versions are `11.1.1` / `12.0.1` / `13.0.1` (the CVE affects everything `<11.1.1`, with no pre-11 backport). All of these ship a `cjs-browser` dist using modern syntax (`??`, `?.`) that the 2.19 `@osd/optimizer` (webpack, in OSD core) cannot parse, since it does not transpile `node_modules`. The PR build fails with `Module parse failed: Unexpected token` in `node_modules/uuid/dist/cjs-browser/*.js` — the same break that hit `securityAnalyticsDashboards` in the 2.19 distribution build.

**Ecosystem check:** No OSD 2.19 dashboards plugin pins uuid to a fixed version — including security-analytics, which hit the identical build error. uuid is left transitive (2.0.3 / 8.3.2) across the board.

**Exploitability:** The flaw only triggers when a `buf` argument is passed to `uuid.v3/v5/v6`. The only consumer here (react-graph-vis → vis-network) generates IDs without `buf`, so it is not practically exploitable.

Leaving uuid deferred on 2.19 (tracked in #2657). The real fix lands on `main` via #2765, where the newer optimizer (OSD 3.8 / Node 22) builds uuid 11 successfully.

PR #2778, opened 2026-07-10
Refine Suggest SLOs page UX

canascar:
- Rename "Back to SLOs" to "Cancel" (navigates back via history)
- Move Preview button to header actions, simplify label
- Remove Refresh button
- Rename CTA from "Create N selected" to "Create N SLOs"
- Disable service checkbox for fully-covered services when unscoped
- Select all respects disabled (covered) services
- Remove scope/combo box UI
- Remove tilde from rule count in service panels
- Move namespace below description text
- Group preview flyout SLOs by service with accordion headers
- Only first 5 service groups expanded by default
- Add horizontal rules between preview groups

### Description
[Describe what this change achieves]

### Issues Resolved
[List any issues this PR will resolve]

### Check List
- [ ] New functionality includes testing.
  - [ ] All tests pass, including unit test, integration test and doctest
- [ ] New functionality has been documented.
  - [ ] New functionality has javadoc added
  - [ ] New functionality has user manual doc added
- [ ] Commits are signed per the DCO using --signoff

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #2783, opened 2026-07-13
[SLO] Refine suggest page

TackAdam:
## Summary

Builds on the original **"Refine Suggest SLOs page UX"** work from **#2778** and significantly expands it by reworking the service-scoping model, extracting a shared service-picker component used across both APM surfaces, moving inline styles into SCSS, addressing several correctness issues around draft selection vs. coverage, and adding extensive test coverage.

> Supersedes **#2778** — the original authored changes are preserved (credit to @canascar); this PR continues and extends that work.

---

### Suggest SLOs page (`slo_suggest_page.tsx`)

- Renamed **"Back to SLOs"** → **"Cancel"** (uses browser history instead of a hard-coded href).
- Moved the preview into the header actions with a simplified label.
- Removed **Refresh/Rediscover**, deleting the epoch-refetch plumbing end-to-end (hook prop + tests).
- Renamed CTA from **"Create N selected"** → **"Create N SLOs"** (with proper pluralization).
- Replaced the scope/combo-box UI with a single searchable service-filter dropdown.
  - Page now starts empty ("Pick one or more services…").
  - SLO drafts are generated only for explicitly selected services.
  - Removed automatic drafting of every discovered service.
- Improved the filter dropdown:
  - Checked services float to the top.
  - Added a **Clear** action.
  - Fully covered services are disabled and display a **"covered"** tag.
- Made the preview flyout resizable:
  - Drag handle with themed hover styling.
  - Width constrained between a minimum size and **45%** of the viewport.
- Removed the tilde (`~`) from rule counts.
- Moved the namespace beneath the description text.

### Preview flyout (`suggest_batch_preview.tsx`)

- Groups SLOs by service using accordion sections.
- Expands only the first five groups by default.
- Adds horizontal separators between groups.
- Groups by service identity instead of list adjacency, preventing duplicate accordions and duplicate React keys when both APM and OpenTelemetry drafts exist.

### Services Home SLO Health panel (`slo_health_panel.tsx`)

- Replaced the single **Suggest** button with the same searchable service picker used on the Suggest page.
- Trigger styling:
  - Primary (filled) while closed.
  - Neutral while open so the footer CTA becomes the primary action.
- Fully covered services are disabled and tagged as **covered**.
- If every service is already covered:
  - The trigger is disabled.
  - An explanatory tooltip is shown.

---

## Refactor

- Extracted shared components:
  - `service_filter_options.ts`
    - Pure `buildServiceFilterOptions()` helper
    - Handles covered-state disabling, checked-state marking, and float-to-top sorting
  - `service_filter_selectable.tsx`
    - Shared `EuiSelectable` implementation
    - Search support
    - Covered badge rendering
- Both APM service pickers now share the same implementation, eliminating duplicated option-building, `EuiSelectable` configuration, and `onChange` logic.
- Moved all inline `style={{}}` usage into SCSS following the directory's kebab-case conventions.
- Replaced magic numbers with named constants and EUI spacing tokens.
- Completed an i18n pass:
  - Wrapped all user-facing strings.
  - Added ICU pluralization where appropriate.

---

## Correctness & Hardening

- **Draft selection vs. coverage**
  - Drafts default to selected unless already covered by:
    - a matching Prometheus recording rule (`existingRuleMatch`), or
    - the service's canonical availability + latency SLO pair (health rollup).
  - Prevents creating duplicate SLOs for already-covered services.

- **Per-key interaction tracking**
  - Default-selection reseeding preserves user-made draft selections.
  - Newly arriving coverage deselects only untouched drafts.
  - Scope changes seed newly discovered drafts without overwriting curated selections.

- **Picker staleness handling**
  - Removes selected services if they disappear from discovery or become covered mid-session.
  - Prevents stale selections from inflating CTA counts or navigating with invalid scope.

- **Popover cleanup**
  - Stores the SLI-mix hover timer in a ref.
  - Cleans it up on unmount to prevent `setState` after unmount.

- **Accessibility**
  - Added meaningful `aria-label`s to the picker search and selectable list.
  - Covered rows communicate disabled state through both text and disabled styling (not color alone).

- **Additional improvements**
  - Lazy `useState` initialization for flyout width.
  - Removed a stale internal milestone comment.
  - Keyed the row-expansion effect using a stable service-name signature to avoid unnecessary reruns during selection changes.

---

## Testing

- ✅ **473 unit tests passing**
- ✅ Typecheck clean
- ✅ Stylelint clean
- ✅ ESLint: **0 errors**

### New / Updated Test Coverage

- Start-empty vs. scoped states
- Covered-draft default selection and supporting text
- Late coverage deselection while preserving manual user toggles
- Preview flyout resize behavior
  - Drag delta calculations
  - Global listener cleanup
- Preview breach/failed badges
- Non-contiguous service grouping
- Service picker:
  - Float-to-top sorting
  - Clear action
  - Primary ↔ neutral trigger behavior
  - Stale selection pruning
- SLI-mix hover popover open/close behavior

No Cypress changes were required—the existing APM Cypress suite does not exercise these SLO workflows.

### Check List

### DIFF
--- a/common/slo/classifier.ts
+++ b/common/slo/classifier.ts
@@ -52,7 +52,13 @@ export function classifySloKind(slo: SloSummary): CanonicalKind | undefined {
   return undefined;
 }
 
-function kindSide(kind: CanonicalKind | undefined): 'availability' | 'latency' | undefined {
+/**
+ * Map a canonical kind to the side of the availability+latency pair it
+ * satisfies. Exported so the Suggest page can decide per-draft coverage from
+ * the same rule the rollup uses (any `-availability` kind covers the
+ * availability side, any `-latency` kind the latency side).
+ */
+export function kindSide(kind: CanonicalKind | undefined): 'availability' | 'latency' | undefined {
   if (!kind) return undefined;
   if (kind.endsWith('-availability')) return 'availability';
   if (kind.endsWith('-latency')) return 'latency';
--- a/public/components/apm/pages/services_home/__tests__/slo_health_panel.test.tsx
+++ b/public/components/apm/pages/services_home/__tests__/slo_health_panel.test.tsx
@@ -4,7 +4,7 @@
  */
 
 import React from 'react';
-import { act, fireEvent, render, screen } from '@testing-library/react';
+import { act, fireEvent, render, screen, waitFor, within } from '@testing-library/react';
 
 // Stub coreRefs so the navigate helpers don't explode on mount. `jest.mock`
 // is hoisted above imports, so the captured variable has to be `mock`-
@@ -17,6 +17,64 @@ jest.mock('../../../../../framework/core_refs', () => ({
     http: { basePath: { prepend: (p: string) => p } },
   },
 }));
+
+// EuiSelectable virtualizes its option list (react-window + AutoSizer), which
+// measures 0×0 in jsdom and renders no rows — so its options can't be queried
+// or clicked in tests. Stub it with a plain checkbox list that preserves the
+// same option/onChange contract our picker relies on, leaving every other EUI
+// component real.
+interface StubOption {
+  label: string;
+  checked?: 'on';
+  disabled?: boolean;
+  append?: React.ReactNode;
+}
+jest.mock('@elastic/eui', () => {
+  const actual = jest.requireActual('@elastic/eui');
+  const ReactLib = jest.requireActual('react');
+  return {
+    ...actual,
+    EuiSelectable: ({
+      options,
+      onChange,
+    }: {
+      options: StubOption[];
+      onChange: (next: StubOption[]) => void;
+    }) =>
+      ReactLib.createElement(
+        'ul',
+        { role: 'listbox' },
+        options.map((opt) =>
+          ReactLib.createElement(
+            'li',
+            {
+              key: opt.label,
+              role: 'option',
+              'aria-selected': opt.checked === 'on',
+              'aria-disabled': opt.disabled ? 'true' : undefined,
+            },
+            ReactLib.createElement(
+              'button',
+              {
+                type: 'button',
+                disabled: opt.disabled,
+                onClick: () =>
+                  onChange(
+                    options.map((o) =>
+                      o.label === opt.label
+                        ? { ...o, checked: o.checked === 'on' ? undefined : ('on' as const) }
+                        : o
+                    )
+                  ),
+              },
+              opt.label
+            ),
+            opt.append
+          )
+        )
+      ),
+  };
+});
 
 import { SloHealthCell, SloHealthPanel } from '../slo_health_panel';
 import { navigateToSloListing, navigateToSloSuggest } from '../../../shared/utils/navigation_utils';
@@ -152,7 +210,7 @@ describe('SloHealthPanel', () => {
     expect(screen.queryByTestId('sloHealthPanelChip-breached')).toBeNull();
   });
 
-  it('primary CTA scopes to services with missingCanonicalPair=true only', () => {
+  it('opens a picker (nothing pre-checked) and scopes to the services the user picks', async () => {
     const bySvc = new Map<string, SloHealthBucket>();
     bySvc.set('foo', makeBucket({ missingCanonicalPair: true }));
     bySvc.set('bar', makeBucket({ missingCanonicalPair: false }));
@@ -169,11 +227,126 @@ describe('SloHealthPanel', () => {
       />
     );
     const cta = screen.getByTestId('sloHealthPanelCta');
-    expect(cta).toHaveTextContent('Suggest SLOs for 2 services');
+    // Default label — nothing picked yet.
+    expect(cta).toHaveTextContent('Suggest SLOs');
+    expect(cta).not.toHaveTextContent('service');
+
+    // Open the popover and pick a subset. `bar` is covered (disabled).
     fireEvent.click(cta);
+    const foo = await screen.findByText('foo');
+    fireEvent.click(foo);
+    fireEvent.click(screen.getByText('baz'));
+
+    // The confirm button reflects the picked count and navigates scoped to it.
+    const confirm = screen.getByTestId('sloHealthPanelCtaConfirm');
+    await waitFor(() => expect(confirm).toHaveTextContent('Suggest SLOs for 2 services'));
+    fireEvent.click(confirm);
     expect(mockNavigateToApp).toHaveBeenLastCalledWith('observability-apm-slo', {
       path: '#/slos/suggest?source=apm&services=foo%2Cbaz',
     });
+  });
+
+  it('floats checked services to the top of the list', async () => {
+    const bySvc = new Map<string, SloHealthBucket>();
+    for (const s of ['ad', 'cart', 'checkout', 'currency']) {
+      bySvc.set(s, makeBucket({ missingCanonicalPair: true }));
+    }
+
+    render(
+      <SloHealthPanel
+        aggregate={makeBucket({ total: 4, ok: 4 })}
+        bySvc={bySvc}
+        allServices={['ad', 'cart', 'checkout', 'currency']}
+        isLoading={false}
+        error={undefined}
+        onRetry={jest.fn()}
+      />
+    );
+    fireEvent.click(screen.getByTestId('sloHealthPanelCta'));
+    // Check two services that aren't first alphabetically.
+    fireEvent.click(await screen.findByText('checkout'));
+    fireEvent.click(screen.getByText('currency'));
+
+    // After selection they sort above the unchecked ones (each group keeps
+    // original order): checkout, currency, then ad, cart.
+    await waitFor(() => {
+      const labels = screen.getAllByRole('option').map((li) => li.textContent);
+      expect(labels).toEqual(['checkout', 'currency', 'ad', 'cart']);
+    });
+  });
+
+  it('clears all picks via the Clear button', async () => {
+    const bySvc = new Map<string, SloHealthBucket>();
+    bySvc.set('foo', makeBucket({ missingCanonicalPair: true }));
+    bySvc.set('baz', makeBucket({ missingCanonicalPair: true }));
+
+    render(
+      <SloHealthPanel
+        aggregate={makeBucket({ total: 2, ok: 2 })}
+        bySvc={bySvc}
+        allServices={['foo', 'baz']}
+        isLoading={false}
+        error={undefined}
+        onRetry={jest.fn()}
+      />
+    );
+    fireEvent.click(screen.getByTestId('sloHealthPanelCta'));
+    fireEvent.click(await screen.findByText('foo'));
+    fireEvent.click(screen.getByText('baz'));
+
+    const clear = screen.getByTestId('sloHealthPanelCtaClear');
+    const confirm = screen.getByTestId('sloHealthPanelCtaConfirm');
+    await waitFor(() => expect(confirm).toHaveTextContent('Suggest SLOs for 2 services'));
+    expect(clear).not.toBeDisabled();
+
+    // Clearing resets the count and disables both the confirm and clear actions.
+    fireEvent.click(clear);
+    await waitFor(() => expect(confirm).toHaveTextContent('Suggest SLOs'));
+    expect(confirm).not.toHaveTextContent('service');
+    expect(confirm).toBeDisabled();
+    expect(clear).toBeDisabled();
+  });
+
+  it('shows a filled (primary) trigger when closed and a neutral one when open', async () => {
+    render(
+      <SloHealthPanel
+        aggregate={makeBucket({ total: 1, ok: 1 })}
+        bySvc={new Map([['foo', makeBucket({ missingCanonicalPair: true })]])}
+        allServices={['foo']}
+        isLoading={false}
+        error={undefined}
+        onRetry={jest.fn()}
+      />
+    );
+    const cta = screen.getByTestId('sloHealthPanelCta');
+    // Closed → primary/filled to draw attention to opening the picker.
+    expect(cta.className).toContain('euiButton--fill');
+    fireEvent.click(cta);
+    // Open → neutral so the footer confirm is the only primary action in view.
+    await waitFor(() => expect(cta.className).not.toContain('euiButton--fill'));
+    expect(screen.getByTestId('sloHealthPanelCtaConfirm').className).toContain('euiButton--fill');
+  });
+
+  it('disables covered services in the picker so they cannot be scoped', async () => {
+    const bySvc = new Map<string, SloHealthBucket>();
+    bySvc.set('foo', makeBucket({ missingCanonicalPair: true }));
+    bySvc.set('bar', makeBucket({ missingCanonicalPair: false }));
+
+    render(
+      <SloHealthPanel
+        aggregate={makeBucket({ total: 2, ok: 2 })}
+        bySvc={bySvc}
+        allServices={['foo', 'bar']}
+        isLoading={false}
+        error={undefined}
+        onRetry={jest.fn()}
+      />
+    );
+    fireEvent.click(screen.getByTestId('sloHealthPanelCta'));
+    // The covered service renders a disabled option carrying the "covered" tag.
+    const barOption = (await screen.findByText('bar')).closest('li');
+    expect(barOption).toHaveAttribute('aria-disabled', 'true');
+    expect(within(barOption as HTMLElement).getByText('covered')).toBeInTheDocument();
   });
 
   it('secondary CTA carries the full service list currently shown', () => {
@@ -209,6 +382,7 @@ describe('SloHealthPanel', () => {
         onRetry={jest.fn()}
       />
     );
+    // Every service is covered → nothing to suggest → the trigger is disabled.
     expect(screen.getByTestId('sloHealthPanelCta')).toBeDisabled();
   });
 
--- a/public/components/apm/pages/services_home/__tests__/slo_health_panel.test.tsx
+++ b/public/components/apm/pages/services_home/__tests__/slo_health_panel.test.tsx
@@ -307,6 +307,51 @@ describe('SloHealthPanel', () => {
     expect(clear).toBeDisabled();
   });
 
+  it('prunes a pick that becomes covered before confirm (no stale/inflated selection)', async () => {
+    const uncovered = ['foo', 'baz'].reduce((m, s) => {
+      m.set(s, makeBucket({ missingCanonicalPair: true }));
+      return m;
+    }, new Map<string, SloHealthBucket>());
+
+    const { rerender } = render(
+      <SloHealthPanel
+        aggregate={makeBucket({ total: 2, ok: 2 })}
+        bySvc={uncovered}
+        allServices={['foo', 'baz']}
+        isLoading={false}
+        error={undefined}
+        onRetry={jest.fn()}
+      />
+    );
+    fireEvent.click(screen.getByTestId('sloHealthPanelCta'));
+    fireEvent.click(await screen.findByText('foo'));
+    fireEvent.click(screen.getByText('baz'));
+    const confirm = screen.getByTestId('sloHealthPanelCtaConfirm');
+    await waitFor(() => expect(confirm).toHaveTextContent('Suggest SLOs for 2 services'));
+
+    // A health refresh marks `foo` as now covered (owns its canonical pair).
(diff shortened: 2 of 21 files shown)


lezzago:
## Code review — max-effort pass (automated)

Reviewed against PR head `ff645635`. 15 findings, ranked most-severe first. Line numbers are in the PR-head files. Findings marked **CONFIRMED** have a concrete trigger + wrong outcome; **PLAUSIBLE** ones have a real mechanism with a less-certain trigger.

### 🔴 Correctness

**1. `toggleServiceSelection` can silently create duplicate SLOs** — `slo_suggest_page.tsx:564` · CONFIRMED
Clicking a service's master checkbox adds *every* draft key to `touchedKeysRef`. The seeding effect then keeps touched keys at `prev.has(key)` instead of `!isDraftCovered(s)` (line 407), so a **late-arriving health-coverage signal can no longer auto-deselect** a now-covered service's drafts. `runCreateSelected` (line 476) has no covered filter, so pressing **Create provisions duplicate Prometheus rules for an already-covered service**.
*Repro:* deep-link to service X that already owns its availability+latency pair → on mount `allServicesCoverage` is empty (async), drafts render selected, checkbox enabled → user clicks the master checkbox → health resolves marking X covered, but the drafts stay checked → Create dual-writes.

**2. `announceOnMount` is not a valid `EuiCallOut` prop** — `slo_suggest_page.tsx:643` (+671, 690, 709) and `slo_health_panel.tsx:310` · CONFIRMED
It's absent from `EuiCallOutProps`/`CommonProps` (no index signature) in this repo's pinned `@elastic/eui`/`@opensearch-project/oui` fork, and exists nowhere in the packages or any repo augmentation. Reproduced **`TS2322`** with the repo's own `tsc` + tsconfig (identical error to a bogus control prop), so this **breaks `yarn typecheck`**. If forced through, OUI spreads it onto the `<div>` as an unknown attribute — the intended screen-reader announcement never fires.

**3. Health-covered drafts are unchecked with no explanation** — `slo_suggest_page.tsx:504` · CONFIRMED
`isDraftCovered` unions per-draft `existingRuleMatch` **and** the health-rollup `allServicesCoverage`, but `coveredCount` / the "N drafts already covered" subline (gated `coveredCount > 0`) / the per-row "N covered" badge count only `existingRuleMatch`. A service covered via the health rollup (no matching recording rule) gets its drafts default-unchecked while the covered messaging never renders — user sees "0 of 2 SLOs" and a disabled Create with no reason given.

**4. "Cancel" is a dead button on deep-link entry** — `slo_suggest_page.tsx:575` · CONFIRMED
Changed from `href="#/slos"` to `onClick={() => history.goBack()}`. The page's primary entry is a cross-app `navigateToApp` deep link (`#/slos/suggest?...`) or a fresh/bookmarked URL, so the HashRouter stack length is 1 and `goBack()` is a no-op (or exits the app). The old href always returned to the listing.

**5. Resize drag leaks document listeners + stuck global cursor** — `slo_suggest_page.tsx:100` · CONFIRMED
`onMouseDown` adds `document` mousemove/mouseup listeners and sets `document.body.style.cursor`/`userSelect`, cleaned up **only** inside `onMouseUp`. Releasing the mouse outside the viewport (common at screen edge), or the flyout unmounting mid-drag (async coverage deselects the last draft → `selectedCount === 0` → flyout unmounts), skips cleanup — leaving the body cursor/user-select stuck **app-wide** plus a leaked mousemove listener. No unmount cleanup exists.

**6. Stale deep-link scope shows an unexplained empty state** — `slo_suggest_page.tsx:329` · CONFIRMED
Deleting `scopeFellThrough` means a scope naming services discovery no longer returns now resolves to `[]` with **no warning** — the user lands on the generic "Pick one or more services" panel. Base behaviour warned ("Scoped services not found — showing all discovered services") and stayed useful. Reachable via the per-row "Suggest" link (`navigateToSloSuggest([serviceName])`) or a bookmarked `?services=cart`.

**7. Filter button count contradicts the body** — `slo_suggest_page.tsx:187` · CONFIRMED
The label counts the raw URL scope (`scopedServices={scope.services}`), not the drafts actually rendered. Two triggers: (a) a stale/partial scope `?services=cart,ghost` reads "Suggest SLOs for 2 services" over 1 rendered row; (b) **a real discovered service that yields zero drafts** (no span-derived RED/OTel metrics) — picking `foo` reads "Suggest SLOs for 1 service" above the empty "pick services" panel with a disabled Create.

**8. First-paint race enables creating covered drafts** — `slo_suggest_page.tsx:500` · CONFIRMED
`useServiceSloHealth`'s `isLoading` is not destructured and not part of the page's `loading` flag (line 572 = config + services + discovery only). A deep link to a health-covered service renders its drafts checked, counted, and creatable during the coverage-fetch window — a fast Create (or confirm-modal) dual-writes before the drafts auto-deselect.

**9. Re-adding a removed service discards manual unchecks** — `slo_suggest_page.tsx:402` · CONFIRMED
The seeding effect prunes `touchedKeysRef` for keys absent from current suggestions. Unchecking a draft, removing its service from scope (suggestions → `[]`, key pruned), then re-adding the service re-seeds that draft as selected — the user's explicit decline silently reverts.

### 🟡 Dead code / correctness (lower severity)

**10. Dead `coveredServices` prop on `ServiceTreeTable`** — `suggest_service_tree_table.tsx:208` · CONFIRMED
The prop → `isCovered` → `disabled={isCovered}` master-checkbox guard is never exercised: the sole caller (`slo_suggest_page.tsx:851`) doesn't pass `coveredServices`, even though `allServicesCoverage` is available at the call site. Either wire it (`coveredServices={allServicesCoverage}`) or delete the prop + `isCovered` + `disabled` usage. Note this also contradicts the filter popover, which *does* disable covered services.

**11. Flyout width never clamped to min / not resize-aware** — `slo_suggest_page.tsx:259` · PLAUSIBLE
`flyoutWidth` seeds once from `window.innerWidth * 0.3` with no clamp to `FLYOUT_MIN_WIDTH` (300) and never recomputes on resize, while the drag handler re-reads `innerWidth` for its max. On an 800px viewport the flyout opens at 240px — below the minimum the drag itself enforces; shrinking a wide window later leaves a stale-too-large panel.

**12. No in-page recovery after a transient ruler failure** — `slo_suggest_page.tsx:343` · PLAUSIBLE
Removing `discoveryEpoch`/"Rediscover" leaves no way to re-run the probes (`useDiscoveryProbes` deps are just `[datasourceId, http]`; toggling the filter only does `history.replace`). A brief ruler outage shows the "could not verify existing recording rules" warning that then persists until a full page reload.

### 🟢 Efficiency / reuse

**13. Extra all-services health fetch on every mount** — `slo_suggest_page.tsx:365` · PLAUSIBLE
`useServiceSloHealth` now fetches over up to 200 discovered services on every mount (incl. the unscoped empty landing), solely to default checkboxes — on servers without the aggregate endpoint it falls back to a paged list fan-out (pageSize up to 800). `existingRuleMatch` (already fetched free by `useDiscoveryProbes`) covers the recording-rule case; this heavyweight call adds only the "already owns a saved pair" signal.

**14. `buildScopeUrl` duplicates `navigateToSloSuggest`** — `slo_suggest_page.tsx:158` · PLAUSIBLE
Both independently build `source=apm&services=csv&from&to`. The URL contract now lives in two places and will drift if a param is added/renamed. A shared `buildSloSuggestSearch(services, timeRange)` consumed by both removes it.

**15. Dead `handleRef`** — `slo_suggest_page.tsx:87` · CONFIRMED
`ResizableFlyout` creates `handleRef` and attaches `ref={handleRef}`, but `.current` is never read (drag math uses only the mouse event + `window.innerWidth`). Removable with no behavior change.

---
<sub>Ruled out (not bugs): `EuiSelectable` search-then-select selection drop (EUI passes the full option list to `onChange`), `EuiFlyout size={number}` (valid — accepts `CSSProperties['width']`), Set-mutation-during-iteration in the seeding effect (spec-safe), the picker pruning effects (return `prev` on no-change, no render loop). The `SloSuggestPicker` prune in `slo_health_panel.tsx` was verified correct.</sub>

TackAdam:
## Summary

Addresses the automated max-effort code review on the Suggest SLOs page (reviewed against `ff645635`). Of the **15 findings**, **12 are fixed** and **3 are intentionally deferred** (with rationale below). Also removes two stray vertical spacers identified during manual QA.

All changes are limited to the APM SLO UI and behavior. No API or saved object changes.

**Verification**

- ✅ `yarn typecheck` passes
- ✅ `yarn lint` clean on changed files
- ✅ 909 tests across 72 APM suites pass

---

## 🔴 Correctness

### #1 — `toggleServiceSelection` could silently create duplicate SLOs — FIXED

The service-level master checkbox marked every draft as "touched", preventing the default-selection effect from deselecting drafts that later became covered. Since `runCreateSelected` also lacked a coverage filter, covered services could be provisioned twice.

**Fix**

`runCreateSelected` now filters out `isDraftCovered(s)` before provisioning, providing a final safety check regardless of selection-state timing.

---

### #2 — `announceOnMount` is not a valid `EuiCallOut` prop — FIXED

This repository aliases `@elastic/eui` to `@opensearch-project/oui@1.22.1`, whose `EuiCallOutProps` does not include `announceOnMount`. The prop therefore produced `TS2322` errors during `yarn typecheck`.

**Fix**

Removed all five occurrences:

- 4 in `slo_suggest_page.tsx`
- 1 in `slo_health_panel.tsx`

This is the same OUI/EUI compatibility issue addressed by the ESLint override in **#2785**.

---

### #3 — Health-covered drafts were unchecked without explanation — FIXED

`isDraftCovered()` combines recording-rule coverage with health-rollup coverage, but the UI messaging (`coveredCount`, "N drafts already covered", and per-row badges) only considered `existingRuleMatch`.

Services covered solely by the health rollup therefore appeared unchecked without any explanation.

**Fix**

Both the page-level and per-row covered counts now use the same `isDraftCovered()` logic that drives default selection, ensuring the UI accurately reflects why drafts are disabled.

---

### #4 — "Cancel" was a dead button for deep-link entry — FIXED

`history.goBack()` becomes a no-op when the browser history contains only a single entry (for example, bookmarked URLs or cross-app deep links).

**Fix**

Replaced it with:

```tsx
history.push('/slos')
```

which always returns the user to the SLO listing.

---

### #5 — Resize drag leaked global listeners and cursor state — FIXED

`mousemove`/`mouseup` listeners and the global `body` cursor/user-select overrides were only cleaned up from `onMouseUp`. Releasing outside the viewport or unmounting during a drag skipped cleanup.

**Fix**

The drag teardown is now stored in a ref and also executed from an unmount cleanup effect, guaranteeing listeners and cursor state are always restored.

---

### #6 — Stale deep-link scope showed an unexplained empty state — FIXED

If every scoped service disappeared from discovery, the page simply rendered the generic empty state without explaining why.

**Fix**

Restored `scopeFellThrough` detection and added a warning callout:

- **Scoped services not found**
- Includes a **Clear scope** action

The messaging has been updated for the new scoped-selection UX ("Clear the scope to pick from all discovered services").

---

### #7 — Filter button count could disagree with rendered services — DEFERRED

The filter button reflects the raw URL scope rather than the services currently rendered, so a partial match can produce text like:

> Suggest SLOs for 2 services

while only a single service is displayed.

**Why deferred**

This is purely cosmetic, has no behavioral impact, and lives in a shared component used elsewhere. I'd prefer to address it separately rather than bundle unrelated UI semantics into this correctness pass.

---

### #8 — First-paint race allowed covered drafts to be created — FIXED

`useServiceSloHealth()` loading state wasn't included in the page loading gate, allowing health-covered drafts to appear selectable before coverage finished loading.

**Fix**

The Create action is now gated on `healthLoading`, eliminating the race. Combined with #1's provisioning filter, this provides defense in depth.

---

### #9 — Re-adding a removed service discards manual draft selections — DEFERRED

Current behavior:

1. User manually deselects a draft.
2. Removes that service from scope.
3. Adds it back.
4. Draft returns to its default selected state.

**Why deferred**

This is arguably the expected behavior: removing a service from scope serves as a logical reset. Preserving selections across removal would require retaining state for absent services, reintroducing the unbounded `touchedKeysRef` growth that pruning was designed to eliminate.

---

## 🟡 Lower Severity

### #10 — Dead `coveredServices` prop on `ServiceTreeTable` — FIXED

The table supported disabling master checkboxes for covered services, but the caller never supplied `coveredServices`, so the path was never exercised.

**Fix**

Passes:

```tsx
coveredServices={allServicesCoverage}
```

at the call site.

---

### #11 — Flyout width wasn't clamped on initial render — FIXED

The initial width (`window.innerWidth * 0.3`) could be smaller than the minimum width enforced during drag resizing.

**Fix**

The initial value now uses the same `[min, max]` clamp as the drag handler.

---

### #12 — No in-page recovery after transient ruler failures — FIXED

Removing the old `discoveryEpoch` mechanism also removed the ability to rediscover services without reloading the page.

**Fix**

Restored:

- `epoch` parameter to `useDiscoveryProbes()`
- dependency tracking
- **Rediscover** header action

The button now calls:

- `refetchServices()`
- increments the discovery epoch

to perform a full rediscovery.

---

## 🟢 Efficiency / Reuse

### #13 — Extra all-services health fetch on every mount — DEFERRED

`useServiceSloHealth()` currently fetches health information for every discovered service, even when the page initially loads with no scope selected.

**Why deferred**

That health information is currently required for the correctness fixes in **#1**, **#3**, and **#8**. Optimizing the fetch strategy (for example, delaying it until services are selected or relying more heavily on `existingRuleMatch`) should be treated as a separate performance-focused change rather than mixed into this correctness PR.

---

### #14 — `buildScopeUrl` duplicated `navigateToSloSuggest` — FIXED

Both implementations independently constructed:

```text
source=apm&services=<csv>&from&to
```

**Fix**

Extracted:

```ts
buildSuggestSearch(services, timeRange)
```

into `slo_suggest_scope.ts`, which already owns URL parsing. Both the in-page filter and `navigateToSloSuggest()` now use the same implementation.

---

### #15 — Dead `handleRef` — FIXED

`ResizableFlyout` created and attached a `handleRef` but never used it.

**Fix**

Removed the unused ref. Drag calculations rely only on the mouse event and `window.innerWidth`.

---

## 🎨 UI Polish

Manual QA surfaced two unnecessary vertical spacers, both of which have been removed:

- Removed the extra `EuiSpacer size="m"` between the intro/namespace section and the content below in the empty state.
- Removed the extra `EuiSpacer size="m"` between the header statistics strip and the service-filter row in the populated state.

lezzago:
## Follow-up review — regressions in `975b8edc` "address comments"

Re-reviewed the fix commit against head `c1b9fca5`. The pass cleanly resolved 12 of the original findings (#7, #9 deferred, #13 left as an efficiency note — all fine). The fix commit introduced **one new count-mismatch bug** plus a cosmetic doc misplacement. Static read only — not verified against a running build.

### 🟠 New: create-time backstop not reconciled with the displayed counts — `slo_suggest_page.tsx:507`

The #1 fix added a create-only filter:

```js
const picks = suggestions
  .filter((s) => selected.has(s.key) && !isDraftCovered(s))   // new: && !isDraftCovered(s)
  .map(applyOverrides);
```

but every "what will be created" surface still counts the raw selection (`selected.has(key)`), so they disagree whenever a covered draft is in `selected`:

| Surface | Line | Counts covered drafts? |
|---|---|---|
| `runCreateSelected` (actual create) | 507 | **No** |
| `selectedCount` (header stat, button) | 532 | Yes |
| `totalRules` (confirm modal "provisions N rules") | 533 | Yes |
| "Create N SLOs" button | 669 | Yes |
| Flyout title / confirm modal count | 988, 1007 | Yes |
| Preview flyout contents | 996 | Yes |
| Batch runner progress + success toast | `suggest_use_batch_create.tsx` | **No** (`picks.length`) |

**Reachable** because the disabling is incomplete: the per-draft inline-row checkbox disables only on `rowStatus` (`suggest_inline_row.tsx:80`), not on covered state — so a user can still explicitly check a covered draft (the original design even documented this as "re-check to create a duplicate on purpose"). The #1 freeze race (root cause unchanged) is a second path into the same state.

**Repro:** `cart` has `cart-avail` (covered) + `cart-lat` (not). User expands `cart` and checks `cart-avail` too → button reads **"Create 2 SLOs"**, confirm modal says **"provisions ~26 rules"**, preview shows both. On Create the backstop drops `cart-avail` → **"Created 1 SLO"**, progress `0/1` — with no message that `cart-avail` was skipped.

Two problems: (1) button/modal/preview/toast contradict each other (reads like a silent partial failure); (2) the blanket `!isDraftCovered` backstop silently overrides *any* covered draft the user deliberately selected, not just freeze-race leftovers — dropping the previously-supported intentional-duplicate case with no feedback. Safe direction (creates fewer, never more — no dual-write), but user-facing.

**Suggested fix:** derive one predicate and use it everywhere —
```js
const creatable = (s) => selected.has(s.key) && !isDraftCovered(s);
```
drive `selectedCount`, `totalRules`, the button/modal/title labels, and the preview `selectedSuggestions` from `creatable` (the same set `runCreateSelected` provisions). If intentional duplicates should stay blocked, also make the covered inline-row checkbox non-interactive so the state can't be entered in the first place.

### 🟢 Cosmetic: orphaned doc comment — `slo_suggest_scope.ts:55`

Inserting `buildSuggestSearch` (62–77) landed it *between* `parseSuggestScopeFromSearch`'s JSDoc (55–61) and the function it documents (79), so the parse function's doc block now floats above the wrong function. Move it back down to just above `parseSuggestScopeFromSearch`. No behavior impact.

### ✅ Clean in the fix commit
`ResizableFlyout` teardown-ref + unmount effect (leak closed), `flyoutWidth` clamp, `buildSuggestSearch` extraction (no circular import — the scope file imports only a type), `epoch?` optional default + `refetch` wiring, and the restored `scopeFellThrough` / Rediscover / covered-count union / `coveredServices` prop.

TackAdam:
Thanks for the thorough re-review — both findings were spot-on. Fixed both, and the count-mismatch investigation turned up a deeper root cause worth calling out.

## 🟠 Count mismatch — fixed, and it exposed a bigger coverage bug

You were right that the `!isDraftCovered` backstop wasn't reconciled with the display surfaces. I took your suggested approach (one predicate everywhere) and chose to block intentional duplicates entirely rather than preserve the "re-check to create a duplicate on purpose" path — creating a known duplicate only ever produces a backend rejection, so there's no real use case worth the footgun.

One creatable predicate now drives every surface:

- `runCreateSelected`
- `selectedCount`
- `totalRules`
- The Create button
- The confirmation modal
- The flyout title
- The preview's `selectedSuggestions`

They can no longer disagree with what actually gets created.

Covered drafts are now non-selectable, so the mismatched state can't be entered in the first place:

- The inline-row checkbox disables when the draft is covered (not just based on `rowStatus`), using an authoritative per-draft `coveredKeys` set from the page rather than re-deriving coverage locally.
- The master checkbox uses a new `selectableCount` (non-covered drafts) as its denominator, so a partially covered service can still reach a fully checked "all" state. The master checkbox disables when nothing selectable remains.
- `toggleServiceSelection` only toggles selectable drafts.

### Deeper root cause found

While testing this against a running build, I found that the original coverage signal was all-or-nothing per service — `missingCanonicalPair` required both availability **and** latency.

That meant creating only a service's availability SLO still left the service marked as "missing its pair," so it was treated as not covered. The availability draft remained selected and could be submitted again, resulting in a Bad Request.

Your reproduction (`cart-avail` covered, `cart-lat` not covered) could not actually be represented correctly under the old logic.

### Fix

Coverage is now per-draft and side-aware:

- Exported `kindSide` from `common/slo/classifier.ts` as the shared source of truth:
  - Any `*-availability` kind covers the availability side.
  - Any `*-latency` kind covers the latency side.
- Replaced the service-level `allServicesCoverage` set with a per-service coverage map:

  ```ts
  {
    availability: boolean;
    latency: boolean;
  }
  ```

- `isDraftCovered` now checks the specific side represented by the draft.
- Creating only the availability SLO marks exactly the availability draft as covered while leaving the missing latency draft selectable.
- Kept a separate `fullyCoveredServices` set (both sides covered) for disabling entire services in the filter, preserving the existing behavior.

Net result: what gets created, what is counted, and what is selectable now all represent the same set. A partial create is reflected correctly in the UI as soon as coverage refreshes.

I did not change the create-time error surface. A genuine create-while-coverage-is-still-loading race could still surface a raw Bad Request. I'm happy to add graceful "already exists" handling as a follow-up, but with covered drafts now blocked, that race window is significantly narrower.

## 🟢 Orphaned doc comment — fixed

Moved `buildSuggestSearch` above the `parseSuggestScopeFromSearch` JSDoc so each function is once again paired with its own documentation block.

lezzago (approved):

PR #2796, opened 2026-07-22
feat(alerting): redesign Prometheus rule creation with Builder/Code toggle

riysaxen-amzn:
### Description

Simplify the "Create metrics rule" flyout for managed Prometheus (AMP) by aligning it with what Prometheus actually offers. The PromQL query itself now defines the alert condition — no separate threshold configuration.

#### Query section (builder-only)

- Point-and-click builder: metric dropdown + label name/operator/value filters, backed by live metadata from the datasource
- Label values are escaped before PromQL interpolation (quotes/backslashes)
- Datasource picker inline in the toolbar
- **For duration** select — the rule's per-rule `for:` clause (condition must hold this long before the alert transitions pending → firing; `0s` fires immediately)

#### Removed (not applicable to managed Prometheus)

- **Trigger condition section** (Operator/Value/Unit) — the PromQL expression is the complete alert condition. Server-side, `operator`/`threshold` are now optional legacy payload fields; when absent, `expr` is the query verbatim
- **Code mode / freeform PromQL editor** (query library, metric browser popovers)
- **Evaluation Settings** (Eval Interval, Pending Period, Firing Period) — these are rule-group-level concerns in AMP, not per-rule

#### Rule group

- Rule Group selector: pick from existing groups (fetched live from the ruler API) or type to create a new one
- Group selection travels through form state as a `_ruleGroup` metadata label, which submission extracts into the payload's `groupName` and strips from persisted labels — it never reaches Alertmanager routing

#### ⚠️ Shared rule-group safety (server-side behavior change)

Making groups user-selectable exposed a latent data-loss hazard: Cortex's rule-group POST is **create-or-replace**, so writing a single rule into an existing shared group would have silently wiped its sibling rules. This PR changes the mutation handlers:

- **Create** merges into the existing group — the new rule replaces any same-named rule, siblings are preserved
- **Delete** accepts an optional `?ruleName=` and splices just that rule out of the group; the group is deleted only when it becomes empty
- **Edit** removes the old rule copy when a rule is renamed or moved between groups (rule-level splice, not group delete)

Covered by the new `prometheus_handlers.test.ts` (expr building, group merge, per-rule splice).

#### Kept

- Labels section (drives Alertmanager routing)
- Annotations section (summary, description, runbook_url)
- Notification routing guidance (Alertmanager-based)
- YAML preview accordion (rule group namespace, expr verbatim, `_ruleGroup` excluded)

#### Monitor detail flyout

- Hide Evaluation Interval / Pending Period for Prometheus rules
- Hide the Threshold row and the "Condition: > threshold for Ns" line — both were parsing artifacts with a misleading `> 0` fallback, not stored rule data
- Show **For Duration** from the rule's `for:` value, normalized for display (`300s` → `5m`)
- Add Notification Routing section for Prometheus rules

#### Edit flow

- Seeds the query verbatim (previously regex-stripped a trailing comparison)
- Round-trips the rule group and For duration; fixed `0s` (fire immediately) being normalized to `1m` on edit

#### Rules table

- Show rule group inline under the rule name (hidden when the dedicated Rule Group column is visible)
- Rename "Group" column to "Rule Group"

All user-facing strings are wrapped in `i18n.translate()`.

### Issues Resolved

https://github.com/user-attachments/assets/a4697af1-3758-4ba3-ba08-c2af8d6d44e1

### Check List


public/components/alerting/create_monitor/prometheus_form_section.tsx
```diff
@@ -65,227 +107,418 @@ export const PrometheusFormSection: React.FC<{
...
+      syncBuilderToQuery();
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [selectedMetric, selectedLabelName, selectedLabelValue, labelOperator]);
 
   const updateThreshold = <K extends keyof ThresholdCondition>(
     key: K,
     value: ThresholdCondition[K]
   ) => {
     onUpdate('threshold', { ...form.threshold, [key]: value });
```
lezzago: 🔴 **"For duration" is now discarded on save.** This correctly drops the old `forDuration → pendingPeriod` sync. But the submission path was not updated to match: `alarms_page.tsx:909` still does `const resolvedForDuration = promForm.pendingPeriod || promForm.threshold.forDuration;`. Since nothing writes `pendingPeriod` anymore and `DEFAULT_PROM_FORM.pendingPeriod` is the truthy default `'5m'`, `resolvedForDuration` is **always `5m`** — the user's selection here never reaches the payload.

Repro: set For duration = `30s` → Save → rule persists with `for: 5m`.

Fix (in `alarms_page.tsx`, both create and edit): use `promForm.threshold.forDuration`, or remove `pendingPeriod` from `PrometheusFormState` / `DEFAULT_PROM_FORM` so the `||` fallback works.

public/components/alerting/create_monitor/prometheus_form_section.tsx
```diff
@@ -65,227 +107,418 @@ export const PrometheusFormSection: React.FC<{
...
+  // Use a ref for form.labels to avoid circular dependency:
+  // handleRuleGroupChange → onUpdate('labels') → parent re-renders → new form.labels → new callback
+  const formLabelsRef = useRef(form.labels);
+  formLabelsRef.current = form.labels;
+
+  const handleRuleGroupChange = useCallback(
+    (value: string) => {
+      setRuleGroupName(value);
+      // Store as a metadata label so it's available during submission.
+      // The parent form's submission handler reads _ruleGroup and strips it before persisting.
```
lezzago: 🔴 **This comment is inaccurate — no such stripping happens.** `handleCreateMonitor` (`alarms_page.tsx:917`) maps every label with a `key && value` filter, so `_ruleGroup` is persisted as a real Prometheus label and will influence Alertmanager routing. It's also never converted into `groupName`: the create path passes no `groupName` at all (only the edit path at `:984` does), so the Rule Group selector has no functional effect on newly created rules.

Please update `handleCreateMonitor` to extract `_ruleGroup` into the payload's `groupName` and filter it out of `labels`.

public/components/alerting/create_monitor/prometheus_form_section.tsx
```diff
@@ -65,227 +107,418 @@ export const PrometheusFormSection: React.FC<{
...
+              />
+              <div style={{ position: 'absolute', top: 4, right: 4, zIndex: 2 }}>
+                <EuiToolTip content="Copy query">
+                  <EuiButtonIcon
+                    iconType="copy"
+                    size="s"
+                    color="subdued"
+                    onClick={() => {
+                      try {
+                        navigator.clipboard.writeText(form.query);
```
lezzago: 🟡 `navigator.clipboard.writeText()` returns a Promise, so this synchronous `try/catch` won't catch a rejection (e.g. clipboard permission denied) — it surfaces as an unhandled rejection. Use `navigator.clipboard.writeText(form.query).catch(() => { /* clipboard unavailable */ })`.

public/components/alerting/create_monitor/prometheus_form_section.tsx
```diff
@@ -65,227 +107,418 @@ export const PrometheusFormSection: React.FC<{
...
+                    setSelectedLabelValue([]);
+                  }}
+                />
+              </EuiFlexItem>
+            </EuiFlexGroup>
+            <EuiSpacer size="s" />
             <EuiText size="xs" color="subdued">
-              {'Example: rate(http_requests_total{job="api"}[5m]) > 100'}
+              {i18n.translate('observability.alerting.prometheusFormSection.builderHelpText', {
+                defaultMessage: 'Select a metric to start.',
```
lezzago: 🟡 i18n: this is one of several strings that were switched to plain literals in this file (also `Labels`, `Annotations`, `Optional`, `Rule Preview (YAML)`, `Metric` / `Label name` / `Label value` / `Datasource` / `Group name`, and the Alertmanager callout copy). Since the `i18n` import is still here, please keep these wrapped in `i18n.translate(...)` for translation coverage.

public/components/alerting/create_monitor/prometheus_form_section.tsx
```diff
@@ -65,227 +107,418 @@ export const PrometheusFormSection: React.FC<{
...
+      .then(({ values }) => {
+        setLabelValueOptions(values.map((v) => ({ label: v })));
+      })
+      .catch(() => {
+        /* non-critical */
+      });
+  }, [datasourceId, selectedLabelName, selectedMetric]);
+
+  // Sync builder selections to PromQL query
+  const syncBuilderToQuery = useCallback(() => {
```
lezzago, line 240: 🟢 Minor: `syncBuilderToQuery` early-returns when no metric is selected, so builder state is never reconciled back from `form.query`, and clearing the metric leaves the last query in place. Acceptable for v1, but a switch from Code → Builder (or clearing the metric) can leave the two views out of sync. Worth a follow-up.

lezzago (changes requested):
Thanks for this — the direction (aligning the flyout with how managed Prometheus actually works) is a clear improvement. I reviewed the diff and traced the submission path in `alarms_page.tsx` (which this PR doesn't touch). Two of the findings are correctness bugs in that untended submission path, so they aren't caught by CI (there's no Jest job in this pipeline). Details inline; summary below.

**Confirmed findings**

1. 🔴 **"For duration" is discarded on save.** `alarms_page.tsx:909` still does `const resolvedForDuration = promForm.pendingPeriod || promForm.threshold.forDuration;`. This PR removes every code path that writes `pendingPeriod`, but `DEFAULT_PROM_FORM.pendingPeriod` is `'5m'` (truthy) and now never changes — so the payload's `forDuration` is hard-pinned to `5m` regardless of what the user selects. Repro: set For duration = `30s` → Save → rule persists with `for: 5m`. Fix in `alarms_page.tsx` (create + edit paths): read `promForm.threshold.forDuration`, or drop `pendingPeriod` from the type/default entirely.

2. 🔴 **`_ruleGroup` leaks into labels, and the group name is never sent on create.** The comment at line 141 says the parent "reads `_ruleGroup` and strips it before persisting" — no such code exists. `handleCreateMonitor` (`alarms_page.tsx:917`) maps *all* labels through a `key && value` filter, so `_ruleGroup` is persisted as a real Prometheus label (it will affect Alertmanager routing). The create path also passes no `groupName` (only the edit path at `:984` does), so the new Rule Group selector has no functional effect on newly created rules. The YAML preview correctly hides it, but that's cosmetic only.

3. 🟡 **i18n regression.** Several strings were de-internationalized to plain literals (`Labels`, `Annotations`, `Optional`, `Rule Preview (YAML)`, all Builder-mode labels, the Alertmanager callout copy, etc.). The `i18n` import is still present, so please wrap the new/reverted user-facing strings.

4. 🟡 **Clipboard error handling can't catch its target.** `navigator.clipboard.writeText()` returns a Promise; the synchronous `try/catch` won't catch a permission rejection → unhandled rejection. Use `.catch(() => {})`.

5. 🟢 **Builder ⇄ Code state isn't reconciled with `form.query`** (minor): builder selections live in local state, and clearing the metric doesn't clear the query. Fine for v1, worth a follow-up.

Also note the tests only exercise `PrometheusFormSection` in isolation with a mocked `onUpdate` — they assert `_ruleGroup` is *emitted* but nothing asserts it's *stripped* on persist, and nothing asserts the saved payload's `forDuration` matches the selection. A `handleCreateMonitor` / `buildPrometheusRulePayload`-level test would catch #1 and #2.

(For reference: the red `enforce-label` and `linkchecker` checks are unrelated to this diff — a missing PR label and pre-existing dead opentelemetry.io links in untouched `getting_started_artifacts/` docs.)

joshuali925:
### Code review

Found 4 issues:

1. Create-merge into a shared rule group silently resets the group's evaluation `interval`. The merge preserves sibling rules but rebuilds the group via `buildRuleGroup(payload)`, whose `interval` comes from the new payload's `evaluationInterval`; on merge only `group.rules` is reassigned, so `existing.interval` is lost. Because the redesigned form removed the interval control, create always sends the default `1m`, so adding a rule to a group configured at `5m` silently changes every sibling rule's cadence. The delete-splice path does it correctly via `{ ...existing, rules: remaining }` — the create path is the asymmetric outlier. (bug: `existing.interval` not preserved on merge)

https://github.com/opensearch-project/dashboards-observability/blob/0a775415758fe465783ec4c8053ddf56b1de312b/server/routes/alerting/mutations/prometheus_handlers.ts#L104-L112

2. Edit-mode builder overwrites seeded complex PromQL, causing data loss. On edit `form.query` is seeded with the existing expression, but the builder state is never parsed back from it. The effect fires `syncBuilderToQuery()` as soon as `selectedMetric` becomes non-empty, unconditionally calling `onUpdate('query', ...)`. The first metric selection replaces a seeded expression like `rate(http_requests_total[5m])` with the bare builder query, and with Code mode removed there is no way to preserve or restore the original. (bug: no round-trip parse of seeded query)

https://github.com/opensearch-project/dashboards-observability/blob/0a775415758fe465783ec4c8053ddf56b1de312b/public/components/alerting/create_monitor/prometheus_form_section.tsx#L161-L181

3. Unguarded async fetches can render stale options (stale-response race). These newly-added `useEffect` fetches have no cancellation, so rapidly switching datasource/metric lets a slower earlier response overwrite a later one. Similar concerns were addressed on prior alerting PRs. (bug: no AbortController / stale-write guard)

https://github.com/opensearch-project/dashboards-observability/blob/0a775415758fe465783ec4c8053ddf56b1de312b/public/components/alerting/create_monitor/prometheus_form_section.tsx#L104-L160

4. The transport-only `_ruleGroup` metadata label leaks into the `LabelEditor`. `form.labels` is passed unfiltered, so `_ruleGroup` is visible, editable, and deletable there — desyncing it from the Rule Group selector — even though submission and the YAML preview both filter it out. (bug: unfiltered labels prop)

https://github.com/opensearch-project/dashboards-observability/blob/0a775415758fe465783ec4c8053ddf56b1de312b/public/components/alerting/create_monitor/prometheus_form_section.tsx#L467-L470

public/components/alerting/create_monitor/prometheus_form_section.tsx
riysaxen-amzn: Fixed in 4c508535 — the submission path now reads `promForm.threshold.forDuration` directly (the `pendingPeriod ||` fallback is gone). The For duration control was re-added in 9a484fca after a design discussion (the `for:` clause is per-rule and can't be expressed in the query). Covered by a test asserting the select updates `threshold.forDuration`.

public/components/alerting/create_monitor/prometheus_form_section.tsx
riysaxen-amzn: Fixed in 4c508535 — `handleCreateMonitor` and `handleEditMonitor` now extract `_ruleGroup` into the payload's `groupName` and filter it out of persisted labels, so it never reaches Alertmanager routing. This also exposed a group-clobbering hazard (Cortex POST is create-or-replace), so the create handler now merges into existing groups and delete splices per-rule — see `prometheus_handlers.test.ts`.

public/components/alerting/create_monitor/prometheus_form_section.tsx
riysaxen-amzn: Moot as of 4c508535 — Code mode (and the copy button) was removed entirely per design review; the form is now builder-only.

public/components/alerting/create_monitor/prometheus_form_section.tsx
riysaxen-amzn: Done in 4c508535 — all user-facing strings in the form section (titles, labels, placeholders, help text, the Alertmanager callout) are wrapped in `i18n.translate()` / `FormattedMessage`.

public/components/alerting/create_monitor/prometheus_form_section.tsx
riysaxen-amzn: Addressed in ec0915ea — builder selections are now seeded by parsing builder-shaped expressions (`metric` or `metric{label op "value"}`) back into the combo boxes, so edits round-trip. Complex expressions leave the builder empty and the builder→query sync inert, so a seeded query is never clobbered on mount. (Code mode itself was removed, so the Code→Builder desync scenario no longer exists.)

riysaxen-amzn:
Thanks @joshuali925 — all 4 addressed: 1. **Interval reset on merge** — fixed in ec0915ea (your review was against 0a77541): the create-merge path now preserves `existing.interval`, matching the delete-splice path. Covered by a test asserting a `1m` payload doesn't overwrite a `300s` group. 2. **Edit-mode builder clobbering seeded PromQL** — fixed in ec0915ea: builder-shaped expressions (`metric` / `metric{label op \"value\"}`) are parsed back into builder selections so they round-trip; complex expressions leave the builder empty and the builder→query sync inert, so the seeded expression is only replaced if the user explicitly picks a new metric. 3. **Stale-response races** — fixed in f2eca859: all three fetch effects now set a `stale` flag in their cleanup, so out-of-order responses from a previous datasource/metric selection can't overwrite current options. 4. **`_ruleGroup` leaking into LabelEditor** — fixed in f2eca859: the transport label is filtered from the editor's `labels` prop and re-appended on every label edit, so it can't be seen, edited, or deleted there (covered by a test).

joshuali925 (approved):

mengweieric (approved):

lezzago:
## Overview

Simplifies the "Create metrics rule" flyout for managed Prometheus (AMP): replaces the freeform PromQL editor + threshold/evaluation sections with a point-and-click builder (metric + single label filter), a user-selectable rule group, and a per-rule `for:` duration. The PromQL expression is now the complete alert condition — `operator`/`threshold` become optional legacy payload fields.

The most consequential change is server-side: because Cortex's rule-group POST is **create-or-replace**, making groups user-selectable exposed a data-loss hazard. The PR reworks create to **merge** into existing groups and delete to **splice** a single rule (deleting the group only when empty). This is the right fix and is well-tested in the new `prometheus_handlers.test.ts`.

Overall a solid, well-documented PR with good test coverage. A few issues below.

## Issues & Risks

### 🔴 Lint break — `monitors_table_columns.tsx:110`
The PR removes the `// eslint-disable-next-line @typescript-eslint/no-explicit-any` comment but **leaves `const cols: Array<Record<string, any>>` unchanged**. The rule is active across this package (7+ sibling suppressions in `alarms_page_helpers.ts`, `notification_routing_panel.tsx`, `monitors_eui_table.tsx`, etc.), so removing the suppression while keeping `any` will fail `yarn lint`. Either restore the comment or remove the `any`.

### 🟠 Hardcoded user-facing strings contradict the PR claim — `monitor_detail_flyout.tsx` (Notification Routing block)
The PR body states *"All user-facing strings are wrapped in `i18n.translate()`"*, but the new Notification Routing block in the detail flyout is not:
- `buttonContent={<strong>Notification Routing</strong>}`
- `"Notifications for Prometheus alerts are managed through Alertmanager..."`
- `<EuiLink href="#/routing">Configure notification routing →</EuiLink>`

The equivalent block in `prometheus_form_section.tsx` *is* fully i18n-wrapped, so this is an inconsistency, not a deliberate exception. Wrap these strings.

Also note the raw `href="#/routing"` — the rest of this codebase deep-links via `coreRefs.application.navigateToApp(...)` (see `toast_helpers.tsx`). A bare hash `href` bypasses basepath handling; prefer the established `navigateToApp` pattern for consistency.

### 🟡 Read-modify-write on shared groups is not atomic — `prometheus_handlers.ts`
`handleCreatePrometheusRule` and the splice path in `handleDeletePrometheusRule` both do `getRuleGroup` → mutate → `upsertRuleGroup`. This fixes the single-client replace hazard, but two concurrent writers targeting the same group can still lose an update (TOCTOU) — the second upsert overwrites the first's merge. Since the whole feature is about shared-group safety, worth a comment acknowledging the remaining window, or a follow-up. Not blocking (the Cortex ruler offers no CAS), but shouldn't be presented as fully safe.

### 🟡 Builder does not clear the query when the metric is cleared — `prometheus_form_section.tsx`
`syncBuilderToQuery` early-returns when `selectedMetric.length === 0`, so clearing the metric combo box leaves the last-built `form.query` stale in form state (and in the YAML preview). This is defensible (avoids clobbering a seeded complex query), but clearing the metric and seeing the old expr persist is surprising. Consider clearing `query` when the builder is emptied *and* was the source of the current query.

### 🟢 Minor: `0s` round-trip depends on string input — `edit_monitor.tsx:185` + `utils/duration.ts`
`normalizeDuration('0s')` correctly yields `'0s'` (matches the new `Immediately (0s)` option). But if `firstRule.for` arrives as the numeric `0`, `String(0)` → `'0'`, which matches neither regex and is returned as `'0'` — not a valid `DURATION_OPTIONS` value, so the select won't reflect it. Prometheus `for:` is normally a string so this is likely fine; confirm the rule payload never delivers a bare number.

## Strengths

- **Excellent server-side test coverage**: `prometheus_handlers.test.ts` covers merge, sibling preservation, interval preservation, same-name replace, per-rule splice, and empty-group deletion — exactly the risky paths.
- **`parseBuilderQuery`** correctly returns `null` for non-representable expressions (aggregations, comparisons, multi-matcher), and the mount-time test confirms complex seeded queries aren't clobbered — good defensive design.
- **Escaping**: label values are escaped for PromQL interpolation and symmetrically unescaped in `parseBuilderQuery`, with a test covering quotes/backslashes.
- **Backward compatibility**: `operator`/`threshold` made `schema.maybe(...)` with legacy append logic kept and tested.
- **Good refactors**: shared `utils/duration.ts`, stale-guard flags on all async `useEffect` fetches, and the `formLabelsRef` pattern to break the `onUpdate → re-render` callback cycle are all handled thoughtfully.

riysaxen-amzn:
Thanks @lezzago — all items from your latest review addressed in 6c1e1cf9: - 🔴 **Lint**: replaced the `any` with `unknown` in `monitors_table_columns.tsx` (matches the function's declared `Array<Record<string, unknown>>` return type), so no suppression is needed at all. - 🟠 **i18n + routing link**: the detail flyout's Notification Routing block is now fully i18n-wrapped, and the deep link uses `coreRefs.application.navigateToApp(...)` + `hashchange` dispatch per the `toast_helpers` pattern instead of a bare `href=\"#/routing\"`. - 🟡 **TOCTOU**: added comments on both the create-merge and delete-splice paths acknowledging the non-atomic read-modify-write window (no CAS on the Cortex ruler) — protects single-writer clobber, not concurrent writers. - �� **Clear query on metric clear**: already fixed in 84a31d9d with exactly the semantics you describe — the query clears only when the builder authored it (tracked via ref), so a seeded complex expression is never wiped. - 🟢 **Numeric `for: 0`**: `normalizeDuration` now treats bare numeric strings as seconds, so a YAML `for: 0` round-trips to the `Immediately (0s)` option; added `duration.test.ts` covering it. And thanks for the shared-group leniency, but no concession needed — the merge/splice fix was small and is fully tested.

lezzago (approved):

PR #2800, opened 2026-07-27
chore: resolutions and dependencies cleanup, address CVEs

tomkdgun:
### Description

There are many leftover resolutions/overrides from past fixes that are no longer relevant and necessary after recent changes.
We need to attempt to remove as many manual resolutions as possible to clean up our dependencies.
Some of them are bringing in modules that are no longer in the dependency tree, override newer versions by older or are not aligned with OpenSearch Dashboard 3.8.0 dependencies.
In some cases resolutions were used instead of dependencies/devDependencies, or completely assume that some specific dependencies will be provided by OpenSearch Dashboard.

Changes:
- Cleaning up resolutions/dependencies/devDependencies
- Align with versions used on OpenSearch Dashboard 3.8
- Check CVEs by `yarn audit` and address these possible to fix

Details:
- adding missing dependency not present on OpenSearch Dashboard and used in code: @egjs/hammerjs
- adding missing peerDependencies related to react-graph-vis: vis-network, vis-utils, vis-data
- align plotly.js-dist version with OpenSearch Dashboard
- adding resolutions for mocha to address CVE `"**/mocha/**/minimatch": "^10.1.1"` and `"diff": "^8.0.4"`
- remaining changes are result of removing not used dependencies, and updates (transitive dependencies change)

Testing:
- `yarn osd bootstrap` with OpenSearch Dashboard 3.8
- `yarn build`
- `yarn test:jest`

### Issues Resolved

### Check List


TackAdam:
Thanks for the cleanup — the direction is solid. A couple of things I'd like addressed before merge:

## 1. `vis-data` is pinned twice, shipping two copies (blocking)

The exact pin:

```json
"vis-data": "7.1.6"
```

doesn't satisfy `react-graph-vis`'s `vis-data@^7.1.2` requirement, so the lockfile now carries both:

```text
vis-data@7.1.6:   version "7.1.6"    ← this PR's direct dependency
vis-data@^7.1.2:  version "7.1.10"   ← react-graph-vis's copy
```

`vis-network` / `react-graph-vis` will use `7.1.10`, not the pinned `7.1.6`, so we end up bundling two versions of `vis-data`.

I'd suggest changing it to:

```json
"vis-data": "^7.1.2"
```

so it dedupes to a single copy (or adding a `resolution` if `7.1.6` genuinely needs to be forced everywhere).

The exact pin is also inconsistent with the `^` ranges used for the other new dependencies.

## 2. Dependency ordering got scrambled (minor)

The `react-*` and `vis-*` entries are currently interleaved:

```json
"react-graph-vis": "^1.0.7",
"vis-network": "^9.1.13",
"react-paginate": "^8.1.3",
"react-plotly.js": "^2.6.0",
"vis-data": "7.1.6",
"vis-util": "^5.0.7",
```

Could you group the three `vis-*` dependencies together (alphabetically, or at the end of the block) to keep the dependency list easier to scan?

## 3. Manual smoke test

Since `service_map.tsx` (Trace Analytics) is the only `react-graph-vis` consumer and is sensitive to `vis-data` version mismatches, could you confirm the Service Map still renders after a `yarn osd bootstrap`?

The listed bootstrap/build/Jest checks wouldn't catch a runtime rendering regression there.

Everything else — the stale resolution removals, `@egjs/hammerjs`, the Plotly/type bumps, and the CVE resolutions — checked out cleanly.

tomkdgun:
@TackAdam Thanks for review, comments addressed, vis-data and vis-network versions are currently untouched in my pull request, I mean yarn.lock versions are same as on `main`

TackAdam (approved):

PR #2804, opened 2026-08-07
Add detector and forecaster management to Alerts Manager

jackiehanyang:
## Summary
Adds detector and forecaster management to Alerts Manager so anomaly detection and forecasting rules can be viewed and managed without jumping back to the legacy plugin pages.

https://github.com/user-attachments/assets/6206dc21-3776-4061-876f-81805c8522a4

## Changes
- Adds anomaly detection and forecasting create flows in the Alerts Manager Rules tab.
- Adds detector and forecaster detail flyouts with edit, start, stop, and delete display/actions aligned with existing rule flyout patterns.
- Adds AOS/AOSS datasource gating for detector and forecaster creation, including AOSS custom result index handling.
- Threads detector/forecaster state, health, and mutation support through the Alerts Manager table and backend service layer.
- Adds tests for detector/forecaster flyouts, monitor table behavior, mutation client behavior, and alert service routing/utilities.

## Validation
- `git diff --check origin/main...HEAD`
- `yarn lint:es` against changed TS/TSX/JS files

Note: this PR branch was rebuilt from current `main` as a single signed-off commit so CI compares against the intended base and excludes unrelated package/build-script artifacts.

### DIFF
--- a/public/components/alerting/create_ad_rule_flyout.tsx
+++ b/public/components/alerting/create_ad_rule_flyout.tsx
@@ -61,6 +61,7 @@ import { toAdApiDataSourceId, withAdApiDataSource } from './utils/ad_api_paths';
 import { useIndexMappings } from './hooks/use_index_mappings';
 import { useIndices } from './hooks/use_indices';
 import { useRuleDetail } from './hooks/use_rule_detail';
+import { isStandardOpenSearchDatasource } from './shared_constants';
 
 export type CreateAdRuleType = 'detector' | 'forecaster';
 type AdRuleFlyoutMode = 'create' | 'edit';
@@ -349,11 +350,11 @@ const parseFilterQuery = (value: string): Record<string, unknown> => {
   return trimmed ? JSON.parse(trimmed) : { match_all: {} };
 };
 
-const getOpenSearchDatasources = (datasources: Datasource[]): Datasource[] =>
-  datasources.filter((datasource) => datasource.type === 'opensearch');
+export const getCreateAdRuleDatasources = (datasources: Datasource[]): Datasource[] =>
+  datasources.filter(isStandardOpenSearchDatasource);
 
 const getInitialDatasourceId = (datasources: Datasource[], selectedDsIds?: string[]): string => {
-  const openSearchDatasources = getOpenSearchDatasources(datasources);
+  const openSearchDatasources = getCreateAdRuleDatasources(datasources);
   const selected = selectedDsIds
     ?.map((id) => openSearchDatasources.find((datasource) => datasource.id === id))
     .find(Boolean);
@@ -531,9 +532,9 @@ const buildFeatureFormState = (
 };
 
 const getFeatureForms = (resource: Record<string, unknown>): CreateAdFeatureFormState[] => {
-  const features = asArray(getField(resource, 'feature_attributes', 'featureAttributes')).map(
-    (feature) => buildFeatureFormState(resource, feature)
-  );
+  const features = asArray(
+    getField(resource, 'feature_attributes', 'featureAttributes')
+  ).map((feature) => buildFeatureFormState(resource, feature));
   return features.length > 0 ? features : [createInitialFeature()];
 };
 
@@ -1175,18 +1176,18 @@ const errorsForStep = (
       step === 0
         ? ['name', 'datasourceId', 'indices', 'timeField', 'filterQuery', 'resultIndex']
         : step === 1
-          ? [
-              'features',
-              'categoryField',
-              'interval',
-              'frequency',
-              'windowDelay',
-              'history',
-              'shingleSize',
-            ]
-          : step === 2
-            ? []
-            : Object.keys(errors);
+        ? [
+            'features',
+            'categoryField',
+            'interval',
+            'frequency',
+            'windowDelay',
+            'history',
+            'shingleSize',
+          ]
+        : step === 2
+        ? []
+        : Object.keys(errors);
   } else {
     stepFields =
       step === 0
@@ -1200,18 +1201,18 @@ const errorsForStep = (
             'categoryField',
           ]
         : step === 1
-          ? [
-              'interval',
-              'windowDelay',
-              'history',
-              'horizon',
-              'shingleSize',
-              'resultIndex',
-              'resultIndexMinAge',
-              'resultIndexMinSize',
-              'resultIndexTtl',
-            ]
-          : Object.keys(errors);
+        ? [
+            'interval',
+            'windowDelay',
+            'history',
+            'horizon',
+            'shingleSize',
+            'resultIndex',
+            'resultIndexMinAge',
+            'resultIndexMinSize',
+            'resultIndexTtl',
+          ]
+        : Object.keys(errors);
   }
 
   return Object.fromEntries(
@@ -1741,8 +1742,8 @@ const SuggestParametersDialog: React.FC<{
         typeof intervalValue === 'number'
           ? intervalValue
           : typeof frequencyValue === 'number'
-            ? frequencyValue
-            : 10;
+          ? frequencyValue
+          : 10;
 
       const historyValue = payload.history;
       const windowDelayValue = payload.windowDelay?.period?.interval;
@@ -1953,8 +1954,14 @@ export const CreateAdRuleFlyout: React.FC<CreateAdRuleFlyoutProps> = ({
   onCreated,
   onUpdated,
 }) => {
-  const openSearchDatasources = useMemo(() => getOpenSearchDatasources(datasources), [datasources]);
   const isEdit = mode === 'edit' && !!editTarget;
+  const openSearchDatasources = useMemo(
+    () =>
+      isEdit
+        ? datasources.filter((datasource) => datasource.type === 'opensearch')
+        : getCreateAdRuleDatasources(datasources),
+    [datasources, isEdit]
+  );
   const isDetectorSettingsEdit =
     isEdit && ruleType === 'detector' && editTarget?.initialStep !== 'model';
   const isDetectorModelEdit =
@@ -2009,12 +2016,12 @@ export const CreateAdRuleFlyout: React.FC<CreateAdRuleFlyoutProps> = ({
           defaultMessage: 'Edit forecasting rule',
         })
     : isDetector
-      ? i18n.translate('observability.alerting.createAdRuleFlyout.detectorTitle', {
-          defaultMessage: 'Create anomaly detection rule',
-        })
-      : i18n.translate('observability.alerting.createAdRuleFlyout.forecasterTitle', {
-          defaultMessage: 'Create forecasting rule',
-        });
+    ? i18n.translate('observability.alerting.createAdRuleFlyout.detectorTitle', {
+        defaultMessage: 'Create anomaly detection rule',
+      })
+    : i18n.translate('observability.alerting.createAdRuleFlyout.forecasterTitle', {
+        defaultMessage: 'Create forecasting rule',
+      });
   const detectorStepTitles = isEdit
     ? [
         isDetectorModelEdit
@@ -2468,12 +2475,12 @@ export const CreateAdRuleFlyout: React.FC<CreateAdRuleFlyoutProps> = ({
               defaultMessage: 'Forecasting rule updated successfully',
             })
         : isDetector
-          ? i18n.translate('observability.alerting.createAdRuleFlyout.detectorCreatedToast', {
-              defaultMessage: 'Anomaly detection rule created successfully',
-            })
-          : i18n.translate('observability.alerting.createAdRuleFlyout.forecasterCreatedToast', {
-              defaultMessage: 'Forecasting rule created successfully',
-            });
+        ? i18n.translate('observability.alerting.createAdRuleFlyout.detectorCreatedToast', {
+            defaultMessage: 'Anomaly detection rule created successfully',
+          })
+        : i18n.translate('observability.alerting.createAdRuleFlyout.forecasterCreatedToast', {
+            defaultMessage: 'Forecasting rule created successfully',
+          });
 
       if (isEdit && isDetector && shouldOfferStartDetectorAfterEdit && editTarget) {
         setStartAfterEditPrompt({
@@ -2521,12 +2528,12 @@ export const CreateAdRuleFlyout: React.FC<CreateAdRuleFlyoutProps> = ({
                 defaultMessage: 'Failed to update forecasting rule',
               })
           : isDetector
-            ? i18n.translate('observability.alerting.createAdRuleFlyout.detectorCreateFailed', {
-                defaultMessage: 'Failed to create anomaly detection rule',
-              })
-            : i18n.translate('observability.alerting.createAdRuleFlyout.forecasterCreateFailed', {
-                defaultMessage: 'Failed to create forecasting rule',
-              }),
+          ? i18n.translate('observability.alerting.createAdRuleFlyout.detectorCreateFailed', {
+              defaultMessage: 'Failed to create anomaly detection rule',
+            })
+          : i18n.translate('observability.alerting.createAdRuleFlyout.forecasterCreateFailed', {
+              defaultMessage: 'Failed to create forecasting rule',
+            }),
         text: message,
       });
     } finally {
@@ -4136,12 +4143,12 @@ export const CreateAdRuleFlyout: React.FC<CreateAdRuleFlyoutProps> = ({
           }
         )
       : isAwaitingDataToInit || isAwaitingDataToRestart
-        ? i18n.translate('observability.alerting.createAdRuleFlyout.cancelForecastToEditTitle', {
-            defaultMessage: 'Cancel forecast to edit?',
-          })
-        : i18n.translate('observability.alerting.createAdRuleFlyout.stopForecastToEditTitle', {
-            defaultMessage: 'Stop forecast to edit?',
-          });
+      ? i18n.translate('observability.alerting.createAdRuleFlyout.cancelForecastToEditTitle', {
+          defaultMessage: 'Cancel forecast to edit?',
+        })
+      : i18n.translate('observability.alerting.createAdRuleFlyout.stopForecastToEditTitle', {
+          defaultMessage: 'Stop forecast to edit?',
+        });
     const forecastModalDescription = (() => {
       if (isInitializingForecast) {
         return i18n.translate(
@@ -4312,14 +4319,14 @@ export const CreateAdRuleFlyout: React.FC<CreateAdRuleFlyoutProps> = ({
                     }
                   )
               : isDetector
-                ? i18n.translate('observability.alerting.createAdRuleFlyout.detectorSubtitle', {
-                    defaultMessage:
-                      'Create an anomaly detection rule using the same model definition fields from the AD workflow.',
-                  })
-                : i18n.translate('observability.alerting.createAdRuleFlyout.forecasterSubtitle', {
-                    defaultMessage:
-                      'Create a forecasting rule using the same data source and model parameter fields from the Forecasting workflow.',
-                  })}
+              ? i18n.translate('observability.alerting.createAdRuleFlyout.detectorSubtitle', {
+                  defaultMessage:
+                    'Create an anomaly detection rule using the same model definition fields from the AD workflow.',
+                })
+              : i18n.translate('observability.alerting.createAdRuleFlyout.forecasterSubtitle', {
+                  defaultMessage:
+                    'Create a forecasting rule using the same data source and model parameter fields from the Forecasting workflow.',
+                })}
           </EuiText>
         </EuiFlyoutHeader>
 
--- a/public/components/alerting/create_ad_rule_flyout.tsx
+++ b/public/components/alerting/create_ad_rule_flyout.tsx
@@ -56,6 +56,7 @@ import {
 import { i18n } from '@osd/i18n';
 import { ADDetector, ADForecaster, Datasource } from '../../../common/types/alerting';
 import { coreRefs } from '../../framework/core_refs';
+import { toAdApiDataSourceId, withAdApiDataSource } from './utils/ad_api_paths';
 import { useIndexMappings } from './hooks/use_index_mappings';
 import { useIndices } from './hooks/use_indices';
 import { useRuleDetail } from './hooks/use_rule_detail';
@@ -139,7 +140,6 @@ interface DetectorSuggestionResponse {
 
 const AD_DETECTOR_API = '/api/anomaly_detectors/detectors';
 const FORECASTER_API = '/api/forecasting/forecasters';
-const LOCAL_OPENSEARCH_DATASOURCE_IDS = new Set(['', 'local', 'local-cluster', 'local_cluster']);
 const OPENSEARCH_SERVERLESS_ENGINE_TYPE = 'OpenSearch Serverless';
 const OPENSEARCH_SERVERLESS_SIGV4_SERVICE = 'aoss';
 const DETECTOR_RUNNING_STATE = 'Running';
@@ -223,16 +223,8 @@ const createInitialForm = (datasourceId: string): CreateAdRuleFormState => ({
   startAfterCreate: true,
 });
 
-const normalizeDataSourceId = (dsId?: string): string =>
-  !dsId || LOCAL_OPENSEARCH_DATASOURCE_IDS.has(dsId) ? '' : dsId;
-
-const withOptionalDatasource = (basePath: string, dsId?: string): string => {
-  const normalized = normalizeDataSourceId(dsId);
-  return normalized ? `${basePath}/${encodeURIComponent(normalized)}` : basePath;
-};
-
-const isServerlessDataSource = async (dsId?: string): Promise<boolean> => {
-  const normalized = normalizeDataSourceId(dsId);
+const isServerlessDataSource = async (dsId: string): Promise<boolean> => {
+  const normalized = toAdApiDataSourceId(dsId);
   if (!normalized || !coreRefs.savedObjectsClient) return false;
 
   try {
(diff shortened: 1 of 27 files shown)


jackiehanyang:
> ## PR Code Suggestions ✨
> Explore these optional code suggestions:
> 
> **Category**	**Suggestion                                                                                                                                    **	**Impact**
> Possible issue	
> Fix stale-state batching when syncing frequency
> **The comparison `form.frequency === form.interval` uses the pre-update `form.interval`, so the frequency-follows-interval behavior only fires when frequency already equals the OLD interval. Also, calling `updateForm` twice invokes `setForm` sequentially — the second call will operate on stale state. Use a single functional `setForm` update to keep frequency synced when it previously matched interval.**
> 
> [public/components/alerting/create_ad_rule_flyout.tsx [2824-2828]](https://github.com/opensearch-project/dashboards-observability/pull/2804/files#diff-86cce8909e6b581220398ba88b0320e9f219a545ee4191ace8fb1b7396a23bacR2824-R2828)
> 
> ```diff
>  onChange={(e) => {
>    const next = parseNumberInputValue(e.target.value);
> -  updateForm('interval', next);
> -  if (form.frequency === form.interval) updateForm('frequency', next);
> +  setForm((prev) => ({
> +    ...prev,
> +    interval: next,
> +    frequency: prev.frequency === prev.interval ? next : prev.frequency,
> +  }));
> +  setSubmitError(null);
>  }}
> ```
> 
> Suggestion importance[1-10]: 7
> __
> 
> Why: Valid concern: the comparison `form.frequency === form.interval` compares against the pre-update value which is confusing, and consecutive `updateForm` calls operate on stale state. The functional update fix is more correct and reliable.
> 
> Medium
> Verify AD lifecycle route path structure
> **The `stopDetector` path builds `/detectors/{id}/stop/false` and then appends `/{dsId}`, producing `/detectors/{id}/stop/false/{dsId}`. If the server route expects the datasource id as a query parameter or before the `false` literal, this URL pattern will 404. Verify that the backend route actually accepts the datasource id appended after the literal `false` (and similarly after `start`), otherwise start/stop will silently fail.**
> 
> [public/components/alerting/mutations/monitor_mutations_client.ts [47-49]](https://github.com/opensearch-project/dashboards-observability/pull/2804/files#diff-1ec8a7e4ead0616abdcb64cc4806cf092d7b031557633c5aea117d3e381dc792R47-R49)
> 
> ```diff
> +private withOptionalDatasource(basePath: string, dsId?: string): string {
> +    return dsId ? `${basePath}/${encodeURIComponent(dsId)}` : basePath;
> +  }
>  
> -
> ```
> 


on the medium item, checked the implementation and believe the current behavior is safe: `updateForm` uses functional `setForm`, so the queued interval and frequency updates are applied in order. The comparison against the previous `form.interval` is intentional because we only want frequency to follow interval when it matched the prior interval value. Since this would be a readability only refactor rather than a behavior fix, I’m leaning toward leaving it unchanged to avoid extra churn on this PR.

TackAdam:
## 🔴 Blocker (must fix before merge)

### 1. Build-breaking type errors: `Datasource.engineType` / `Datasource.version` don't exist

**File:** `public/components/alerting/monitors_table/index.tsx:173, 234, 236, 237`

The new `AnalyticEngine`/version gating reads `d.engineType` and `d.version` on values typed `Datasource[]` (`selectedDatasources` is explicitly narrowed to `Datasource` at line 215). The `Datasource` interface (`common/types/alerting/unified_types.ts`) declares neither field. I confirmed with an isolated compile using the repo's own `tsc`:

```text
error TS2339: Property 'engineType' does not exist on type 'Datasource'.
error TS2339: Property 'version' does not exist on type 'Datasource'.
```

Two consequences:

- Clean CI typecheck will fail (the local pass was a stale-cache artifact).
- Even if forced through, these fields are also never populated by the datasource service, so the `'OpenSearch Serverless'` and `semver.gte(version, 3.5.0)` branches are dead at runtime — every remote OS datasource (`mdsId` set) falls through to `if (!d.version) return false`, disabling Logs create unless `AnalyticEngine` happens to be on.

**Fix:** Add `engineType?: string` and `version?: string` to `Datasource` (and populate them where datasources are built), or drop the dead branches.

---

## 🟠 High

### 2. A failed auto-start reports "Failed to create" and orphans the created detector

**File:** `create_ad_rule_flyout.tsx:2174-2182` + catch at `2215`

When `startAfterCreate` is on, the follow-up `POST .../start` is awaited inside the same `try` as the create. If start fails (common while the cluster initializes), control jumps to the `catch` → `"Failed to create anomaly detection rule"` toast, and `onCreated()`/success toast never fire — but the detector already exists.

User retries → duplicate detector.

The start call should be in its own `try` with a distinct **"created but couldn't start"** message.

### 3. Edit payloads send duplicate `snake_case` + `camelCase` keys

**File:** `create_ad_rule_flyout.tsx:544-587, 635-654`

The detail API returns `snake_case` (`alert_detail.ts:74-84` spreads `anomaly_detector`). On edit, `buildRulePayload` spreads `stripReadOnlyResourceFields(existingResource)` (carrying `filter_query`, `feature_attributes`, `category_field`, `detection_interval`, `result_index`, …) and then overrides only the `camelCase` equivalents.

So the payload carries both old `category_field` and new `categoryField`, etc. Edits like **"disable categorical field"** or **"change the filter"** may silently not take effect depending on which key the API honors.

### 4. Detail flyout body goes stale after start/stop

**File:** `hooks/use_rule_detail.ts:27-48`, consumed by both flyouts

`useRuleDetail`'s effect deps are:

```ts
[service, dsId, ruleId, definitionType]
```

None change on a lifecycle action.

After Stop/Start, the header badge flips (optimistic prop update) but the body's **"Real-time job"**, **"Current state"**, and **"Enabled time"** rows keep pre-action values until the flyout is reopened. The header/body can visibly disagree (also flagged as the detail agent's #1).

**Fix:** Needs a refetch trigger tied to the mutation.

---

## 🟡 Medium

### M1. `anomalyDirection` is collected but never sent

`create_ad_rule_flyout.tsx:2718-2721` writes the `both`/`above`/`below` selection, but `buildFeatureAttributes` never reads it and the detector payload always sends:

```ts
rules: []
```

at `:675`.

The user's directional criteria is silently discarded.

### M2. Missing numeric validation → silent coercion

`horizon`/`history` have no `validateForm` checks. Forecaster inputs use `Number(e.target.value)` so clearing a field yields `0`, and `buildRulePayload` clamps to `Math.max(1, …)`.

For example:

- Clearing **Horizon** silently submits `horizon: 1`.
- Typing `history: 5` (below the documented minimum `40`) passes and the API rejects it.
- `errorsForStep` even lists `'horizon'` as expecting an error that's never produced.

### M3. Dynamic `defaultMessage` under a shared i18n ID

**File:** `create_ad_rule_flyout.tsx:2338, 2347`

Uses:

```ts
defaultMessage: isDetector ? '…detector' : '…forecaster'
```

`@osd/i18n` expects a static literal, and one ID can't hold two translations. Not a CI break here (this plugin isn't in the i18n scan scope I could find), but it's non-idiomatic and detector/forecaster copy can never diverge in translation.

**Fix:** Use two separate IDs.

### M4. N+1 runtime-profile fetch on every rule list

**File:** `alert_service.ts:1217, 1238`

`enrichAD*RuntimeStates` issues one `_profile` request per detector/forecaster. It's concurrency-limited with per-item fallback (good), but adds meaningful latency to the rules list as AD resource counts grow.

Consider whether a single bulk profile call exists.

### M5. Optimistic update immediately clobbered by refetch

**File:** `alarms_page.tsx:658-676`

`setRules(optimistic)` is followed unconditionally by `refetchRules()`.

The optimistic state is either:

- Pointless with a fast refetch, or
- Causes a double status flicker with a slow refetch.

Also, there's no unmount guard around the post-`await` `setRules`/`addToast`.

### M6. Test coverage gap on the actual new feature

The detail-flyout tests (`__tests__/detector_detail_flyout.test.tsx:112`, `forecaster_detail_flyout.test.tsx:111`) only exercise edit handlers — they never pass `onStart`/`onStop`, so the start/stop branch selection, loading state, and error path are untested.

Fixtures use:

```ts
status: 'active'
```

which isn't in either action-status set, so `isAdResourceRunning` falls through to `enabled` and only the **Stop** branch could ever render.

No test covers:

- Mixed monitor + detector bulk-delete routing.
- Detector/forecaster create-disabled gating.

---

## 🔵 Low / cleanup

- **Dead abstraction:** `monitors_table_columns.tsx:79` — `isSelectableRuleDefinition` is now `(_item) => true`. Inline it or remove.
- **Dead value:** `index.tsx:227` — the reason string in `const [logsCreateDisabled] = useMemo(...)` (`'no_selection'`, etc.) is computed but never consumed.
- **Duplication:** `index.tsx:252-258` — `detectorCreateDisabled` and `forecasterCreateDisabled` are byte-identical despite the comment claiming they're "kept separate to diverge." Confirm AD/Forecasting have no minimum-version requirement (Logs requires `≥3.5.0`; these require none).
- **Unmounted `setState`:** In `handleSubmit` `finally` (`create_ad_rule_flyout.tsx:2236`) and in `runLifecycleAction` `finally` (both flyouts) — cosmetic React warnings.
- **Fire-and-forget lifecycle buttons:** `monitors_main_panel.tsx` Start/Stop use `void onStart…`, allowing a double-click to fire two mutation batches before selection clears.

---

## ✅ Verified correct (don't worry about these)

- The automated bot's **"duplicate `EuiFieldNumber`/`EuiFormRow` import"** and **"`onCreateMonitor` called without null-guard"** are both false positives — the duplicate import was already fixed in the CI-fix commit, and every `onCreateMonitor(...)` call site sits inside the `{onCreateMonitor && (…)}` guard at `monitors_main_panel.tsx:187`.
- `runWithConcurrencyLimit`, per-datasource error isolation, `STATUS_COLORS` exhaustiveness for the 20+ new statuses, and the AOS/AOSS serverless detection are all sound.
- Backend routing tests and `alert_utils` mapping tests are solid and meaningful.

public/components/alerting/mutations/monitor_mutations_client.ts
```diff
@@ -29,13 +29,25 @@ export interface AcknowledgeAlertResponse {
...
+}
+
 export class MonitorMutationsClient {
   private requireHttp() {
     const http = coreRefs.http;
     if (!http) throw new Error('HTTP client not available');
     return http;
   }
 
+  private withOptionalDatasource(basePath: string, dsId?: string): string {
```
lezzago: 🔴 **High — `local-cluster` sentinel not normalized here, so bulk AD/forecaster actions break on non-MDS clusters.**

This `withOptionalDatasource` only checks truthiness, so the synthetic `'local-cluster'` id (used on single-cluster / non-MDS deployments) is truthy and gets appended. `stopDetector` then builds `/api/anomaly_detectors/detectors/<id>/stop/false/local-cluster`.

The create flyout has its own `withOptionalDatasource` (`create_ad_rule_flyout.tsx:226-232`) that normalizes `''`/`local`/`local-cluster`/`local_cluster` to the base path via `normalizeDataSourceId`. So **create works but start/stop/delete of the same resource fail** — those `/api/anomaly_detectors/*` and `/api/forecasting/*` routes are owned by the external anomaly-detection-dashboards-plugin, which treats the trailing segment as a real MDS `dataSourceId` (no `local-cluster` awareness, unlike `getAlertingClient` in this repo's `server/routes/alerting/index.ts:220`).

Affects `deleteDetector`, `startDetector`, `stopDetector`, and all three forecaster methods. The existing test at `mutations/__tests__/monitor_mutations_client.test.ts:91` only exercises a non-sentinel id, so it doesn't catch this.

**Fix:** share the flyout's `normalizeDataSourceId` + `withOptionalDatasource` in one module both import, and add a test asserting the sentinel ids produce the base path with no trailing segment.

_Note: the exact 404 symptom is inferred — the consuming routes live in an external plugin — but the flyout deliberately stripping the sentinel before hitting the identical endpoints is conclusive that appending it is wrong._

public/components/alerting/create_ad_rule_flyout.tsx
```diff
@@ -0,0 +1,3892 @@
...
+  ruleType: CreateAdRuleType,
+  form: CreateAdRuleFormState,
+  options: {
+    customResultIndexRequired?: boolean;
+    existingResource?: ADDetector | ADForecaster;
+  } = {}
+) => {
+  const existingResource = asRecord(options.existingResource);
+  const commonPayload = {
+    ...stripReadOnlyResourceFields(options.existingResource),
```
lezzago: 🔴 **High — on edit, the payload carries both snake_case and camelCase config keys.**

`buildRulePayload` spreads `stripReadOnlyResourceFields(existingResource)` (`544-587`), which strips only lifecycle/identity/job fields, then layers camelCase config keys on top. `existingResource` is the raw snake_case `_source` (`editDetail.raw`, set in `server/services/alerting/alert_detail.ts` from the native AD API, never camel-mapped).

Result: the emitted JSON contains both `filter_query`+`filterQuery`, `category_field`+`categoryField`, `detection_interval`+`detectionInterval`, `time_field`+`timeField`, etc.

**Concrete data-loss case:** disabling the category field sets `categoryField: undefined` (dropped by `JSON.stringify`) while the stale `category_field: [...]` survives from the spread — so the disable is silently lost for any backend honoring snake_case.

The broader duplicate-key impact depends on the external plugin's key handling (not in this repo), but the category case doesn't. **Fix:** strip config fields too, or don't spread the raw resource — send a single canonical payload.

public/components/alerting/monitors_table/index.tsx
```diff
@@ -139,23 +163,99 @@ export const MonitorsTable: React.FC<MonitorsTableProps> = ({
...
+      if (isAnalyticEngineEnabled(d.id)) return true;
+      if (!d.version) return false;
+      const coerced = semver.coerce(d.version);
+      return coerced ? semver.gte(coerced, BASE_PPL_ALERTING_SUPPORTED_VERSION) : false;
+    });
+    if (!hasSupportedDs) return [true, 'version_unsupported'] as const;
+    return [false, ''] as const;
+  }, [selectedDatasources, isAnalyticEngineEnabled]);
+
+  const metricsCreateDisabled = useMemo(() => {
```
lezzago: 🟠 **Medium — Metrics-rule gating guard was dropped (impact is conditional).**

The prior logic disabled the Metrics option when every selected datasource was OpenSearch (`selected.every(d => d.type === 'opensearch')`). Now any non-empty selection returns `false`, so an OpenSearch-only selection enables “Metrics alert rule” → `onCreateMonitor('metrics')` → opens the Prometheus create flyout.

Impact is **conditional**: the flyout's datasource dropdown is built from the full datasource list (not the selection), so the genuinely-broken empty-flyout case only occurs when the cluster has **zero** Prometheus connections.

Also: the justifying comment here (Prometheus “may exist without a saved object”) is contradicted by `hooks/use_datasources.ts`, which discovers Prometheus via the `data-connection` saved-object type.

server/services/alerting/alert_utils.ts
```diff
@@ -827,6 +919,8 @@ export function adDetectorToUnifiedRuleSummary(
   const indices = detector.indices || [];
   const evalInterval = formatPeriod(detector.detection_interval) || 'unknown interval';
   const windowDelay = formatPeriod(detector.window_delay) || 'none';
+  const status = getRuntimeStatus(detector) || 'Running';
```
lezzago: 🟠 **Medium — detector defaults to `Running` when runtime state is absent.**

`getRuntimeStatus(detector) || 'Running'` — when the per-detector `_profile` enrichment fails (rejection swallowed per-resource in `alert_service.ts`, original detector returned with no `curState`) or when a brand-new never-run detector returns no `state`, status falls back to `Running`, `enabled` → true, and healthStatus → healthy. A stopped/never-started detector is then shown as Running with a Stop action offered.

Note the forecaster path (`~994-996`) is more defensive — it checks `enabled === false` → `'Inactive stopped'` first, only defaulting to `Running` when enabled is true/undefined. Consider the same guard here.

server/services/alerting/alert_service.ts
```diff
@@ -1194,10 +1203,53 @@ export class MultiBackendAlertService {
...
       id: hit._id,
     }));
+    return this.enrichADForecasterRuntimeStates(client, forecasters);
+  }
+
+  private async enrichADDetectorRuntimeStates(
+    client: AlertingOSClient,
+    detectors: ADDetector[]
+  ): Promise<ADDetector[]> {
+    const results = await runWithConcurrencyLimit(
```
lezzago, line 1217: 🟡 **Low — per-resource `_profile` fan-out on a hot listing path.**

`enrichADDetectorRuntimeStates` / `enrichADForecasterRuntimeStates` issue one `_profile` transport request per detector and per forecaster (N+M), on top of the two `_search` calls, on every unified-rules listing fetch, per OpenSearch datasource. Bounded by `runWithConcurrencyLimit` (default 5), but materially heavier than a single search. Worth considering if the profile enrichment can be batched or made lazy.

public/components/alerting/monitors_table/index.tsx
```diff
@@ -139,23 +163,99 @@ export const MonitorsTable: React.FC<MonitorsTableProps> = ({
...
+    if (selectedDatasources.length === 0) return true;
+    // Always allow Metrics creation when at least one datasource is selected —
+    // the Prometheus data-connection may exist on the cluster without a
+    // corresponding MDS saved object (discovered via SQL plugin API, not SO).
+    return false;
+  }, [selectedDatasources]);
+
+  const detectorCreateDisabled = useMemo(() => {
```
lezzago: 🟡 **Low (style) — `detectorCreateDisabled` and `forecasterCreateDisabled` are byte-identical `useMemo`s.**

Both compute `!selectedDatasources.some(d => d.type === 'opensearch')` with the same deps. The shared predicate is correct for both AD rule types (not a bug), but a future gating change would have to be made in both places or they'd silently diverge — collapse to one shared boolean.

public/components/alerting/create_ad_rule_flyout.tsx
```diff
@@ -0,0 +1,3892 @@
...
+
+  const renderStepContent = () => {
+    if (startAfterEditPrompt) return renderStartAfterEditPrompt();
+
+    if (isDetector) {
+      if (isDetectorModelEdit) return renderDetectorModelStep();
+      if (isDetectorSettingsEdit) return renderDefineStep();
+      if (currentStep === 0) return renderDefineStep();
+      if (currentStep === 1) return renderDetectorModelStep();
+      if (isEdit) return renderReviewStep();
```
lezzago: 🟡 **Low — unreachable dead code.**

`if (isEdit) return renderReviewStep();` can never execute for a detector in edit mode: `isDetectorModelEdit` and `isDetectorSettingsEdit` are exhaustive complements on `editTarget?.initialStep`, so line 3670 or 3671 always returns first. Reaching this line forces `isEdit === false`. Safe to remove.

lezzago (commented):
## Code review

Reviewed the AD/forecaster management changes. 7 findings below, anchored inline. Two high-severity items I'd suggest addressing before merge; the rest are lower severity (a couple are style/DRY).

| # | Sev | Finding | Location |
|---|-----|---------|----------|
| 1 | 🔴 High | `local-cluster` sentinel not normalized in bulk AD/forecaster actions | `monitor_mutations_client.ts:47` |
| 2 | 🔴 High | Edit payload carries both snake_case and camelCase config keys | `create_ad_rule_flyout.tsx:637` |
| 3 | 🟠 Med | Metrics-rule gating guard dropped (impact conditional) | `monitors_table/index.tsx:244` |
| 4 | 🟠 Med | Detector defaults to `Running` when profile enrichment fails | `alert_utils.ts:922` |
| 5 | 🟡 Low | Per-resource `_profile` fan-out on hot listing path | `alert_service.ts:1217` |
| 6 | 🟡 Low | Duplicated `useMemo` bodies (style) | `monitors_table/index.tsx:252` |
| 7 | 🟡 Low | Unreachable dead code in detector edit branch | `create_ad_rule_flyout.tsx:3674` |

Details inline.

public/components/alerting/mutations/monitor_mutations_client.ts
jackiehanyang: Alerts Manager uses local-cluster as its local sentinel. Added shared AD API path helpers used by create, update, start, stop, and delete; the sentinel is omitted while real MDS saved-object IDs remain unchanged. Added local and MDS route tests.

server/services/alerting/alert_service.ts
jackiehanyang: no bulk profile endpoint is available and the calls are concurrency-limited with per-resource fallback. we can track batching as future optimization.

public/components/alerting/create_ad_rule_flyout.tsx
jackiehanyang: Normalized the raw resource into one canonical update object before applying edits. Unmodeled settings, query DSL, metadata, and concurrency fields are preserved, while disabled category/result-index fields are removed

jackiehanyang:
> ## 🔴 Blocker (must fix before merge)
> ### 1. Build-breaking type errors: `Datasource.engineType` / `Datasource.version` don't exist
> **File:** `public/components/alerting/monitors_table/index.tsx:173, 234, 236, 237`
> 
> The new `AnalyticEngine`/version gating reads `d.engineType` and `d.version` on values typed `Datasource[]` (`selectedDatasources` is explicitly narrowed to `Datasource` at line 215). The `Datasource` interface (`common/types/alerting/unified_types.ts`) declares neither field. I confirmed with an isolated compile using the repo's own `tsc`:
> 
> ```
> error TS2339: Property 'engineType' does not exist on type 'Datasource'.
> error TS2339: Property 'version' does not exist on type 'Datasource'.
> ```
> 
> Two consequences:
> 
> * Clean CI typecheck will fail (the local pass was a stale-cache artifact).
> * Even if forced through, these fields are also never populated by the datasource service, so the `'OpenSearch Serverless'` and `semver.gte(version, 3.5.0)` branches are dead at runtime — every remote OS datasource (`mdsId` set) falls through to `if (!d.version) return false`, disabling Logs create unless `AnalyticEngine` happens to be on.
> 
> **Fix:** Add `engineType?: string` and `version?: string` to `Datasource` (and populate them where datasources are built), or drop the dead branches.
> 
> ## 🟠 High
> ### 2. A failed auto-start reports "Failed to create" and orphans the created detector
> **File:** `create_ad_rule_flyout.tsx:2174-2182` + catch at `2215`
> 
> When `startAfterCreate` is on, the follow-up `POST .../start` is awaited inside the same `try` as the create. If start fails (common while the cluster initializes), control jumps to the `catch` → `"Failed to create anomaly detection rule"` toast, and `onCreated()`/success toast never fire — but the detector already exists.
> 
> User retries → duplicate detector.
> 
> The start call should be in its own `try` with a distinct **"created but couldn't start"** message.
> 
> ### 3. Edit payloads send duplicate `snake_case` + `camelCase` keys
> **File:** `create_ad_rule_flyout.tsx:544-587, 635-654`
> 
> The detail API returns `snake_case` (`alert_detail.ts:74-84` spreads `anomaly_detector`). On edit, `buildRulePayload` spreads `stripReadOnlyResourceFields(existingResource)` (carrying `filter_query`, `feature_attributes`, `category_field`, `detection_interval`, `result_index`, …) and then overrides only the `camelCase` equivalents.
> 
> So the payload carries both old `category_field` and new `categoryField`, etc. Edits like **"disable categorical field"** or **"change the filter"** may silently not take effect depending on which key the API honors.
> 
> ### 4. Detail flyout body goes stale after start/stop
> **File:** `hooks/use_rule_detail.ts:27-48`, consumed by both flyouts
> 
> `useRuleDetail`'s effect deps are:
> 
> ```ts
> [service, dsId, ruleId, definitionType]
> ```
> 
> None change on a lifecycle action.
> 
> After Stop/Start, the header badge flips (optimistic prop update) but the body's **"Real-time job"**, **"Current state"**, and **"Enabled time"** rows keep pre-action values until the flyout is reopened. The header/body can visibly disagree (also flagged as the detail agent's #1).
> 
> **Fix:** Needs a refetch trigger tied to the mutation.
> 
> ## 🟡 Medium
> ### M1. `anomalyDirection` is collected but never sent
> `create_ad_rule_flyout.tsx:2718-2721` writes the `both`/`above`/`below` selection, but `buildFeatureAttributes` never reads it and the detector payload always sends:
> 
> ```ts
> rules: []
> ```
> 
> at `:675`.
> 
> The user's directional criteria is silently discarded.
> 
> ### M2. Missing numeric validation → silent coercion
> `horizon`/`history` have no `validateForm` checks. Forecaster inputs use `Number(e.target.value)` so clearing a field yields `0`, and `buildRulePayload` clamps to `Math.max(1, …)`.
> 
> For example:
> 
> * Clearing **Horizon** silently submits `horizon: 1`.
> * Typing `history: 5` (below the documented minimum `40`) passes and the API rejects it.
> * `errorsForStep` even lists `'horizon'` as expecting an error that's never produced.
> 
> ### M3. Dynamic `defaultMessage` under a shared i18n ID
> **File:** `create_ad_rule_flyout.tsx:2338, 2347`
> 
> Uses:
> 
> ```ts
> defaultMessage: isDetector ? '…detector' : '…forecaster'
> ```
> 
> `@osd/i18n` expects a static literal, and one ID can't hold two translations. Not a CI break here (this plugin isn't in the i18n scan scope I could find), but it's non-idiomatic and detector/forecaster copy can never diverge in translation.
> 
> **Fix:** Use two separate IDs.
> 
> ### M4. N+1 runtime-profile fetch on every rule list
> **File:** `alert_service.ts:1217, 1238`
> 
> `enrichAD*RuntimeStates` issues one `_profile` request per detector/forecaster. It's concurrency-limited with per-item fallback (good), but adds meaningful latency to the rules list as AD resource counts grow.
> 
> Consider whether a single bulk profile call exists.
> 
> ### M5. Optimistic update immediately clobbered by refetch
> **File:** `alarms_page.tsx:658-676`
> 
> `setRules(optimistic)` is followed unconditionally by `refetchRules()`.
> 
> The optimistic state is either:
> 
> * Pointless with a fast refetch, or
> * Causes a double status flicker with a slow refetch.
> 
> Also, there's no unmount guard around the post-`await` `setRules`/`addToast`.
> 
> ### M6. Test coverage gap on the actual new feature
> The detail-flyout tests (`__tests__/detector_detail_flyout.test.tsx:112`, `forecaster_detail_flyout.test.tsx:111`) only exercise edit handlers — they never pass `onStart`/`onStop`, so the start/stop branch selection, loading state, and error path are untested.
> 
> Fixtures use:
> 
> ```ts
> status: 'active'
> ```
> 
> which isn't in either action-status set, so `isAdResourceRunning` falls through to `enabled` and only the **Stop** branch could ever render.
> 
> No test covers:
> 
> * Mixed monitor + detector bulk-delete routing.
> * Detector/forecaster create-disabled gating.
> 
> ## 🔵 Low / cleanup
> * **Dead abstraction:** `monitors_table_columns.tsx:79` — `isSelectableRuleDefinition` is now `(_item) => true`. Inline it or remove.
> * **Dead value:** `index.tsx:227` — the reason string in `const [logsCreateDisabled] = useMemo(...)` (`'no_selection'`, etc.) is computed but never consumed.
> * **Duplication:** `index.tsx:252-258` — `detectorCreateDisabled` and `forecasterCreateDisabled` are byte-identical despite the comment claiming they're "kept separate to diverge." Confirm AD/Forecasting have no minimum-version requirement (Logs requires `≥3.5.0`; these require none).
> * **Unmounted `setState`:** In `handleSubmit` `finally` (`create_ad_rule_flyout.tsx:2236`) and in `runLifecycleAction` `finally` (both flyouts) — cosmetic React warnings.
> * **Fire-and-forget lifecycle buttons:** `monitors_main_panel.tsx` Start/Stop use `void onStart…`, allowing a double-click to fire two mutation batches before selection clears.
> 
> ## ✅ Verified correct (don't worry about these)
> * The automated bot's **"duplicate `EuiFieldNumber`/`EuiFormRow` import"** and **"`onCreateMonitor` called without null-guard"** are both false positives — the duplicate import was already fixed in the CI-fix commit, and every `onCreateMonitor(...)` call site sits inside the `{onCreateMonitor && (…)}` guard at `monitors_main_panel.tsx:187`.
> * `runWithConcurrencyLimit`, per-datasource error isolation, `STATUS_COLORS` exhaustiveness for the 20+ new statuses, and the AOS/AOSS serverless detection are all sound.
> * Backend routing tests and `alert_utils` mapping tests are solid and meaningful.

fixes done for each findings:
1. Added optional engine/version metadata to Datasource and populated it from saved objects in both client and server mappings. Logs, Metrics, AD, and Forecasting gating now use the correct datasource metadata.
2. Separated create and start outcomes. After creation succeeds, onCreated is invoked even if auto-start fails, and the user receives a “created but could not start” warning instead of a create failure.
3. Normalized the raw resource into one canonical update object before applying edits. Unmodeled settings, query DSL, metadata, and concurrency fields are preserved, while disabled category/result-index fields are removed
4. Exposed refetch from useRuleDetail and refresh detector/forecaster details after a successful lifecycle action.

M1. Implemented AD-compatible suppression rules for above and below, infer the direction during edit, preserve unrelated rules, and remove rules belonging to deleted features.
M2. Empty numeric fields remain invalid. Detector/forecaster history is validated as 40–10000, and forecaster horizon as 1–180.
M3. Replaced conditional messages with separate static detector and forecaster translation IDs.
M4 (deferred). No bulk profile endpoint is currently available. Requests remain concurrency-limited with per-resource failure isolation; batching or lazy loading can be tracked separately.
M5(not changed). This was retained so the UI updates immediately and then reconciles against backend state. It should not be described as fixed; we can track it if observable status flicker remains.
M6. Added coverage for start/stop branches, detail refetch, datasource gating, local/MDS route handling, payload normalization, rules, validation, and status fallback. A dedicated mixed monitor+detector deletion test was not added.

public/components/alerting/monitors_table/index.tsx
jackiehanyang: Restored Prometheus-only Metrics gating. Metrics creation is enabled only when a selected datasource is Prometheus; OpenSearch-only and empty selections remain disabled.

server/services/alerting/alert_utils.ts
jackiehanyang: Status now uses explicit runtime state or job-enabled information. If neither exists, the detector displays `Inactive not started` with `no_data`, never `Running`.

public/components/alerting/monitors_table/index.tsx
jackiehanyang: Consolidated the identical calculations into one shared adCreateDisabled boolean used by both options.

public/components/alerting/create_ad_rule_flyout.tsx
jackiehanyang: Removed the unreachable `isEdit` review step branch.

jackiehanyang:
unrelated test failure caused panels_test ci fail:
```
Testing a panel
      ✓ Opens visualization flyout from empty panel (4785ms)
      ✓ Redirects to correct page on breadcrumb click (1959ms)
      ✓ Duplicate the open panel (2966ms)
      ✓ Rename the open panel (5068ms)
      - Change date filter of the panel
      ✓ Add existing visualization #1 (3993ms)
      ✓ Add existing visualization #2 (4797ms)
      1) Add ppl filter to panel
      ✓ Drag and drop a visualization (6876ms)
      ✓ Resize a visualization (7003ms)
      ✓ Delete a visualization (6956ms)
      ✓ Duplicate a visualization (7570ms)
      ✓ Replace a visualization (8930ms)
      ✓ Add new visualization to panel (32381ms)
      ✓ Check visualization edit button (24717ms)


  27 passing (5m)
  1 pending
  1 failing

  1) Panels testing with Sample Data
       Testing a panel
         Add ppl filter to panel:

      Timed out retrying after 10000ms
      + expected - actual

      -'were Carrier = \'OpenSearch-Air\' | where Dest = \'Munich Airport\''
      +'where Carrier = \'OpenSearch-Air\' | where Dest = \'Munich Airport\''
      
      at Context.eval (webpack://observability-dashboards/./.cypress/integration/panels_test/panels.spec.ts:409:0)
```

public/components/alerting/monitors_table/index.tsx
```diff
@@ -139,23 +163,85 @@ export const MonitorsTable: React.FC<MonitorsTableProps> = ({
...
+  const logsCreateDisabled = useMemo(() => {
+    if (selectedDatasources.length === 0) return true;
+    const osSelected = selectedDatasources.filter((d) => d.type === 'opensearch');
+    if (osSelected.length === 0) return true;
+    // Check if any selected OS datasource supports PPL alerting
+    const hasSupportedDs = osSelected.some((d) => {
+      if (!d.mdsId) return true; // Local cluster — no version metadata available
+      if (d.engineType === 'OpenSearch Serverless') return true;
+      if (isAnalyticEngineEnabled(d.id)) return true;
+      if (!d.version) return false;
```
lezzago: **Logs-create gating hard-requires datasource `version` metadata**

```js
if (!d.version) return false;
```

For a remote (MDS) OpenSearch datasource, `hasSupportedDs` now returns `false` whenever the datasource saved object has no `dataSourceVersion` — so the **Logs alert rule** option is disabled even for a valid >= 3.5 cluster. This is a regression from the prior behavior where any OpenSearch datasource enabled Logs. It can happen when the SO was created before version tracking or the connection test hasn't run.

Note this only affects the **Logs alert rule** option — anomaly detector / forecaster creation is gated separately via `isStandardOpenSearchDatasource` (`shared_constants.ts`), which doesn't check `version`, so those remain enabled.

Could we confirm MDS OpenSearch saved objects reliably carry `dataSourceVersion`? If not, consider treating an unknown version as *enabled* (fail-open) rather than disabling the option.

TackAdam (approved):

lezzago (approved):

PR #2805, opened 2026-08-07
APM setup wizard for one-click onboarding

TackAdam:
## Summary

Adds a 4-page **Set up Application Monitoring** wizard to the APM Services page
(`app/observability-apm-services`). It replaces the "figure it out yourself"
onboarding gap: instead of manually creating datasets and wiring them together
in APM Settings, a user opens the wizard from the empty state and is walked
through detecting, reusing, or auto-creating each required object, ending with
the same persisted APM config the manual "APM Settings" flow produces.

The wizard covers the three things APM needs:
1. **Traces** dataset (+ correlated logs)
2. **Service map** dataset (v2)
3. **RED metrics** — an existing Prometheus (direct-query) data source

It detects what already exists, lets the user reuse it, auto-creates what's
missing (traces/services only), validates by fields, and gates the flow so you
can't finish with a broken configuration.

## Entry point

- Launched from the APM **empty-state CTA** ("Get started") on the Services page
  when no APM config exists.
- On finish, writes the linking `correlations` APM-config saved object via the
  existing `OSDSavedApmConfigClient` — identical persisted state to the manual
  "APM Settings" flow. No new saved-object type.

## The four pages

1. **Overview** — explains APM, shows the Application Telemetry Flow diagram
   (reused from APM Settings, with light/dark swap via the `theme:darkMode`
   uiSetting), lists what the wizard sets up, and an "Auto-create requirements"
   callout clarifying the naming conventions are only needed for auto-create.
2. **Traces** — data source picker + reuse/auto-create/refresh for a traces
   dataset; surfaces correlated logs (read-only) once a valid dataset is chosen.
3. **Services** — data source picker + reuse/auto-create/refresh for the v2
   service map dataset.
4. **RED metrics** — checks existing direct-query Prometheus data sources for the
   required span-derived metrics and lets the user pick a matching one. Never
   creates a data source.

---

### Reuse strategy: port, don't import
The trace/log auto-create logic already existed in the `explore` plugin. Rather
than take a cross-plugin bundle dependency, we **ported and generalized** it into
plugin-local utils (`setup_wizard/utils/apm_auto_detect.ts`,
`apm_auto_create.ts`), extended to also cover the v2 service map. Files note the
`explore` origin in a header comment.

### Validation is field-based, not name-based
Index naming conventions (`otel-v1-apm-span*`, `otel-v2-apm-service-map*`) are
**only** used for auto-create (we only create conventionally-named datasets).
Reuse/validity is gated purely on **required fields**, via a shared
`fieldNamesSatisfy(fieldNames, required)` helper that matches an exact field name
OR a nested child (`sourceNode.keyAttributes.name` satisfies `sourceNode`), and
does not treat prefix collisions as matches (`spanIdentifier` ≠ `spanId`).

- **Traces** required fields: `traceId`, `spanId`, `serviceName`.
- **Service map** required fields: `sourceNode`, `targetNode`, `sourceOperation`,
  `targetOperation`, `nodeConnectionHash`, `timestamp`.
- **RED metrics** required (span-derived): `request`, `fault`,
  `latency_seconds_bucket` — checked by exact membership against the Prometheus
  metadata endpoint (substring matches like `app_request_total` don't count).

Rationale: a DataView can be tagged/created before its index has the fields
(e.g. via the API). Trusting `signalType` or a title alone let broken datasets
pass and advance the wizard — field validation is the reliable gate.

### Auto-create targets conventions; reuse is permissive
- **Auto-create** only ever creates conventionally-named datasets
  (`otel-v1-apm-span*` traces with `signalType: traces` + correlated
  `logs-otel-v1*`; `otel-v2-apm-service-map*` as a plain index-pattern, time
  field `timestamp`, no signalType/schemaMappings — matching the real v2 shape).
- **Reuse** does not require the convention. Traces reuse lists
  `signalType: traces` datasets; **service-map reuse lists ALL datasets on the
  source** (any type) because service maps have no reliable type discriminator
  and a user may build one from a logs/traces dataset. Validity is decided by
  fields in both cases.

### Multi-data-source support
- Detection probes **every** data source (capped at 50, in concurrent chunks),
  returning a result per source whether or not anything was found — so every
  source is a selectable picker option.
- The synthetic "Local Cluster" entry is only shown when there are **no**
  registered data sources (otherwise a `local_cluster` data source already
  covers it — avoids the confusing duplicate).
- Each page picks its own data source independently; a source that lacks the
  page's signal is shown disabled with a "— no APM index" label + tooltip.
- **Services defaults to the Traces source** (common case: both on one cluster)
  until the user explicitly overrides the Services picker. Consequence: the final
  config can legitimately mix sources (traces from A, service map from B).

### Correlated logs
Auto-create of traces also creates the `logs-otel-v1*` dataset + `trace-to-logs`
correlation **when logs are detected** (mirrors explore's behavior). The Traces
page surfaces a read-only "Correlated Logs" accordion (like APM Settings) once a
valid trace dataset is selected, with a link out to manage correlations.

### Prometheus / RED metrics: detect-and-reuse only
The wizard never creates a Prometheus data source (that needs URL + credentials).
It lists direct-query datasources that expose all required RED metrics and lets
the user pick one. Note: the metadata endpoint keys on the **connection id**
(`directQueryName`), while the APM config stores the **saved-object id** — the
step handles both.

### Wizard navigation gating
- Step nav (1–2–3–4) is clickable **backward** to any visited step and
  **forward only to already-validated steps**; unreachable steps are disabled
  (no dead clicks). Next/Finish are gated on each step being complete
  (`exists`/`created`) with its id captured.
- Fixed content min-height + single scroll container to avoid modal resize
  flicker and double scrollbars.

---

## Per-step flow / status states

Each data step resolves to one status, scoped to the selected data source:

- **`exists`** — a valid dataset is already present → reuse dropdown (valid
  first), compact "ready to use" confirmation, Next enabled.
- **`created`** — auto-created this session → same as exists, held via a ref
  until the datasets list refresh catches up (prevents a race that clobbered the
  fresh selection).
- **`invalid`** — user selected a dataset missing required fields → warning
  callout + **"Refresh fields"** (re-pulls the DataView's field list in case the
  index gained them after creation) + Next gated. If a valid dataset exists,
  nudges the user to pick it; if not and raw data is detected, offers auto-create.
- **`missing`** — no dataset yet: if raw data detected → "Auto-create" button;
  else "No data found" + Refresh (re-probe).
- **`error`** — create failure surfaced in a danger callout.

### Situation matrix (how the wizard behaves)

| Situation | Behavior |
|---|---|
| Trace dataset exists (valid), service map does not | Traces step = reuse; Services step defaults to same source, shows auto-create if raw v2 data detected, else "no data / pick another source". |
| Trace dataset exists, but on a different data source than selected | Existence is scoped per source, so it only counts on its own source; other sources evaluate independently. |
| Trace-tagged dataset exists but missing required fields | Listed as "— missing required fields", selectable → `invalid` → Refresh fields; Next gated until valid. |
| Only invalid trace datasets, raw span data present | `invalid` branch also offers **Auto-create** (fresh valid dataset). |
| Multiple valid trace datasets | Dropdown lists all (valid first); user picks; "Found N — choose which one" helptext. |
| Service map on a logs/traces-tagged dataset (manual build) | Services lists all datasets on the source; validity by fields; user can select + Refresh a logs/traces dataset into a valid service map. |
| No Prometheus data source has the RED metrics | RED metrics step shows "no matching data source" + which metrics each is missing; Next gated (no auto-create). |
| One Prometheus source matches | Auto-selected; user can change via dropdown. |
| Data source lacks a page's signal | Shown disabled in the picker with "— no APM index" + tooltip. |
| Registered data sources exist | Synthetic "Local Cluster" hidden to avoid duplicate with `local_cluster`. |

---

## Files

**New — `public/components/apm/setup_wizard/`**
- `apm_setup_wizard_modal.tsx` — 4-step modal shell, step gating, Finish → config write.
- `steps/overview_step.tsx`, `traces_step.tsx`, `services_step.tsx`, `metrics_step.tsx`
- `components/data_source_picker.tsx`, `requirement_callout.tsx`
- `hooks/use_apm_detection.ts`, `hooks/use_prometheus_metric_check.ts`
- `utils/apm_auto_detect.ts` (+ `fieldNamesSatisfy`), `utils/apm_auto_create.ts`
  (+ exported `refreshAndPersistFields`)
- `constants.ts`, `types.ts`, `setup_wizard.scss`

**Changed**
- `components/apm/services.tsx` — mounts the wizard, opens from empty-state CTA.
- `components/apm/common/apm_empty_state.tsx` — CTA opens the wizard; spacing.
- `shared/hooks/use_apm_config.ts` — `useDatasets` options now expose
  `datasourceId` and `fieldNames` (used for per-source scoping and field-based
  validity). No behavior change for existing consumers.

---

## Testing

- **28 unit tests** (Jest) pass across 3 suites:
  - `apm_auto_detect.test.ts` — detection (traces/logs/v2 service-map incl.
    nested fields), across-data-sources behavior, and `fieldNamesSatisfy` incl.
    nested match + prefix-collision guard.
  - `apm_auto_create.test.ts` — create/reuse of trace, log, correlation, and
    service-map datasets; `refreshAndPersistFields` (happy path, best-effort
    swallow, rethrow-on-error).
  - `use_prometheus_metric_check.test.ts` — RED-metric checking incl. exact
    membership (substring collision rejected), partial/missing, and error paths.
- Lint, stylelint, and typecheck clean.

### Check List


public/components/apm/setup_wizard/steps/traces_step.tsx
```diff
@@ -0,0 +1,713 @@
...
+import { navigateToDatasetCorrelations } from '../../shared/utils/navigation_utils';
+import { RequirementCallout } from '../components/requirement_callout';
+import { DataSourcePicker } from '../components/data_source_picker';
+import { createApmTraceDatasets, refreshAndPersistFields } from '../utils/apm_auto_create';
+import { fieldNamesSatisfy } from '../utils/apm_auto_detect';
+import { APM_TRACES_INDEX_PATTERN, APM_TRACES_REQUIRED_FIELDS } from '../constants';
+import { APM_TRACES_DOCS_URL, APM_CORRELATIONS_DOCS_URL } from '../../common/constants';
+import { ApmDetectionResult, StepState } from '../types';
+
+interface DetectionHook {
```
ps48: ### p1: extract the shared dataset-step logic

`traces_step` and `services_step` are about 200 lines of near-identical logic (the reconcile effect, `handleSelectExisting`, `handleRefreshFields`, `datasetOptions`, the `justCreatedIdRef` race, and the exists/invalid/no-data render branches), differing only by required fields and the traces-only correlated-logs accordion. The `DetectionHook` interface is also declared verbatim in both. Should we extract a `useDatasetStep({ requiredFields, scopeFilter, create })` hook, and export `ReturnType<typeof useApmDetection>` instead of redeclaring `DetectionHook`? Today the two reconcile effects have to be kept in lockstep by hand.

public/components/apm/setup_wizard/hooks/use_apm_detection.ts
```diff
@@ -0,0 +1,62 @@
...
+    })();
+
+    return () => {
+      cancelled = true;
+    };
+  }, [refreshToken]);
+
+  const refresh = useCallback(() => setRefreshToken({}), []);
+
+  return { detections, loading, error, refresh };
```
ps48, line 61: ### p1: detection error is never surfaced

`error` is returned here but no consumer reads `detection.error`, so a detection failure renders as "no trace/service map data found", which is indistinguishable from an empty cluster. Can we surface it via a danger `EuiCallOut`, or drop the return value if that is intended? The same pattern exists in `useDatasets` and `usePrometheusDataSources` in `shared/hooks/use_apm_config.ts`, whose `error` the step consumers also drop.

public/components/apm/setup_wizard/hooks/use_prometheus_metric_check.ts
```diff
@@ -0,0 +1,116 @@
...
+      setResults([]);
+      setLoading(false);
+      return;
+    }
+
+    let cancelled = false;
+    setLoading(true);
+
+    (async () => {
+      const checked = await Promise.all(
```
ps48: ### p1: bound the probe fan-out, and the `error` field is dead

This fires `candidates x 3` requests with no concurrency cap, unlike the deliberate `PROBE_CONCURRENCY` chunking in `apm_auto_detect.ts`. Should we reuse a shared `mapInChunks` helper here? Also the `error` captured on the result (line 97) is never rendered (`metrics_step` shows only `missing`), so a failed probe looks identical to metrics being genuinely absent.

public/components/apm/setup_wizard/utils/apm_auto_create.ts
```diff
@@ -0,0 +1,270 @@
...
+          timeFieldName: detection.logTimeField,
+          signalType: 'logs',
+          schemaMappings: { ...APM_LOGS_SCHEMA_MAPPINGS },
+          // @ts-expect-error TS2322 see note above on trace dataset.
+          dataSourceRef,
+        },
+        effectiveDataSourceId
+      );
+    } catch (createError) {
+      console.warn('Failed to create log dataset:', createError);
```
ps48, line 221: ### p1: correlated-logs and correlation failures are swallowed

The log-dataset create here and the correlation create (line 225) only `console.warn` on failure, and trace creation still reports success. The user is told traces were created and never learns that correlated logs did not wire up. Can we return a partial-failure flag and surface a non-blocking `addWarning` toast?

public/components/apm/setup_wizard/utils/apm_auto_create.ts
```diff
@@ -0,0 +1,270 @@
...
+    try {
+      result.traceDatasetId = await createOrReuseDataView(
+        savedObjectsClient,
+        dataViews,
+        {
+          title: detection.tracePattern,
+          displayName: `Trace Dataset${dataSourceSuffix}`,
+          timeFieldName: detection.traceTimeField,
+          signalType: 'traces',
+          // @ts-expect-error TS2322 IndexPatternSpec types dataSourceRef as SavedObjectReference
```
ps48: ### p1: type `dataSourceRef` instead of suppressing

The `@ts-expect-error TS2322` on `dataSourceRef` appears three times (lines 171, 193, 261), alongside duplicated `effectiveDataSourceId` / `dataSourceRef` boilerplate in both create functions. Should we extract a typed `buildDataSourceContext(detection, dataSourceId)` helper, so a future upstream type change surfaces as a real error rather than being suppressed in three places?

public/components/apm/setup_wizard/utils/apm_auto_detect.ts
```diff
@@ -0,0 +1,205 @@
...
+    required
+  );
+
+/**
+ * Detect APM data for a single data source following OpenTelemetry conventions.
+ * Probes traces (otel-v1-apm-span*), correlated logs (logs-otel-v1*) and the v2
+ * service map (otel-v2-apm-service-map*), validating required fields for each.
+ */
+export async function detectApmData(
+  savedObjectsClient: SavedObjectsClientContract,
```
ps48: ### p1: unused `savedObjectsClient` parameter

`detectApmData` takes `savedObjectsClient` (threaded from `detectApmDataAcrossDataSources`) but never uses it; only `indexPatternsService` is referenced. Can we drop it, or prefix it `_` if it is kept for signature symmetry?

public/components/apm/setup_wizard/steps/metrics_step.tsx
```diff
@@ -0,0 +1,199 @@
...
+      return;
+    }
+
+    if (matching.length === 0) {
+      onStateChange({ status: 'missing' });
+      onPrometheusDataSourceIdChange('');
+      return;
+    }
+
+    if (matching.length === 1 && state.existingId !== matching[0].id) {
```
ps48, line 76: ### p1: multi-match leaves the step in `checking`

When more than one Prometheus source matches, neither effect branch fires, so `metricsState` stays `checking` while the picker is shown. Finish is still correctly gated, so this is cosmetic, but should we set an explicit "select one" or `missing` status? The effect also omits `state` from its deps and relies on the eslint-disable.

public/components/apm/setup_wizard/utils/apm_auto_create.ts
```diff
@@ -0,0 +1,270 @@
...
+  if (result.traceDatasetId && result.logDatasetId) {
+    try {
+      const correlationResponse = await savedObjectsClient.create(
+        'correlations',
+        {
+          title: `trace-to-logs_${detection.tracePattern}`,
+          correlationType: `${CORRELATION_TYPE_PREFIXES.TRACE_TO_LOGS}${detection.tracePattern}`,
+          version: '1.0.0',
+          entities: [
+            { tracesDataset: { id: 'references[0].id' } },
```
ps48, line 240: ### p1: confirm the literal `references[N].id` placeholders

`entities` stores the literal strings `'references[0].id'` and `'references[1].id'` rather than the ids (the real ids live in `references`). This matches the ported explore convention and reads correctly, but a drifted copy here would fail silently. Worth a one-line comment noting the resolve-against-references contract.

public/components/apm/setup_wizard/utils/apm_auto_create.ts
```diff
@@ -0,0 +1,270 @@
...
+  DuplicateDataViewError,
+  IndexPatternSpec,
+} from '../../../../../../../src/plugins/data/public';
+import { APM_LOGS_SCHEMA_MAPPINGS, APM_SERVICE_MAP_TIME_FIELD } from '../constants';
+import { ApmDetectionResult } from '../types';
+
+export interface CreateApmDatasetsResult {
+  traceDatasetId: string | null;
+  logDatasetId: string | null;
+  serviceMapDatasetId: string | null;
```
ps48: ### p1: dead field on the trace-create result

`CreateApmDatasetsResult.serviceMapDatasetId` is set to `null` here and never assigned (the service map is created by a separate function), and callers read only `traceDatasetId` and `correlationId`. Should we drop it from this result type so it does not imply it is populated here?

public/components/apm/setup_wizard/constants.ts
```diff
@@ -0,0 +1,66 @@
...
+ *
+ * The wizard requires BOTH the naming convention AND validation that the
+ * required fields exist before it will claim a match or offer to auto-create a
+ * dataset. Values here were confirmed against a live OpenTelemetry demo cluster
+ * (indices otel-v1-apm-span-*, logs-otel-v1-*, otel-v2-apm-service-map-*) and
+ * the working DataViews on the datasets page.
+ */
+
+/** Traces: OpenTelemetry span index pattern (Data Prepper otel-v1). */
+export const APM_TRACES_INDEX_PATTERN = 'otel-v1-apm-span*';
```
ps48: ### p1: consider `apm/common` for the domain constants

The index patterns and required-field lists are APM domain knowledge (the manual APM Settings flow validates the same datasets), and `public/components/apm/common/constants.ts` already exists. Worth moving the shareable constants there? `types.ts` is genuinely wizard-local, so that one is fine where it is.

ps48 (commented):
### apm setup wizard review

Nice onboarding addition. The wizard is well structured and the core logic holds up: field validation is dot-anchored so `spanIdentifier` correctly does not satisfy `spanId`, RED-metric checks use exact membership (no substring false positives), step gating and the created-vs-exists race handling are sound, and the datasource / saved-object surface is clean (no new saved-object type, everything goes through DataViews and saved objects, no new backend routes, no new dependencies). State management is coherent (hooks plus one context, no redux or rxjs mixing), and cypress uses real backend data.

No `p0` blockers. Themes across the `p1` inline comments:

- error surfacing: several caught errors are captured into state or only logged, and never rendered. Detection failure, dataset-fetch failure, Prometheus-list failure, and a swallowed correlated-logs / correlation create all degrade into a misleading "no data found" or a silent success. These should surface via a callout or toast.
- duplication: `traces_step` and `services_step` are about 200 lines of near-identical logic, and `apm_auto_create` repeats the `dataSourceRef` boilerplate plus three `@ts-expect-error` suppressions.
- testing: the 28 unit tests are good, but the new 4-page wizard has no cypress / e2e coverage. The existing spec still drives the old manual "APM Settings" modal. Since this is marked WIP, flagging so it is not lost.

Two pre-existing hooks (`useDatasets`, `usePrometheusDataSources` in `shared/hooks/use_apm_config.ts`) capture an `error` that the new wizard consumers drop, the same surfacing gap, worth addressing while this feature starts depending on them.

Findings are inline on the code.

TackAdam:
Thanks for the thorough pass, @ps48. All 11 inline `p1` comments are addressed — summary of what changed for each, in the order you left them:

1. **Extract the shared dataset-step logic** — Pulled all of the shared logic into a new `hooks/use_dataset_step.ts` (`useDatasetStep`): the single reconcile effect, the just-created selection race, `datasetOptions`/`selectedOptions`, and the select / refresh-fields / auto-create handlers now live in one place, parameterized by `{ requiredFields, scope, createDataset, … }`. Both `traces_step` and `services_step` are reduced to their own JSX (traces keeps the correlated-logs accordion). Also exported `ApmDetectionHook = ReturnType<typeof useApmDetection>` and removed both duplicated `DetectionHook` interfaces. No more keeping two reconcile effects in lockstep by hand.

2. **Detection error never surfaced** — Added a danger `EuiCallOut` (with a Retry button) in the traces and services steps gated on `detection.error`, so a detection failure no longer degrades into a misleading "no data found". Also surfaced the previously-dropped `usePrometheusDataSources` error in the metrics step.

3. **Bound the probe fan-out + dead error field** — Exported `mapInChunks`/`PROBE_CONCURRENCY` from `apm_auto_detect.ts` and used them to bound the RED-metric probe the same way the data-source probe is bounded. The metrics step now renders `probe failed: …` when a per-source `error` is set, so a failed probe is distinct from metrics being genuinely absent.

4. **Swallowed correlated-logs/correlation failures** — `createApmTraceDatasets` now returns a `correlatedLogsFailed` flag; when correlated logs were detected but the log dataset or the correlation fails, the traces step fires a non-blocking `addWarning` toast instead of reporting a clean success. Trace creation stays non-fatal as before. Added two unit tests.

5. **Type `dataSourceRef` instead of suppressing** — Extracted `buildDataSourceContext(detection, dataSourceId)` used by both create functions. The three `@ts-expect-error` sites are gone, replaced by a single localized cast in the helper; the `effectiveDataSourceId`/suffix/ref boilerplate is no longer duplicated.

6. **Unused `savedObjectsClient` parameter** — Dropped it from `detectApmData`; updated the caller and the tests.

7. **Multi-match leaves the step in `checking`** — Added an explicit multi-match branch in the metrics step so it no longer sits in `checking` while the picker is shown (Finish stays gated until a source is selected).

8. **`references[N].id` placeholders** — Confirmed intentional; added a comment noting they're placeholders resolved against the `references` array at read time, so they aren't "fixed" into literal ids later.

9. **Dead field on the trace-create result** — Removed `serviceMapDatasetId` from `CreateApmDatasetsResult` (it was never assigned; the service map is created by `createApmServiceMapDataset`).

10. **Domain constants location** — Moved the index patterns, required-field lists, time fields, schema mappings, and RED metrics into `apm/common/constants.ts`, deleted the wizard-local `constants.ts`, and repointed all importers — so the wizard and the manual APM Settings flow share one source of truth. `types.ts` stays wizard-local.

11. **Summary themes** — Error surfacing (1 detection + dataset/Prometheus list + swallowed correlated logs), duplication, and the `@ts-expect-error` collapse are all covered above. **Testing:** unit tests are green (30 passing, +2 for the partial-failure flag); cypress/e2e coverage for the full 4-page flow is tracked as a follow-up before this comes out of WIP.

Verified: **30/30 unit tests pass, ESLint clean on all touched files.**

---

## Additional improvement — slow wizard load (N+1 saved-object lookups)

Separate from the review comments, addressed the wizard's slow initial load, which shared the root cause described in [OpenSearch-Dashboards#12630](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/12630) (N+1 saved-object lookups on Explore page load). That PR's files live in `src/plugins/explore` (core) and don't apply here, and our `apm_auto_detect.ts` already avoids the loop (it probes via `getFieldsForWildcard` per data source, not a per-pattern `get`). But the same anti-pattern lived in the wizard's own `useDatasets` (`public/components/apm/shared/hooks/use_apm_config.ts`).

- **Batched the dataset lookups.** Replaced the serial `for … await dataViews.get(id)` loop over every index pattern with a single `Promise.all`. `dataViews.get` routes through the core saved-objects client, which batches concurrent gets into one `_bulk_get` (and resolves each data source once) — so ~60 serial round-trips collapse into one batch. Output (options, order, AnalyticEngine filtering, error/abort handling) is unchanged. Added a unit test asserting the lookups now fire concurrently (it fails on the old serial loop).

- **Hoisted `useDatasets` to the wizard modal.** The traces and services steps each mounted their own `useDatasets`, re-fetching the whole dataset list on every step visit. It's now loaded once at the modal level (like detection) and passed to both steps as a `datasets` prop, so the list resolves once per wizard session instead of once per stage.

Impact scales with (index-pattern count × latency): negligible on a small/local workspace, large on a many-pattern remote workspace (the upstream PR measured ~14.8s → ~1.9s page load in a comparable ~60-pattern scenario).

Verified: **308 APM tests pass, ESLint clean.**

public/components/apm/setup_wizard/hooks/use_dataset_step.ts
```diff
@@ -0,0 +1,338 @@
...
+        return; // still waiting for the list refresh — keep the created selection
+      }
+    }
+
+    // If the user has selected an invalid dataset on this source, keep it
+    // selected (so they can refresh its fields) unless it has since become valid.
+    if (state.status === 'invalid') {
+      const stillPresent = allScopedDatasets.find((d) => d.value?.id === state.existingId);
+      const nowValid = scopedDatasets.find((d) => d.value?.id === state.existingId);
+      if (nowValid) {
```
ps48, line 176: ### p1: document the intentional fall-through

This `invalid -> nowValid` branch deliberately does not `return`; it falls through to the reuse branch below (`onDatasetIdChange`) to capture the id, which is what keeps `canFinish` correct after a "Refresh fields" promotion. `state` is also intentionally left out of the effect deps. Could we add a one-line comment noting the intentional fall-through, so a future edit does not add a `return` here and silently break finish-gating?

ps48 (commented):
### re-review: "address comments, optimize" (d2f982d9)

Thanks for the thorough pass. All ten inline comments from the previous review are addressed, and I verified the refactor introduced no regressions.

**Resolved and verified:**
- shared step logic extracted into `useDatasetStep`; both steps consume it, and the traces-vs-services differences (scope `traces` vs `all`, the correlated-logs UI, the created-vs-exists race, step gating) are preserved.
- `DetectionHook` is now `ReturnType<typeof useApmDetection>` (`ApmDetectionHook`).
- detection, prometheus, and per-probe errors now surface via callouts.
- probe fan-out is bounded via `mapInChunks(..., PROBE_CONCURRENCY)`.
- correlated-logs / correlation failures set `correlatedLogsFailed` and surface a non-blocking warning toast.
- the three `@ts-expect-error` are replaced by a single typed `buildDataSourceContext` cast.
- unused `savedObjectsClient` param, dead `serviceMapDatasetId` field, and the multi-match `checking` state are cleaned up.
- domain constants moved to `apm/common/constants.ts`; the `references[N].id` placeholder now carries the explaining comment.

**Verified end to end locally** against a real OpenTelemetry stack (OpenSearch + Data Prepper + Cortex) with security enabled, on a from-source OSD build:
- reuse path (existing datasets): Finish persists the APM-config and the Services page renders the live service catalog.
- traces auto-create: creates the trace dataset, the correlated log dataset, and the `trace-to-logs` correlation.
- service-map auto-create: creates the v2 `otel-v2-apm-service-map*` dataset (no signalType, `timestamp` time field).
- the invalid / missing branches, forward step-gating, and RED-metrics exact-membership matching all behave as intended.

No `p0` blockers. The remaining item is automated coverage: the new wizard still has no cypress / e2e test (the spec drives the old manual "APM Settings" modal). Since this is WIP, flagging so it is not lost. One minor inline note.

Nice work.

ps48 (approved):

PR #2810, opened 2026-08-11
[SLO] Update

TackAdam:
## Description

This PR fixes several correctness and UX issues in the SLO/SLI experience (Suggest SLOs page + SLO listing), and extracts a shared UI component. The changes fall into five areas, plus a small refactor for reuse.

Issues:
#2813 #2815

### 1. Suggest SLOs: coverage is workspace-scoped (saved objects), not ruler rules

**Problem.** The Suggest page marked a draft "already covered" whenever a matching Prometheus recording rule existed on the ruler (`existingRuleMatch`). Because a single AMP datasource can be shared across workspaces, and recording rules are written to a per-workspace namespace (`slo-generated-<workspaceId>`) but the ruler exposes all namespaces, a workspace that had no SLO of its own could still see another workspace's rules — and was wrongly blocked from creating the SLO. It also blocked creating a second SLO with a different target.

**Fix.** `isDraftCovered` now depends only on the workspace-scoped SLO health rollup (saved objects). `existingRuleMatch` is demoted to an informational "rules exist — will reuse" badge. Creation is workspace-scoped and recording-rule writes are idempotent + refcounted by fingerprint, so reusing/duplicating rules is safe. The SLI fingerprint excludes the objective target (`slo_sli_fingerprint.ts`), so a second SLO with a different target reuses the same recording group and only adds its own alert group — now unblocked as well.

### 2. SLO listing: state and canonicalKind filters returned 0 results despite matches

**Problem.**

- `state` was pushed to the saved-object filter clause via the `cachedState` projection, but `cachedState` is only written by a best-effort writeback after an SLO is surfaced on a page — so fresh/never-viewed SLOs were wrongly excluded and stale ones wrongly kept.
- `canonicalKind` isn't projected on the saved object at all, so it was filtered post-fetch on the current page slice, corrupting totals and producing empty pages when matches lived on later pages.

**Fix.** `SloQueryService.paginate` routes any state- or `canonicalKind`-filtered request through the materialize-and-slice `list()` path, which applies both filters over the full result set after live-status fold-in. Removed the dead `cachedState` pushdown clause and the post-fetch `canonicalKind` block. The fast SO-pushdown path is retained for all other facets.

### 3. SLO listing: datasource filter returned "No SLOs match"

**Problem.** The datasource facet emitted the data-connection saved-object UUID as the filter value, but SLOs persist and are resolved by the connection name (`spec.datasourceId`). The server couldn't resolve the UUID → empty result.

**Fix.** The facet now selects by connection name (what the server and `_aggregate` resolve against). The active-filter chip falls back to an id→name map for legacy URLs.

### 4. SLO listing: landing datasource scope + per-tab memory

**New behavior.** On landing, the listing resolves the datasource scope once:

- honor an explicit `?datasourceId=` from the URL (shared/bookmarked links), else
- restore this tab's remembered scope (`sessionStorage`, per workspace), else
- auto-select the up-to-5 datasources with the most SLOs (ranked via a lightweight per-datasource count probe).

The remembered scope is a tri-state (`all` / `scope[ids]` / `absent`) so a return visit restores exactly what the user had — including an explicit "show all" or "show nothing" — instead of re-running the auto-select. The first fetch is gated until the scope resolves, so there's no flash of an unscoped list.

### 5. SLO listing: empty datasource scope shows nothing

**New behavior.** Unchecking every datasource is an explicit "no datasource in scope" (`[]`, distinct from "no filter" = `undefined`). The listing short-circuits client-side (no server call) and renders a tailored "No datasources selected" empty state with a "Show all datasources" recovery button (which re-checks the top datasources by SLO count; label reflects the 5-cap when more exist).

### 6. Refactor / UX polish

- Extracted Alert Manager's `TruncatedLabel` into a shared component (`public/components/common/truncated_label/`) and reused it in both the AM facet panel and the SLO datasource facet — single implementation, consistent behavior.
- The tooltip now renders through a portal to `document.body` so it's instant on both pages (a `position: fixed` tooltip was mispositioned/clipped inside the SLO listing's `EuiResizableContainer` transformed ancestor).
- `TruncatedLabel` inherits font metrics by default so the datasource label matches sibling facets (AM keeps its explicit `12/18`).
- Removed the stray per-row divider and horizontal scrollbar in the datasource facet; names truncate with an ellipsis + full-name-on-hover.

---

### Suggest page — is a draft selectable?

```text
draft
├─ SLO saved object exists for (service, side) in THIS workspace?  ── yes ─▶ COVERED (disabled, "SLO exists")
└─ no
   ├─ matching recording rule on the ruler (existingRuleMatch)? ── yes ─▶ SELECTABLE + info badge "rules exist — will reuse"
   └─ no                                                        ──────▶ SELECTABLE (normal)
```

### Listing landing — datasource scope resolution (runs once)

```text
wait for datasource catalog to load
├─ URL has ?datasourceId= ?           ─▶ use URL scope
├─ sessionStorage scope for workspace?
│  ├─ {kind:'all'}                    ─▶ no datasource constraint (show all)
│  └─ {kind:'scope', ids}             ─▶ restore ids (may be [] = show nothing)
└─ nothing remembered
   ├─ 0 datasources                   ─▶ no scope
   └─ ≥1 datasources                  ─▶ auto-select top-5 by SLO count (count>0)
→ then: mark scopeReady, run first fetch
```

### Listing — datasource filter value → results

```text
filters.datasourceId
├─ undefined  ─▶ no datasource constraint → show all SLOs
├─ []         ─▶ explicit empty scope → short-circuit, show nothing ("No datasources selected")
└─ [names]    ─▶ server resolves names → scoped results
```

### Listing — paginate routing

```text
request
├─ filters on state OR canonicalKind ?  ─▶ materialize-and-slice via list() (correct totals/pages, live status)
└─ otherwise                            ─▶ SO filter pushdown + page-slice status fold-in (fast path)
```

---

## Files changed

- **Suggest page (Issue 1):** `slo_suggest_page.tsx`, `suggest_inline_row.tsx`, `suggest_service_tree_table.tsx`
- **Listing filters (Issue 2):** `common/slo/slo_query_service.ts`, `server/services/slo/slo_saved_object_store.ts`
- **Listing datasource scope (Issues 3–5):** `slo_listing_page.tsx`, `slo_list_filter_panel.tsx`, `slo_list_filter_panel.scss`, `slo_datasource_scope_cache.ts` (new)
- **Shared component (refactor):** `common/truncated_label/{truncated_label.tsx, truncated_label.scss, index.ts}` (new), `alerting/facet_filter_panel.tsx`, `alerting/alerting.scss`
- **Tests:** `slo_query_service_paginate.test.ts`, `slo_suggest_page.test.tsx`, `suggest_inline_row.test.tsx`, `slo_list_filter_panel.test.tsx`, `slo_listing_page_datasource_scope.test.tsx` (new), `slo_datasource_scope_cache.test.ts` (new), `truncated_label.test.tsx` (new)

---

## Testing

- **Unit/Jest:** full SLO surface green (public components, `common/slo`, `server/services/slo`, `server/routes/slo`). New regression tests cover: rule-match drafts staying selectable vs saved-object coverage; `state`/`canonicalKind` correctness with a store double that mimics the SO pushdown gaps; datasource selection by name; the landing auto-select/URL-wins/cache-restore/empty-scope paths; and the shared `TruncatedLabel`.
- **Lint:** ESLint + Stylelint clean.
- **Live verification of Issue 1 (two-workspace scenario):** created a throwaway second workspace on the same AMP datasource; its ruler still exposed all 12 ad-labeled rules, but the coverage aggregate returned not-covered (0 saved objects) — i.e. the Suggest page now treats those drafts as creatable there — while the owning workspace stayed covered. Workspace deleted afterward.
- **Existing SLO Cypress specs reviewed:** no spec toggles the datasource facet, asserts the renamed badges, or intercepts `/v1/slos`; empty-state subjects (`slosEmptyFilteredZero`/`slosEmptyFilteredClear`) are unchanged, so no local spec is broken.

---

## Backward compatibility & caveats

- Server changes are behavior-only (no API/schema/mapping changes; no saved-object migration). The `cachedState` projection/writeback is left in place but no longer used for filtering.
- With ≥6 datasources, "Show all datasources" and the landing auto-select cap at 5 (by SLO count); the label reflects the cap.
- Follow-ups (not in this PR): projecting `canonicalKind` on the SO for index-level pushdown; deriving Service/Team/Tier facet options from the full catalog rather than the current page.

### Check List


public/components/apm/pages/slos/slo_list_filter_panel.tsx
```diff
@@ -384,37 +394,39 @@ const DatasourceFacet: React.FC<DatasourceFacetProps> = ({
...
+              // under a single datasource. The label truncates with an ellipsis.
+              overflowX: 'hidden',
               paddingRight: 4,
               marginRight: -4,
             }}
             data-test-subj="slosFilterAccordionDatasourceList"
           >
             {visible.map((ds) => {
-              const isSelected = selectedSet.has(ds.id);
+              const isSelected = selectedSet.has(ds.name);
```
lezzago, line 416: **Legacy-URL inconsistency.** Selection is keyed on `ds.name` here (and in `toggle` just below), which is the right call for freshly-created filters. But the active-filter badge in `slo_listing_page.tsx` tolerates a legacy saved-object id via `nameById.get(value) ?? value`. So an old bookmarked/shared link carrying `?datasourceId=<so-id>` renders the matching datasource **unchecked** (the set holds the id, not the name), and clicking it produces a mixed `[<so-id>, <name>]` filter for the same datasource.

Since the badge path already anticipates legacy ids, consider mirroring that here — e.g. resolve incoming ids to names once at hydration, or reverse-map with the same `nameById` when computing `isSelected`/`toggle` — so the checkbox state stays consistent with the badge.

common/slo/slo_query_service.ts
```diff
@@ -278,23 +295,10 @@ export class SloQueryService {
...
-          keptStates.push(cachedStates[i]);
-        }
-      }
-      docs = kept;
-      cachedStates = keptStates;
-    }
+    // `state` / `canonicalKind` requests never reach this branch — they route
+    // through `legacyPaginate` above — so the SO pushdown result is final.
+    const docs = paginated.docs;
+    const cachedStates = paginated.cachedStates;
```
lezzago: **Possibly-orphaned writeback.** Now that `state` pushdown was removed from `buildFilterKuery`, `cachedState` no longer feeds any filter path. Yet the best-effort `writeBackChangedStates(...)` a few lines below still fires on every SO-pushdown page to keep this `cachedStates` projection fresh. With no remaining consumer for the projection, that looks like dead maintenance work — repeated async SO updates on hot list paths.

If nothing else reads `cachedState`, it may be worth dropping the writeback (and the `cachedStates` plumbing) — or adding a note about the intended future consumer if it's being kept deliberately.

lezzago (commented):
A couple of follow-up observations from a closer read of the datasource-scope and pagination-routing logic. The core refactors look correct — routing `state`/`canonicalKind` through the materialize `list()` path is a solid fix for the pagination-totals bug, and keying the facet on connection name is right. Two edge cases below.

public/components/apm/pages/slos/slo_list_filter_panel.tsx
TackAdam: Good catch — fixed. You're right that keying the facet on `ds.name` left a legacy `?datasourceId=` link with the box unchecked and produced a mixed `[, ]` filter on click. There's also a consequence worth noting: since the server resolves the connection name (not the saved-object id), a legacy link didn't just look wrong — it returned 0 results.

I addressed it in both layers so display and query stay consistent:

- **Hydration (`slo_listing_page.tsx`):** the URL-wins branch of the scope-resolve effect now maps any incoming id → connection name (using the same id→name map the badge uses), dedupes, and rewrites `filters.datasourceId` before the first fetch. So the server query resolves and the URL heals to the canonical name form (no wrong-query, no mixed filter).
- **Facet (`slo_list_filter_panel.tsx`):** `selectedSet` reverse-maps ids → names, so the checkbox renders checked even for a transient/residual id, and toggling emits the name-keyed (healed) selection. Unknown values pass through unchanged.

**Regression tests added:**
- facet reverse-maps a legacy id to a checked box, and toggling it off emits the healed (empty) selection;
- the listing heals `?datasourceId=id-a` → queries by name `a` with no count-probes (URL-wins path).

common/slo/slo_query_service.ts
TackAdam: Agreed — removed the dead code. `state` is live-computed, so the cached projection had no real consumer. Removed the per-page `writeBackChangedStates` call + the helper (and its tests), `updateCachedState`, and the `cachedStates` plumbing from the stores and `SloPaginateResult`.

I deliberately left the `cachedState` keyword in the `slo-definition` mapping: removing a mapped field changes the saved-object type hash and forces a `.kibana` migration on upgrade, which isn't worth it for a now-unused field. It's annotated as a reserved/dormant keyword (no reader or writer).

Net effect: the maintenance work on hot list paths is gone, no saved-objects migration, and `state` resolves live via the materialize path. Unit tests pass (964).

lezzago (approved):

PR #2817, opened 2026-08-13
feat(alerting): unify Create metrics rule flyouts across Alert Manager and Metrics page

riysaxen-amzn:
### Description

Unifies the two "Create metrics rule" surfaces — the Alert Manager Rules page flyout and the Metrics (Explore) page's Create alert rule action — into **a single shared component**, so the two can never drift again.

 #### One flyout component (`CreateMetricsMonitor`) 

The Alert Manager Rules page now mounts the same `CreateMetricsMonitor` component the Metrics Explore page uses. Two optional props adapt it to the Alert Manager context (both inert on the Metrics page): - `datasources`: selectable Prometheus datasource list for the query panel picker, defaulting to the first entry (the Metrics page stays pinned to its Explore context datasource) - `isNameTaken`: duplicate-name guard surfacing an inline error on the Rule name row and disabling Create The flyout persists the rule itself; the Rules page reconciles with an optimistic row plus immediate and 15s-delayed refetches (Cortex propagation). Logs (PPL) creates keep the `CreateMonitor` shell, which also still serves Prometheus **edit** seeding (edit-flyout unification is a noted follow-up).

 #### Shared building blocks (used by both mount points) - 

**`PromQueryBuilder`** — point-and-click metric/label/operator/value selectors backed by live datasource metadata, with stale-response guards, PromQL escaping, and seeding from an existing expression (complex expressions leave the builder inert so a seeded query is never clobbered) - **`RuleGroupSelector`** — rule group combo box offering existing groups fetched live from the ruler API; type to create a new one - **`QueryPreviewResults`** — the Run preview results block (sample data for now; live range query as follow-up) - **`LabelEditor`** — shared label rows (equal-width key/value inputs, `=` separator, dynamic bolt toggle, quick-add chips) 

#### Flyout structure

 **Rule details**: read-only namespace (`observability-alerting`), rule group selector, rule name, hierarchy explanation, optional description (persisted as the `description` annotation — previously silently dropped on save; fixed) - **Query**: shared builder + per-rule `for:` duration + Run preview - **Labels / Annotations / Rule Preview (YAML)** - **Removed**: Trigger condition (the PromQL expression is the complete alert condition), Evaluation settings (rule-group-level in AMP), Matched notification actions and the Notification routing guidance section (routing is Alertmanager's job, driven by labels) 

#### Condition model & payload -

The Metrics page flyout no longer sends `operator`/`threshold` — previously the server appended them to the expr, which would silently turn `rate(...)` into `rate(...) > 0.05` - YAML previews single-quote templated label values, matching the server's js-yaml serialization 

#### Severity —

Aligned with Prometheus conventions - Severity is a routing **label** (Alertmanager matches its literal value); both flyouts seed `severity: warning` by default so a new rule never pages anyone unless deliberately escalated (metrics flyout previously seeded `critical`; Alert Manager previously mis-flagged the static default as a dynamic template) - Dynamic (templated) label values round-trip: edit seeding infers the dynamic flag from `{{ }}` syntax 

#### Misc -

 Alert Manager gains a **Build query in metrics →** link (same round-trip pattern as the logs flyout) - Detail flyout: routing link renamed to **View notification routing →** (read-only surface) 

### Testing -

 311 unit tests across 42 alerting suites pass, including new coverage for: builder seeding/round-trip, Run preview toggle, rule group propagation, dup-name blocking, datasource-list defaulting, payload shape (asserts `operator`/`threshold` absent), and template quoting - TypeScript and lint clean; formatting verified against both prettier 2.x and 3.x 

https://github.com/user-attachments/assets/09177b28-0e92-4805-bfdc-72411d0aee26

 ### Check List

 - [x] New functionality includes testing. - [x] All tests pass, including unit test, integration test and doctest - [ ] New functionality has been documented. - [x] Commits are signed per the DCO using --signoff By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

### DIFF
--- a/public/components/alerting/__tests__/create_metrics_monitor.test.tsx
+++ b/public/components/alerting/__tests__/create_metrics_monitor.test.tsx
@@ -95,8 +95,6 @@ describe('CreateMetricsMonitor', () => {
     expect(body).toMatchObject({
       name: 'my-test-rule',
       query: expect.any(String),
-      operator: '>',
-      threshold: expect.any(Number),
       forDuration: expect.any(String),
       evaluationInterval: expect.any(String),
       enabled: true,
@@ -104,6 +102,10 @@ describe('CreateMetricsMonitor', () => {
     });
     expect(body).toHaveProperty('labels');
     expect(body).toHaveProperty('annotations');
+    // The PromQL expression is the complete alert condition — no separate
+    // operator/threshold is sent (Trigger condition section was removed)
+    expect(body).not.toHaveProperty('operator');
+    expect(body).not.toHaveProperty('threshold');
 
     // Should call onSave and show success toast
     expect(onSave).toHaveBeenCalled();
--- a/public/components/alerting/__tests__/create_metrics_monitor.test.tsx
+++ b/public/components/alerting/__tests__/create_metrics_monitor.test.tsx
@@ -96,6 +96,25 @@ describe('CreateMetricsMonitor', () => {
     expect(createBtn.disabled).toBe(true);
   });
 
+  it('shows the "Build query in metrics" link only when requested (Alert Manager)', () => {
+    const { unmount } = render(
+      <CreateMetricsMonitor
+        onCancel={jest.fn()}
+        onSave={jest.fn()}
+        datasourceId="prom-1"
+        showBuildInMetricsLink
+      />
+    );
+    expect(
+      document.querySelector('[data-test-subj="alertManagerOpenInMetricsLink"]')
+    ).not.toBeNull();
+    unmount();
+
+    // Metrics Explore page context: link must be absent (circular hop)
+    render(<CreateMetricsMonitor onCancel={jest.fn()} onSave={jest.fn()} datasourceId="prom-1" />);
+    expect(document.querySelector('[data-test-subj="alertManagerOpenInMetricsLink"]')).toBeNull();
+  });
+
   it('POSTs the correct payload shape on save', async () => {
     const mockPost = jest.fn().mockResolvedValue({});
     const onSave = jest.fn();
--- a/public/components/alerting/__tests__/create_metrics_monitor.test.tsx
+++ b/public/components/alerting/__tests__/create_metrics_monitor.test.tsx
@@ -24,6 +24,14 @@ jest.mock('../promql_editor', () => ({
 jest.mock('../metric_browser', () => ({
   MetricBrowser: () => <div data-test-subj="metricBrowserMock" />,
 }));
+// Stub the shared builder with a button that emits a query, so tests can
+// simulate an explicit builder selection (the form seeds query: '' and the
+// Create button stays disabled until the builder produces one)
+jest.mock('../create_monitor/prom_query_builder', () => ({
+  PromQueryBuilder: ({ onQueryChange }: { onQueryChange: (q: string) => void }) => (
+    <button data-test-subj="mockBuilderSetQuery" onClick={() => onQueryChange('up{job="api"}')} />
+  ),
+}));
 
 global.ResizeObserver = jest.fn().mockImplementation(() => ({
   observe: jest.fn(),
@@ -131,9 +139,11 @@ describe('CreateMetricsMonitor', () => {
       />
     );
 
-    // Fill in required fields: monitorName
+    // Fill in required fields: monitorName + an explicit builder selection
+    // (the form seeds query: '' — no invisible default expression)
     const nameInput = document.querySelector('input[aria-label="Rule name"]') as HTMLInputElement;
     fireEvent.change(nameInput, { target: { value: 'my-test-rule' } });
+    fireEvent.click(document.querySelector('[data-test-subj="mockBuilderSetQuery"]')!);
 
     // Click Create button
     const createBtn = document.querySelector(
--- a/public/components/alerting/__tests__/create_metrics_monitor.test.tsx
+++ b/public/components/alerting/__tests__/create_metrics_monitor.test.tsx
@@ -57,6 +57,45 @@ describe('CreateMetricsMonitor', () => {
     expect(createBtn!.disabled).toBe(true);
   });
 
+  it('defaults to the first Prometheus datasource from a provided list (Alert Manager)', () => {
+    render(
+      <CreateMetricsMonitor
+        onCancel={jest.fn()}
+        onSave={jest.fn()}
+        datasources={[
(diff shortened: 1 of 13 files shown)


TackAdam:
## Verdict

Solid, well-documented refactor. The shared-component extraction (`PromQueryBuilder`, `RuleGroupSelector`, `QueryPreviewResults`, `LabelEditor`) is clean, the payload/severity changes are correct, and the whole alerting suite passes (312 tests). I verified the two claims that matter most:

- Payload shape matches the server route schema (`prometheus_routes.ts` — `operator`/`threshold` are `schema.maybe`, now correctly omitted). ✅
- Optimistic severity mapping in `handleMetricsRuleSaved` is an exact mirror of the server's `promSeverityFromLabels`. ✅

## Findings worth addressing

### 1. (Medium — real footgun) The default query is submittable but invisible in the builder.

`create_metrics_monitor.tsx:719` seeds `query: DEFAULT_PROMQL = rate(http_requests_total{status=~"5.."}[5m])`. That expression is not builder-representable, so `parseBuilderQuery` returns `null` and the builder renders empty and inert. But `isValid` only checks `form.query.trim() !== ''`, so a user who types just a name and clicks Create persists a rule querying `http_requests_total` — a metric they never chose, with an empty-looking builder. Your own test (`create_metrics_monitor.test.tsx:123`, `"POSTs the correct payload shape"`) does exactly this and passes, which codifies the behavior. Consider starting `query: ''` (and letting the placeholder live in the builder/preview), or blocking Create until the builder produces a query.

### 2. (Low) `isDirty` in `CreateMetricsMonitor` misses most fields.

`create_metrics_monitor.tsx:732` only compares `monitorName`, `query`, and label/annotation lengths. Editing the description, rule group, `for:` duration, datasource, or the content of a label/annotation won't trigger the discard-confirm modal — the user can lose those edits on Esc/overlay-click silently. Note the `CreateMonitor` shell (`index.tsx:238`) does this properly with a JSON snapshot; worth mirroring that here.

### 3. (Low) Dead/misleading namespace state.

`MetricsMonitorFormState.namespace` defaults to `'default'` (line 717) but the UI hard-codes/display `observability-alerting` and the payload never sends `namespace`. `form.namespace` is passed to `onSave` but ignored by both mount points. Either drop the field or set it to the real namespace to avoid confusing a future reader.

## Minor / non-blocking

- Stale file header: `create_metrics_monitor.tsx:6-11` still lists `"Trigger Condition, Evaluation Settings, Matched Notification Actions"` as sections — all removed in this PR.
- Duplicate static DOM ids: `QueryPreviewResults` uses `id="prom-preview-results"` and both flyouts reuse accordion ids like `prom-labels`. Harmless today (only one flyout mounts at a time), but a latent trap if that ever changes.
- Annotation templates aren't single-quoted in the YAML preview (`RulePreviewSection`, line 640) the way label templates are (line 634). Annotations lack the `isDynamic` flag, so a `{{ $value }}` annotation renders double-quoted in the preview — valid YAML, but slightly inconsistent with the "match server serialization" rationale you applied to labels.

lezzago:
Two issues in the shared "create metrics rule" flyout that I don't see covered by the existing bot suggestions or review notes:

### 1. Hardcoded `evaluationInterval: '1m'` can reset an existing rule group's evaluation cadence

`public/components/alerting/create_metrics_monitor.tsx:801`

The EvaluationSettings UI was removed, but `form.evalInterval` stays fixed at `'1m'` with no way to change it. When a user targets an **existing** rule group via `RuleGroupSelector`, the POST body carries `groupName` + `evaluationInterval: '1m'`. Since a Cortex/Prometheus group write replaces the whole group (interval included), a group that previously evaluated at, say, 30s is silently reset to 1m for **every** rule in it — not just the newly created one.

Suggestion: read the target group's existing interval when an existing `groupName` is selected, or surface the interval in the UI so the user isn't unknowingly rewriting it.

### 2. YAML Rule Preview omits the `description` annotation that the save path actually persists

`public/components/alerting/create_metrics_monitor.tsx:624`

`buildYaml` constructs annotations from `form.annotations` only, but the Description field is stored separately in `form.description` and is merged into annotations only inside `handleSave` (~line 811). A user who fills in Description sees a preview with no `description:` annotation, yet the created rule contains one — so the preview doesn't match what gets saved.

(This is distinct from the earlier bot suggestion about a user-typed `description` annotation key being overwritten by the Description field via spread order — that's an annotation-key collision on the save path; this one is a preview-vs-saved-rule divergence.)

riysaxen-amzn:
Thanks @TackAdam and @lezzago — all findings addressed in d3155d76: **@TackAdam:** 1. **Invisible default query (medium)** — fixed exactly as suggested: the form now seeds `query: ''`, so Create stays disabled until the builder produces an expression. No more silently persisting `http_requests_total` for a metric the user never chose. The payload test now makes an explicit builder selection instead of codifying the old behavior. 2. **isDirty coverage** — switched to a JSON snapshot (mirroring the `CreateMonitor` shell), so description/group/duration/datasource/label-content edits all arm the discard-confirm modal. 3. **Dead namespace state** — `form.namespace` now carries the real `observability-alerting` value (shared constant, also used by the read-only field). 4. Minors: file header refreshed; `CreateMetricsMonitor` accordion ids namespaced (`cmm-*`) and `QueryPreviewResults` takes an `id` prop so the two flyouts can't collide; templated annotation values are single-quoted in both YAML previews, consistent with labels. **@lezzago:** 1. **evaluationInterval resetting existing groups** — this is already guarded server-side by the #2796 create-merge path: `handleCreatePrometheusRule` preserves `existing.interval` when writing into a non-empty group (covered by the `'preserves the existing group evaluation interval when merging'` test in `prometheus_handlers.test.ts`). The client's `'1m'` only ever applies to brand-new groups. Surfacing the interval in the UI is a reasonable follow-up, but there's no reset hazard today. 2. **Preview omits the Description annotation** — real gap, fixed: the Rule Preview now merges `form.description` into annotations exactly the way the save path does (including the field winning over a manually-typed `description` annotation), so preview and saved rule always match. **CI:** the two failing checks are pre-existing on main, unrelated to this diff — `panels_test` (`Add ppl filter to panel`) has failed the last 5 consecutive main runs, and the FTR E2E failure is in `6_notebooks.spec.js` (notebooks editor timeouts); this PR touches only `components/alerting/**`.

TackAdam (approved):

lezzago (approved):

PR #2819, opened 2026-08-13
feat(observability): generic error classification & surfacing layer

lezzago:
## Description

Adds a framework-agnostic **error-classification and surfacing layer** so opaque failures in the SLO and unified-alerts features stop collapsing to `HTTP 500 {"message":"An internal error occurred"}` and instead show a clear **category, cause, remediation, and correlation id**.

### Core — `common/error/` (pure TS; no OSD/React/EUI imports)
- Provider-neutral taxonomy + message catalog, redaction, and a priority-ordered classifier registry with a translator seam and a detail-enricher seam.
- Default classifiers: upstream-wrapped (inspects the **inner** cause), ruler, state (rule_health / routing), timeout, and generic HTTP-status. The `UNKNOWN` fallback surfaces a **redacted** message rather than a bare generic string.
- **Registration-only extension model** — downstream consumers add higher-priority classifiers/enrichers with zero core edits (in-repo generic example + README).

### Server
- Boundary adapter mints a correlation id, logs full detail server-side, and applies a client-exposure policy (`observability.errors.exposeSensitiveErrorDetail`, default **off**).
- Prometheus rule create/delete and SLO handler paths classify instead of collapsing to a generic 500; a wrapped inner 409 conflict is reclassified from 503.
- **Backward compatible**: legacy `message`, `error`, `code`, `httpStatus`, and `rawBody` fields are preserved; `errorDetail` + `correlationId` are added alongside them.
- **Response-status behavior change (worth calling out):** on the Prometheus rule create/delete route, a `RULER_UNREACHABLE` failure previously produced HTTP 500 (via the generic `toHandlerResult` fallback). It now returns the classified status — for a real upstream unreachable, that's **503** — which is semantically more accurate. A client asserting `=== 500` on that specific failure needs updating; body shape is unchanged and legacy fields still populate. The SLO route preserves its existing status via the `legacyStatus` guard, so its wire status is unaffected.

### Client
- Toast + inline callout render adapter (localized via `@osd/i18n`, redacted by default); a "See the full error" toast opens a modal with the copyable safe diagnostic.
- Correlation id shown only for escalation categories.
- Wired into rule creation (metrics + explore), the alerts page, routing status, and the SLO wizard (verbatim raw body now redacted).

### Safety
Sensitive raw upstream text never reaches the browser unless explicitly opted in (a registered enricher or the default-off config flag). The full detail is always retrievable server-side via the correlation id.

## Testing
Unit tests for classifiers, redaction, the registry, a downstream-override example, the server adapter, and the client extract/callout helpers.

## Check List

### DIFF
--- a/common/error/README.md
+++ b/common/error/README.md
@@ -1,0 +1,143 @@
+# Error classification & surfacing (`common/error`)
+
+A small, **framework-agnostic** layer that turns opaque failures (the classic
+`HTTP 500 {"message":"An internal error occurred"}`) into a **classified,
+sanitized** result with a stable category, code, user-facing title/message,
+remediation, retryability, and a correlation id — so both features (SLOs and
+the unified alerts view) surface a clear cause instead of a generic wall.
+
+The core is pure TypeScript: **no** imports from OpenSearch Dashboards, Kibana,
+React, or EUI. Rendering (toasts/callouts) and i18n resolution live in
+framework-specific adapters. This is what lets the whole `common/error/`
+directory drop into a downstream fork unchanged; the fork only re-implements
+the thin adapters and registers its own extras.
+
+## Taxonomy
+
+Provider-neutral categories (`ErrorCategory`):
+
+| Category | Meaning |
+|---|---|
+| `VALIDATION` | The request was rejected as invalid (bad input, invalid query). |
+| `NOT_FOUND` | The target resource does not exist. |
+| `CONFLICT` | The request conflicts with current state (e.g. "already exists"). |
+| `PERMISSION_DENIED` | Not authenticated (`AUTH_REQUIRED`) or not allowed (`PERMISSION_DENIED`). |
+| `UPSTREAM_UNAVAILABLE` | A backend could not be reached / did not complete. |
+| `TIMEOUT` | The request did not finish in time / was aborted. |
+| `RATE_LIMITED` | The backend is throttling requests. |
+| `PRECONDITION_FAILED` | The resource changed since it was loaded. |
+| `PARTIAL_STATE` | A successful response in an unusable state (missing rule groups, unknown routing status). |
+| `UNKNOWN` | Unmatched — surfaced with a **redacted** message, never a raw one. |
+
+`code` narrows a category to a specific, actionable situation (e.g.
+`RULE_GROUP_CONFLICT`, `RULE_BACKEND_UNAVAILABLE`, `RULE_CONFIG_INVALID`,
+`RULES_MISSING`). Wording for each code lives in `messages.ts`.
+
+## Pipeline
+
+```
+RawErrorContext ──▶ classifyError() ──▶ ClassifiedError ──▶ toClientPayload() ──▶ wire
+                     (highest-priority        (resolved,        (exposure policy)
+                      matching classifier)     localized)
+```
+
+- `classifyError(ctx)` picks the highest-`priority` classifier whose `match`
+  returns true (else the `UNKNOWN` fallback), resolves wording from the catalog
+  (or the classifier's inline overrides) through the registered `Translator`,
+  runs any enrichers, and echoes the correlation id.
+- `toClientPayload(err, { exposeSensitive })` enforces the exposure policy
+  before anything reaches the browser.
+
+## The three-tier exposure guarantee
+
+Raw upstream text can leak deployment topology, so exposure is tiered:
+
+1. **Server logs** — the full, un-redacted detail is always logged with the
+   correlation id at the server boundary. Never dropped.
+2. **`safe` details** — a **redacted** excerpt (`redactForDisplay`) is attached
+   and shown by default (toast text / "View details" callout).
+3. **`sensitive` details** — verbatim upstream text. **Stripped from client
+   payloads by default.** Revealed only when opted in:
+   - an operator sets `observability.errors.exposeSensitiveErrorDetail: true`, or
+   - a downstream `ErrorDetailEnricher` opts in.
+
+**Invariant:** the browser never receives raw upstream text unless explicitly
+opted in. Even `UNKNOWN` surfaces only a redacted message.
+
+`redactForDisplay` scrubs URLs, hostnames, IPs, ARNs, UUIDs, long opaque ids,
+account-number-like runs, and long mixed tokens. It is best-effort
+defense-in-depth, not a security boundary — the exposure gate above is. Known
+gaps: a **single-label** hostname with no port and no dotted TLD (e.g. a bare
+`cortex-ruler`) can pass through; a forker that needs stricter scrubbing should
+add rules via its own enricher rather than editing core.
+
+## Extension model — registration only, no core edits
+
+A downstream fork extends behavior **at startup**, touching no file in
+`common/error/`:
+
+```ts
+import {
+  registerDefaultClassifiers,
+  registerErrorClassifier,
+  registerErrorDetailEnricher,
+  setTranslator,
+} from 'common/error';
+
+registerDefaultClassifiers();               // core defaults
+setTranslator(myFrameworkTranslate);        // @osd/i18n here, @kbn/i18n in the fork
+
+// (a) higher-priority classifier overrides a default and can inline its own wording
+registerErrorClassifier({
+  name: 'myenv.ruleBackendUnavailable',
+  priority: 200,                            // > any core priority (max 100)
+  match: (ctx) => ctx.operation === 'rule.create.metric' && ctx.upstreamCode === 'RULER_UNREACHABLE',
+  classify: () => ({
+    category: 'UPSTREAM_UNAVAILABLE',
+    code: 'MYENV_RULE_BACKEND_UNAVAILABLE',
+    retryable: true,
+    messages: { title: { id: '…', defaultMessage: '…' }, message: { id: '…', defaultMessage: '…' } },
+  }),
+});
+
+// (b) enricher adds environment-specific detail (safe values are re-redacted;
+//     it cannot remove safety guarantees)
+registerErrorDetailEnricher({
+  name: 'myenv.supportHint',
+  enrich: (err) => ({ ...err, details: [...(err.details ?? []), /* … */] }),
+});
+```
+
+A complete, provider-neutral example lives in
+`examples/example_downstream_registration.ts` (exercised by
+`__tests__/downstream_registration.test.ts`). Provider-specific names,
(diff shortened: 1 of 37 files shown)


TackAdam:
## Verdict

This is a well-designed, well-tested layer. The three-tier exposure guarantee (server log full → safe redacted → sensitive gated by flag) is coherent, the classifier priority ordering is correct for the documented cases, and the registration-only extension model genuinely means downstream forks avoid core edits. Tests exercise both the pipeline internals and the extension seam, and the server adapter test covers the correlation-id log path plus the sensitive-strip default.

Legacy fields (`error`, `code`, `httpStatus`, `rawBody`) are preserved in every changed response, so existing clients keep working while new ones can consume `errorDetail`/`correlationId`.

## Findings

### 1. `timeoutClassifier` (priority 60) can hijack a legitimate 4xx validation error (`common/error/classifiers/timeout.ts:19`)

```ts
return typeof ctx.message === 'string' && /timed out|timeout/i.test(ctx.message);
```

An HTTP 400 whose upstream message happens to contain the word `"timeout"` (e.g. an OpenSearch parse error mentioning a timeout field name) is scored higher than `httpStatusClassifier` (priority 40) and lands in `TIMEOUT` with `retryable: true`. A retry would still fail with the same 400. `ruler`/`state`/`upstream_wrapped` outrank it, so this is a corner case — worth tightening to also require a status/errorName signal, e.g. `httpStatus === 408 || 504 || errorName in {AbortError,TimeoutError}` OR require both `httpStatus >= 500` and message match.

### 2. Prometheus rule route status is now 503 where it was 500 (`server/routes/alerting/mutations/prometheus_routes.ts:92,133`)

Old `toHandlerResult` treated a `SloRulerError` as an unclassified `Error`, falling through to the generic 500. New path returns whatever `clientErr.httpStatus` says — for `RULER_UNREACHABLE` with a real upstream 503, that's 503. Semantically more accurate, but a subtle contract change if any client checks `=== 500`. Worth calling out in the PR description under "Backward compatible" (which currently only mentions body shape). SLO route already preserves legacy behavior via the `legacyStatus` guard.

### 3. `rawBody` on the SLO route body is still un-redacted server-side (`server/routes/slo/handlers.ts:98–102`)

The three-tier guarantee applies to `errorDetail`. The legacy `rawBody` field only goes through `sanitizeRulerRawBody`, which truncates length but keeps URLs/hostnames/IPs. `slo_wizard_page.tsx` now wraps it in `redactForDisplay` at render time, but any other consumer of this route (e.g. a downstream fork's alternate wizard, a copy-diagnostic-to-clipboard action, DevTools) sees the un-redacted body. Either redact server-side for defense in depth, or drop `rawBody` and require consumers to read `errorDetail.details`.

### 4. Toast id collisions on concurrent errors (`public/components/common/toast/index.tsx:19`)

`id: new Date().toISOString()` — pre-existing, but the new error-classification paths make simultaneous toasts more likely (a form submit that triggers two failure toasts in the same ms merges into one EUI toast entry). Not required for this PR, but the file's already being touched — swap to `crypto.randomUUID()` (already used in `classified_error.ts`) or a monotonic counter.

### 5. `routingStatus` treats `settling` as unavailable (`public/components/alerting/notification_routing_panel.tsx:353`)

`classifyRoutingStatus` returns non-null for anything not exactly `'ready'`, so an Alertmanager cluster in the valid transient state `'settling'` (starting up, gossip in progress) now surfaces as `"Routing status unavailable"` instead of the literal `settling`. Health color also flipped from danger→warning for this branch, so it's still clearly non-green — but the title is slightly misleading. Consider whitelisting the known transient states (`settling`, `disabled`) as their own labels, or fall back to the literal status when it's a recognized non-error value.

### 6. `extractClassifiedError` assumes attributes wrapper (`public/components/common/error/extract.ts:29–41`)

Reads `err.body.attributes.errorDetail`. This matches `toErrorBody` (alerting) and the SLO route's `{message, attributes: result.body}` wrap, so it works today. If any future route registers `errorDetail` at `body.errorDetail` directly (which reads naturally from `classifyToHandlerResult`'s body shape), the client silently falls back to the generic toast. Cheap fix: try `body.errorDetail` if `body.attributes.errorDetail` is absent.

### 7. Nitpicks

- `common/error/redact.ts:53–57` — FQDN pattern requires a recognized TLD suffix; single-label hostnames (`cortex-ruler`, `alertmanager-0`) as commonly seen in Kubernetes DNS pass through un-redacted. Consider a broader `"dotted-word:port"` fallback OR accept and document the gap.
- `common/error/classifiers/upstream_wrapped.ts:25–27` — `hasInnerConflict` matches on message text; risks a false positive if a 5xx wraps a validation error whose text contains `"already exists"` as a noun (e.g. `"'name' field already exists in the schema"`). Tighten to require the 409 signal alongside `"already exists"` (currently `||`).
- `common/error/registry.ts:189–191` — `localizeClassified` re-translates the label but not the value. That's fine when the value is data (an excerpt or count), but if a downstream classifier stores localizable text there, it'll surface in the server's translation. Worth a comment.
- Registering the built-in defaults from both `server plugin.ts:369` and `client plugin.tsx:223` is correct given they run in different bundles, but a downstream fork that imports the barrel in a shared test-harness could hit the `defaultsRegistered` short-circuit. The `__resetDefaultsFlagForTests` seam already covers this — call it out in the README.

### 8. Non-issues I checked and cleared

- No circular imports through `common/error/index.ts` — classifiers import `../registry`, not the barrel.
- `configureErrorExposure` module state is per-process; tests reset it in `beforeEach`.
- `randomUUID` from `node:crypto` is fine on Node 22.
- The wrapped-inner-409 reclassification test (`classified_error.test.ts:74`) proves the priority-100 override defeats the outer 503 → HTTP 409 to the client. Good.
- Server always logs full raw upstream + correlation id regardless of client exposure — the invariant holds.

## Suggested wording tweaks

- README, `"Portability checklist"`: the `"Do not change"` list should mention the priority scale (`0–100` for core, `200+` for downstream) so forks don't accidentally collide.
- `classified_error.ts:113–115` — the log message concatenates `ctx.rawBody` verbatim; if it's a large object, log lines can balloon. Consider capping at ~4KB with an ellipsis (same threshold as `RULER_RAW_BODY_MAX_CHARS` in `handlers.ts`) for log hygiene, since we already sanitize the client copy.

lezzago:
Thanks for the thorough review, @TackAdam. Force-pushed `8dea5e13` addressing each finding — mapped to your numbering:

1. **`timeoutClassifier` hijack of 4xx (`common/error/classifiers/timeout.ts`)** — tightened so a message-only match is only trusted when there's no client-error status (or the status is 5xx). AbortError / TimeoutError / 408 / 504 still match on their own. Regression test in `classifiers.test.ts` proves a 400 mentioning "timeout" now falls through to VALIDATION.
2. **500→503 on the Prometheus route (`server/routes/alerting/mutations/prometheus_routes.ts`)** — called out explicitly in the PR description now (see the new "Response-status behavior change" bullet). Body shape and legacy fields unchanged; SLO route still uses `legacyStatus` guard so its wire status is unaffected.
3. **Un-redacted `rawBody` in the SLO handler (`server/routes/slo/handlers.ts`)** — the legacy `rawBody` now goes through `redactForDisplay` before `sanitizeRulerRawBody`, so *any* consumer of that route (not just the wizard) gets scrubbed text. Full body is still logged server-side under the correlation id. Regression test in `handlers_repair_and_health.test.ts`.
4. **Toast id collisions (`public/components/common/toast/index.tsx`)** — swapped to `${new Date().toISOString()}-${++seq}`. Small monotonic counter; no dependency on `crypto.randomUUID` since this bundle also runs in insecure contexts.
5. **`settling` mislabeled as unavailable (`common/error/classifiers/state.ts`)** — added `NON_ERROR_ROUTING_STATES = {ready, settling}`; both surface as their literal status. Test in `classifiers.test.ts`.
6. **`extractClassifiedError` attributes-only (`public/components/common/error/extract.ts`)** — now falls back to `body.errorDetail` when the `attributes` wrapper is absent. Test in `extract.test.ts`.

Nits:
- **Single-label host redaction (`common/error/redact.ts`)** — added a `name:port` rule so bare `alertmanager-0:9093`-style k8s DNS is scrubbed. README updated to acknowledge the remaining gap (single-label host with no port).
- **`upstream_wrapped` "already exists" false positive (`common/error/classifiers/upstream_wrapped.ts`)** — now requires a 409 marker alongside "already exists"; explicit HTTP-409 markers still match on their own. Test proves a 400 with `'name' field already exists in the schema` no longer promotes to CONFLICT.
- **`localizeClassified` value/label semantics (`common/error/registry.ts`)** — added the comment explaining detail values are data, not localizable text, and calling out how a downstream should handle localized detail text.
- **`__resetDefaultsFlagForTests` seam** — mentioned in the README's Portability checklist.
- **Log-line ballooning (`server/routes/alerting/classified_error.ts`)** — capped at 4096 chars with an ellipsis (`LOG_RAW_MAX_LEN`) so a large upstream body can't blow up a log line.
- **README priority scale** — added: core = 0–100, downstream should use 200+.

New tests: **203 total (up from 198)**. Lint clean.

lezzago:
Follow-up push (`2d92c86d`):

**Rebased onto latest `main`** — resolved the conflict with #2804 (Add detector/forecaster management) in `public/components/alerting/alarms_page.tsx`. The conflict was purely in the state-hook declarations; kept #2804's new `createAdRuleType` / `editAdTarget` state and my classified-error toast wiring in `handleCreateMonitor` (both verified present, no markers).

**Two additional issues found via an independent self-review of the review-response changes, now fixed:**
1. **`redact.ts` over-redaction** — the new single-label `host:port` rule scrubbed ordinary prose like `at line:42` / `column:15`. Tightened so the host must look network-ish (contain a hyphen or digit); e.g. `alertmanager-0:9093` is redacted, `line:42` is preserved. Test added.
2. **`upstream_wrapped.ts` regression from the "require 409" tightening** — a genuine conflict masked by a 5xx whose body says "already exists" but doesn't echo the literal `409` would have fallen through to a retryable `UPSTREAM_UNAVAILABLE`. Reworked: an explicit inner 409 marker is conclusive; otherwise "already exists" is a conflict only when the outer status is 409 or 5xx/unknown (a plain 4xx that merely mentions it stays `VALIDATION`; a bare 409 with no such body still maps to the generic `RESOURCE_CONFLICT`). Three focused tests cover these branches.

Unit tests: **205 pass**. Lint clean.

TackAdam (approved):

PR #2820, opened 2026-08-13
feat(alerting): surface unified-alerts errors as toasts + per-datasource indicators

lezzago:
### Description

Move the Alert Manager page-level error surface off of stacked `EuiCallOut` banners onto event-driven toasts, and add a persistent per-datasource error indicator in the filter facet so users can drill from _"something is wrong"_ to _"which datasource, and why"_ without leaving the page. Also fixes a broken sentence in the "new alerting experience" intro callout.

**Behavior — before vs. after**

Before, the Alerts and Rules tabs stacked up to four `EuiCallOut` banners at the top of the page ("Error loading alerts", "Error loading data", "Alerting plugin not detected", "Some datasources could not be reached"), plus two inline warnings inside the alerts dashboard ("Search incomplete — too many alerts", "Showing current alerts only"). Datasource connection failures were only summarized in the aggregate banner; the affected row in the datasource facet had no visual marker. Users could dismiss a banner but a re-render brought it back.

After, each of those six conditions fires exactly once per transition as an `EuiToast` (`danger` for hook errors, `warning` for the four degraded-state conditions). Per-datasource connection failures also decorate the affected row in the filter facet with an `alert` icon that opens a popover with the datasource name and the underlying error text. The banner strip is gone; the visualization row is anchored to the top of the panel. First-mount duplicate-fire is prevented by ref-tracked transition keys.

**Client changes**

- `hooks/use_alerting_page_toasts.ts` (new): centralized transition-aware toast hook. Sorts + joins each incoming set into a stable key and only fires when the key changes, so a datasource that stays down does not spawn a fresh toast on every re-render. Six conditions covered.
- `facet_filter_panel.tsx`: adds optional `errorMap: Record<string, string>` to `FacetFilterGroup`. When set, affected options render a trailing `EuiIcon type="alert" color="danger"` button that toggles an `EuiPopover` showing "{displayLabel} — {error text}". Extracted a `FacetErrorIndicator` sub-component so each row can own its own popover state (hook-in-map safe).
- `alarms_page.tsx`: builds a merged per-datasource issue list from `alertsWarnings` + `rulesWarnings` + the alerting-plugin probe's `unavailableDsIds` (skipped when every OS DS failed — that's the aggregate "plugin not detected" toast's job). Deduped by ds id; first error wins. Emits `datasourceIssues` for the toast hook and a name-keyed `datasourceErrorMapByName` for the facet indicator. Grammar fix on the intro callout: `"Welcome to the new alerting experience view your..."` → `"Welcome to the new alerting experience. View your..."` (missing period + trailing period).
- `alerts_dashboard.tsx`: drops the inline `truncated` and `fallbackHints` callouts and their corresponding props; accepts `datasourceErrorMap` and threads it to the datasource facet. Toasts for those two conditions now come from `useAlertingPageToasts`.
- `monitors_table/index.tsx` + `monitors_filters_panel.tsx`: thread the same `datasourceErrorMap` through so the Rules tab's datasource facet gets the identical indicator.
- **Deleted:** `alarms_page_callouts.tsx` and its `.altCalloutSpaced` SCSS block — no other consumers.

**Design notes**

- Toast + indicator, not either/or. The filter panel can be collapsed or scrolled off; the toast is the one-shot notification, and the indicator is the persistent home for the details.
- The plugin-probe adorns individual rows only when the probe is done AND we're not in the "every OS DS failed" aggregate state — otherwise every row would carry a redundant icon on top of the aggregate toast.
- Per-datasource error text is keyed by NAME (not id) in the facet's `errorMap` because `FacetFilterGroup` uses the display label as the option key. The internal issue list is keyed by id for stable dedup.

### Screenshots

**Alerts tab — before:** banners stacked at the top, ungrammatical welcome text, no per-datasource error indicator.

Alerts tab before

**Alerts tab — after:** banners replaced by toasts in the bottom-right, red alert icon next to the affected datasource, grammatical welcome text.

Alerts tab after

**Datasource error popover — after:** clicking the indicator opens a popover framed with "Could not connect" plus the datasource name and underlying error.

Popover after

**Rules tab — before:** same two banners plus the ungrammatical welcome text.

Rules tab before

**Rules tab — after:** banners gone, indicator + grammatical welcome text.

Rules tab after

### Issues Resolved

N/A — UX cleanup surfaced during error-classification integration.

### Check List

  - [x] All tests pass, including unit test, integration test and doctest — full alerting suite (327 tests) green.

**Test plan**

- [x] `yarn jest public/components/alerting/__tests__/alarms_page.test.tsx public/components/alerting/__tests__/alerts_dashboard.test.tsx public/components/alerting/__tests__/facet_filter_panel.test.tsx` — 47/47 passing
- [x] Full alerting suite (`public/components/alerting/`) — 327/327 passing
- [x] `yarn lint` on all changed files — clean
- [x] Typecheck — zero errors under `dashboards-observability` (unrelated preexisting errors in `src/plugins/*` unaffected)
- [x] Manual verification against a running observability stack: banners replaced by toasts; datasource indicator appears next to failing DS; popover shows underlying error; grammar fix visible on the intro callout; both Alerts and Rules tabs

**Tests updated / added**

- New `facet_filter_panel.test.tsx` case: `errorMap` renders the alert icon only on affected options and reveals the popover contents on click.
- `alarms_page.test.tsx`: rewrote three obsolete banner-render assertions as toast + prop assertions (unreachable DS, Prom fallback, OS truncation).
- `alerts_dashboard.test.tsx`: removed the four callout-render tests (truncated ×2, fallback ×2) — the callouts are gone.


TackAdam:
## Verdict

Clean, well-reasoned refactor. The banner→toast migration is coherent, the transition-dedup design is sound, the name-keyed `errorMap` correctly lines up with how both facets key options (`label: ds.name` in `alerts_dashboard.tsx:498` and `monitors_table/index.tsx:191`), and the deletion of `alarms_page_callouts.tsx` + its SCSS is complete with no dangling consumers. Tests pass.

## Findings, most important first

### 1. Hook's core logic (transition dedup) has no direct test — Medium

`use_alerting_page_toasts.ts` is 215 lines and its entire reason to exist is "fire once per transition, never on re-render." Yet there's no `__tests__/use_alerting_page_toasts.test.ts`. The `alarms_page` tests only assert that a toast fires, not the hard part: that a sustained failure across re-renders doesn't spawn duplicate toasts, and that recover→fail re-fires. This is the regression the whole hook was built to prevent, and it's untested. I'd ask for a `renderHook`-based test that rerenders with a stable prop set and asserts `setToast` is called exactly once.

### 2. Duplicate toast on datasource-list hydration — Low

In `alarms_page.tsx:1397`, the dedup key uses `const id = ds?.id ?? dsName`. Before datasources hydrates, the key is name-based (`dsName|error`); after hydration it becomes `id|error`. If a warning is present across that transition, `datasourceIssuesKey` changes and the toast fires twice. In practice warnings usually only appear after datasources load (selection drives the fetch), so it's unlikely — but it's a real edge given the async ordering isn't guaranteed. Worth a comment or keying consistently.

### 3. Toasts now fire regardless of active tab — Low (intentional?)

The old `alertsErrorMessage` banner was gated on `activeTab === 'alerts'`, and the truncated/fallback callouts only rendered inside `AlertsDashboard` (alerts tab only). `useAlertingPageToasts` is now called unconditionally at page level, so an alerts-derived truncation/fallback/error toast can fire while the user is sitting on the Rules tab. Arguably fine (the error is real either way), but it's a behavior change not called out in the description — worth confirming it's intended.

### 4. Style nits — Low

- `FacetErrorIndicator` uses a raw `<button>` with a block of inline styles. `CLAUDE.md` calls for SCSS with a 3-letter prefix, and `EuiButtonIcon` (supports `onClick` + `aria-label`) would be the idiomatic, theme-consistent choice. The `stopPropagation`/hook-in-map rationale is valid, but neither forces raw inline styles.
- `data-test-subj={\`facetGroup-${facetId}-error-${option}\`}` interpolates the datasource *name*, which can contain spaces/special chars — `CLAUDE.md` wants `data-test-subj` values camelCase. Functional, just non-conforming and slightly brittle for selectors.

TackAdam (approved):

lezzago:
Thanks for the thorough review @TackAdam (and the bot). Addressed in 4a27ca6b — kept as a separate commit so the delta is easy to see. Summary:

**#1 (Medium) — hook had no direct test.** Added `public/components/alerting/hooks/__tests__/use_alerting_page_toasts.test.ts` (`renderHook`) that asserts the actual contract, not just "a toast fired":
- a sustained condition across multiple re-renders (fresh prop objects each time) toasts **exactly once**;
- recover → recur **re-fires**;
- a datasource-issue set with new array/object identity but unchanged `(name, error)` does **not** re-toast;
- the datasource key stays stable across the id hydration (see #2);
- tab gating (see #3).

7 new tests; full alerting suite now 341 passing.

**#2 (Low) — duplicate toast on datasource-list hydration.** Fixed by keying the dedup on the datasource **name** instead of the derived id. Since the call site uses `id = ds?.id ?? name`, the id flips name→id once datasources hydrate; the name is always present, so the key is now stable across that transition. Covered by the "id hydrates from name to real id" test.

**#3 (Low) — toasts fired regardless of active tab.** Good catch — restored the prior behavior. The three alerts-derived toasts (alerts fetch error, truncation, Prometheus fallback) are now gated to the Alerts tab. The transition ref is advanced only when a toast actually fires, so if the condition still holds when the user switches **to** the Alerts tab, it fires then rather than being silently consumed. Plugin-missing, rules-fetch error, and per-datasource connection toasts remain cross-cutting (matching the old callouts, which rendered above the tab bar / on both tabs).

**#4 (Low) — style nits.** `FacetErrorIndicator` now uses `EuiButtonIcon` instead of a raw `<button>` with inline styles; spacing/width moved to `.altFacetErrorBtn` / `.altFacetErrorPopover` in `alerting.scss` (3-letter `alt` prefix). On the `data-test-subj`: I left the `facetGroup-${id}-error-${option}` shape because it matches the pre-existing convention already used in this component (`facetGroup-${id}-icon-${opt}`, `facetGroup-${id}-<option>`), so it stays consistent with the surrounding code rather than introducing a one-off.

**Bot — keyboard propagation.** Added `onKeyDown`/`onKeyUp` propagation stops alongside `onClick`/`onMouseDown` so activating the indicator via Enter/Space never bubbles to toggle the datasource checkbox.

**Bot — "always update ref to detect recurrence."** No change needed — the datasource-issues effect already unconditionally advances the ref after the `if`, and the (now name-based) key resets to `''` when issues clear, so recover→recur re-fires. The new hook test locks this in.

PR #2821, opened 2026-08-14
feat(observability): SLO onboarding + Suggest toolbar + tidier rows; fix older monitor types in Alerts Rules tab

lezzago:
## Description

Improves the **SLO listing** onboarding/empty state, adds a **"Suggest SLOs" toolbar button**, tidies up the **SLO catalog table rows**, and fixes how the **unified Alerts Rules tab** handles the older OpenSearch monitor types (composite/workflow support, simplified type labels, and a richer bucket-level flyout).

### SLO listing
- Replaced the generic *"No SLOs yet"* onboarding with a **services-aware empty state** (`SloNoSlosEmptyState`):
  - **Services discovered → leads with "Suggest SLOs"**, which auto-drafts availability + latency objectives for each service (the body shows the discovered-service count).
  - **No services yet → "Set up services to get started"**, guiding the user to set up service monitoring first.
  - "Create manually" remains the secondary path in both cases. The discovery query is mount-gated so it runs only when there are zero SLOs.
- Added a **"Suggest SLOs" toolbar button** (left of "Create SLO") once the workspace already has SLOs, to steer users to the easier batch flow. It is **disabled with an explanatory tooltip** when no APM services are discovered.
- **Tidied the catalog rows:** the Name column now truncates to a single line (full name on hover + via the detail link) and the text columns have deterministic widths, so long SLO names no longer word-break into cramped multi-line rows on narrower screens.

### SLO listing — empty state (no services configured)

| Before | After |
|---|---|
|  |  |

### SLO listing — empty state (services discovered)
Leads with the **Suggest SLOs** batch-onboarding flow.

### SLO listing — "Suggest SLOs" toolbar button
Added left of "Create SLO" once SLOs exist. **Enabled** when services are set up; **disabled with a tooltip** when they aren't.

| Before (toolbar) | Disabled — no services (tooltip) | Enabled — services set up |
|---|---|---|
|  |  |  |

### SLO catalog — Name column no longer scrunched
Long names used to word-break across ~4 lines on narrower widths; now they stay on one line and elide with an ellipsis.

| Before | After |
|---|---|
|  |  |

The full name is available on hover via a tooltip (and on the detail page):

## Unified Alerts — Rules tab & monitor flyout
Tested creating all five older OpenSearch monitor types (per-query, per-bucket, per-document, cluster-metrics, composite) and viewing each in the Rules tab + flyout.

**Composite (workflow) monitors** were returned by the monitors search with no `monitor_type`, so they were coerced to query-level and shown as **"Metric"**, with a broken flyout: an empty `{}` query, a nonsensical `Threshold > 0`, and **Edit/Clone/Delete/Disable enabled** (acting on them would corrupt the workflow). Now recognized structurally (`composite_input`) → labeled **"Composite"** (which also auto-gates Edit/Enable-Disable), with an **"Associated monitors"** list of member monitors and Clone/Delete disabled. Because a composite is a workflow (its detail can't be fetched from the monitors endpoint), the flyout drops the "could not load details" banner and the fields that don't apply (threshold, pending period, condition-preview chart), showing only the member monitors + evaluation interval.

| Composite flyout — before (broken) | After (clean) |
|---|---|
|  |  |

**Type labels** were guessed from index names (per-document → "Log", per-bucket → "Infrastructure", per-query → "APM"/"Metric"), which described a data domain rather than the monitor mechanism. Consolidated so per-query / per-bucket / per-document all surface as **"Log"**, with **"Cluster Metrics"** and **"Composite"** kept distinct.

| Type column — before | After |
|---|---|
|  |  |

**Bucket-level flyout** previously showed only the raw query JSON; it now renders a structured view — target indices, group-by fields, and the per-bucket condition — with the full aggregation query behind a toggle. Detail-load failures on cluster-metrics/doc/bucket monitors now show a plain "configuration unavailable" note instead of a malformed query block.

| Bucket flyout — before | After |
|---|---|
|  |  |

## Testing
Verified against a live observability stack (OTel demo data) via a dev server:
- SLO empty state shows the "Set up services" branch with no services, and the "Suggest SLOs" branch when services are discovered.
- With real SLOs present, the toolbar shows "Suggest SLOs" left of "Create SLO" — enabled when services are configured, disabled + tooltip otherwise.
- Long SLO names stay on a single line and truncate with a tooltip.

Automated:
- `slo_listing_page.test.tsx` — covers both empty-state branches and the enabled/disabled toolbar button.
- `slo_listing_page_datasource_scope.test.tsx` + `slo_listing_page_pagination.test.tsx` — updated to stub `useServices` for the new toolbar button.
- Full plugin Jest suite (512 tests) + ESLint pass locally.

## Check List
- [x] All tests pass
- [x] New functionality has been documented / covered by tests

### DIFF
--- a/common/types/alerting/opensearch_types.ts
+++ b/common/types/alerting/opensearch_types.ts
@@ -82,13 +82,25 @@ export type OSMonitorInput =
         queries: Array<{ id: string; name: string; query: string; tags: string[] }>;
       };
     }
+  | {
+      // Composite (workflow) monitors sequence other monitors. Their documents
+      // are returned by the monitors search alongside regular monitors, so the
+      // UI must recognize this input shape to label + render them correctly
+      // rather than mis-detecting them as query-level monitors.
+      composite_input: {
+        sequence: {
+          delegates: Array<{
+            order: number;
+            monitor_id: string;
+            chained_monitor_findings?: { monitor_id?: string } | null;
+          }>;
+        };
+      };
+    }
   | OSPPLInput;
 
 export type OSMonitorType =
-  | 'query_level_monitor'
-  | 'bucket_level_monitor'
-  | 'doc_level_monitor'
-  | 'ppl_monitor';
+  'query_level_monitor' | 'bucket_level_monitor' | 'doc_level_monitor' | 'ppl_monitor';
 
 export interface OSMonitor {
   id: string;
--- a/public/components/alerting/__tests__/monitor_detail_flyout.test.tsx
+++ b/public/components/alerting/__tests__/monitor_detail_flyout.test.tsx
@@ -22,18 +22,20 @@ global.ResizeObserver = jest.fn().mockImplementation(() => ({
 
 // MonitorDetailFlyout instantiates AlertingOpenSearchService internally
 // via `useMemo(() => new AlertingOpenSearchService(), [])` and calls
-// `getRuleDetail(dsId, ruleId)` on mount. Mock the class so the
-// constructor returns a stubbed instance with `getRuleDetail` resolving
-// to `null` — the flyout falls back to the monitor summary in that
-// case, which is what these render tests exercise.
+// `getRuleDetail(dsId, ruleId)` on mount. Mock the class so every instance
+// shares one `getRuleDetail` spy the tests can drive per-case: resolve `null`
+// (the default — flyout falls back to the summary), resolve a full detail
+// (exercises the structured bucket view), or reject (exercises the
+// detail-unavailable / suppressed-error-banner branches).
+const mockGetRuleDetail = jest.fn();
 jest.mock('../query_services/alerting_opensearch_service', () => ({
   AlertingOpenSearchService: jest.fn().mockImplementation(() => ({
-    getRuleDetail: jest.fn().mockResolvedValue(null),
+    getRuleDetail: mockGetRuleDetail,
   })),
 }));
 
 import { MonitorDetailFlyout } from '../monitor_detail_flyout';
-import type { UnifiedRuleSummary } from '../../../../common/types/alerting';
+import type { UnifiedRule, UnifiedRuleSummary } from '../../../../common/types/alerting';
 
 const mockMonitor: UnifiedRuleSummary = {
   id: 'mon-1',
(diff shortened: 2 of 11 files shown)


riysaxen-amzn (approved):

jackiehanyang (commented):
I found a functional issue to address before merging.

[P2] Empty SLO onboarding CTA in no-SLOs state navigates unscoped and generates no draft objectives

In [slo_no_slos_empty_state.tsx](https://github.com/opensearch-project/dashboards-observability/blob/5fad083b737c1050b0ae91add6d9a860e976801a/public/components/apm/pages/slos/slo_no_slos_empty_state.tsx), when discovered services exist, the CTA calls . That produces a suggest-page URL without , and  treats missing scope services as , which causes the draft list to be empty.

This conflicts with the empty-state copy (“we’ve discovered services and can draft SLOs”) and means users land on an empty state with no auto-generated proposals.

Suggested fix: collect discovered service names (serviceName in services data) and pass them to , and assert in tests that the expected array is passed.

Thank you.

jackiehanyang (commented):
overall lgtm, just the above findings

TackAdam:
### Test-coverage gap on the flyout

The substantial, safety-relevant new branches — composite rendering, disabled Clone/Delete, the structured bucket view, and `detailUnavailableForStructuredKind` — have zero test coverage (`grep` for `composite`/`Associated monitors`/`detailUnavailable` in the flyout test → 0 hits). The existing 4 tests still pass but exercise none of this.

Given the gating is what prevents workflow corruption, a test asserting Clone/Delete are disabled for a composite would be cheap insurance. The PR only added tests for the SLO listing side.

TackAdam (approved):

lezzago:
Thanks @jackiehanyang — fixed in 8c652243.

The no-SLOs empty-state CTA now collects the discovered service names and passes them to `navigateToSloSuggest(...)`, so the Suggest page lands pre-scoped and auto-drafts availability + latency objectives for each service (matching the onboarding copy) instead of dropping the user on the empty "pick services" prompt.

I also caught a latent bug in the test mock while adding coverage: it built `{ name }` but the component reads `.serviceName`, so the empty array was slipping through unnoticed. Fixed the mock and added an assertion that the expected `['payments-api', 'checkout']` array is passed.

Left the `Suggest SLOs` toolbar button (shown once SLOs exist) landing on the service picker unchanged — there the user is curating additional/uncovered services rather than onboarding from zero, so the picker is the intended entry. Happy to scope that too if you'd prefer consistency.

lezzago:
Good catch @TackAdam — the flyout branches were untested. Added coverage in 1cd7609b.

New `monitor_detail_flyout.test.tsx` tests, driven by a controllable `getRuleDetail` mock:

- **Composite gating (the workflow-corruption guard):** renders `Associated monitors` + member ids, and asserts Clone / Delete / Edit are all disabled **and inert** (`onClone`/`onDelete` are not called on click). Also asserts the expected-404 detail error is suppressed (no error banner) and that threshold / pending-period / condition-preview don't render for composites.
- **Structured bucket view:** asserts target indices, the group-by badge extracted from `terms.field`, the per-bucket condition, and the aggregation-query toggle render from the raw monitor body.
- **Detail-unavailable fallback:** on a structured kind whose detail fetch rejects, asserts the plain 'Detailed configuration is unavailable' note renders instead of the abbreviated summary string.

7/7 tests pass; lint clean.

PR #2822, opened 2026-08-14
Improve Alerts Manager empty state

jackiehanyang:
## Description

Improves the Alerts Manager empty state so the unified Alerting, Anomaly Detection, and Forecasting capabilities are visible when no datasource or rules are available.

- Replaces the generic empty prompt with capability cards for Alerting, Anomaly Detection, and Forecasting.
- Opens the existing anomaly detector and forecaster creation flyouts from their cards.
- Adds a Logs versus Metrics choice before opening an alert rule creation flyout.
- Improves the All Alerts empty-table messages to mention alerts and detected anomalies.
- Removes the introductory banner and link to the classic alerting experience.

The primary Alerts Manager navigation remains Alerts, Rules, and Routing; this change does not add a Forecasts tab.

## Validation

- `alerts_dashboard.test.tsx`: 12 tests passed.
- Changed files pass `git diff --check`.
- Changed files pass Prettier 3.5.3 formatting checks using the OpenSearch Dashboards configuration.
- Verified the empty-state cards, Logs/Metrics chooser, existing creation flyouts, navigation tabs, and banner removal on a local dashboard.

### DIFF
--- a/public/components/alerting/__tests__/alerts_dashboard.test.tsx
+++ b/public/components/alerting/__tests__/alerts_dashboard.test.tsx
@@ -97,6 +97,10 @@ const baseProps = {
   rulesTotal: 1,
   defaultDatasources: [],
   onGoToRules: jest.fn(),
+  onCreateLogsRule: jest.fn(),
+  onCreateMetricsRule: jest.fn(),
+  onCreateAnomalyDetection: jest.fn(),
+  onCreateForecasting: jest.fn(),
   startMs: NOW - HOUR_MS,
   endMs: NOW,
   pickerStart: 'now-24h',
@@ -113,16 +117,72 @@ describe('AlertsDashboard', () => {
   it('renders "no alerts in range" empty state when rules exist but no alerts', () => {
     const { getByText } = render(<AlertsDashboard {...baseProps} />);
     expect(getByText('No alerts in the selected time range')).toBeInTheDocument();
+    expect(
+      getByText('No alerts or anomalies were detected in the selected time range.')
+    ).toBeInTheDocument();
   });
 
   it('renders "no datasource" empty state when selection is empty', () => {
     const { getByText } = render(<AlertsDashboard {...baseProps} selectedDsIds={[]} />);
-    expect(getByText('No datasource selected')).toBeInTheDocument();
+    expect(
+      getByText('Define rules, detect anomalies, and forecast from one place')
+    ).toBeInTheDocument();
+    expect(getByText('Alerting')).toBeInTheDocument();
+    expect(getByText('Anomaly detection')).toBeInTheDocument();
+    expect(getByText('Forecasting')).toBeInTheDocument();
+    expect(
+      getByText('Select a datasource to view alerts and detected anomalies.')
+    ).toBeInTheDocument();
+  });
+
+  it('offers logs and metrics creation from the Alerting capability card', () => {
+    const onCreateLogsRule = jest.fn();
+    const onCreateMetricsRule = jest.fn();
+    const onCreateAnomalyDetection = jest.fn();
+    const onCreateForecasting = jest.fn();
+    const { getByText } = render(
+      <AlertsDashboard
+        {...baseProps}
+        selectedDsIds={[]}
+        onCreateLogsRule={onCreateLogsRule}
+        onCreateMetricsRule={onCreateMetricsRule}
+        onCreateAnomalyDetection={onCreateAnomalyDetection}
+        onCreateForecasting={onCreateForecasting}
+      />
+    );
+
+    fireEvent.click(getByText('Alerting'));
+    expect(screen.getByTestId('alertsEmptyAlertingRuleTypeModal')).toBeInTheDocument();
+    fireEvent.click(screen.getByTestId('alertsEmptyCreateLogsRule'));
+    expect(onCreateLogsRule).toHaveBeenCalledTimes(1);
+
+    fireEvent.click(getByText('Alerting'));
+    fireEvent.click(screen.getByTestId('alertsEmptyCreateMetricsRule'));
+    expect(onCreateMetricsRule).toHaveBeenCalledTimes(1);
+
+    fireEvent.click(getByText('Anomaly detection'));
+    expect(onCreateAnomalyDetection).toHaveBeenCalledTimes(1);
+
+    fireEvent.click(getByText('Forecasting'));
+    expect(onCreateForecasting).toHaveBeenCalledTimes(1);
   });
 
   it('renders "no rules or detectors" empty state when rulesTotal is 0', () => {
     const { getByText } = render(<AlertsDashboard {...baseProps} rulesTotal={0} />);
-    expect(getByText('No rules or detectors have been created')).toBeInTheDocument();
+    expect(
+      getByText('Define rules, detect anomalies, and forecast from one place')
+    ).toBeInTheDocument();
+    expect(
+      getByText(
+        'Create an alerting, anomaly detection, or forecasting rule for the selected datasource.'
+      )
+    ).toBeInTheDocument();
+    expect(
+      getByText(
+        'Create a rule to get started. Triggered alerts and detected anomalies will appear here.'
+      )
+    ).toBeInTheDocument();
+    expect(getByText('Go to Rules')).toBeInTheDocument();
   });
 
   it('renders alert table when alerts provided', () => {
--- a/public/components/alerting/__tests__/alerts_dashboard.test.tsx
+++ b/public/components/alerting/__tests__/alerts_dashboard.test.tsx
@@ -153,6 +153,7 @@ describe('AlertsDashboard', () => {
 
     fireEvent.click(getByText('Alerting'));
     expect(screen.getByTestId('alertsEmptyAlertingRuleTypeModal')).toBeInTheDocument();
+    expect(document.querySelectorAll('.euiOverlayMask, .ouiOverlayMask')).toHaveLength(1);
     fireEvent.click(screen.getByTestId('alertsEmptyCreateLogsRule'));
     expect(onCreateLogsRule).toHaveBeenCalledTimes(1);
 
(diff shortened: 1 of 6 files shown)


TackAdam:
## Summary

Solid, well-scoped change. It replaces the generic empty prompts with a unified `UnifiedSignalsEmptyState` (Alerting / Anomaly detection / Forecasting capability cards), adds a Logs-vs-Metrics chooser modal, improves the empty-table messages, and removes the "classic experience" callout. Test coverage is good and the diff is mostly clean refactoring plus the new component.

## One substantive finding

### AD/Forecaster empty-state cards bypass the OpenSearch-datasource guard — `alarms_page.tsx:1323-1324`

The Rules tab routes detector/forecaster creation through `handleCreateADResource` (`alarms_page.tsx:1358-1359`), which calls `resolveOpenSearchDatasourceId()` first. That guard checks a selected OpenSearch datasource exists and, if not, shows the toast `"Select an OpenSearch datasource before creating AD resources"` and refuses to open the flyout (`alarms_page.tsx:984-990`).

The new empty-state cards wire straight to the setter, skipping the guard:

```tsx
onCreateAnomalyDetection={() => setCreateAdRuleType('detector')}
onCreateForecasting={() => setCreateAdRuleType('forecaster')}
```

Consequences, which land squarely in the scenarios this empty state targets:

- **no-ds mode** (no datasource selected — the main case for these cards): the flyout opens anyway. Via `getInitialDatasourceId` (`create_ad_rule_flyout.tsx:356-362`) it either silently defaults to an arbitrary first OpenSearch datasource the user never selected, or — if none exists — opens with an empty datasource that fails validation (`"OpenSearch datasource is required"`).
- **no-rules mode with only a Prometheus datasource selected:** same bypass; the flyout opens instead of the guided toast.

The logs/metrics cards don't have this problem — they replicate the Rules-tab logic exactly. Only the AD/forecaster path diverges.

**Suggested fix (reuses the existing guard):**

```tsx
onCreateAnomalyDetection={() => handleCreateADResource('detector')}
onCreateForecasting={() => handleCreateADResource('forecaster')}
```

## Minor notes (non-blocking)

- Empty-prompt title heading went from `<h4>` to `<h3>` while the sibling no-alerts prompt still uses `<h4>` — minor heading-level inconsistency within the same panel.
- The removed `NEW_EXPERIENCE_CALLOUT_DISMISSED_KEY` leaves stale `localStorage` entries in existing users' browsers. Harmless, not worth cleanup.
- The `[React Intl] Could not find required intl object` console noise in the test run comes from a pre-existing test (line 329), not the new tests — unrelated to this PR.

lezzago:
### Metrics card is a dead end when no Prometheus datasource exists — `alarms_page.tsx:1319`

This is distinct from the `no-ds`-mode concerns already raised. `CreateMetricsMonitor` filters datasources to `type: 'prometheus'`, so in **any** deployment with no Prometheus datasource configured — not just when nothing is selected — the Metrics alert-rule card opens a flyout with an empty dropdown and nothing to pick. The Rules-tab replication doesn't cover this case, because the precondition is a missing datasource *type*, not a missing selection. Disabling the Metrics card when no Prometheus datasource is configured should be enough to close this.

### On the AD/Forecaster guard bypass (already flagged)

Just adding the root cause behind that finding: the four empty-state handlers re-implement the type→flyout-state mapping that already lives in the Rules-tab `onCreateMonitor` switch (`alarms_page.tsx:1351-1361`), and diverge from it by calling `setCreateAdRuleType` directly instead of `handleCreateADResource`. Routing all four cards through a single shared handler (detector/forecaster via `handleCreateADResource`) fixes the guard bypass and removes the duplicate mapping that let the paths drift in the first place — one change addresses both.

jackiehanyang:
> ## One substantive finding
> ### AD/Forecaster empty-state cards bypass the OpenSearch-datasource guard — `alarms_page.tsx:1323-1324`
> The Rules tab routes detector/forecaster creation through `handleCreateADResource` (`alarms_page.tsx:1358-1359`), which calls `resolveOpenSearchDatasourceId()` first. That guard checks a selected OpenSearch datasource exists and, if not, shows the toast `"Select an OpenSearch datasource before creating AD resources"` and refuses to open the flyout (`alarms_page.tsx:984-990`).
> 
> The new empty-state cards wire straight to the setter, skipping the guard:
> 
> ```tsx
> onCreateAnomalyDetection={() => setCreateAdRuleType('detector')}
> onCreateForecasting={() => setCreateAdRuleType('forecaster')}
> ```
> 
> Consequences, which land squarely in the scenarios this empty state targets:
> 
> * **no-ds mode** (no datasource selected — the main case for these cards): the flyout opens anyway. Via `getInitialDatasourceId` (`create_ad_rule_flyout.tsx:356-362`) it either silently defaults to an arbitrary first OpenSearch datasource the user never selected, or — if none exists — opens with an empty datasource that fails validation (`"OpenSearch datasource is required"`).
> * **no-rules mode with only a Prometheus datasource selected:** same bypass; the flyout opens instead of the guided toast.
> 
> The logs/metrics cards don't have this problem — they replicate the Rules-tab logic exactly. Only the AD/forecaster path diverges.
> 
> **Suggested fix (reuses the existing guard):**
> 
> ```tsx
> onCreateAnomalyDetection={() => handleCreateADResource('detector')}
> onCreateForecasting={() => handleCreateADResource('forecaster')}
> ```
> 


Thanks for calling this out. The bypass is a valid concern, but I found the suggested handleCreateADResource guard relies on a page-level selected datasource, while Alerts Manager supports opening the creation flow before a datasource is selected.
I updated the flow so the AD and Forecasting flyouts only list standard OpenSearch datasources, never silently select an arbitrary datasource, and remain unselected unless a compatible datasource is already selected. If no compatible OpenSearch datasource is configured, the Anomaly detection and Forecasting cards are disabled with an explanation that an OpenSearch datasource is required. This prevents the invalid submit path while preserving datasource selection within the creation flow.

jackiehanyang:
> ### Metrics card is a dead end when no Prometheus datasource exists — `alarms_page.tsx:1319`
> This is distinct from the `no-ds`-mode concerns already raised. `CreateMetricsMonitor` filters datasources to `type: 'prometheus'`, so in **any** deployment with no Prometheus datasource configured — not just when nothing is selected — the Metrics alert-rule card opens a flyout with an empty dropdown and nothing to pick. The Rules-tab replication doesn't cover this case, because the precondition is a missing datasource _type_, not a missing selection. Disabling the Metrics card when no Prometheus datasource is configured should be enough to close this.
> 
> ### On the AD/Forecaster guard bypass (already flagged)
> Just adding the root cause behind that finding: the four empty-state handlers re-implement the type→flyout-state mapping that already lives in the Rules-tab `onCreateMonitor` switch (`alarms_page.tsx:1351-1361`), and diverge from it by calling `setCreateAdRuleType` directly instead of `handleCreateADResource`. Routing all four cards through a single shared handler (detector/forecaster via `handleCreateADResource`) fixes the guard bypass and removes the duplicate mapping that let the paths drift in the first place — one change addresses both.

Thanks, this is a valid gap. I updated the Metrics alert-rule card to require an available Prometheus datasource. When none is configured, the card is disabled and explains the requirement instead of opening CreateMetricsMonitor with an empty datasource selector. I also added regression coverage for that path.
On the AD/Forecaster routing point, the duplicate mapping is worth reducing, but the existing handleCreateADResource requires a datasource selected in the page filter. The empty-state flow intentionally allows creation before that selection, with the datasource chosen in the flyout. Reusing that handler unchanged would block valid creation. The updated AD/Forecasting flow instead limits the flyout to standard OpenSearch datasources, leaves the datasource blank when none is selected, and disables the cards only when no compatible datasource exists at all.

TackAdam (approved):

lezzago (approved):

lezzago:
### Follow-up on the two prior findings

Thanks for the quick turnaround. Re-reviewed the latest commits — one fix is complete, the other still has a gap.

**✅ Metrics card dead-end — fixed.** The Metrics card is now `isDisabled={!hasPrometheusDatasource}` with a clear "A Prometheus datasource is required…" description, and `EuiCard isDisabled` suppresses the click. No longer a dead end.

**⚠️ AD/Forecaster guard bypass — only partially fixed.** The fix disables the card instead of routing the CTA through the shared `handleCreateADResource` handler, and the disable check diverges from the real guard:

- The empty-state cards still call `setCreateAdRuleType('detector'|'forecaster')` directly (`alarms_page.tsx:1323-1324`), bypassing `resolveOpenSearchDatasourceId()` — unlike the Rules-tab path at `alarms_page.tsx:1359`.
- The new disable guard keys off **system-wide** datasource existence (`datasources.some(isStandardOpenSearchDatasource)`), whereas the real guard checks the **selected** datasource.

The remaining hole is: an OpenSearch datasource exists in the system but isn't selected.

- In `no-ds` mode (nothing selected) or `no-rules` mode (only a Prometheus DS selected), the AD/Forecasting cards stay **enabled**, the flyout opens, and the guidance toast never fires.
- The `create_ad_rule_flyout.tsx` change swapped the old `?? openSearchDatasources[0]?.id` fallback for `?? ''` (lines 356-363), and the `EuiSelect` has no `hasNoInitialSelection`/placeholder — so it visually shows the first OpenSearch DS while `form.datasourceId` is `''`, producing a "Datasource is required" error on an apparently-filled field.
- Worst case: with exactly one OpenSearch datasource (single local cluster) that isn't selected, the select shows that lone option already highlighted while state is `''`; re-picking the already-displayed option in a native select doesn't fire `onChange`, so the form can't be completed.

The new test only covers the "no OpenSearch DS anywhere" case (serverless + Prometheus), so it doesn't exercise this gap.

**Suggested fix:** route the empty-state CTAs through `handleCreateADResource` (same as the Rules tab) instead of `setCreateAdRuleType`, and/or base the card disable on the *selected* datasource rather than system-wide existence; plus restore an auto-preselect (or add `hasNoInitialSelection` + a placeholder) so the select value can't silently be `''`.

Minor: the Metrics card has the same system-vs-selected divergence (Prometheus DS exists but isn't selected → card enabled), but the create-monitor save path has a backstop toast, so it's not a dead end — lower priority.

PR #2823, opened 2026-08-14
Alerts: classic-experience escape hatches for unsupported monitor types

TackAdam:
### Description

The new alerting experience doesn't yet support every monitor type (per-bucket, per-cluster-metrics, per-document, composite) or edit them. This PR adds two escape hatches to the classic alerting app so users aren't blocked, plus a backdrop fix for the confirmation modals.

**1. "Open the classic experience" link in the Create logs rule flyout**

The Create logs rule flyout only builds per-query PPL rules. Added a link in the flyout header (under the subtitle, create-mode logs only) to the classic create-monitor flow for the types we don't cover yet, with a tooltip explaining why. Placed at the flow level in the header — deliberately separated from the query-scoped "Build query in logs" link so the two don't compete.

Wording leads with the new term and keeps the old one in parentheses ("Need another rule (monitor) type?") since existing users know these as monitors while we transition to "rule".

**2. Edit redirect for non-PPL rule types in the monitor detail flyout**

Previously the Edit button was disabled for every non-PPL/non-metric rule with a "only supported for PPL" tooltip. Now those OpenSearch monitor types show a clickable Edit that opens a confirmation and then deep-links to the classic alerting app's edit form:

`/app/monitors#/monitors/{id}?action=edit-monitor&monitorType={type}&dataSourceId={id}`

- `{id}` is the monitor's OpenSearch document `_id`, carried through unchanged from the backend (`mapMonitor` → `osMonitorToUnifiedRuleSummary`).
- `{type}` is derived from the preserved backend `monitor_type`/`monitor_kind` labels (cluster-metrics is special-cased since it's stored as `query_level_monitor`).
- `{dataSourceId}` is always included so the classic app can resolve its data source on a cold deep link (omitting it caused "DataSource was not set" + a `DataSourceView` crash when MDS is enabled). MDS monitors use their data-source saved-object id; the local-cluster sentinel maps to an empty value.

PPL/Prometheus-metric rules still edit in place; detectors/forecasters and non-OpenSearch rules keep the disabled tooltip.

**3. Remove redundant overlay masks**

`EuiConfirmModal` renders its own `EuiOverlayMask`; the edit-redirect and discard-changes modals were each wrapped in a second one, double-dimming the backdrop. Removed the redundant wrappers so the modals use the standard single backdrop.

### Testing

- Added unit tests for the edit-redirect: URL construction (`id`, `type`, `dataSourceId`), cluster-metrics type mapping, local-cluster empty `dataSourceId`, and that Cancel doesn't navigate.
- [ ] Manually verify the classic edit redirect against both a local-cluster and an MDS-data-source rule.
- [ ] Verify the "Open the classic experience" link from the Create logs rule flyout.

### Check List

### DIFF
--- a/public/components/alerting/__tests__/monitor_detail_flyout.test.tsx
+++ b/public/components/alerting/__tests__/monitor_detail_flyout.test.tsx
@@ -32,8 +32,19 @@ jest.mock('../query_services/alerting_opensearch_service', () => ({
   })),
 }));
 
+// Stub the shared core refs so the Edit-redirect path can assert on
+// `navigateToApp` without a real Core start contract.
+jest.mock('../../../framework/core_refs', () => ({
+  coreRefs: {
+    application: { navigateToApp: jest.fn() },
+  },
+}));
+
 import { MonitorDetailFlyout } from '../monitor_detail_flyout';
+import { coreRefs } from '../../../framework/core_refs';
 import type { UnifiedRuleSummary } from '../../../../common/types/alerting';
+
+const navigateToApp = coreRefs.application!.navigateToApp as jest.Mock;
 
 const mockMonitor: UnifiedRuleSummary = {
   id: 'mon-1',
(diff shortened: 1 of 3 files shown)


lezzago:
**Minor / defensive — not a blocker.**

In `public/components/alerting/monitor_detail_flyout.tsx` (`goToClassicEdit`), the classic deep link interpolates `monitor.id`, `classicMonitorType`, and `dataSourceIdParam` raw into the URL:

```js
path: `#/monitors/${monitor.id}?action=edit-monitor&monitorType=${classicMonitorType}&dataSourceId=${dataSourceIdParam}`,
```

None are `encodeURIComponent`-wrapped. If any value ever contains a URL-reserved character (`&`, `#`, `=`, `%`, space), it'd be parsed as URL structure rather than data and the classic editor could receive a truncated id or an injected param.

In practice the risk is low: `classicMonitorType` is a backend enum (`query_level_monitor`, etc.), `dataSourceIdParam` is a saved-object id, and `monitor.id` is a system-generated OpenSearch `_id` (URL-safe). So this is defensive hardening, not a live bug — but since `monitor.id` flows through unchanged from the backend, wrapping the three values is cheap insurance:

```js
path: `#/monitors/${encodeURIComponent(monitor.id)}?action=edit-monitor&monitorType=${encodeURIComponent(classicMonitorType)}&dataSourceId=${encodeURIComponent(dataSourceIdParam)}`,
```

Fine to defer if you'd rather keep the diff tight.

lezzago (approved):

TackAdam:
Good call — addressed. Wrapped all three interpolated values in encodeURIComponent:
```
  const id = encodeURIComponent(monitor.id);
  const monitorTypeParam = encodeURIComponent(classicMonitorType);
  const dsParam = encodeURIComponent(dataSourceIdParam);
  coreRefs?.application?.navigateToApp(CLASSIC_MONITORS_APP_ID, {
    path: `#/monitors/${id}?action=edit-monitor&monitorType=${monitorTypeParam}&dataSourceId=${dsParam}`,
  });
```

PR #2826, opened 2026-08-26
[Alerting] Shared primitives: i18n enum labels, tz-aware timestamps, theme color tokens

lezzago:
## What & why
Foundation for the batch. Adds a single home for how raw alert/rule enums are rendered as human text: i18n severity/state labels, `getMonitorStateLabel` for rule status/health (passing AD/forecaster sentence-statuses through unchanged), a timezone-aware `formatTimestamp` that names the zone, an `EMPTY_VALUE` placeholder, and `euiThemeVars`-backed severity/state color tokens. Pure additions — no behaviour change on its own.

**Findings:** CLAR1 (label maps), CLAR9 (tz formatter), M6 groundwork (color tokens).

## Before / after

**Before / after — the effect the module produces (its call sites are #2825, #2827, #2829, #2839)**

Before / after — the effect the module produces (its call sites are #2825, #2827, #2829, #2839)

**Flow**

Flow

## Testing
464 lines of unit tests across `enum_labels`/`time_format`/`alert_colors`. Verified centrally (127 suites / 1294 tests green).

## Review
Independent review agent: **APPROVE** — confirmed `getMonitorStateLabel` never re-cases already-cased statuses, i18n ids unique, theme keys valid.

## Regression check

## Dependency
**Merge this first.** #2825, #2827, #2829, and #2839 import these primitives and currently include this change on their branch; once this merges, a rebase drops it from their diffs.

### DIFF
--- a/public/components/alerting/__tests__/alert_colors.test.ts
+++ b/public/components/alerting/__tests__/alert_colors.test.ts
@@ -1,0 +1,63 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import {
+  ALERT_KIND_HEX,
+  SEVERITY_HEX,
+  STATE_HEX,
+  UNKNOWN_HEX,
+  getSeverityHex,
+  getStateHex,
+} from '../alert_colors';
+import type { UnifiedAlertSeverity, UnifiedAlertState } from '../../../../common/types/alerting';
+
+const ALL_SEVERITIES: UnifiedAlertSeverity[] = ['critical', 'high', 'medium', 'low', 'info'];
+const ALL_STATES: UnifiedAlertState[] = [
+  'active',
+  'pending',
+  'acknowledged',
+  'silenced',
+  'resolved',
+  'error',
+];
+
+describe('alert_colors', () => {
+  it('resolves every severity to a real theme color', () => {
+    ALL_SEVERITIES.forEach((severity) => {
+      // A typo in a euiThemeVars key yields `undefined`, which silently renders
+      // as "no color" — assert we got an actual CSS color back.
+      expect(SEVERITY_HEX[severity]).toMatch(/^#[0-9a-f]{3,8}$/i);
+    });
+  });
+
+  it('resolves every state to a real theme color', () => {
+    ALL_STATES.forEach((state) => {
+      expect(STATE_HEX[state]).toMatch(/^#[0-9a-f]{3,8}$/i);
+    });
+  });
+
+  it('resolves both row kinds and keeps anomaly distinct from a high-severity alert', () => {
+    expect(ALERT_KIND_HEX.alert).toMatch(/^#[0-9a-f]{3,8}$/i);
+    expect(ALERT_KIND_HEX.anomaly).toMatch(/^#[0-9a-f]{3,8}$/i);
+    expect(ALERT_KIND_HEX.anomaly).not.toBe(SEVERITY_HEX.high);
+  });
+
+  it('gives critical and active the danger tone', () => {
+    expect(getSeverityHex('critical')).toBe(SEVERITY_HEX.critical);
+    expect(getStateHex('active')).toBe(STATE_HEX.active);
+    expect(getStateHex('error')).toBe(STATE_HEX.active);
+  });
+
+  it('resolves the anomaly row kind through the state getter', () => {
+    expect(getStateHex('anomaly')).toBe(ALERT_KIND_HEX.anomaly);
+  });
+
+  it('falls back to a subdued tone for missing or unknown values', () => {
+    expect(getSeverityHex(undefined)).toBe(UNKNOWN_HEX);
+    expect(getSeverityHex('sev0')).toBe(UNKNOWN_HEX);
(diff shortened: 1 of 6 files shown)


TackAdam:
## Worth addressing

1. **`formatTimestamp` drops stringified-epoch input — `time_format.ts:70`**

   `const date = value instanceof Date ? value : new Date(value);`

   `new Date('1756145045000')` → `Invalid Date`, so a backend that serializes a long timestamp as a JSON string renders `—` instead of the real time. `moment.tz(Number(value), tz)` would parse it.

   ISO strings and numeric input are handled fine (and tested) — whether this matters depends entirely on what shape the alerting backends emit. Cheap to guard: coerce all-digit strings to a number before the `Date` normalization.

2. **`humanizeUnknown` shouts on upper/mixed-case unknowns — `enum_labels.ts:70-74`**

   `return spaced.charAt(0).toUpperCase() + spaced.slice(1);`

   It upper-cases the first char but never lower-cases the rest. An unmapped `AT_RISK` renders as `AT RISK`, not `At risk` — defeating the "polished prose" purpose for exactly the unknown-value path this function exists to handle.

   Note the AD/forecaster pass-through case (`"Awaiting data to init"`) must stay intact, so any fix needs to only down-case when the token looks machine-style (e.g. no existing spaces, or all-caps). Worth a moment's care.

## Nits (fine to defer)

3. **Theme colors are frozen at module-load — `alert_colors.ts:27-55`**

   The maps are const-evaluated from `euiThemeVars` at import time, so a runtime light/dark toggle (no reload) keeps the stale palette. This matches the existing `apm/common/constants.ts` pattern (relies on reload-on-theme-change), so it's low severity — but the module doc claims the values "resolve against whichever theme is active," which over-promises. Either soften the comment or make them getters.

4. **`getStateHex` vs `getStateLabel` disagree on unknowns — `alert_colors.ts:67`**

   An unmapped state gets a humanized label but a subdued "unknown" swatch — same cell reads as recognized in text yet greyed-out in color. Looks intentional; just confirm it's the desired behavior.

5. **Doc/format mismatch — `time_format.ts:38-40`**

   JSDoc says the offset looks like `+05:30`, but `moment.format('z')` emits `+0545` (your own test at `time_format.test.ts:97` asserts this). Doc-only.

6. **Minor duplication — `enum_labels.ts:73`**

   The capitalize-first idiom already exists in `shared_constants.ts` (`formatDatasourceType`). A shared `capitalizeFirst` would keep them in sync, but not blocking.

lezzago:
Thanks @TackAdam — addressed in `e2e921f5`:

1. **`formatTimestamp` stringified epoch** ✅ — all-digit strings are now coerced to a number before the `Date` normalization, so `"1756145045000"` parses like the numeric path instead of rendering `—`. Test added.
2. **`humanizeUnknown` shouting** ✅ — machine-style tokens (separated or all-caps) are now sentence-cased, so `AT_RISK` → "At risk". Space-containing sentences (AD/forecaster statuses like "Awaiting data to init") and mixed-case no-separator tokens (`inProgress`) pass through unchanged, and the all-separators fallback is preserved. Tests added for the shouting fix and the internal-casing preservation.
3. **Theme colors frozen at load** ✅ (doc) — softened the module comment to state the palette is resolved once at import and relies on reload-on-theme-change, matching the sibling `apm/common/constants.ts`.
4. **`getStateHex` vs `getStateLabel` on unknowns** ✅ — confirmed intentional and documented on `UNKNOWN_HEX`: an unmapped state is still humanized to text but gets a neutral swatch rather than a guessed semantic color (danger/success), since the UI can't infer severity.
5. **Doc/format mismatch (`+05:30` vs `+0545`)** ✅ — corrected the JSDoc to `+0530` and noted moment's `z` token emits the offset without a colon.
6. **Shared `capitalizeFirst` with `formatDatasourceType`** — deferred (your "fine to defer"); the two now differ slightly (this one down-cases the remainder for machine tokens), so a shared helper would need a flag. Left as a follow-up to avoid widening this PR into `shared_constants.ts`.

ps48 (commented):
### clean shared-primitives foundation

Reviewed at `e2e921f`. This is a pure-additive module (i18n enum labels, tz-aware `formatTimestamp`, `euiThemeVars`-backed color tokens) with no behavior change on its own, and it holds up.

Verified all cross-references resolve against the current tree: the alert enum types live in `common/types/alerting/unified_types.ts`, `uiSettingsService` in `common/utils`, and `moment-timezone` / `@osd/i18n` / `euiThemeVars` are all already-present deps, so there are no new package.json or manifest additions. Placement at the `alerting/` root matches existing siblings like `shared_constants.ts` (the dir has no `common/` subfolder). No feature-flag leaks (these are leaf helpers; the gated call sites are the stacked PRs), no new backend routes, no raw OpenSearch/Prometheus access, no saved objects, no stray console/logger.

Logic and coverage check out: `humanizeUnknown` sentence-cases machine tokens while passing AD/forecaster prose through unchanged, `formatTimestamp` coerces stringified epochs and returns the placeholder instead of "Invalid date", and the three test files cover every union member plus the unknown/fallback paths. i18n ids are unique across the `stateLabel`/`monitorStateLabel` namespaces.

No p0/p1 findings. Nice foundation.

TackAdam (approved):

PR #2827, opened 2026-08-26
[Alerting] Alerts list: labeled severity, Title-cased state, responsive columns, a11y

lezzago:
## What & why
Severity was a bare colour dot under an abbreviated 'Sev' header with no text label; State/Message showed raw enums; empty cells drifted between `—`/`---`. Now: labelled `EuiHealth`/badges under a full 'Severity' header, Title-cased state, a single em-dash empty glyph, a Message tooltip, ≥24px targets, `aria-expanded`, and responsive column widths.

**Findings:** B1/A11Y1, CLAR1, CLAR2, CLAR4, CLAR9, A11Y3, A11Y4, A11Y6, RSP1, M6.

## Before / after

**Before / after**

Before / after

**Flow**

Flow

## Testing
`alerts_dashboard.test.tsx` (+ new assertions). Centrally green.

## Review
Independent review agent: **APPROVE** (mutation-checked tests).

## Regression check

## Dependency
Stacked on #2826 (shared primitives). This branch **includes** #2826's files so it builds and tests standalone in CI; merge #2826 first, then a rebase drops those files from this diff.

### DIFF
--- a/common/services/alerting/__tests__/filter.test.ts
+++ b/common/services/alerting/__tests__/filter.test.ts
@@ -3,7 +3,14 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import { emptyFilters, filterAlerts, matchesFilters, matchesSearch, sortRules } from '../filter';
+import {
+  alertMatchesSearch,
+  emptyFilters,
+  filterAlerts,
+  matchesFilters,
+  matchesSearch,
+  sortRules,
+} from '../filter';
 
 describe('filter', () => {
   describe('emptyFilters', () => {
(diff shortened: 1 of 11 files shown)


lezzago:
Added in `81569a60`: the Alerts tab now honors a `slo_id:<id>` label deep-link. `alarms_page` already forwarded `deepLink.q` to `AlertsDashboard`, but alert search was a plain substring match so `slo_id:<id>` never matched. A new shared `alertMatchesSearch` treats a single `key:value` term (no whitespace) as a label match and keeps exact substring behavior for everything else (free-text search unchanged); used in both `filterAlerts` and the facet-count search. This is what makes the SLO detail 'View alerts' pivot (#2831) scope to the SLO — resolving a scoping point @TackAdam raised on #2831, kept here since this PR owns `alerts_dashboard.tsx` / `filter.ts`.

PR #2828, opened 2026-08-26
[Alerting] Alerts timeline: stop clamping pre-window alerts into bucket 0; name the y-axis

lezzago:
## What & why
Alerts that started before the selected window were clamped into the first bucket, producing a false left-edge spike; the y-axis was unlabelled. Now pre-window alerts are excluded and surfaced as an honest 'N started before this window' note, and the y-axis is named 'Alerts'.

**Findings:** DVZ-AT1, DVZ-AT2.

## Before / after

**Before / after**

Before / after

**Flow**

Flow

## Testing
`alerts_charts.test.tsx` — asserts bucket-0 is not inflated by a pre-window alert. Centrally green.

## Review
Independent review agent: **PASS** — verified bucketing math and boundary behaviour; refuted 3 candidate findings.

## Regression check

## Dependency
None.

### DIFF
--- a/public/components/alerting/alerts_charts.tsx
+++ b/public/components/alerting/alerts_charts.tsx
@@ -117,48 +117,92 @@ export const AlertTimeline: React.FC<AlertTimelineProps> = ({ alerts, startMs, e
     const bucketCount = clamp(rawBucketCount, MIN_BUCKETS, MAX_BUCKETS);
     const bucketDuration = rangeMs / bucketCount;
 
-    const buckets: Array<{
-      label: string;
-      critical: number;
-      high: number;
-      medium: number;
-      low: number;
-      info: number;
-    }> = [];
+    const tz = resolveDisplayTz();
 
-    // Parse each alert's start time once (perf: avoid re-parsing per bucket).
+    // Parse each alert's start time once, then bucket in a single O(alerts)
+    // pass. The previous approach re-scanned the whole alerts array per bucket
+    // and ran five more severity `.filter` passes inside each — O(buckets ×
+    // alerts × 6). A single pass that computes the bucket index per alert is
+    // O(alerts), which matters once the list runs into the thousands.
     const alertStartMs = alerts.map((a) => new Date(a.startTime).getTime());
+
+    const buckets = Array.from({ length: bucketCount }, (_, i) => ({
+      label: formatBucketLabel(startMs + i * bucketDuration, rangeMs, tz),
+      critical: 0,
+      high: 0,
+      medium: 0,
+      low: 0,
+      info: 0,
+    }));
 
     // This histogram counts alert *starts* per time bucket. The OS backend's
     // interval-overlap filter (opensearch_backend.ts) also returns alerts that
     // began BEFORE the picked window but are still firing / recently resolved
     // inside it. Those alerts are real, but they did not *start* in any visible
     // bucket — the previous `Math.max(startMs, ...)` clamp forced them into
-    // bucket 0, painting a false spike at the left edge of the timeline. We
-    // instead leave the true start time unclamped (so the bucket filter below
-    // naturally excludes anything before `startMs`) and surface the excluded
-    // count in the chart title, so the data is honest rather than silently
-    // dropped and users can still reconcile it against the summary cards.
-    const excludedBeforeWindow = alertStartMs.filter((t) => t < startMs).length;
-
-    const tz = resolveDisplayTz();
-
-    for (let i = 0; i < bucketCount; i++) {
-      const bucketStart = startMs + i * bucketDuration;
-      const bucketEnd = bucketStart + bucketDuration;
-      const label = formatBucketLabel(bucketStart, rangeMs, tz);
-      const inBucket = alerts.filter(
-        (_, idx) => alertStartMs[idx] >= bucketStart && alertStartMs[idx] < bucketEnd
-      );
-      buckets.push({
-        label,
-        critical: inBucket.filter((a) => a.severity === 'critical').length,
-        high: inBucket.filter((a) => a.severity === 'high').length,
-        medium: inBucket.filter((a) => a.severity === 'medium').length,
-        low: inBucket.filter((a) => a.severity === 'low').length,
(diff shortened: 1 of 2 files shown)


TackAdam:
No high-severity or correctness bugs. The findings below are all low severity and mostly pre-existing edges the diff brushes against.

## Worth a look

1. **Unparseable `startTime` breaks the reconciliation guarantee the new subtitle promises — `alerts_charts.tsx:130,142`**

   A malformed/missing `startTime` → `new Date(...).getTime()` is `NaN`. `NaN < startMs` is false (not counted in `excludedBeforeWindow`) and `NaN >= bucketStart` is false (lands in no bucket). So it's counted in the summary cards but appears in neither the bars nor the "N alerts started before this window" note — which is exactly the "users can reconcile against the summary cards" invariant the new comment advertises.

   This particular drop is pre-existing (old `Math.max(startMs, NaN)` was also `NaN`), but the PR newly makes a promise the code doesn't fully keep.

   A cheap fix: count `Number.isNaN` starts into a separate "N with unknown start time" bucket, or just fold them into the excluded note.

2. **Half-open final bucket can silently drop an alert right at `endMs` — `alerts_charts.tsx:142,151`**

   `bucketDuration = rangeMs / bucketCount` is a float, so the last `bucketEnd` may land a hair below `endMs`. An alert starting in that sliver satisfies neither `< bucketEnd` (dropped from bars) nor `t < startMs` (not in the excluded count) — it vanishes with no acknowledgement.

   Narrow: the backend's overlap filter shouldn't return alerts starting `>= endMs`, so this only bites the float-rounding sliver. Making the final bucket inclusive of `endMs` (`idx === bucketCount - 1 ? <= bucketEnd : < bucketEnd`) would close it.

## Nits (fine to defer)

3. **Bucketing is `O(buckets × alerts × 6)` — `alerts_charts.tsx:150`**

   The outer loop re-scans the full alerts array per bucket, then runs five more `.filter` passes per bucket for severity. A single pass — `idx = floor((alertStartMs[i] - startMs) / bucketDuration)`, increment the matching severity — is `O(alerts)`.

   Pre-existing, re-touched by the diff, and the removed comment had even claimed a "perf win," so worth a mention. Only matters at thousands of alerts.

4. **Y-axis title spacing — `alerts_charts.tsx:213`**

   `nameGap: 30` against `grid.left: 44` leaves little room; 3-digit tick labels (counts `≥ 100`) at `fontSize 9` plus the rotated name could clip the left edge.

   Cosmetic — bump `grid.left` a touch if you see overlap.

## Style note

The hardcoded hex colors (`#98A2B3`, `#EDF0F5`) are exactly what the sibling PR #2826 introduces theme tokens for — not this PR's job, but a candidate follow-up once those primitives land.

lezzago:
Thanks @TackAdam — addressed in `88810c8a`:

1. **Unparseable `startTime` breaks the reconciliation promise** ✅ — NaN starts are now counted separately and surfaced as an "{n} alert(s) have an unknown start time" note next to the "started before this window" note (joined with a middot). Every alert is now reconcilable against the summary cards: in a bar, before the window, or unknown. Test added.
2. **Half-open final bucket drops an alert at `endMs`** ✅ — a start at (or at the float edge of) `endMs` is now folded into the last bucket. A start *strictly* after `endMs` stays excluded from the bars (backend overlap filter shouldn't emit those, and folding them would paint a false right-edge spike — the existing post-window test still holds). Test added for the `endMs` edge.
3. **Bucketing was `O(buckets × alerts × 6)`** ✅ — replaced with a single `O(alerts)` pass that computes each alert's bucket index directly (`floor((t - startMs) / bucketDuration)`) and increments the matching severity.
4. **Y-axis title spacing** ✅ — bumped `grid.left` 44 → 50 so 3-digit count ticks + the rotated name have room.

Style note (hardcoded hex → the tokens #2826 introduces): agreed, that's a clean follow-up once #2826 lands — left out of this PR to keep it file-disjoint.

public/components/alerting/alerts_charts.tsx
```diff
@@ -116,41 +117,92 @@ export const AlertTimeline: React.FC<AlertTimelineProps> = ({ alerts, startMs, e
...
+      }
+      // Bucket index. `bucketDuration` is a float, so a start at exactly `endMs`
+      // computes an index of `bucketCount` (out of range). Fold a start at (or
+      // right at the float edge of) `endMs` into the last bucket so it isn't
+      // dropped, but leave a start strictly after `endMs` excluded from the
+      // bars — the backend's overlap filter shouldn't return those, and folding
+      // them in would paint a false spike on the right edge.
+      let bucketIdx = Math.floor((t - startMs) / bucketDuration);
+      if (bucketIdx >= bucketCount) {
+        if (t > endMs) continue;
```
ps48, line 169: ### p2: post-window alerts vanish, contradicting the reconciliation comment

The comment just above promises every alert ends up either in a bar, before the window, or unknown, rather than silently vanishing from both the bars and the notes. But a start strictly after `endMs` hits `if (t > endMs) continue;` and is counted in no bar and no note. The `post-window alerts are excluded` test pins this as intended, so it is a comment/consistency nit, not a bug: could we either add an `afterWindow` counter + note, or soften the invariant comment to acknowledge the deliberate after-window drop?

ps48 (commented):
### reviewed: alerts timeline pre-window clamp fix

Read at 88810c8. The core fix is right: replacing the `Math.max(startMs, ...)` clamp with a single O(alerts) pass that buckets by actual start time, excludes pre-window and unknown-start alerts from the bars, and surfaces their counts as an honest subtitle is the correct call, and it keeps the bars reconcilable against the summary cards. The `endMs` float-edge fold into the last bucket is handled and tested, timezone only touches labels (bucketing is on raw epoch ms), and the severity guard matches the full closed union. echarts throughout, no vega/legacy charts, no new deps/routes/saved objects, i18n is namespaced and extractable, and both the clamp fix and the y-axis get real regression tests plus unknown-start and edge coverage.

One optional nit inline about the counting-pass comment vs the deliberate after-window drop. No p0/p1.

TackAdam (approved):

joshuali925 (approved):

PR #2830, opened 2026-08-26
[SLO] Listing: humanized state, stable numeric formatting, server-side budget sort, skeleton rows

lezzago:
## What & why
The Health cell rendered the raw store enum ('ok'); per-row precision was ad-hoc; sort re-ordered only the current page. Now: humanized state label, 2-decimal target via the shared `SLO_PRECISION` policy, tabular-aligned numerals, server-side sort via existing `sortField`/`sortOrder`, and skeleton rows on first load.

**Findings:** M3/CLAR7, M4, M5, M7, m5.

## Before / after

**Before / after**

Before / after

**Flow**

Flow

## Testing
listing + query-service tests. Centrally green.

## Review
Independent review agent: **APPROVE**; applied one in-scope comment fix (a stale sort comment that M7 had made inaccurate).

## Regression check

## Dependency
No dependencies — can merge independently.

### DIFF
--- a/common/slo/state.ts
+++ b/common/slo/state.ts
@@ -3,6 +3,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+import { i18n } from '@osd/i18n';
 import type { SloHealthState } from './slo_types';
 
 export const SLO_HEALTH_COLOR: Record<SloHealthState, string> = {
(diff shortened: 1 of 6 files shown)


TackAdam:
## Worth addressing

1. **Third copy of the state-label map — `slo_listing_page.tsx:249`**

   This PR adds an i18n `STATE_LABEL` map inline. There's already:

   - `STATE_LABEL` in `slo_list_filter_panel.tsx:54` (i18n, different message IDs), and
   - `SLO_HEALTH_LABEL` + `getSloHealthLabel` in `common/slo/state.ts:46` (not localized — plain English).

   So the same `SloHealthState → label` mapping now lives in three places with three sets of strings. The author's own comment acknowledges the overlap.

   The clean fix is to localize the shared `getSloHealthLabel` in `common/slo/state.ts` and route all three call sites through it — otherwise adding a future state (or fixing wording) means editing three maps and the plain-English one silently drifts.

   Not a blocker, but this PR is the natural place to consolidate rather than deepen the fork. Medium-low.

## Nit

2. **`STATE_LABEL[state] ?? state` leaks the raw enum on an unknown state — `slo_listing_page.tsx:184,185`**

   Since `state` is typed `SloHealthState`, every key is covered today, so the fallback is dead in practice — but if a backend ever sends a state outside the union, the cell shows the raw machine token, which is exactly what this change set out to prevent.

   A `getSloHealthLabel`-style `"Unknown"` fallback (see #1) would close it.

## Latent / out of scope

Pre-existing in `slo_query_service.ts`, not introduced here.

The PR only edits a comment in this file, but since it's touched, flagging what the automated pass surfaced — none are caused by this PR:

- **`MAX_CURSOR_PAGE` ceiling (~line 235):** At the clamp, `hasMore`/`nextCursor` still get emitted, so past 500 pages the "next" cursor decodes back to the same clamped page — a potential infinite re-fetch loop that contradicts the "stop with a clear message" comment.

- **Cursor `ps` (stored page size) is never read:** Changing `pageSize` mid-cursor recomputes page boundaries against the new size, which could skip rows.

Both predate this change and deserve their own issue rather than scope-creep here.

lezzago:
Thanks @TackAdam — addressed in `56a3f183`:

1. **Third copy of the state-label map** ✅ — did the **full consolidation**: localized the shared `getSloHealthLabel` in `common/slo/state.ts` (en output byte-identical to the old plain-English labels, so existing callers like the detail header are unaffected) and routed **all three** call sites through it — this page's inline `STATE_LABEL`, `slo_list_filter_panel.tsx`, and the detail header. Both inline maps are deleted, so adding a future state / fixing wording is now a one-file edit. (Both extra files are touched by no other open PR, so this stays file-disjoint and #2830 remains independently mergeable.)
2. **`STATE_LABEL[state] ?? state` leaks the raw enum** ✅ — closed by the same change: `getSloHealthLabel` maps an out-of-union state to `"Unknown"` rather than the raw token. Tests cover the per-state labels and the Unknown fallback.

**Latent / out of scope** (`MAX_CURSOR_PAGE` infinite-refetch ceiling, unread cursor `ps`): agreed these predate this change and deserve their own issue — left out to avoid scope-creep, as you suggested.

ps48 (commented):
### slo listing cleanup: humanized state, fixed precision, server-order, skeleton

Scoped, well-tested cleanup verified at the head SHA. The Health cell and state filter badge now render `getSloHealthLabel(state)` (out-of-union states map to "Unknown", not a leaked token), the duplicated `STATE_LABEL` maps are removed in favor of the single `common/slo/state.ts` source, `formatTargetPct` routes through `formatPct(..., SLO_PRECISION.target)` for consistent 2dp with tabular nums, and the page-local `worstBudgetRemaining` re-sort is removed so the real server order is rendered (the right call, since error-budget-remaining is a derived per-read value the store can't index). `EuiLoadingContent` skeleton replaces the centered spinner. No new deps, routes, or saved objects.

One thing to reconcile: the title says "server-side budget sort", but the change does not sort by budget server-side, it removes the client re-sort and renders the existing `_id`-ordered page. Consider tightening the title. One low-pri i18n note inline. No p0/p1.

common/slo/state.ts
```diff
@@ -38,25 +39,35 @@ export function getSloHealthColor(state: SloHealthState | string | undefined | n
...
 const SLO_HEALTH_LABEL: Record<SloHealthState, string> = {
-  breached: 'Breached',
-  warning: 'Warning',
-  ok: 'Healthy',
-  no_data: 'No data',
-  source_idle: 'Source idle',
-  stale: 'Stale',
-  disabled: 'Disabled',
-  rules_missing: 'Rules missing',
+  breached: i18n.translate('observability.slo.state.label.breached', {
```
ps48, line 50: ### p2: module-level i18n.translate may freeze labels to the english defaults

`SLO_HEALTH_LABEL` is a module-level const, so each `i18n.translate` here resolves once at import time. If this module (now in `common/`, reachable server-side via `getSloHealthColor`/`SLO_HEALTH_ORDER`) is imported before the locale bundle registers, the map captures the English `defaultMessage` for the process lifetime, which would undercut the stated goal that non-en locales translate. It matches the `STATE_LABEL` map it replaces, so not a regression, but was resolving inside `getSloHealthLabel` (so the lookup honors the active locale at call time) considered?

TackAdam (approved):

joshuali925 (approved):

PR #2831, opened 2026-08-26
[SLO] Detail: header layout, not-found state, re-poll rule health, 'View alerts' pivot

lezzago:
## What & why
Fixes the real cause of the clipped title at 1024px (`responsive={false}` + min-width), adds a proper 'SLO not found' EmptyPrompt before the null return, spells out 'pp', re-polls rule health (suppressing the callout during the post-create propagation window), and pivots 'View alerts' to firing alerts.

**Findings:** CLAR14/BUG-S1, CLAR16, CLAR12, F-CRUD2, OBS1.

## Before / after

**Before / after (header pivot)**

Before / after (header pivot)

**Flow**

Flow

## Testing
`slo_detail_page.test.tsx`. Centrally green.

## Review
Independent review agent: **APPROVE** (review commit applied).

## Regression check

## Dependency
No dependencies — can merge independently. Note: #2838 (SLO edit path) is stacked on this and should merge after it.

### DIFF
--- a/public/components/apm/pages/slos/slo_detail_page.tsx
+++ b/public/components/apm/pages/slos/slo_detail_page.tsx
@@ -680,9 +680,24 @@ export const SloDetailPage: React.FC<SloDetailPageProps> = ({
   // leak or poll forever.
   useEffect(() => {
     if (!doc) return undefined;
-    const calloutState = deriveSloCalloutState(ruleHealth?.state, doc.liveStatus.state);
-    const missing = calloutState === 'rules_missing' || calloutState === 'rules_partial';
-    if (!missing) {
+    // Keep re-probing while EITHER the live probe OR the persisted live-status
+    // flag reads missing — deliberately NOT `deriveSloCalloutState`, which lets
+    // a healthy probe win for the *callout* (so a fresh-create SLO never shows a
+    // false destructive alarm). For *polling* we must be more eager: the
+    // rule-health probe is cached server-side (~90s TTL), so right after a real
+    // regression the first probe can still read a stale `ok` while the persisted
+    // state already flipped to `rules_missing`. If we stopped polling on that
+    // stale `ok`, we'd never re-check and the genuine "missing" callout would
+    // never surface. Polling while `liveState` is missing lets a later probe
+    // catch the fresh `rules_missing` and escalate; a genuinely-healthy probe
+    // just runs the bounded budget out harmlessly without ever escalating.
+    const probeState = ruleHealth?.state;
+    const probeMissing =
+      probeState === 'rules_missing' ||
+      probeState === 'rules_partial' ||
+      probeState === 'ruler_unreachable';
+    const keepProbing = probeMissing || doc.liveStatus.state === 'rules_missing';
+    if (!keepProbing) {
       if (ruleHealthRetries !== 0) setRuleHealthRetries(0);
       return undefined;
     }
--- a/public/components/apm/pages/slos/slo_detail_page.tsx
+++ b/public/components/apm/pages/slos/slo_detail_page.tsx
@@ -829,7 +829,7 @@ export const SloDetailPage: React.FC<SloDetailPageProps> = ({
                 <p>
                   {i18n.translate('observability.apm.slo.detail.notFound.body', {
                     defaultMessage:
-                      "We couldn't find an SLO with the id \"{id}\". It may have been deleted, or the link may be out of date.",
+                      'We couldn\'t find an SLO with the id "{id}". It may have been deleted, or the link may be out of date.',
                     values: { id },
                   })}
                 </p>
@@ -1104,9 +1104,12 @@ export const SloDetailPage: React.FC<SloDetailPageProps> = ({
                     isLoading={ruleHealthLoading}
                     data-test-subj="slosDetailRulePropagatingCheckNow"
                   >
-                    {i18n.translate('observability.apm.slo.detail.rulesPropagatingCallout.checkNow', {
-                      defaultMessage: 'Check now',
-                    })}
+                    {i18n.translate(
+                      'observability.apm.slo.detail.rulesPropagatingCallout.checkNow',
+                      {
+                        defaultMessage: 'Check now',
+                      }
+                    )}
                   </EuiButtonEmpty>
                 </EuiCallOut>
                 <EuiSpacer size="m" />
--- a/public/components/apm/pages/slos/slo_detail_page.tsx
+++ b/public/components/apm/pages/slos/slo_detail_page.tsx
@@ -610,6 +610,25 @@ export const SloDetailPage: React.FC<SloDetailPageProps> = ({
     setRuleHealthRetries(0);
   }, [id]);
 
+  // Stale-write guard for the auto-poll below. Its re-probe resolves
+  // asynchronously; if we've navigated to a different SLO or unmounted before
+  // it lands, the result belongs to a stale id and must be dropped rather than
+  // clobbering the current SLO's rule-health state. A retry-count increment is
+  // NOT stale (`id` is unchanged then), so the guard keys on id + mount only —
+  // never on the retry counter. The mount effect re-arms `mounted` on remount
+  // so a StrictMode double-mount doesn't wedge it permanently false.
+  const pollLiveIdRef = useRef(id);
+  const pollMountedRef = useRef(true);
+  useEffect(() => {
+    pollLiveIdRef.current = id;
+  }, [id]);
+  useEffect(() => {
+    pollMountedRef.current = true;
+    return () => {
+      pollMountedRef.current = false;
+    };
+  }, []);
+
   // While rule health reads as missing/partial, re-probe on a bounded
   // interval (F-CRUD2). A freshly-created SLO's rule groups take time to
   // propagate through the Cortex/AMP ruler, so a "missing" reading is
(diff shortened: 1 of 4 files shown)


TackAdam:
## Must fix

1. **The grace window delays the false "rules missing" alarm rather than preventing it — the exact F-CRUD2 case — `slo_detail_page.tsx:128-143`, poll at `:640`**

   `deriveSloCalloutState` does not let a concrete healthy probe win, despite its comment ("the probe wins whenever it reports a concrete state"):

   `if (ruleHealthState === 'rules_missing' || 'rules_partial' || 'ruler_unreachable') return ruleHealthState;`

   `return liveState === 'rules_missing' ? 'rules_missing' : null;   // 'ok' probe falls through to here`

   So `deriveSloCalloutState('ok', 'rules_missing')` returns `'rules_missing'` — a fresh `ok` probe is ignored when the persisted `liveStatus.state` is `rules_missing`.

   And the poll effect only re-calls `loadRuleHealth` — it never reloads `doc`, so `liveStatus` stays stale `rules_missing` forever.

   Net: a freshly-created SLO whose rules have propagated (probe → `ok`, but `liveStatus` not yet recomputed server-side) keeps `missing === true`, exhausts the 5 retries, and after ~25s shows the destructive Restore/Delete alarm on a healthy SLO. That's precisely the scenario F-CRUD2 set out to eliminate.

   The reason the "recovers during grace window" test passes is that it uses `liveStatusState: 'ok'` — there's no test for `liveStatus: 'rules_missing'` + probe recovering to `ok`, which is the real fresh-create shape.

   **Fix:** Let a concrete `ok`/healthy probe override the stale `liveState` (make the probe genuinely win), and add that test case.

2. **The "SLO not found" empty state (CLAR16) is unreachable in production — `slo_detail_page.tsx:552`, `:808`**

   `apiClient.get(id)` returns `Promise<SloDocument & …>` (non-nullable) and `http.get` rejects on a 404. `load`'s catch (`:552-555`) sets `error` for any failure with no 404 special-casing — so a deleted/stale-link SLO renders the generic "Unable to load SLO" error branch with a raw message, never the friendly `slosDetailNotFound` prompt.

   The new branch only fires when `get` resolves with a falsy body, which the backend/type contract says won't happen — only the synthetic `mockResolvedValue(null)` test exercises it.

   To actually ship CLAR16, detect 404 in the catch (there's an `isNotFoundError` helper in `slo_health_summary.ts`) and route it to the not-found state.

## Worth addressing

3. **"View alerts" is no longer scoped to this SLO — `:1010-1030`**

   The pivot deep-links to `#/alerts?q=slo_id:<id>&ds=<dsId>`, but — per the code's own comment — the Alerts tab ignores `q` (only the Rules tab/`MonitorsTable` honors `deepLink.q`).

   So the operator lands on all firing alerts for the datasource, not just this SLO's, whereas the old `#/rules` link filtered precisely.

   This is framed as intentional (OBS1), but it's a real scoping regression until the Alerts tab honors `q`. Confirm the product intent, or gate the pivot on the Alerts tab actually supporting the filter.

4. **Repeated probe failures spam danger toasts — `:584`**

   `loadRuleHealth`'s catch fires `addDanger` on every failure, then `setRuleHealth(null)` leaves `derivedCalloutState` at the `rules_missing` fallback, so `missing` stays true and the poll re-fires every 5s — up to 6 "Could not load rule health" toasts in ~25s for a flapping endpoint.

   De-dupe (toast once) or suppress during the bounded poll.

## Nit

5. **`pp` renders a double tooltip — `:315-333`**

   An `EuiToolTip` wraps an `<abbr>` that also carries a native `title` with the same text, so hovering shows both the EUI popover and the browser's native tooltip, and the `percentagePointsTooltip` message is translated twice.

   Pick one mechanism.

lezzago:
Thanks @TackAdam — thorough catches. All addressed:

**Must fix**
1. **Grace window delayed but didn't prevent the false alarm** ✅ (`de4801f8`) — you're right: `deriveSloCalloutState('ok', 'rules_missing')` returned `'rules_missing'` (the healthy probe fell through), and the poll never reloads `doc`, so a fresh-create SLO whose probe recovered to `ok` still escalated after ~25s. Fixed so a **concrete probe result always wins** over the stale `liveState` — a concrete healthy probe now returns `null`. Added the real fresh-create test (`liveStatus` stays `rules_missing`, probe recovers to `ok` → never escalates), which the old `liveStatusState: 'ok'` test didn't cover.
2. **"SLO not found" unreachable** ✅ (`de4801f8`) — `get()` rejects on 404, so the catch hit the generic error branch. Added a local `isNotFoundError` and route a 404 to the not-found state; the test now rejects with a `{ response: { status: 404 } }` shape (the production path), not the synthetic resolved-null.

**Worth addressing**
3. **"View alerts" not scoped to this SLO** ✅ — rather than revert the OBS1 pivot, I taught the **Alerts tab to honor the `slo_id:<id>` deep-link** in #2827 (which owns `alerts_dashboard.tsx` + `filter.ts`; `alarms_page` already forwards `deepLink.q` to `AlertsDashboard`). A new `alertMatchesSearch` treats a single `key:value` term as a label match, so the pivot now filters to just this SLO's firing alerts. Soft ordering: this scopes fully once **#2827** merges; if #2831 merges first the pivot degrades gracefully to the datasource-wide view. No file overlap between the two PRs.
4. **Probe-failure toast spam** ✅ (`de4801f8`) — the bounded re-poll now passes `silent: true`, so a flapping endpoint can't emit ~6 identical danger toasts in the grace window; the initial/manual probe still toasts once.

**Nit**
5. **`pp` double tooltip** ✅ (`de4801f8`) — dropped the duplicate native `title` on the `<abbr>`; the wrapping `EuiToolTip` already provides the hover text (and the message is no longer translated twice).

public/components/apm/pages/slos/slo_detail_page.tsx
```diff
@@ -830,38 +1058,37 @@ export const SloDetailPage: React.FC<SloDetailPageProps> = ({
...
+        // (today only MonitorsTable seeds its search from `deepLink.q`); it
+        // is harmless until then. The recording-rule list on this page stays
+        // informational — recording rules never surface in Alert Manager.
         const params = new URLSearchParams({
           q: `slo_id:${doc.id}`,
           ds: doc.spec.datasourceId,
         });
         coreRefs?.application?.navigateToApp(observabilityAlertingID, {
-          path: `#/rules?${params.toString()}`,
+          path: `#/alerts?${params.toString()}`,
```
ps48, line 1085: ### p2: does 'view alerts' still scope to this slo?

The pivot changes the deep-link from `#/rules?q=slo_id:<id>&ds=<ds>` to `#/alerts?q=slo_id:<id>&ds=<ds>`. Tracing `alarms_page.tsx`: `ds` is applied globally, but `deepLink.q` is only consumed on the Rules tab (`initialSearchQuery` on MonitorsTable); the Alerts tab receives no search-query prop. So 'View alerts' now lands on the Alerts tab scoped to the SLO's datasource but with no `slo_id` filter, showing every firing alert on that datasource rather than just this SLO's. The code comment acknowledges it. Is an unscoped alerts view the intended contract for this button, or should the Alerts tab honor `q` before the label is repointed? A tooltip note or follow-up issue would avoid surprising operators.

ps48 (commented):
### slo detail: header wrap, not-found, bounded rule-health re-poll, alerts pivot

Verified against head `caa58c8`; read the full `slo_detail_page.tsx` plus `alarms_page.tsx` to trace the pivot. The rule-health re-poll is bounded (`RULE_HEALTH_MAX_RETRIES = 5`) and cleaned up on unmount/navigation via `clearTimeout` plus stale-write refs, so it can't leak or run forever, and escalation to the destructive callout only fires after the budget is spent. Errors are not swallowed on the path that matters (initial and manual probes toast; only the auto-poll is silent to avoid duplicate toasts in the grace window). Not-found routes 404 to the prompt behind the loading/error guards. Header fix is pure OUI, data goes through `SloApiClient` (no raw client, no new route/saved object), and the new surface has unit + real-Cortex cypress coverage.

One business-logic question inline about the "View alerts" pivot scoping. No p0/p1.

TackAdam (approved):

public/components/apm/pages/slos/slo_detail_page.tsx
lezzago: Good catch — and you traced it exactly right: on this branch the Alerts tab ignores `deepLink.q` (only `MonitorsTable` seeds `initialSearchQuery`), so `#/alerts?q=slo_id:<id>` lands datasource-scoped, not SLO-scoped. @TackAdam raised the same point.

Intended contract: the pivot **should** scope to the SLO, and the Alerts tab is taught to honor the `slo_id:<id>` deep-link in **#2827** — a new `alertMatchesSearch` helper treats a single `key:value` term as a label match (used in both `filterAlerts` and the facet-count search), and `alarms_page` already forwards `deepLink.q` to `AlertsDashboard`. So once #2827 merges, this button filters to just this SLO’s firing alerts. I kept that change in #2827 rather than here because it owns `alerts_dashboard.tsx` / `filter.ts` and the batch is intentionally file-disjoint.

Merge-order/behavior:
- **#2827 before or with #2831** → fully SLO-scoped (the intended contract).
- **#2831 first** → degrades gracefully to the datasource-scoped Alerts view (still useful — the operator sees what is firing on this datasource; the SLO’s own alerts panel on the detail page already shows this SLO’s firing alerts) until #2827 lands. No error, just a wider result set.

I’ll also tighten the inline comment here to reference #2827 by number instead of the vague “for the day the Alerts tab honours…”, so the dependency is explicit in-code. Happy to gate the label on #2827 if you’d prefer the button not repoint until the filter support is in — let me know.

PR #2832, opened 2026-08-26
[SLO] Charts: derive window from type, no false-green on missing burn data, dataviz polish

lezzago:
## What & why
Derives the window from `window.type` (kills a hard-coded '30d'), renders a null/empty window as `no_data` instead of false-green, dashes the reference line, adds non-hue status encoding, UTC + delta-vs-target tooltips, and `smooth:false` over step data.

**Findings:** M1, M2, CLAR5, CLAR6, m1, m2, m3, m4.

## Before / after

**Before / after**

Before / after

## Testing
chart/panel tests (`classifyTier` null-window, etc.). Centrally green. **Note:** the local stack has no burn data, so these data-dependent deltas are near-identical on screen — correctness is covered by jest and flagged for managed-env re-verify.

## Review
Independent review agent: **PASS** — verified all eight fixes; refuted two out-of-scope observations.

## Regression check

## Dependency
None.

### DIFF
--- a/public/components/apm/pages/services_home/__tests__/slo_budget_sparkline.test.tsx
+++ b/public/components/apm/pages/services_home/__tests__/slo_budget_sparkline.test.tsx
@@ -11,7 +11,11 @@
 
 import React from 'react';
 import { render, screen } from '@testing-library/react';
-import { buildAggregateErrorRatioQuery, SloBudgetSparkline } from '../slo_budget_sparkline';
+import {
+  buildAggregateErrorRatioQuery,
+  buildOption,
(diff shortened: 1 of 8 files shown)


lezzago:
Cross-referencing @TackAdam's precision finding on #2836 (item 1): the budget-remaining chart tooltip/axis/threshold label and the budget sparkline live in this PR's files, so the fix landed here in `99d87f07` — they now render via `formatPct(v, { decimals: SLO_PRECISION.budget })` (2 decimals), matching `slo_budget_panel.tsx`. Tests updated for the 2-decimal output.

PR #2833, opened 2026-08-26
[SLO] Overview: consistent firing/breached, clamp budget hero at 0, KPI a11y

lezzago:
## What & why
The overview tile read 'Firing' and could disagree with 'Breached'; the budget hero could go negative. Now relabelled 'Alerts firing' (gated on the same status feed, counts instances), the budget hero clamps at 0, and the KPI tiles expose `aria-pressed` and handle Space.

**Findings:** M12, CLAR8, A11Y2, A11Y5.

## Before / after

**Before / after (overview panel)**

Before / after (overview panel)

**Flow**

Flow

## Testing
`slo_overview_panel.test.tsx`. Centrally green.

## Review
Independent review agent: **PASS** — verified the firing/breached disambiguation and budget clamp; refuted alternatives.

## Regression check

## Dependency
None.

### DIFF
--- a/public/components/apm/pages/slos/slo_overview_panel.tsx
+++ b/public/components/apm/pages/slos/slo_overview_panel.tsx
@@ -19,7 +19,7 @@
  *            the listing filtered to remaining < 25%.
  */
 
-import React, { useMemo } from 'react';
+import React, { useId, useMemo } from 'react';
 import {
   EuiButtonEmpty,
   EuiFlexGroup,
@@ -27,6 +27,7 @@ import {
   EuiIcon,
   EuiLink,
   EuiPanel,
+  EuiScreenReaderOnly,
   EuiText,
   EuiToolTip,
 } from '@elastic/eui';
@@ -165,15 +166,29 @@ const KpiCell: React.FC<{
   dataTestSubj?: string;
 }> = ({ value, label, accent, tooltip, onClick, active, dataTestSubj }) => {
   const clickable = Boolean(onClick);
+  // A non-clickable readout tile (e.g. "Alerts firing") that still carries an
+  // explanatory tooltip must stay reachable by keyboard/SR — otherwise the
+  // tooltip is mouse-hover-only, unlike every clickable tile (ps48 review).
+  // Make it focusable (so EuiToolTip also shows on focus) and expose the
+  // tooltip text via aria-describedby to a screen-reader-only note, without a
+  // misleading button/toggle role.
+  const describable = !clickable && Boolean(tooltip);
+  const descId = useId();
+  const accessibleName =
+    typeof value === 'string' || typeof value === 'number' ? `${value} ${label}` : label;
   const content = (
     <div
       role={clickable ? 'button' : undefined}
-      tabIndex={clickable ? 0 : undefined}
+      tabIndex={clickable || describable ? 0 : undefined}
       // Expose toggle state to assistive tech so SR users can tell which KPI
       // filter is active — mirrors how HealthRail's segment buttons do it
       // (A11Y2). The tile's accessible name comes from its text content (the
       // value + label spans), e.g. "3 Breached".
       aria-pressed={clickable ? Boolean(active) : undefined}
+      // A focusable readout has no widget role to name it from content, so give
+      // it an explicit name + description for keyboard/SR users.
+      aria-label={describable ? accessibleName : undefined}
+      aria-describedby={describable ? descId : undefined}
       onClick={onClick}
       onKeyDown={(e) => {
         if (!clickable) return;
@@ -198,6 +213,11 @@ const KpiCell: React.FC<{
         minWidth: 72,
       }}
     >
+      {describable && (
+        <EuiScreenReaderOnly>
+          <span id={descId}>{tooltip}</span>
+        </EuiScreenReaderOnly>
+      )}
       <span
         style={{
           width: 3,
--- a/public/components/apm/pages/slos/slo_overview_panel.tsx
+++ b/public/components/apm/pages/slos/slo_overview_panel.tsx
@@ -388,7 +388,9 @@ export const SloOverviewPanel: React.FC<SloOverviewPanelProps> = ({
       // contributes 0 here. The two therefore measure different units — the
       // tile is labelled "Alerts firing" (not "Firing") so a value of 0 next
       // to a non-zero breached count is unambiguous rather than contradictory
-      // (M12). Guard against absent/NaN status so one bad row can't NaN the sum.
+      // (M12). Guard only a NaN/absent `firingCount` so one bad row can't NaN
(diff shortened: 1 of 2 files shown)


TackAdam:
## Worth fixing in this PR

1. **Clicking "Alerts firing" highlights the Breached tile, not itself — and this PR's new `aria-pressed` now announces that wrong state to screen readers — `slo_overview_panel.tsx:626`, parent at `slo_listing_page.tsx:232,238`**

   The chain:

   - `toggle('firing')` → `stateTileToFilterState('firing')` returns `['breached']` (intentional — there's no server-side `firingCount > 0` facet, so it maps to the closest real one).
   - `filterStateToTile(['breached'])` returns `'breached'`, which flows back as `activeStateFilter`.
   - So `active={activeStateFilter === 'firing'}` on the firing tile is unreachable — clicking "Alerts firing" tints and sets `aria-pressed="true"` on Breached instead.

   The filter behavior is defensible (documented at `slo_listing_page.tsx:224-226`), but the highlight landing on the wrong tile is confusing, and this is squarely in scope:

   - The PR's whole theme is firing/breached consistency (M12).
   - The `aria-pressed` you just added makes the mismatch worse for AT users — a firing click now announces "Breached, pressed."

   **Options:** Give the firing tile no toggle/pressed affordance (it's really a readout, not a filter), or teach `filterStateToTile` to round-trip the firing pseudo-state.

   Pre-existing, but this PR is the right place.

## Latent / consistency (pre-existing, out of diff scope)

2. **"No data / disabled" tile counts both but filters only `no_data` — `slo_overview_panel.tsx:630,639`**

   Value is `stats.noData + stats.disabled` with label `"No data / disabled"`, but `onClick={toggle('no_data')}` filters to `no_data` only — so a tile reading `"5"` can filter down to 3 rows, silently dropping the disabled ones.

   Same class of count/filter mismatch as 1. Not touched by this diff; worth a follow-up.

## Nit

3. **Comment overstates the NaN guard — `slo_overview_panel.tsx:388-391`**

   The comment says it guards against `"absent/NaN status"`, but only `firingCount` is guarded — if `s.status` itself were `undefined`, `s.status.firingCount` throws before `Number.isFinite` runs (and again at `s.status.state`).

   `status` is assumed present everywhere else, so this is low risk, but the comment is misleading. Either narrow the comment to `"NaN firingCount"` or guard `s.status?.firingCount`.

lezzago:
Thanks @TackAdam — addressed in `5db67eeb`:

1. **Firing click highlights/announces the wrong tile** ✅ — made **'Alerts firing' a read-only stat**: dropped its `onClick`/`active` so `KpiCell` renders it non-interactive (no `role=button`, `tabIndex`, or `aria-pressed`). Since there's no server-side firing-only facet, a click could only map to `breached` and tint/announce the Breached tile — misleading for keyboard/SR users. The Breached tile stays the way to filter. Added a test asserting the firing tile exposes no pressed state. (The `filterStateToTile('firing')` round-trip logic lives in `slo_listing_page.tsx`, owned by #2830 — it's now harmless dead-code for this tile and I left it untouched to keep the PRs file-disjoint.)
3. **NaN-guard comment overstates the guard** ✅ — narrowed it to say it guards only a NaN/absent `firingCount`; `s.status` is assumed present (the unguarded `s.status.state` read just below makes that explicit).

2. **'No data / disabled' tile counts both but filters only `no_data`** — agreed it's the same count/filter-mismatch class; it's pre-existing and out of this diff's scope, so I've left it for a dedicated follow-up rather than widening this PR.

public/components/apm/pages/slos/slo_overview_panel.tsx
```diff
@@ -584,16 +615,21 @@ export const SloOverviewPanel: React.FC<SloOverviewPanelProps> = ({
...
               stats.firing > 0 ? euiThemeVars.euiColorDanger : euiThemeVars.euiColorMediumShade
             }
             tooltip={i18n.translate('observability.apm.slo.overviewPanel.kpi.firingTooltip', {
-              defaultMessage: 'Total MWMBR burn-rate alerts currently firing',
+              defaultMessage:
+                'Burn-rate alert instances currently firing across all SLOs — a count of alerts, not SLOs. A breached SLO with no alert rules configured contributes 0, so this can read 0 even while SLOs are breached; use the Breached tile for the SLO count.',
             })}
-            onClick={toggle('firing')}
-            active={activeStateFilter === 'firing'}
+            // Read-only stat, not a filter: there is no server-side
```
ps48, line 647: ### p2: the firing tile tooltip is now keyboard-unreachable

The firing tile drops `onClick`, so `clickable` is false and it loses `tabIndex`/`role`, making it non-focusable. Its `EuiToolTip` is the most explanatory one on the strip (it describes why a 0 can sit beside a non-zero Breached), so it becomes mouse/hover-only for keyboard and screen-reader users while every clickable KPI tile keeps its tooltip reachable. Given the a11y intent here, could we surface that explanation another way, e.g. an `aria-describedby` visually-hidden note, or keep the tile focusable without a toggle role?

ps48 (commented):
### verified: firing/breached disambiguation and 0-clamp both hold

Confirmed against the head-ref source, not a stale main diff. `firingCount` is summed as alert instances with a finite guard, and the tile is relabelled "Alerts firing", so a 0 beside a non-zero Breached is honest rather than a contradiction. The budget hero clamp is `Math.max(0, avg)` at format time only: the accent still runs on the raw average into the danger band and the Breached tile still counts the breach, so the clamp cleans up a nonsensical negative percentage without hiding a real breach. `aria-pressed` is exposed only on the clickable filter tiles, and Space/Enter now `preventDefault` and fire exactly once on the div. Tests cover each change.

One low a11y question inline: the now-non-clickable firing tile is no longer focusable, so its explanatory tooltip is mouse-only. No p0/p1.

TackAdam (approved):

public/components/apm/pages/slos/slo_overview_panel.tsx
lezzago: Fixed in `9438d62f`. You're right that dropping `onClick` made the firing readout non-focusable and its tooltip mouse-only. In `KpiCell`, a non-clickable tile that still carries a tooltip is now treated as "describable": it gets `tabIndex={0}` (so `EuiToolTip` also opens on keyboard focus) plus an explicit `aria-label` and `aria-describedby` pointing at an `EuiScreenReaderOnly` note with the tooltip text — so keyboard and screen-reader users get the same explanation, without reintroducing a misleading `button`/`aria-pressed` role. Went with your `aria-describedby` visually-hidden-note suggestion. Added a test asserting the firing tile is focusable, non-toggle, and describable. (This also benefits the aggregate-budget readout, which had the same mouse-only tooltip.)

joshuali925 (approved):

PR #2834, opened 2026-08-26
[a11y] Remove nested interactive controls; enforce >=24px targets; aria-expanded on tree

lezzago:
## What & why
The facet error button was nested inside a checkbox `<label>` and the section header was a `div role=button` with a stopPropagation hack. Now the error button is a sibling, the header is a real `<button>` (hack removed), targets are ≥24px, and the tree expander exposes `aria-expanded`.

**Findings:** AXE2, AXE3, A11Y4.

## Before / after

**Before / after (facet panel; pixels match by design — see note)**

Before / after (facet panel; pixels match by design — see note)

**Flow**

Flow

## Testing
RTL tests assert no nested interactive/label and `aria-expanded`. **Note:** these are DOM/ARIA changes, not pixel changes.

## Review
Independent review agent: **APPROVE** — confirmed the hacks are removed (not retained) and keyboard behaviour is preserved by the native button.

## Regression check

## Dependency
None.

### DIFF
--- a/public/components/alerting/facet_filter_panel.tsx
+++ b/public/components/alerting/facet_filter_panel.tsx
@@ -260,7 +260,12 @@ export const FacetFilterGroup: React.FC<FacetFilterGroupProps> = ({
             iconSide="left"
             onClick={() => onToggleCollapse(id)}
             aria-expanded={!isCollapsed}
-            aria-controls={`facetGroup-${id}-region`}
+            // Only reference the region while it's actually in the DOM. The
+            // options region renders under `!isCollapsed`, so pointing
+            // `aria-controls` at it while collapsed would be a dangling
+            // reference (axe `aria-valid-attr-value`). `aria-expanded` already
+            // conveys the collapsed state; `aria-controls` is optional here.
+            aria-controls={isCollapsed ? undefined : `facetGroup-${id}-region`}
             data-test-subj={`facetGroup-${id}-toggle`}
           >
             <strong>{label}</strong>
--- a/public/components/alerting/facet_filter_panel.tsx
+++ b/public/components/alerting/facet_filter_panel.tsx
@@ -140,7 +140,7 @@ const FacetErrorIndicator: React.FC<FacetErrorIndicatorProps> = ({
           iconType="alert"
           color="danger"
           size="xs"
-          className="altFacetErrorBtn"
+          className="altFacetErrorBtn altFacetTarget"
           aria-label={ariaLabel}
           onClick={() => setOpen((v) => !v)}
           data-test-subj={`facetGroup-${facetId}-error-${option}`}
@@ -260,6 +260,7 @@ export const FacetFilterGroup: React.FC<FacetFilterGroupProps> = ({
             iconSide="left"
             onClick={() => onToggleCollapse(id)}
             aria-expanded={!isCollapsed}
+            aria-controls={`facetGroup-${id}-region`}
             data-test-subj={`facetGroup-${id}-toggle`}
           >
             <strong>{label}</strong>
@@ -288,14 +289,8 @@ export const FacetFilterGroup: React.FC<FacetFilterGroupProps> = ({
               color="primary"
               onClick={() => onChange([])}
               data-test-subj={`facetGroup-${id}-clear`}
-              style={{
-                display: 'inline-flex',
-                alignItems: 'center',
-                justifyContent: 'center',
-                minWidth: 24,
-                minHeight: 24,
-                padding: '0 4px',
-              }}
+              className="altFacetTarget"
+              style={{ padding: '0 4px' }}
             >
               <EuiText size="xs">
                 <FormattedMessage
@@ -308,7 +303,12 @@ export const FacetFilterGroup: React.FC<FacetFilterGroupProps> = ({
         )}
       </EuiFlexGroup>
       {!isCollapsed && (
-        <div style={{ paddingLeft: 4 }}>
+        <div
+          id={`facetGroup-${id}-region`}
+          role="region"
+          aria-label={label}
+          style={{ paddingLeft: 4 }}
+        >
           {searchable && (
             <>
               <EuiFieldSearch
--- a/public/components/alerting/facet_filter_panel.tsx
+++ b/public/components/alerting/facet_filter_panel.tsx
@@ -15,8 +15,10 @@ import {
   EuiFlexGroup,
   EuiFlexItem,
   EuiIcon,
+  EuiButtonEmpty,
   EuiButtonIcon,
   EuiText,
+  EuiTextColor,
   EuiBadge,
   EuiCheckbox,
   EuiHealth,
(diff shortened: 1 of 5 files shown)


TackAdam:
## Worth addressing

1. **The "≥24px targets" claim is only half-implemented — the facet error button is still `size="xs"` (~16px) with no min-size — `facet_filter_panel.tsx:143`, `alerting.scss:14`**

   The Clear link got explicit `minWidth`/`minHeight: 24` (`:296-301`), but the per-option error indicator — `EuiButtonIcon size="xs"`, the other small interactive target in the same panel, and the one the whole error-indicator refactor is about — got nothing.

   `.altFacetErrorBtn` is just `margin-left` + `flex-shrink`, so it renders at the EUI `xs` size (~16px), below the WCAG 2.5.8 (AA) 24px floor the PR title advertises. And no test guards any target size, so the gap is invisible to CI.

   Either bump it to `size="s"` or give `.altFacetErrorBtn` a `min-width`/`min-height: 24px` (+ centering), and ideally add a test.

2. **`aria-expanded` has no `aria-controls`, and the collapsible region has no `id`/`role` — `facet_filter_panel.tsx:262` → region at `:311`**

   Since adding `aria-expanded` is the point of this PR, wiring `aria-controls={regionId}` on the toggle and `id={regionId} role="region"` on the `{!isCollapsed && <div>}` is the natural completion — otherwise a screen reader announces an expandable control with no associated region.

   Same applies to the SLO tree expander. Medium.

## Nits

3. **Option count can be clipped inside the button — `facet_filter_panel.tsx:248`**

   The title + `optionCount` now live inside `EuiButtonEmpty`, whose `__text` span applies `text-overflow: ellipsis; white-space: nowrap`.

   In a narrowed `EuiResizableContainer` alongside the active-count badge + Clear link, the count can be truncated away — cardinality the plain-`EuiText` header didn't clip.

   Worth a quick visual check in a narrow panel.

4. **Target-size rule lives inline rather than in `alerting.scss` — `facet_filter_panel.tsx:296`**

   The file's own header comment and the `.altFacet*` classes establish that facet styling belongs in SCSS. The inline `24×24` object (rebuilt every render) scatters the WCAG target-size logic across TS + SCSS — which is exactly how finding 1's error button got missed.

   Fold both targets into a shared `.altFacetTarget` class.

5. **`data-test-subj` uses hyphens, not camelCase — `:263` (`facetGroup-${id}-toggle`) and the SLO expander**

   `CLAUDE.md` says `data-test-subj` should be camelCase; these match the file's pre-existing `facetGroup-*` pattern, so it's consistency-preserving, but a literal deviation for newly-added attributes.

   Low.

lezzago:
Thanks @TackAdam — addressed in `0d902b8d`:

1. **≥24px only half-implemented (error button still ~16px)** ✅ — added a shared `.altFacetTarget` SCSS class (min 24×24, centered) and applied it to the per-option error `EuiButtonIcon` (`min-width`/`min-height` win over the `size="xs"` intrinsic box). Test asserts the class is present.
2. **`aria-expanded` with no `aria-controls` / region has no id/role** ✅ — the toggle now sets `aria-controls={facetGroup-<id>-region}`, and the collapsible options region carries `id` + `role="region"` + `aria-label`. Applied the same wiring to the SLO service tree expander. Tests added for both surfaces.
4. **Target-size rule inline rather than in SCSS** ✅ — folded into the same `.altFacetTarget` class and removed the inline 24×24 object from the Clear link (both targets now share one SCSS rule — which is exactly what would have kept the error button from being missed).

3. **Option count can clip inside the button** — this is display-only and needs a narrow-panel visual check; I've left it as a follow-up rather than restructuring the button's text span in this a11y-focused PR.
5. **`data-test-subj` hyphens vs camelCase** — kept as-is: these match the file's established `facetGroup-*` / `slosSuggest*-<name>` pattern, so switching just these would make the attributes inconsistent within the same file. Worth a separate sweep if the repo wants to enforce camelCase on the existing ones too.

public/components/alerting/facet_filter_panel.tsx
```diff
@@ -242,38 +235,48 @@ export const FacetFilterGroup: React.FC<FacetFilterGroupProps> = ({
...
+            // center`, which centers the arrow + title instead of leaving them
+            // flush-left. Pin the content to the start so the header aligns
+            // with the plain-text groups (e.g. "Labels") at every panel width.
+            // `flush="left"` only strips padding, not alignment.
+            contentProps={{ style: { justifyContent: 'flex-start', width: '100%' } }}
+            iconType={isCollapsed ? 'arrowRight' : 'arrowDown'}
+            iconSide="left"
+            onClick={() => onToggleCollapse(id)}
+            aria-expanded={!isCollapsed}
+            aria-controls={`facetGroup-${id}-region`}
```
ps48: ### p2: aria-controls points at a region absent while collapsed

When `isCollapsed` is true the toggle keeps `aria-controls="facetGroup-${id}-region"`, but that region only renders under `!isCollapsed`, so the reference dangles while collapsed. This is an a11y PR, and axe's `aria-valid-attr-value` flags controls pointing at a missing id. Could we always render the region wrapper (toggling its contents or `hidden`) so the target always exists, or drop `aria-controls` when collapsed? The tests only assert the expanded case, so the collapsed state is unguarded.

public/components/apm/pages/slos/suggest_service_tree_table.tsx
```diff
@@ -265,6 +265,8 @@ export const ServiceTreeTable: React.FC<ServiceTreeTableProps> = ({
                           )
                     }
                     iconType={isExpanded ? 'arrowDown' : 'arrowRight'}
+                    aria-expanded={isExpanded}
+                    aria-controls={`slosSuggestServiceExpanded-${row.serviceName}`}
```
ps48: ### p2: expander aria-controls points at a panel that only exists when expanded

Same pattern as the facet header: `aria-controls="slosSuggestServiceExpanded-${row.serviceName}"` is always present on the expander, but the target `div` renders only when `isExpanded`, so a collapsed row has a dangling reference. Could the region be present (empty or `hidden`) whenever the row exists so the control always resolves? The added test covers only the expanded row.

ps48 (commented):
### a11y facet + tree refactor looks solid, one dangling aria-controls to tidy

Confirmed against the head SHA, not main. The nested-interactive removal is done right: the `stopPropagation` hack is gone (not retained), the error button is a real sibling of the checkbox instead of living inside the `<label>`, the header is a native `EuiButtonEmpty` with `aria-expanded` wired to `!isCollapsed`, Clear sits outside the toggle button, the count swap to `EuiTextColor` avoids a block element inside a button, and the tree expander is a real `EuiButtonIcon` with `aria-expanded` bound to `isExpanded`. 24px targets come from a single shared `.altFacetTarget` class. No new deps, flags, routes, or logger noise, OUI throughout, and the RTL tests assert the new ARIA and the nested-interactive absence.

One thing before merge: both `aria-controls` values are always rendered, but the regions they point at only mount when expanded, so a collapsed group/row has a dangling reference that axe's `aria-valid-attr-value` will flag. Details inline. No p0/p1.

TackAdam (approved):

public/components/alerting/facet_filter_panel.tsx
lezzago: Fixed in `7f677738` — the toggle now only sets `aria-controls` when expanded: `aria-controls={isCollapsed ? undefined : `facetGroup-${id}-region`}`. While collapsed the region isn't rendered, so there's no dangling reference for axe `aria-valid-attr-value`; `aria-expanded={!isCollapsed}` conveys the state and `aria-controls` is optional in the disclosure pattern. Added a test asserting the collapsed toggle has no `aria-controls` and mounts no region.

public/components/apm/pages/slos/suggest_service_tree_table.tsx
lezzago: Fixed in `7f677738` — same treatment on the expander: `aria-controls={isExpanded ? `slosSuggestServiceExpanded-${row.serviceName}` : undefined}`, so a collapsed row no longer references the unmounted panel. Added a test for the collapsed row (no `aria-controls`, no region in the DOM).

ps48 (approved):

PR #2836, opened 2026-08-26
[SLO] Localize percent formatting; make SLO_PRECISION the single precision policy

lezzago:
## What & why
Localizes the `%` suffix and the `—` fallback and makes `SLO_PRECISION` the single source of precision. en-US output is unchanged for SLO value ranges (`|value| < 10`, i.e. percentages `< 1000%`, which covers all real SLO data); the change surfaces under non-en locales (localized grouping/decimal separators) and adds a grouping comma only for percentages `≥ 1000%`.

**Findings:** CLAR10, CLAR3.

## Before / after

**Before / after (SLO listing; identical under en-US by design — see note)**

Before / after (SLO listing; identical under en-US by design — see note)

**Flow**

Flow

## Testing
`format.test.ts` incl. a `de`-locale assertion. **Note:** no visible change under en-US.

## Review
Independent review agent: **APPROVE** — `SLO_PRECISION` values byte-identical to main; i18n id unique.

## Regression check

## Dependency
No dependencies — can merge independently. Consumers migrate to `SLO_PRECISION` in #2830 and #2833.

### DIFF
--- a/common/slo/format.ts
+++ b/common/slo/format.ts
@@ -22,13 +22,41 @@ export const EMPTY_VALUE_FALLBACK = i18n.translate('observability.slo.format.emp
   defaultMessage: '—',
 });
 
+/**
+ * Cache of percent formatters keyed by `${locale}:${decimals}`. Constructing an
+ * `Intl.NumberFormat` is markedly heavier than the old `toFixed`, and
+ * `formatPct` is wired into ECharts axis/tooltip formatters that fire on every
+ * render and hover — so we build each (locale, decimals) formatter once and
+ * reuse it. Keying on the locale means a runtime `setLocale` still resolves to
+ * the correct formatter (a different key) rather than a stale one. The map is
+ * bounded by (locales × the handful of `decimals` values) so it can't grow
+ * unboundedly.
+ */
+const percentFormatterCache = new Map<string, Intl.NumberFormat>();
+
+function getPercentFormatter(decimals: number): Intl.NumberFormat {
+  const locale = i18n.getLocale();
+  const key = `${locale}:${decimals}`;
+  let formatter = percentFormatterCache.get(key);
+  if (!formatter) {
+    formatter = new Intl.NumberFormat(locale, {
+      style: 'percent',
+      minimumFractionDigits: decimals,
+      maximumFractionDigits: decimals,
+    });
+    percentFormatterCache.set(key, formatter);
+  }
+  return formatter;
+}
+
 export function formatPct(value: number, options: FormatPctOptions = {}): string {
   const { decimals = 1, fallback = EMPTY_VALUE_FALLBACK } = options;
   if (!Number.isFinite(value)) return fallback;
   // `Intl.NumberFormat` percent style places the `%` sign (and any grouping)
   // per locale and multiplies by 100 itself, so we pass the raw ratio — NOT
   // `value * 100` — and pin both fraction-digit bounds to `decimals` for a
-  // fixed number of decimals.
+  // fixed number of decimals. The formatter is memoized per (locale, decimals)
+  // — see getPercentFormatter — because this runs on every chart render/hover.
   //
   // Rounding note: `Intl` uses half-expand (round half away from zero), which
   // can differ in the last digit from the old `(value*100).toFixed(decimals)`
@@ -37,11 +65,7 @@ export function formatPct(value: number, options: FormatPctOptions = {}): string
   // This is intentional — standard rounding, and no caller parses the string
   // back to a number — so display may shift by one unit in the last decimal
   // versus the pre-i18n formatter.
-  return new Intl.NumberFormat(i18n.getLocale(), {
-    style: 'percent',
-    minimumFractionDigits: decimals,
-    maximumFractionDigits: decimals,
-  }).format(value);
+  return getPercentFormatter(decimals).format(value);
 }
 
 /**
--- a/common/slo/format.ts
+++ b/common/slo/format.ts
@@ -27,8 +27,16 @@ export function formatPct(value: number, options: FormatPctOptions = {}): string
   if (!Number.isFinite(value)) return fallback;
   // `Intl.NumberFormat` percent style places the `%` sign (and any grouping)
   // per locale and multiplies by 100 itself, so we pass the raw ratio — NOT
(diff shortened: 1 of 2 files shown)


TackAdam:
## Worth addressing

1. **The precision policy isn't actually consistent yet — budget-remaining still renders at two different precisions — `slo_budget_remaining_chart.tsx:132,156` & `slo_budget_sparkline.tsx:93`**

   These call `formatPct(v)` with no decimals, so they get the default `1`, while `slo_budget_panel.tsx:500` renders the same budget-remaining value at `SLO_PRECISION.budget = 2`.

   So `0.5` reads `50.0%` in the chart tooltip/sparkline but `50.00%` in the panel — the exact "four different precisions" inconsistency the new `SLO_PRECISION` doc says "every surface must read from this map" to eliminate.

   The PR body says consumers migrate in 2830/2833, but those cover the listing/overview — these budget-chart surfaces don't obviously fall under either. Worth confirming they're in a migration PR, or fixing here.

2. **`EMPTY_VALUE_FALLBACK` is introduced but almost nothing uses it (`CLAR10` largely unrealized) — `format.ts:21`**

   Dozens of hardcoded `'—'` literals remain across the SLO surfaces, and two callers (`slo_detail_page.tsx:148`, `probe_sli_panel.tsx:327`) pass a literal `'—'` as the null path rather than the new constant.

   A translator swapping `observability.slo.format.emptyValue` changes almost nothing today. Fine as a foundation if adoption lands in a follow-up — but flag it so the constant doesn't sit dead.

## Nits

3. **"en-US byte-for-byte unchanged" is slightly overstated — `format.ts`**

   It's true for `|value| < 10` (percentages `< 1000%`), which covers all real SLO data — but `formatPct(12.3456)` now gains a grouping comma (`1,234.56%`), and non-en locales get a comma decimal + NBSP-before-`%`.

   The author already pins this in the test (`:42-44`), so it's known; just tighten the PR wording to "unchanged for SLO value ranges."

4. **`EMPTY_VALUE_FALLBACK` is frozen at module import — `format.ts:21`**

   `i18n.translate(...)` runs once at import, so a runtime `setLocale` (or a translation registered after this module loads) won't be reflected.

   Same pattern as the theme-token-at-load note from 2826; low risk since the em-dash `defaultMessage` is locale-agnostic anyway.

5. **Per-call `Intl.NumberFormat` allocation — `format.ts:40`**

   A new formatter is constructed on every call, and `formatPct` is wired into ECharts axis/tooltip formatters (`slo_budget_remaining_chart.tsx:156`) that fire on every render/hover.

   `Intl.NumberFormat` construction is markedly heavier than the old `toFixed`. A small memo keyed on `(locale, decimals)` would erase the regression. Low priority.

6. **`de` test assumes full ICU — `format.test.ts:66`**

   Passes here and in CI (Node 22 ships full ICU), but under a small-ICU build `de` falls back to `en` and the assertions fail.

   Consider guarding or noting the ICU dependency.

lezzago:
Thanks @TackAdam:

1. **Precision not consistent (budget-remaining chart + sparkline)** ✅ — those files are owned by #2832 (charts), so I fixed it there (commit `99d87f07`): the budget-remaining tooltip/axis/threshold label and the budget sparkline now render via `formatPct(v, { decimals: SLO_PRECISION.budget })`, matching the budget panel. So `0.5` reads `50.00%` everywhere. Kept out of this PR to preserve file-disjointness.
3. **"en-US byte-for-byte unchanged" overstated** ✅ — tightened the PR body to "unchanged for SLO value ranges (`|value| < 10`)" and noted the grouping comma only appears for percentages ≥ 1000%.
5. **Per-call `Intl.NumberFormat` allocation** ✅ — memoized the formatter per `(locale, decimals)` in `format.ts` (commit `ef0e16b6`); keying on locale keeps a runtime `setLocale` correct.
6. **`de` test assumes full ICU** ✅ — the test now detects real `de` support (comma separator) and skips the divergence assertions on a small-ICU build; CI's full-ICU Node 22 still exercises the real path.

2 & 4. **`EMPTY_VALUE_FALLBACK` adoption / module-load freeze** — acknowledged as foundation: broad migration of the remaining hardcoded `'—'` literals is a follow-up (touches many files across other PRs' surfaces), and the module-load freeze is low-risk since the em-dash `defaultMessage` is locale-agnostic. Flagged so the constant doesn't sit dead.

common/slo/format.ts
```diff
@@ -30,9 +84,24 @@ export const TABULAR_NUMS_STYLE = {
 };
 
 /**
- * SLO numeric precision policy (audit P1 #12). Render contexts share a
- * precision so decimals line up column-to-column — a budget card mixing
- * `100%` / `100.0%` / `100.00%` was the original offender.
+ * SLO numeric precision policy (audit P1 #12, CLAR3). THE single source of
+ * truth for how many decimal places each SLO surface renders — pass the
+ * relevant key as `formatPct`'s `decimals` so the same value reads identically
```
ps48, line 89: ### p2: the precision-policy doc tells callers to pass burnRate through formatPct, which misrenders

The added policy comment says to pass the relevant `SLO_PRECISION` key as `formatPct`'s `decimals`, and lists `burnRate` (a `3.2x` multiplier). But `formatPct` forces percent style and multiplies by 100, so `formatPct(3.2, { decimals: SLO_PRECISION.burnRate })` renders `320.0%` instead of `3.2x`. Could we scope the 'pass to formatPct' instruction to the percent surfaces, and note that `burnRate` (and `eventsRatio` if ever shown as a raw ratio) is a precision-only key formatted elsewhere?

ps48 (commented):
### slo percent formatting review

Verified at head `ef0e16b`, scoped to `common/slo/format.ts` and its new test. `formatPct` moves to `Intl.NumberFormat` percent style with the raw ratio and pinned fraction digits (no double-multiply), `EMPTY_VALUE_FALLBACK` is an i18n-translatable glyph, and `SLO_PRECISION` is the single precision policy, all correctly kept in `common/` where server and client share them. The per-`locale:decimals` formatter cache is a sensible bounded perf guard. Tests cover en defaults, fixed precision, the no-double-multiply invariant, grouping, non-finite fallback, and a real `de`-locale divergence. No new deps, routes, console, or dead code.

One small doc nit inline. No p0/p1.

TackAdam (approved):

PR #2837, opened 2026-08-26
[Alerting] Align the forecaster edit gate with the running-state predicate

lezzago:
## What & why
The forecaster edit gate decided "is it running?" with a hand-maintained state list that had drifted from `isAdResourceRunning` — the shared predicate the detail flyout uses for its Stop button. Generic states like `Initializing` slipped through, so the backend rejected the edit with a raw internal string. This makes the gate reuse the shared predicate as a deliberate **superset** — it blocks everything `isAdResourceRunning` treats as running, **plus** failure states (`FORECAST_FAILURE`) that predicate classifies as not-running but that still can't be edited in place — and swaps the raw error for guidance.

**Code before / after** (no runtime screenshot — see note):

_Before_ — gate used its own drifting list; the raw backend error reached the user:
```ts
const message = buildErrorMessage(error); // "Job is running: forecast-3f9c…"
```
_After_ — gate reuses the shared predicate; the leak is replaced with actionable copy:
```ts
if (isAdResourceRunning(asForecasterRunningInput(resource))) return true; // catches "Initializing" too
// …
const message = humanizeAdUpdateError(buildErrorMessage(error), ruleType);
// → "Stop the forecaster before editing its configuration, then restart it after your changes are saved."
```

**Findings:** BUG-AD1.

## Testing
`forecaster_edit_gate.test.ts` asserts the safety-critical direction — the gate blocks **every** status `isAdResourceRunning` treats as running (so the backend never rejects a permitted edit) — plus the failure states the gate additionally blocks, and that the raw 'Job is running' string never reaches the user. **Note:** no runtime before/after screenshot — the local Docker stack has no running forecaster (the forecast `_search` proxy returns 'No Living connections'); flagged for managed-env visual verification.

## Review
Independent review agent (earlier round): review commit applied.

## Regression check
This PR ships no runtime screenshot (see the Testing note); regression safety rests on the unit tests above plus central verification: the change is file-disjoint from the other in-flight audit fixes (no file overlap, so it merges cleanly with them), and the affected plugin test suites pass with **0 new type errors** vs `main`. It can be reviewed, merged, and reverted independently.

## Dependency
None.

### DIFF
--- a/public/components/alerting/__tests__/forecaster_edit_gate.test.ts
+++ b/public/components/alerting/__tests__/forecaster_edit_gate.test.ts
@@ -54,10 +54,17 @@ const summaryForStatus = (status: MonitorStatus, enabled = false): UnifiedRuleSu
     enabled,
   }) as unknown as UnifiedRuleSummary;
 
+// A real forecaster `curState` is the backend UPPER_SNAKE enum (`RUNNING`,
+// `INITIALIZING_FORECAST`), NOT the prose `MonitorStatus`. Feed that enum so the
+// gate's raw→prose normalization + the shared `isAdResourceRunning` predicate are
+// what's actually exercised — the prose form is a domain the gate never receives
+// in production, so feeding it would prove a code path that doesn't run.
+const toRawEnum = (status: MonitorStatus): string => status.toUpperCase().replace(/ /g, '_');
+
 const rawForecasterForStatus = (status: MonitorStatus, enabled = false): ADForecaster =>
   ({
     id: 'forecast-1',
-    curState: status,
+    curState: toRawEnum(status),
     enabled,
   }) as unknown as ADForecaster;
 
@@ -100,6 +107,23 @@ describe('forecaster edit gate vs shared running-state predicate (BUG-AD1)', () 
     expect(
       getEditLifecycleBlocker('forecaster', rawForecasterForStatus('Stopped', true), false)
     ).not.toBeNull();
+  });
+
+  it('engages the shared predicate on the raw UPPER_SNAKE curState a forecaster actually carries', () => {
+    // Explicit guard for the production shape: cur_state is the enum, not prose.
+    // The raw→prose normalization must let isAdResourceRunning fire so these block.
+    const raw = (curState: string, enabled = false): ADForecaster =>
+      ({ id: 'forecast-1', curState, enabled }) as unknown as ADForecaster;
+    expect(getEditLifecycleBlocker('forecaster', raw('RUNNING'), false)).not.toBeNull();
+    expect(getEditLifecycleBlocker('forecaster', raw('INITIALIZING'), false)).not.toBeNull();
+    expect(
+      getEditLifecycleBlocker('forecaster', raw('INITIALIZING_FORECAST'), false)
+    ).not.toBeNull();
+    expect(
+      getEditLifecycleBlocker('forecaster', raw('AWAITING_DATA_TO_INIT'), false)
+    ).not.toBeNull();
+    // A genuinely stopped, disabled forecaster carries no running enum → editable.
+    expect(getEditLifecycleBlocker('forecaster', raw('STOPPED'), false)).toBeNull();
   });
 });
 
--- a/public/components/alerting/__tests__/forecaster_edit_gate.test.ts
+++ b/public/components/alerting/__tests__/forecaster_edit_gate.test.ts
@@ -5,7 +5,11 @@
 
 import { getEditLifecycleBlocker, humanizeAdUpdateError } from '../create_ad_rule_flyout';
 import { isAdResourceRunning } from '../shared_constants';
-import type { ADForecaster, MonitorStatus, UnifiedRuleSummary } from '../../../../common/types/alerting';
+import type {
+  ADForecaster,
+  MonitorStatus,
+  UnifiedRuleSummary,
+} from '../../../../common/types/alerting';
 
 /**
  * BUG-AD1: the forecaster edit gate (getEditLifecycleBlocker) decided "is it running"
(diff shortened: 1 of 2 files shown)


TackAdam:
## Minor notes (no code change strictly required)

1. **The "two predicates agree for every status" claim is imprecise (prose + comment).**

   They don't fully agree — the gate is a deliberate superset. Failure states like `Forecast failure` / `Initialization failure` are classified by `isAdResourceRunning` as not running (they're in `AD_RESOURCE_START_ACTION_STATUSES`), yet the explicit list still blocks editing on `FORECAST_FAILURE`.

   The `||` masks the disagreement, and the test only asserts the safety-critical direction (running ⟹ blocked, `forecaster_edit_gate.test.ts:71-74`), which is right. But the PR body and the `:715` comment read as full equivalence.

   Consider softening to: "the gate is a superset of `isAdResourceRunning` — it blocks everything the backend would reject, plus failure states."

   Functionally correct as-is.

2. **`asForecasterRunningInput` name/type mismatch — `:706`**

   It accepts `ADDetector | ADForecaster` but hardcodes `monitorType`/`definitionType: 'forecaster'`.

   Only called from the forecaster path, so harmless, but the broad parameter type is slightly misleading. A `resource?: ADForecaster` param (or a comment) would tighten it.

3. **PR-description boilerplate is inconsistent (not code).**

   The "Regression check" paragraph claims a "full-viewport capture" screenshot, while the Testing section correctly states there's no runtime screenshot (code-only before/after, Docker stack has no live forecaster).

   Worth cleaning up the templated paragraph so reviewers aren't misled — and the flagged managed-env visual verification is a reasonable ask before merge given there's no local repro.

lezzago:
Thanks @TackAdam — all three addressed (`882fae08` + PR body):

1. **"Two predicates agree for every status" is imprecise** ✅ — reworded the `:715` comment and the PR body to state the gate is a deliberate **superset** of `isAdResourceRunning`: it blocks everything that predicate treats as running, **plus** failure states (`FORECAST_FAILURE`) it classifies as not-running but that still can't be edited in place. The Testing note now says the test asserts the safety-critical direction (running ⟹ blocked) rather than full equivalence.
2. **`asForecasterRunningInput` name/type mismatch** ✅ — added a doc note: the broad `ADDetector | ADForecaster` param exists only because the sole caller holds that union; the adapter is forecaster-only and intentionally hardcodes the `'forecaster'` kind (detectors use a different edit lifecycle and never reach this path).
3. **PR-description boilerplate inconsistent** ✅ — the generated "Regression check" paragraph no longer claims a full-viewport screenshot for screenshot-less PRs; it now rests on the unit tests + central verification, consistent with the Testing note. The managed-env visual verification remains flagged as a reasonable pre-merge ask.

public/components/alerting/create_ad_rule_flyout.tsx
```diff
@@ -691,7 +697,37 @@ const isForecasterTestInitializing = (resource?: ADDetector | ADForecaster): boo
...
+ * The parameter type is the broad `ADDetector | ADForecaster` only because the single
+ * caller (`isForecasterActiveForEdit`) holds that union; this adapter is forecaster-only
+ * and intentionally hardcodes the `'forecaster'` kind (detectors use a different edit
+ * lifecycle and never reach this path).
+ */
+const asForecasterRunningInput = (resource?: ADDetector | ADForecaster): UnifiedRuleSummary =>
+  ({
+    monitorType: 'forecaster',
+    definitionType: 'forecaster',
+    status: getResourceState(resource) as MonitorStatus,
```
ps48: ### p1: the isAdResourceRunning reuse never matches a real curState

`getResourceState` returns `normalizeStateValue(curState)`, and `normalizeStateValue` only trims the raw value, so this hands `isAdResourceRunning` the backend's uppercase enum (`RUNNING`, `INITIALIZING_FORECAST`, ...). But `isAdResourceRunning` keys its Sets on the human-readable `MonitorStatus` values (`Running`, `Initializing`, `Initializing forecast`), so for a real forecaster this branch matches nothing and falls through. The reuse is effectively dead, and the comment above ("the same human-readable domain the shared status sets are keyed on") is inaccurate. The gate still blocks correctly via the `enabled` flag and the superset `stateMatches` list below (which matches `state.toUpperCase() === value`, so it does catch the enum forms). Could we either map `curState` through the status map before passing it in, so the reuse genuinely fires and the two predicates share a domain, or drop the reuse branch and rely on the superset list, and correct the comment? As written, any running/initializing state that only `isAdResourceRunning` would catch but that is not in the superset list would slip through.

public/components/alerting/__tests__/forecaster_edit_gate.test.ts
```diff
@@ -0,0 +1,127 @@
...
+    monitorType: 'forecaster',
+    definitionType: 'forecaster',
+    status,
+    enabled,
+  }) as unknown as UnifiedRuleSummary;
+
+const rawForecasterForStatus = (status: MonitorStatus, enabled = false): ADForecaster =>
+  ({
+    id: 'forecast-1',
+    curState: status,
```
ps48: ### p2: the test feeds the human-readable domain the gate never actually receives

`rawForecasterForStatus` sets `curState: status` where `status` is drawn from `ALL_STATUSES`, the human-readable `MonitorStatus` values (`Running`, `Initializing`, ...). But a real `raw.curState` carries the uppercase enum. So the suite goes green by feeding `isAdResourceRunning` exactly the domain it accepts, which is the one branch that does not run in production, while never exercising the superset `stateMatches` list that does the real blocking. Could `ALL_STATUSES` (or a companion case) iterate the enum forms a real `curState` holds so the raw path is what is proven?

ps48 (commented):
### the isAdResourceRunning reuse does not actually engage on a real curState

The intent (reuse `isAdResourceRunning` so the edit gate stops drifting from the detail flyout) is right, and because the gate is a strict superset it cannot newly permit an edit the old gate blocked. But verified against head `882fae0`: `getResourceState` returns `normalizeStateValue(curState)`, and `normalizeStateValue` only trims (`stringValue(value).trim()`), so `asForecasterRunningInput` hands `isAdResourceRunning` the backend's uppercase enum (`RUNNING`, `INITIALIZING_FORECAST`, ...). `isAdResourceRunning` keys its Sets on the human-readable `MonitorStatus` values (`Running`, `Initializing`, ...), so that branch matches nothing for a real forecaster and is effectively dead. The gate still blocks running/initializing forecasters via the `enabled` flag and the superset `stateMatches` list (which compares `state.toUpperCase() === value`, so it does catch the enum forms), so this is not a functional regression for the states in that list. The issues are that the stated goal of sharing the predicate is not achieved, the adapter comment ("the same human-readable domain the shared status sets are keyed on") is inaccurate, and any state only `isAdResourceRunning` would catch but that is missing from the superset list would slip. Two asks inline. No p0.

jackiehanyang:
my findings are already covered by the existing P1/P2 comments left by @ps48 on the raw curState normalization mismatch and the matching test coverage gap.

TackAdam (approved):

public/components/alerting/create_ad_rule_flyout.tsx
lezzago: Great catch — you're right the reuse was effectively dead: `getResourceState` returns the raw UPPER_SNAKE `cur_state` (`RUNNING`, `INITIALIZING_FORECAST`), while `isAdResourceRunning` keys its Sets on the prose `MonitorStatus` (`Running`, `Initializing forecast`), so it matched nothing and fell back to `enabled`. In production a generic `INITIALIZING` forecaster could then slip the gate — the exact BUG-AD1 regression — with the test masking it by feeding prose.

Fixed in `7535c289` with your option (a): `asForecasterRunningInput` now runs `curState` through a new `toMonitorStatus` (lowercase, `_`→space, capitalize) before `isAdResourceRunning`, so the shared predicate genuinely engages for every running/initializing state (including the generic `Initializing`). For every AD/forecaster state the prose form is exactly the enum humanized, so this is a faithful normalization (idempotent for values already in prose). Corrected the inaccurate adapter comment. The `stateMatches` superset list is kept as defense-in-depth for the `enabled` flag + the failure states `isAdResourceRunning` classifies as not-running (e.g. `FORECAST_FAILURE`).

public/components/alerting/__tests__/forecaster_edit_gate.test.ts
lezzago: Fixed in `7535c289`. `rawForecasterForStatus` now feeds the UPPER_SNAKE enum a real `curState` carries (`toRawEnum`), so the suite exercises the raw→prose normalization + shared predicate rather than the prose domain the gate never receives. Added an explicit test asserting the raw enum path engages `isAdResourceRunning` (`RUNNING`/`INITIALIZING`/`INITIALIZING_FORECAST`/`AWAITING_DATA_TO_INIT` block; a stopped+disabled `STOPPED` stays editable). Also taught `isForecasterTestInitializing` the `INITIALIZING_TEST` enum so the test-initializing state still routes to `forecaster-test` on the raw path.

ps48 (approved):

PR #2838, opened 2026-08-26
[SLO] Add the missing edit path: #/slos/:id/edit with wizard prefill + update

lezzago:
## What & why
SLO detail had no way to edit an existing SLO. Adds an `#/slos/:id/edit` route + an Edit action, prefilling the wizard from the saved SLO and calling the existing update path. The datasource is pinned read-only because the deployed rules are keyed to it.

**Findings:** BUG-S3.

## Before / after

**Before / after (Edit action appears)**

Before / after (Edit action appears)

**Flow**

Flow

**New edit wizard, prefilled**

New edit wizard, prefilled

## Testing
`slo_edit_wizard.test.tsx` + `wizard_state.test.ts` + detail test. Centrally green.

## Review
Independent review agent: **APPROVE with fixes applied** — caught a real bug: editing a **paused** SLO silently re-enabled it (`buildCreateInput` hardcodes `enabled:true` and the server merges the PUT over stored). Fixed by preserving the loaded `enabled` flag. Also fixed two type errors in the test file (tsc had not been run).

## Regression check

## Dependency
**Stacked on #2831 (SLO detail) — merge #2831 first.** This branch includes #2831's commits; once it merges, a rebase drops them from this diff.

### DIFF
--- a/public/components/apm/pages/slos/__tests__/slo_detail_page.test.tsx
+++ b/public/components/apm/pages/slos/__tests__/slo_detail_page.test.tsx
@@ -304,6 +304,15 @@ describe('SloDetailPage — View alerts pivot (OBS1)', () => {
   });
 });
 
+describe('SloDetailPage — Edit affordance (BUG-S3)', () => {
+  it('renders an Edit action linking to the SLO edit route', async () => {
+    renderPage({ get: jest.fn().mockResolvedValue(makeDoc()) });
+
+    const editButton = await screen.findByTestId('slosDetailEdit');
+    expect(editButton).toHaveAttribute('href', '#/slos/slo-1/edit');
+  });
+});
+
 describe('SloDetailPage — rule-health grace window + re-poll (F-CRUD2)', () => {
   beforeEach(() => {
     jest.useFakeTimers();
--- a/public/components/apm/pages/slos/__tests__/slo_detail_page.test.tsx
+++ b/public/components/apm/pages/slos/__tests__/slo_detail_page.test.tsx
@@ -311,6 +311,18 @@ describe('SloDetailPage — Edit affordance (BUG-S3)', () => {
     const editButton = await screen.findByTestId('slosDetailEdit');
     expect(editButton).toHaveAttribute('href', '#/slos/slo-1/edit');
   });
+
+  it('hides Edit for an SLO the wizard cannot edit (no dead-end round trip)', async () => {
+    // Calendar-window SLO is unsupported by the wizard, so the detail page must
+    // not offer Edit (it would only land on "This SLO can't be edited here").
+    const doc = makeDoc();
+    doc.spec.window = { type: 'calendar', period: 'month' };
+    renderPage({ get: jest.fn().mockResolvedValue(doc) });
+
+    // The page rendered (Delete is always present), but Edit is absent.
+    expect(await screen.findByTestId('slosDetailDelete')).toBeInTheDocument();
+    expect(screen.queryByTestId('slosDetailEdit')).not.toBeInTheDocument();
+  });
 });
 
 describe('SloDetailPage — rule-health grace window + re-poll (F-CRUD2)', () => {
--- a/public/components/apm/pages/slos/__tests__/slo_edit_wizard.test.tsx
+++ b/public/components/apm/pages/slos/__tests__/slo_edit_wizard.test.tsx
@@ -1,0 +1,294 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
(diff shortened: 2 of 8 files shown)


TackAdam:
## What I verified as correct

- **Enabled-flag preservation** — the server merges `{ ...existing.spec, ...input.spec }` (shallow top-level, `slo_lifecycle_service.ts:439`), and `buildCreateInput` hardcodes `enabled: true`, so the submit's `{ ...input.spec, enabled: editEnabled }` is exactly right. The earlier review's catch was real and the fix is correct. ✅

- **Optimistic concurrency** — version sent, bails if null, advances after save so repeat saves don't `409`. ✅

- **Non-representable window** — `toEditableWindowDuration` returns `null` → "unsupported" gate rather than silently coercing (which the shallow merge would persist). Correct and well-reasoned. ✅

- **Secondary owner teams** — carried and deduped so a multi-team SLO isn't truncated. ✅

- **Route wiring** — both routes exact, distinct paths; `useParams` wrapper keeps the wizard route-shape-agnostic. ✅

- **Round-trip completeness** — I checked every top-level `SloSpec` field the builder emits against the shallow-merge risk:
  - `alarms` (only `{ enabled }` / `{ enabled, forDuration }` exist — fully re-emitted)
  - `alerting` (single MWMBR strategy, `burnRates` hydrated)
  - `canonicalKind` (not emitted by the builder → preserved by the merge)

  All safe. ✅

## Worth fixing

### 1. Editing silently flattens array-valued labels to a CSV string

`wizard_state.ts:662` (`recordToEntries`) ↔ `wizard_builders.ts` (`entriesToRecord`)

`SloSpec.labels` is `Record<string, string | string[]>`. Hydration joins an array label `{ region: ['us', 'eu'] }` into `'us,eu'`; the builder re-emits `Record<string, string>`; and because `buildCreateInput` does emit `labels`, the server's shallow merge replaces the whole `labels` object.

So opening the edit wizard on an API-created SLO with array labels, changing any unrelated field, and saving permanently rewrites `region` to the scalar `'us,eu'` — silent data mutation the user never asked for.

`annotations` is string-only, so it's unaffected — this is labels-specific.

The wizard can't author array labels, so this only hits API-created SLOs, but it's a real fidelity regression.

**Fix:** Preserve array-valued labels through hydration (keep them out-of-band like `ownerTeamsSecondary`), or have the builder round-trip the original shape.

## Nits

### 2. Edit is offered for SLOs that can't be edited

The detail page always renders the fill **Edit** button; clicking it on a composite/non-Prometheus SLO loads the wizard only to show **"This SLO can't be edited here."**

Since the detail page has `doc`, it could hide/disable **Edit** for unsupported SLIs and avoid the dead-end round trip.

Also consider whether **Edit** should be `fill` (primary) alongside the existing **Enable/Disable** action — two filled buttons compete.

### 3. Read-only datasource shows the raw ID

The disabled `EuiFieldText` renders `state.datasourceId` (e.g. `ds-1`) rather than the friendly name the interactive `DatasourceSelect` shows.

Minor, but the read-only view is slightly less legible than the editable one it replaces.

lezzago:
Thanks @TackAdam — thorough pass, and the array-label catch is a real one. All three addressed in `6f1ca1ab`:

**1. Array-valued labels flattened to CSV (worth fixing)** ✅ — exactly right: `SloSpec.labels` is `Record<string, string | string[]>`, hydration joined `{region:['us','eu']}` to `'us,eu'`, and since `buildCreateInput` emits the whole `labels` object the server's shallow merge would persist the rewrite. Fixed by carrying array-valued labels **out-of-band** (`preservedArrayLabels`, mirroring `ownerTeamsSecondary`): hydration keeps them out of the scalar key/value editor, and `buildCreateInput` re-emits them verbatim. Test covers hydrate-split + round-trip (`{region:['us','eu']}` stays an array). `annotations` is string-only so it's untouched.

**2. Edit offered for non-editable SLOs (nit)** ✅ — added `isSloEditable(doc)` (mirrors `hydrateFromDoc`'s single-Prometheus + representable-rolling-window guards); the detail page now only renders **Edit** for editable SLOs, so composite / non-Prometheus / calendar-window SLOs no longer get a dead-end button. Also dropped `fill` so Edit isn't a competing second primary. Test asserts Edit is hidden for a calendar-window SLO.

**3. Read-only datasource shows the raw id (nit)** ✅ — the read-only field now resolves the connection id to the friendly name via `useDatasources` (same source as the interactive `DatasourceSelect`), falling back to the id if the list hasn't resolved.

Thanks also for the detailed round-trip verification (enabled flag, optimistic concurrency, window, secondary teams, `alarms`/`alerting`/`canonicalKind` under the shallow merge) — that matches my understanding of the merge contract.

TackAdam (approved):

PR #2842, opened 2026-08-26
fix(apm): stop service graphs re-rendering on latency filter change #2626

cnoramut:
### Description
Services detail Dependencies and Operations tabs remounted the whole table and re-fetched the three expanded-row charts whenever the latency percentile changed, because the table was keyed on latencyPercentile.

- Remove `latencyPercentile` from the table key so switching percentile updates the latency column in place via the memoized columns prop instead of remounting. Pagination and expanded-row charts are preserved.
- Gate the latency range filter on `latencyUserModified` so it only engages when the user actually adjusts the slider. This avoids a false positive on the render right after a percentile switch, where the stale range was compared against the new larger bounds and briefly filtered out rows.

#### Before

https://github.com/user-attachments/assets/ea29701a-d4ca-4cde-9eb2-23a9ebdf6b71

#### After

https://github.com/user-attachments/assets/5bcde003-c1fa-4d1c-80e7-a8964f289723

### Issues Resolved
[#2626](https://github.com/opensearch-project/dashboards-observability/issues/2626)

### Check List

### DIFF
--- a/public/components/apm/pages/service_details/service_operations.tsx
+++ b/public/components/apm/pages/service_details/service_operations.tsx
@@ -255,7 +255,11 @@ export const ServiceOperations: React.FC<ServiceOperationsProps> = ({
   }, [timeRange]);
 
   // Fetch operations list from PPL
-  const { data: operationsData, isLoading: opsLoading, error: opsError } = useOperations({
+  const {
+    data: operationsData,
+    isLoading: opsLoading,
+    error: opsError,
+  } = useOperations({
     serviceName,
     environment,
     startTime: parsedTimeRange.startTime,
@@ -408,8 +412,8 @@ export const ServiceOperations: React.FC<ServiceOperationsProps> = ({
           latencyPercentile === 'p99'
             ? op.p99Duration
             : latencyPercentile === 'p90'
-            ? op.p90Duration
-            : op.p50Duration;
+              ? op.p90Duration
+              : op.p50Duration;
         if (opLatency < latencyRange[0] || opLatency > latencyRange[1]) {
           return false;
         }
@@ -702,8 +706,8 @@ export const ServiceOperations: React.FC<ServiceOperationsProps> = ({
           latencyPercentile === 'p99'
             ? 'p99Duration'
             : latencyPercentile === 'p90'
-            ? 'p90Duration'
-            : 'p50Duration',
+              ? 'p90Duration'
+              : 'p50Duration',
         name: (
           <EuiFlexGroup gutterSize="xs" alignItems="center" responsive={false}>
             <EuiFlexItem grow={false}>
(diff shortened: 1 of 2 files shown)


TackAdam:
Nice, well-scoped fix — removing the table key and gating the filter on `latencyUserModified` both check out, and the default-case flash is fully addressed. A couple of optional nits, none blocking:

1. **Residual one-render flash when a latency filter is already active during a percentile switch.**  
   The `latencyUserModified.current` gate cleanly fixes the default (no-filter) case. But if the user *has* dragged the slider and then switches percentile, `latencyUserModified.current` is still `true` on the switch render (the reset `useEffect` runs afterward), so the stale range is compared against the new bounds for one frame. Still strictly better than before — just flagging that the "avoids a false positive after a percentile switch" claim only fully holds when no filter is active.

2. **No test coverage.**  
   A small RTL test asserting pagination + expanded-row state persist across a percentile switch would lock this in against future regressions.

3. **Pre-existing (fine to skip in this PR):** `service_dependencies.tsx` guards `depLatency === undefined` in the latency filter, but `service_operations.tsx` has no equivalent guard on `opLatency`. Worth aligning at some point if an operation row can ever lack a duration.

TackAdam (approved):

public/components/apm/pages/service_details/service_operations.tsx
```diff
@@ -129,8 +127,10 @@ const OperationsTablePanelUI: React.FC<OperationsTablePanelProps> = ({
         <p>{operationsCount === 0 ? noDataMessage : noFilteredDataMessage}</p>
       </EuiText>
     ) : (
+      // Don't key the table on latencyPercentile
```
ps48, line 130: ### p1: add a regression test for the re-render fix

This is exactly the kind of subtle behavior that silently regresses: a future change re-adds a key or a dep and the table remounts / the filter falsely activates again, with nothing to catch it. Could we add a test that (a) switches the latency percentile and asserts the table is not remounted (pagination / expanded-row state preserved, charts not re-fetched), and (b) asserts the latency filter stays inactive after a percentile switch until the slider is actually moved? `data-test-subj="latencyPercentileSelector"` already exists to drive it. Same applies to the dependencies tab.

public/components/apm/pages/service_details/service_operations.tsx
```diff
@@ -399,16 +403,18 @@ export const ServiceOperations: React.FC<ServiceOperationsProps> = ({
     return textFilteredOperations.filter((op) => {
       // Latency range filter (only if range has been adjusted)
       const isLatencyFilterActive =
-        latencyRange[0] > latencyBounds.min || latencyRange[1] < latencyBounds.max;
+        // Gating on latencyUserModified avoids a false positive after a percentile switch
+        latencyUserModified.current &&
```
ps48, line 407: ### p1: filter gate reads a mutable ref inside useMemo

`latencyUserModified.current` is read in the `filteredOperations` memo but is not (and cannot be) a dep, so the memo only re-evaluates the gate when another dep (`latencyRange` / `latencyPercentile` / `latencyBounds`) changes. It works today because every mutation of `.current` is paired with a `setState` that re-triggers the memo, but that coupling is implicit and easy to break. Worth a short comment noting the invariant, or promoting `latencyUserModified` to state so the gate cannot desync. Same pattern in the dependencies tab.

public/components/apm/pages/service_details/service_operations.tsx
```diff
@@ -909,15 +918,15 @@ export const ServiceOperations: React.FC<ServiceOperationsProps> = ({
                 style={{ paddingTop: '8px', paddingRight: '8px' }}
               >
                 <OperationFilterSidebar
-                  availabilityThresholds={(AVAILABILITY_THRESHOLD_OPTIONS as unknown) as string[]}
+                  availabilityThresholds={AVAILABILITY_THRESHOLD_OPTIONS as unknown as string[]}
```
ps48, line 921: ### p1: unrelated reformatting in a bugfix PR

The `(x as unknown) as string[]` to `x as unknown as string[]` change and the ternary re-indentation across both files look like prettier output unrelated to the re-render fix. If the repo's lint requires it, fine; otherwise consider reverting so the bugfix diff stays focused and easy to bisect.

ps48 (commented):
### stop service graphs re-rendering on latency filter change

Nice, well-reasoned fix, and the root-cause writeup is clear. Both parts check out: dropping `latencyPercentile` from the table `key` stops the remount (so pagination and the expanded-row charts survive a percentile switch, updated in place via the memoized `columns`), and gating `isLatencyFilterActive` on `latencyUserModified.current` removes the false-positive filter on the render right after a switch. I verified the ref logic end to end: the slider handler sets `.current = true`, the percentile `onChange` and the bounds-change effect both reset it to `false`, and the gate is applied consistently in both the filter memo and the active-filter-badge memo. The added `opLatency === undefined` guard in operations is a good catch that brings it in line with dependencies. No new dependencies, routes, saved objects, or stray `console.*`, and the two tabs stay in sync.

No `p0`. The main ask is a regression test, since this is a subtle render/filter behavior with nothing guarding it. Two minor notes inline.

Nice work.

ps48 (approved):
the comments are nits and good for follow up

PR #2848, opened 2026-08-27
Fix/notification channel picker get configs

riysaxen-amzn:
### Description

Fixes the notification channel picker in the PPL alerting rule flyout (Create/Edit logs rule) showing no channels even when channels exist on the datasource.

**Root cause**

The picker's service method only called the plugin's backend route (`/api/alerting/opensearch/{dsId}/destinations`), which relies on the OSD server-side client reaching the Notifications plugin REST API directly. In some deployments that call fails or returns empty, leaving the picker permanently blank — the browser never issues a `get_configs` request at all. The route also lives under the legacy alerting "destinations" naming; the alerting destinations APIs were removed in the 3.0 release, so nothing should be built on that path.

**Fix**

- Renamed `listDestinations` → `getNotificationConfigs` to reflect what it actually fetches.
- It now calls only `GET /api/notifications/get_configs` — the notifications-dashboards plugin's own route (the same one alerting-dashboards-plugin uses for its channel picker), which carries the correct per-environment (MDS-aware) client routing. There is no fallback to the alerting destinations route.
- The synthetic `local-cluster` / `local_cluster` ids omit the `dataSourceId` query param, matching how the notifications route addresses the local cluster.
- Errors from `get_configs` propagate directly to the picker's inline error state.

**Before:** channel dropdown empty, no `get_configs` call in the network tab.
**After:** channels load via `get_configs` (`config_type=slack|chime|webhook|email|sns|microsoft_teams`, sorted by name), same as the classic alerting UI.

Note: the now-unused server-side `/api/alerting/opensearch/{dsId}/destinations` route and `osBackend.getDestinations` are left in place to keep this diff focused; happy to remove them here or in a follow-up.

### Testing

- Unit tests for `getNotificationConfigs`: request shape and channel-type filters, `dataSourceId` omission for local-cluster ids, and that a `get_configs` failure propagates without any call to the legacy destinations route.
- Updated `use_destinations` hook tests to the renamed method; 17 tests pass across both suites.
- Verified manually against a managed OpenSearch domain: the picker now lists the configured SNS channel that previously did not appear.
Image 8-27-26 at 4 55 PM

### Issues Resolved

N/A

### Check List

### DIFF
--- a/public/components/alerting/query_services/alerting_opensearch_service.ts
+++ b/public/components/alerting/query_services/alerting_opensearch_service.ts
@@ -103,27 +103,48 @@ export class AlertingOpenSearchService {
   }
 
   /**
-   * List notification destinations for a datasource. Used by the create/edit
-   * flyout to populate the action destination picker. Returns a thin
+   * List notification channel configs for a datasource. Used by the
+   * create/edit flyout to populate the action channel picker. Returns a thin
    * summary — id, name, and type are all the picker needs.
    *
-   * The upstream alerting API isn't paginated; the server caps at size=200
-   * and surfaces `truncated`/`totalDestinations` so the UI can hint when
-   * older entries are unreachable.
+   * Calls `/api/notifications/get_configs` — the notifications-dashboards
+   * plugin's own route (the same one alerting-dashboards-plugin uses for its
+   * channel picker). It carries the correct per-environment client routing
+   * (MDS-aware). The legacy alerting destinations APIs were removed in the
+   * 3.0 release and are intentionally not used here.
    */
-  async listDestinations(dsId: string): Promise<DestinationsListResult> {
-    const resp = (await this.requireHttp().get(
-      `/api/alerting/opensearch/${encodeURIComponent(dsId)}/destinations`
-    )) as {
-      destinations: DestinationSummary[];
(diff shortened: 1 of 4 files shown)


public/components/alerting/query_services/alerting_opensearch_service.ts
```diff
@@ -112,6 +112,64 @@ export class AlertingOpenSearchService {
    * older entries are unreachable.
    */
   async listDestinations(dsId: string): Promise<DestinationsListResult> {
+    // Primary path: /api/notifications/get_configs — the notifications-dashboards
+    // plugin's own route (the same one alerting-dashboards-plugin uses for its
+    // channel picker). It carries the correct per-environment client routing
+    // (MDS-aware), so it's the most reliable source across deployments.
+    //
+    // Fallback: our backend /destinations route. Needed when get_configs is
```
lezzago: alerting destination api have been deprecated since OpenSearch 2.0 and moved over to notification channels, so we should never call the alerting destination apis.

riysaxen-amzn: addressed

lezzago (approved):

Issue #2849, opened 2026-08-27
[BUG] Service details table jumps back to page 1 when switching the latency percentile


cnoramut:
**What is the bug?**
On the APM `Service` details page, the `Operations` and `Dependencies` tabs each have a latency percentile selector (P99 / P90 / P50) that only controls which value the "Latency" column displays. Switching it while paginated snaps the table back to page 1.

**How can one reproduce the bug?**
1. Open `Services` with more than 10 Operations or Dependencies.
2. Go to the `Operations` or `Dependencies` tab.
3. In the table, navigate to any page beyond the first.
4. Change the latency selector to any other value.
5. See the table jump back to page 1, even though the row order is unchanged and only the "Latency (Pxx)" column values differ.

**What is the expected behavior?**
The table should stay on the current page when the latency percentile changes. Only the Latency column's displayed value should update and pagination position should be preserved.

**What is your host/environment?**
 - OS: MacOS Tahoe
 - Version: 3.9.0
 - Plugins: ashboards-observability (APM / Service details)

**Do you have any screenshots?**
https://github.com/user-attachments/assets/1e95853e-e55f-4b8b-bcac-4f99fb8d8ea3

**Do you have any additional context?**
Appears tied to how the table's `items` array is rebuilt when the percentile changes, which makes the table reset its page.

PR #2851, opened 2026-08-28
fix(apm): preserve table page across latency percentile switch (#2849)

cnoramut:
### Description
The service-details **Operations** and **Dependencies** tables reset to page 1 whenever the latency percentile (P50/P90/P99) is switched while on page 2+.

Root cause: switching the percentile rebuilds the filtered-rows array (`filteredOperations`/`filteredDependencies`) as a new array reference, and `EuiInMemoryTable` resets an *uncontrolled* `pageIndex` to 0 on any change to the `items` reference.

Fix: make pagination controlled - `pageIndex` and `pageSize` are lifted into component state and updated via `onTableChange`, so the current page and page size survive a percentile switch. The page index is additionally clamped during render, so a filter that shrinks the row count below the current page never lands the table on an empty, out-of-range page.

#### Before

https://github.com/user-attachments/assets/4493f161-bc57-4e04-9c00-7df5b42334e6

#### After

https://github.com/user-attachments/assets/66dea5f1-7693-469a-9417-1857df599959

### Issues Resolved
[2849](https://github.com/opensearch-project/dashboards-observability/issues/2849)

### Check List

### DIFF
--- a/public/components/apm/pages/service_details/__tests__/service_dependencies.test.tsx
+++ b/public/components/apm/pages/service_details/__tests__/service_dependencies.test.tsx
@@ -85,11 +85,10 @@ const props = {
   serviceMapDataset: 'otel-v1-apm-span',
 };
 
-// EuiSuperSelect is not a native <select> (and the dependencies one has no data-test-subj):
-// click the control to open the popover, then click the option.
+// EuiSuperSelect is not a native <select>: click the control to open the popover, then click
+// the option. Targeting the stable data-test-subj avoids grabbing an unrelated EuiSuperSelect.
 const switchPercentile = (label: 'P99' | 'P90' | 'P50') => {
-  const control = document.querySelector('.euiSuperSelectControl') as HTMLElement;
-  fireEvent.click(control);
+  fireEvent.click(screen.getByTestId('dependencyLatencyPercentileSelector'));
   const listbox = screen.getByRole('listbox');
   fireEvent.click(within(listbox).getByText(label));
 };
--- a/public/components/apm/pages/service_details/__tests__/service_dependencies.test.tsx
+++ b/public/components/apm/pages/service_details/__tests__/service_dependencies.test.tsx
@@ -1,0 +1,157 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import React from 'react';
+import { fireEvent, render, screen, waitFor, within } from '@testing-library/react';
+
+// Counts expanded-row chart mounts to detect a table remount on a percentile switch
+// (the pre-#2626 bug). See service_operations.test.tsx for the full rationale.
+const mockChartState = { mounts: 0 };
+
+jest.mock('../../../shared/hooks/use_dependencies', () => ({ useDependencies: jest.fn() }));
+jest.mock('../../../shared/hooks/use_dependency_metrics', () => ({
+  useDependencyMetrics: jest.fn(),
+}));
+jest.mock('../../../shared/hooks/use_chart_step_window', () => ({
+  useChartStepWindow: () => ({ window: '1m' }),
+}));
+jest.mock('../../../shared/hooks/use_debounced_value', () => ({
+  useDebouncedValue: (value: unknown) => value,
+}));
+jest.mock('../../../shared/components/dependency_filter_sidebar', () => ({
+  DependencyFilterSidebar: () => <div data-test-subj="dependencyFilterSidebar" />,
+}));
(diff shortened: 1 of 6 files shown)


TackAdam:
Nice fix on the core bug — the keying-vs-controlled tradeoff is well reasoned, and the inline comments explaining why pagination is controlled are appreciated. One behavioral refinement and a few smaller items below.

## 1. Pagination: preserve across percentile switch, but reset to page 1 on filter changes

Right now the controlled-pagination mechanism treats every `items` change identically (preserve `pageIndex`, then clamp). But there are really two distinct triggers, and they want opposite behavior:

- **Latency percentile switch** (the bug this PR targets) — same rows, just a different latency column. Preserving the page is correct. ✅
- **Filter changes** (search box, sidebar thresholds, latency/requests range sliders, operation selection) — these produce a new result set, and the user expects to land on page 1 to see the top of the filtered results. Currently they stay on whatever page they were on.

Before this PR, filtering reset to page 1 for free: pagination was uncontrolled, so EUI's `getDerivedStateFromProps` did `nextPageIndex = pagination?.pageIndex || 0` on every `items` change → `0`. Now that `pageIndex` is controlled, that same code re-applies the stored index instead, so a new search/filter leaves you on your old page.

**Suggested approach:** encode the intent explicitly — reset `pageIndex` to `0` inside the filter handlers, and deliberately not in the percentile `onChange`:

- `service_operations.tsx`: search `onChange` (`:898`), `setSelectedOperations`/threshold setters, `onLatencyRangeChange` (`:252`), `onRequestsRangeChange` (`:257`), `handleClearAllFilters` (`:561`) — leave the percentile `onChange` at `:909` untouched.
- `service_dependencies.tsx`: the equivalent handlers.

> ⚠️ **Trap to avoid:** Don't do this with a `useEffect` keyed on `latencyRange`/`requestsRange`. Switching the percentile recomputes `latencyBounds`, and the effect at `service_operations.tsx:389-407` resets `latencyRange` to the new bounds — so an effect watching `latencyRange` would fire a page-reset on percentile switch too, defeating the fix. Resetting inside explicit user-action handlers sidesteps that.

Keep the `clampedPageIndex` clamp (`service_operations.tsx:466`, `service_dependencies.tsx:526`) — with filters resetting to page 1, its remaining job is the one path that's neither a filter nor a percentile switch: a background data refetch (`refreshTrigger`) that returns fewer rows while you're on a high page.

## 2. Stale `pageIndex` can resurrect an out-of-range page

Related to the above and worth calling out explicitly: `clampedPageIndex` is used only for display and never written back to state. So if a filter shrinks the list, the clamp shows a valid page but state keeps the old high index; when the list grows again the old index re-applies and jumps the user.

Adopting the reset-on-filter behavior in 1 dissolves this — filters reset to `0`, so there's no stale index to resurrect. This affects both files.

## 3. Duplication between the two files

The entire controlled-pagination mechanism — `pageIndex`/`pageSize` state + `onTableChange` (`service_operations.tsx:218-226`, `service_dependencies.tsx:252-259`) and the clamp (`:466-467` / `:526-527`) — is byte-for-byte identical apart from the generic type.

Extracting a small `useControlledPagination(filteredCount)` hook returning `{ pageIndex, pageSize, onTableChange }` (and, once 1 lands, the reset helper) would keep the two tables from drifting and mean the reset-on-filter logic is implemented and tested once.

## 4. Fragile percentile selector in the dependencies test

`service_dependencies.test.tsx:91` selects the control via `document.querySelector('.euiSuperSelectControl')`, which grabs the first `EuiSuperSelect` on the page. It only passes today because `DependencyFilterSidebar` is mocked to a bare `div` (`:24`). If another `EuiSuperSelect` ever renders, it silently drives the wrong control.

The operations test targets a stable `data-test-subj="latencyPercentileSelector"`. The dependencies `EuiSuperSelect` (`service_dependencies.tsx:973-982`) has no `data-test-subj`; adding one and switching the test to it would make it robust.

cnoramut:
Thanks @TackAdam for the review, all four are addressed

1. Filter changes reset to page 1, percentile switch keeps the page. One effect handles the discrete filters and their badge removals, the range sliders and clear-all reset directly. The effect does not watch `latencyRange`, so a percentile switch never resets the page, with a comment saying why.
2. Also closed the case your dissolve did not cover, a refetch that shrinks then grows the rows. The hook persists the clamp during render so the stale page cannot come back.
3. Extracted `useControlledPagination`, shared by both tables.
4. Added a stable `data-test-subj` to the dependencies latency select and retargeted the test off the `.euiSuperSelectControl` query.

Tests cover reset-on-filter on both tabs and the grow-back clamp, plus the existing percentile tests.

TackAdam (approved):

PR #2855, opened 2026-09-01
fix(observability): surface the missing permission on ruler 403 create failures

### LINKED ISSUE #2814: [BUG] Errors are not surfaced to the user when creating metric rules / SLOs
**What is the bug?**
When creating metric rules or SLOs fails, the actual underlying error is not surfaced to the user. The failure message is generic and does not convey what actually went wrong, leaving users unable to understand or self-diagnose the problem. For example, a failure caused by a missing "create rule group namespace" permission surfaces only as a generic error with no indication of the real cause — but the problem is broader than any single error type: underlying errors in general are not being surfaced.

1. Attempt to create a metric rule or an SLO under a condition that causes the backend request to fail (for example, missing the "create rule group namespace" permission on the Prometheus datasource, or another backend error).
2. Observe the creation fails.
3. See that the error shown to the user is generic and does not reflect the actual underlying error returned by the backend.

The UI should surface the actual underlying error returned by the backend so the user understands what went wrong and can act on it. Error messages should be specific and actionable rather than generic — for permission failures this includes indicating the missing permission, and more generally the real cause of any failure should be shown.

Found during testing of metric rule / SLO creation. The create operation itself works when the request succeeds — this is specifically about surfacing failures clearly. Generic/opaque error messaging here caused significant confusion during testing (e.g. a permission error that was not identifiable as such).

cnoramut:
### Description
The ruler `RULER_AUTH_FAILED` branch previously returned no `details` for either 401 or 403, so a 403 on metric rule / SLO create surfaced only a generic toast that never told the user which permission was missing.

This change attaches `rawDetails(ctx.rawBody)` to the 403 branch, the same redaction and sensitive-stripping path the `RULER_VALIDATION_FAILED` branch already uses. The redacted `safe` detail keeps the actionable permission string, for example `no permissions for [cluster:admin/opensearch/directquery/write]`, while URLs, hosts, IPs, and token-shaped content are scrubbed by `redactForDisplay`. The verbatim `sensitive` copy is dropped at the client boundary unless the operator sets `observability.errors.exposeSensitiveErrorDetail`, which defaults to `false`.

The 401 branch stays detail-free on purpose. A 401 means the credential itself was rejected, which is the response most likely to echo a bad token or header, so no upstream body is surfaced there.

### Issues Resolved
Fixes [#2814](https://github.com/opensearch-project/dashboards-observability/issues/2814)
Builds on [#2819](https://github.com/opensearch-project/dashboards-observability/pull/2819)

### Check List

### DIFF
--- a/common/error/__tests__/classifiers.test.ts
+++ b/common/error/__tests__/classifiers.test.ts
@@ -110,11 +110,22 @@ describe('default classifiers — coverage table', () => {
       operation: 'rule.create.metric',
       upstreamCode: 'RULER_AUTH_FAILED',
       httpStatus: 403,
+      rawBody:
+        'no permissions for [cluster:admin/opensearch/directquery/write] and User [name=test] at https://host.internal:9090',
     });
     expect(forbidden.code).toBe('PERMISSION_DENIED');
 
-    // Auth failures carry no raw upstream detail.
+    // 401 re-auth carries no raw upstream detail.
     expect(unauthorized.details).toBeUndefined();
+
+    // 403 surfaces the missing permission (redacted) so the user can
+    // self-diagnose, while topology (host/URL) is still scrubbed.
+    const safe = forbidden.details?.find((d) => d.sensitivity === 'safe');
+    expect(safe?.value).toContain(
+      'no permissions for [cluster:admin/opensearch/directquery/write]'
+    );
+    expect(safe?.value).not.toContain('host.internal');
+    expect(safe?.value).toContain('<redacted-url>');
   });
 
   it('UNKNOWN: unmatched error still surfaces a redacted message + keeps raw sensitive', () => {
--- a/common/error/classifiers/ruler.ts
+++ b/common/error/classifiers/ruler.ts
@@ -31,20 +31,23 @@ export const rulerClassifier: ErrorClassifier = {
           details: rawDetails(ctx.rawBody),
         };
       case 'RULER_AUTH_FAILED':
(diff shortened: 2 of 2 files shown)


TackAdam:
The `401` branch correctly carries no details, but the `403` branch attaches `rawDetails(ctx.rawBody)`, which emits two details: the sensitive verbatim body (stripped by default — good) and a safe redacted excerpt that's shown by default. That safe excerpt is the concern.

`redactForDisplay` is explicitly documented as **"best-effort defense-in-depth, not a security boundary,"** and a `401`/`403` body is exactly the kind of content it wasn't built to sanitize.

Concrete shapes that slip through today:

- Short / all-letter credentials (`TOKEN_PATTERN` requires length ≥24 + a letter and a digit), e.g. `Authorization: Basic dGVzdHVzZXI`
- Bare single-label hosts without a port, e.g. `... on node opensearch-node-1`
- The RBAC principal `User [name=...]`, which on a server-side dual-write is the Dashboards service/proxy account, not the caller

Since this branch previously returned no details, the change opens a default-shown exposure path gated only by that best-effort redactor.

## Proposal

For `403`, emit only the sensitive raw detail and skip the default-shown safe excerpt.

The missing-permission diagnostic stays retrievable via the explicit `exposeSensitiveErrorDetail` opt-in and server logs, but nothing derived from the upstream auth body reaches the browser by default.

If we'd rather not maintain a partial-detail path, matching the `401` branch (no details) is the simpler fallback.

PR #2856, opened 2026-09-01
fix(apm): promote range-filter gate flags from refs to state + add re-render regression tests

cnoramut:
### Description

Follow up to the review on [#2842 ](https://github.com/opensearch-project/dashboards-observability/pull/2842) after it merged. 

- Promote the range-filter gate flags from refs to state. `latencyUserModified` and `requestsUserModified` on the operations and dependencies tabs, and `latencyUserModified` and `throughputUserModified` on the services home page, were refs read inside the filter and `activeFilters` `useMemo` bodies. A ref is not a valid memo dependency, so the gate could desync from the flag. They are now `useState` and listed in the relevant memo dependency arrays. Behavior is unchanged, since every write was already paired with a `setState` in the same handler. This removes the ref-in-memo anti-pattern so the gate cannot desync if a dependency's memoization changes later.
- Add regression tests for the percentile re-render fix. The tests assert that switching the latency percentile does not remount the operations and dependencies tables, so pagination and expanded-row charts are preserved and not re-fetched, and that a percentile switch clears the latency range filter and re-engages it when the slider moves again.

### Issues Resolved
Follow up to review comments on [#2842](https://github.com/opensearch-project/dashboards-observability/pull/2842). Related to [#2626](https://github.com/opensearch-project/dashboards-observability/issue/2626) 

### Check List

### DIFF
--- a/public/components/apm/pages/service_details/service_dependencies.tsx
+++ b/public/components/apm/pages/service_details/service_dependencies.tsx
@@ -243,8 +243,10 @@ export const ServiceDependencies: React.FC<ServiceDependenciesProps> = ({
   // Range filter states
   const [latencyRange, setLatencyRange] = useState<[number, number]>([0, 0]);
   const [requestsRange, setRequestsRange] = useState<[number, number]>([0, 0]);
-  const latencyUserModified = useRef(false);
-  const requestsUserModified = useRef(false);
+  // Track whether the user has explicitly interacted with range filters.
+  // Use as state, so the memos that read them re-run when a flag flips.
+  const [latencyUserModified, setLatencyUserModified] = useState(false);
+  const [requestsUserModified, setRequestsUserModified] = useState(false);
 
   // Parse time range
   const parsedTimeRange = useMemo(() => {
@@ -421,7 +423,7 @@ export const ServiceDependencies: React.FC<ServiceDependenciesProps> = ({
   // Step 3: Reset slider ranges when bounds change
   // Use functional update to prevent unnecessary re-renders when values haven't changed
   useEffect(() => {
-    latencyUserModified.current = false;
+    setLatencyUserModified(false);
     setLatencyRange((prev) => {
       if (prev[0] === latencyBounds.min && prev[1] === latencyBounds.max) {
         return prev; // Return same reference to avoid re-render
@@ -431,7 +433,7 @@ export const ServiceDependencies: React.FC<ServiceDependenciesProps> = ({
   }, [latencyBounds.min, latencyBounds.max]);
 
   useEffect(() => {
-    requestsUserModified.current = false;
+    setRequestsUserModified(false);
     setRequestsRange((prev) => {
       if (prev[0] === requestsBounds.min && prev[1] === requestsBounds.max) {
         return prev; // Return same reference to avoid re-render
@@ -458,7 +460,7 @@ export const ServiceDependencies: React.FC<ServiceDependenciesProps> = ({
       // Latency range filter (only if range has been adjusted)
       const isLatencyFilterActive =
         // Gating on latencyUserModified avoids a false positive after a percentile switch
-        latencyUserModified.current &&
+        latencyUserModified &&
         (latencyRange[0] > latencyBounds.min || latencyRange[1] < latencyBounds.max);
       if (isLatencyFilterActive) {
         // Use the selected percentile's duration for filtering
@@ -491,6 +493,7 @@ export const ServiceDependencies: React.FC<ServiceDependenciesProps> = ({
     });
   }, [
     textFilteredDependencies,
+    latencyUserModified,
     latencyRange,
     requestsRange,
     latencyBounds,
@@ -507,7 +510,7 @@ export const ServiceDependencies: React.FC<ServiceDependenciesProps> = ({
   // Range-slider changes reset the page to 1 (a filter change), unlike a percentile switch.
   const onLatencyRangeChange = useCallback(
     (val: [number, number]) => {
-      latencyUserModified.current = true;
+      setLatencyUserModified(true);
       setLatencyRange(val);
       resetPage();
     },
@@ -516,7 +519,7 @@ export const ServiceDependencies: React.FC<ServiceDependenciesProps> = ({
 
   const onRequestsRangeChange = useCallback(
     (val: [number, number]) => {
-      requestsUserModified.current = true;
+      setRequestsUserModified(true);
       setRequestsRange(val);
       resetPage();
     },
@@ -611,7 +614,7 @@ export const ServiceDependencies: React.FC<ServiceDependenciesProps> = ({
 
     // Latency range filter badge (only if user has interacted and modified from default bounds)
     const isLatencyModified =
-      latencyUserModified.current &&
+      latencyUserModified &&
       (latencyRange[0] > latencyBounds.min || latencyRange[1] < latencyBounds.max);
     if (isLatencyModified) {
       badges.push({
@@ -621,7 +624,7 @@ export const ServiceDependencies: React.FC<ServiceDependenciesProps> = ({
         }),
         values: [`${latencyRange[0].toFixed(0)}-${latencyRange[1].toFixed(0)}ms`],
         onRemove: () => {
-          latencyUserModified.current = false;
+          setLatencyUserModified(false);
           setLatencyRange([latencyBounds.min, latencyBounds.max]);
           resetPage();
         },
@@ -630,7 +633,7 @@ export const ServiceDependencies: React.FC<ServiceDependenciesProps> = ({
 
     // Requests range filter badge (only if user has interacted and modified from default bounds)
     const isRequestsModified =
-      requestsUserModified.current &&
+      requestsUserModified &&
       (requestsRange[0] > requestsBounds.min || requestsRange[1] < requestsBounds.max);
     if (isRequestsModified) {
       badges.push({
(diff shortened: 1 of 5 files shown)


TackAdam:
Nice fix on the ref→state promotion — the badge memos now correctly re-run when the gate flips, and the dependency arrays / callback identities all check out. One thing to finish before merge: the `userModified` gate is now applied to the latency filter and to the badges, but the requests/throughput filters are still ungated. That leaves the exact filter↔badge asymmetry this PR set out to remove, and it produces a one-frame empty-table flicker on load. All three are the same class of issue:

## 1. `service_dependencies.tsx:483` — requests filter not gated

    const isRequestsFilterActive =
      requestsRange[0] > requestsBounds.min || requestsRange[1] < requestsBounds.max;

Unlike the latency filter right above it (which gates on `latencyUserModified`), this runs during the initial-load / bounds-change transient. On first data load `requestsBounds` becomes e.g. `[0,100]` in the same render while `requestsRange` is still `[0,0]`, so `isRequestsFilterActive` is `true` and every row with `requestCount !== 0` is filtered out for one painted frame until the bounds `useEffect` resets `requestsRange`.

**Suggested:**

    const isRequestsFilterActive =
      requestsUserModified &&
      (requestsRange[0] > requestsBounds.min || requestsRange[1] < requestsBounds.max);

And add `requestsUserModified` to the memo's dependency array.

## 2. `service_operations.tsx:424` — same, plus a live badge mismatch

    const isRequestsFilterActive =
      requestsRange[0] > requestsBounds.min || requestsRange[1] < requestsBounds.max;

Same `[0,0]` transient as above. Worse here: the requests badge is already gated on `requestsUserModified`, so during that window (and any time the gate is `false`) the filter can hide rows while no "requests" badge is shown — the precise inconsistency this PR is fixing for the other dimensions.

Apply the same `requestsUserModified &&` gate and add it to the deps array.

## 3. `services_home.tsx:670,683` — neither latency nor throughput filter gated

    const isLatencyFilterActive =
      latencyRange[0] > metricRanges.latencyMin || latencyRange[1] < metricRanges.latencyMax;

    ...

    const isThroughputFilterActive =
      throughputRange[0] > metricRanges.throughputMin ||
      throughputRange[1] < metricRanges.throughputMax;

`displayedServices` gates neither filter on the `userModified` flags, while the corresponding badges do. On hydration `latencyRange` / `throughputRange` start at `[0,0]`; once `metricRanges` computes non-zero maxes the filters go active and drop every service until the `metricRanges` `useEffect` (line 653) resets the ranges → empty/flickering list. And as in 2, a filter can hide services with no active-filter badge.

**Suggested:**

    const isLatencyFilterActive =
      latencyUserModified &&
      (latencyRange[0] > metricRanges.latencyMin || latencyRange[1] < metricRanges.latencyMax);

    const isThroughputFilterActive =
      throughputUserModified &&
      (throughputRange[0] > metricRanges.throughputMin ||
        throughputRange[1] < metricRanges.throughputMax);

And add `latencyUserModified` / `throughputUserModified` to `displayedServices`' dependency array.

## Why this belongs in this PR

These pre-date the change, but the PR promotes the gate flags to state and applies them symmetrically to the badge memos — leaving the filter memos asymmetric is the one gap that keeps the filter/badge state out of sync and keeps the load-time flicker.

Gating the filters the same way finishes the fix. Worth a quick test that asserts the table isn't empty on first render while bounds/ranges are still settling.

cnoramut:
Thanks, all three addressed. Gated each filter memo on its `*UserModified` flag to match the badges and added the flags to the dependency arrays:

- `service_dependencies.tsx` requests filter now gates on `requestsUserModified`
- `service_operations.tsx` requests filter now gates on `requestsUserModified`
- `services_home.tsx` `displayedServices` gates latency on `latencyUserModified` and throughput on `throughputUserModified`

This removes the filter/badge asymmetry and the one-frame empty-table flicker on load, where the raw range-vs-bounds check read active while the range was still settling from `[0,0]`.

Tests: added a `React.Profiler` regression test to the operations and dependencies specs asserting the table shows rows on every commit during the bounds/range settle. A profiler is needed because `render()` flushes the settling effect before returning, which hides the transient from a post-render query. Confirmed it goes red without the gate.

TackAdam (approved):

public/components/apm/pages/services_home/services_home.tsx
```diff
@@ -652,8 +653,8 @@ export const ServicesHome: React.FC<ServicesHomeProps> = ({
   useEffect(() => {
     setLatencyRange([metricRanges.latencyMin, metricRanges.latencyMax]);
     setThroughputRange([metricRanges.throughputMin, metricRanges.throughputMax]);
-    latencyUserModified.current = false;
-    throughputUserModified.current = false;
+    setLatencyUserModified(false);
```
ps48, line 656: Not introduced by this PR, but worth flagging since it is the same symptom class: this effect resets `latencyUserModified` / `throughputUserModified` (and the ranges) on every `metricRanges` change, not just on first paint. At high service cardinality metrics populate in several waves, so a filter the user applies while metrics are still streaming in can be silently cleared by the next wave: the selection/badge disappears with no user action. The gating fixes the first-paint transient but not this later-wave wipe. Consider deriving the ranges only until metrics settle (or not resetting the flags once the user has interacted) so an in-progress selection survives subsequent metric updates.

public/components/apm/pages/services_home/services_home.tsx
```diff
@@ -652,8 +653,8 @@ export const ServicesHome: React.FC<ServicesHomeProps> = ({
   useEffect(() => {
     setLatencyRange([metricRanges.latencyMin, metricRanges.latencyMax]);
```
ps48, line 654: Minor / consistency: this always allocates fresh `[min, max]` array references, whereas `service_operations.tsx` and `service_dependencies.tsx` (edited in this same PR) use a functional update with a reference-equality guard so they bail out when the numeric bounds are unchanged. Because this one always sets new refs, `displayedServices` (and the table) recompute on every metrics wave even when the rounded bounds did not change. Aligning this with the sibling files would avoid the extra re-renders.

public/components/apm/pages/service_details/service_operations.tsx
```diff
@@ -925,7 +932,7 @@ export const ServiceOperations: React.FC<ServiceOperationsProps> = ({
             options={LATENCY_OPTIONS}
             valueOfSelected={latencyPercentile}
             onChange={(value) => {
-              latencyUserModified.current = false;
+              setLatencyUserModified(false);
```
ps48, line 936: Nit (pre-existing, latent): this sets `latencyUserModified` false but leaves resetting `latencyRange` to the bounds effect keyed on `[latencyBounds.min, latencyBounds.max]`. If the new percentile's rounded bounds happen to equal the previous ones, that effect's deps do not change and a previously narrowed `latencyRange` lingers in state. It is masked today because the flag is false (filter gated off), but the "range == full bounds after a percentile switch" invariant is technically violated. A direct `setLatencyRange(fullBounds)` here would make it robust.

public/components/apm/pages/service_details/service_dependencies.tsx
```diff
@@ -993,7 +1000,7 @@ export const ServiceDependencies: React.FC<ServiceDependenciesProps> = ({
             options={LATENCY_OPTIONS}
             valueOfSelected={latencyPercentile}
             onChange={(value) => {
-              latencyUserModified.current = false;
+              setLatencyUserModified(false);
```
ps48, line 1004: Same as the operations page: `setLatencyUserModified(false)` here relies on the bounds effect to reset `latencyRange`, which will not fire if the new percentile's rounded bounds match the old, leaving a stale narrowed range in state (masked while the flag is false). Resetting the range directly here would close the gap.

ps48 (commented):
Reviewed the full change (both page components and the two new test files) at head 340f419. Overall this is a clean, well-scoped follow-up to #2842: the ref -> useState conversion is complete (no lingering `.current`), the new tests use the real sidebar prop names so they genuinely exercise the gating, and it legitimately removes several on-load filter-blanking transients. I did not find any newly-introduced correctness bug, so this looks good to merge. The inline notes are mostly pre-existing issues in the touched functions rather than regressions.

Two things worth a look:

- The `[metricRanges]` effect on services_home still clears the user-modified flags on every metrics wave, so the on-load blanking can still recur at high service cardinality when metrics stream in over time (inline comment). That is the same symptom class this PR targets, so it may be worth folding in.
- Since these flags moved from refs into `useState` and into dependency arrays, could you confirm there is no extra re-render/refetch churn at scale (for example a quick network-panel check on the services table)? Keeping them as refs previously kept them out of the memo deps.

Heads up on overlap: #2860 (open) edits these same three files and the same `*UserModified` gating as part of the scale-hardening work. Whichever lands first, the other will need a rebase on this gating logic, so flagging it here so we can sequence the two and avoid a surprise conflict. I am happy to rebase #2860 onto this once it merges.

public/components/apm/pages/service_details/service_operations.tsx
cnoramut: Fixed. The handler now calls `setLatencyRange([latencyBounds.min, latencyBounds.max])` directly. Verified with a fixture where p99 and p90 round to the same `{min, max}` so the bounds effect never fires. Without the fix the range stays narrowed.

public/components/apm/pages/service_details/service_dependencies.tsx
cnoramut: Fixed. Same as above.

ps48 (approved):

PR #2860, opened 2026-09-03
[APM] Harden Services, Overview, Operations, Dependencies, and Topology pages for scale

ps48:
## Summary

This PR hardens all five APM pages (Services home, Service overview, Service operations, Service dependencies, Topology map) at high service and environment cardinality (hundreds of services across many environments), where query construction, request concurrency, and rendering issues surface. The changes came from an audit of every PromQL and PPL query builder plus the data hooks and page components.

### 1. Oversized PromQL queries
The Services home and Topology pages built a `service=~"a|b|c|..."` filter from the entire service list and embedded it into each metric query. The failure rate query embedded it three times, so at a few hundred services it crossed the datasource query-length limit and returned HTTP 400. That blanked the failure rate column, and because throughput shared an all-or-nothing `Promise.all` with it, throughput went blank too.

Fix: rely on the existing `sum by (service)` grouping instead of a service list filter, so query length is constant regardless of fleet size. The service filter is now optional and omitted when empty.

### 2. All-or-nothing metric fetching
Services home, Operations, Dependencies, and Topology fetched their metric queries with `Promise.all`, so one failing query blanked every column, not just its own.

Fix: `Promise.allSettled` with per-metric handling, so a single failure only affects its own column.

### 3. Racy renders and stale responses
No APM data hook cancelled or guarded in-flight requests, so a response from a superseded refresh or time range change could overwrite fresh state.

Fix: AbortController based stale-response guards in every data hook (matching the existing pattern in `use_apm_config.ts`), with an optional `AbortSignal` threaded through the PromQL and PPL search services. Also fixed a Topology refresh that fetched the map twice.

The related "catalog collapses to a single service on refresh" filter-blanking is fixed upstream in #2856, which gates the range-filter sync on explicit user interaction; this PR builds on that rather than re-implementing it. Because the visible-page sparkline loading (item 8) gives `metricsMap` a new identity as each page's series arrive, this PR also makes that range-sync effect depend on the primitive metric bounds, so paging or sorting no longer resets an active latency/throughput filter.

### 4. Query correctness and safety
- Application and group builders now filter `remoteService=""` so server span metrics are not double counted with client spans, and latency histograms are not mixed across span kinds.
- Added PromQL label, PromQL regex, and PPL string escape helpers and applied them to interpolated service, environment, operation, and remote values.
- Added a row cap to the unbounded PPL list and service map queries, and removed two dead PPL builders.
- Reconciled the environment value passed to the Operations list and metrics hooks.

### 5. Multi-environment metrics and failure-rate accuracy
- Services home metric queries grouped by service only, so a service running in multiple environments had its throughput, latency, and failure rate summed into one value shown on every environment row. The catalog builders now group by `(environment, service)` and the metrics map keys on both.
- The catalog failure rate averaged the per-step ratio (over-weighting low-traffic steps); it now uses a ratio of summed totals over the range.

### 6. Service overview P99 latency card
The P99 card averaged per-step P99 values (labelled "Avg"), which is not meaningful for a percentile. It now shows the P99 over the selected range and drops the misleading subtitle.

### 7. Topology rendering
The graph rebuilt all nodes and ran a full dagre relayout on every metrics tick and every edge selection.

Fix: memoize the structural layout separately from the metric and selection overlays so a metrics update or edge click no longer triggers relayout, plus a node cap with a clear notice above it.

### 8. Catalog sparklines fetched per visible page
The per-step range queries that back the row sparklines were fetched for every service, even though sparklines are only shown for the rows on the current page. Instant metrics (numbers, sort, filter) still load for all services; the range/sparkline queries now run only for the visible page via a bounded, escaped `service=~` filter.

The visible-page fetch is cached and incremental: revisiting a page whose series are already in memory issues no new request, only the not-yet-fetched services on a page are queried, page changes are debounced so skimming past pages does not fire a request per intermediate page, and superseded fetches are aborted. A time range, percentile, or refresh change clears the cache so no stale series remain.

## Follow ups (intentionally not in this PR)
These were found in the audit but need product decisions or live data validation and are better as separate changes: aligning the throughput unit across all surfaces, windowing the remaining Service overview RED cards over the selected range, and reconciling the availability (5xx only) vs failure ratio (4xx and 5xx) definitions.

## Testing
- Built and ran the plugin against a local OpenSearch stack fed with synthetic services (hundreds of services across several environments).
- Verified with Chrome DevTools network capture that the Services page loads at that scale, the previously failing queries are now small and correctly formed and stay under the datasource query-length limit, and per-environment rows show distinct values.
- Verified the sparkline fetch behaviour on the network panel: page load fetches instant metrics for all services plus range queries scoped to the visible page; paging fetches only the new page's range queries; revisiting a cached page fetches nothing; skimming across several pages fetches only the page that is landed on; and a refresh clears the cache and refetches.
- Retested end to end after rebasing on latest main (catalog loads at scale with populated per-environment columns, no oversized-query failures, pagination sparklines scoped correctly).
- Updated the metric-hook unit tests for the instant-vs-visible-page fetch model, the `(service, environment)` map keying, and the `Promise.allSettled` partial-failure behaviour; added unit tests for the PromQL/PPL escape helpers. eslint clean on all changed files.

## Notes
- Rebased on latest main. This PR now relies on #2856's user-modified range-filter gating; the earlier ref-based gating that was here was dropped to avoid duplicating it.
- No user facing strings changed beyond the Topology "too many services" notice and the Service overview P99 card subtitle.

### DIFF
--- a/public/components/apm/common/constants.ts
+++ b/public/components/apm/common/constants.ts
@@ -348,6 +348,15 @@ export const APPLICATION_MAP_CONSTANTS = {
 
   /** Map container minimum height */
   MAP_MIN_HEIGHT: 500,
+
+  /**
+   * Max nodes rendered in the Services / groupByValue graph. Above this the
+   * graph switches to a "narrow your selection" notice instead of laying out
+   * the full graph (dagre layout is O(N+E)). Set as a pathological-case guard;
+   * layout only recomputes on structural change now, so typical large fleets
+   * (a few hundred services) still render.
+   */
+  MAX_RENDERED_NODES: 500,
 } as const;
 
 // Platform utility functions moved to shared/utils/platform_utils.ts
--- a/public/components/apm/common/types/prometheus_types.ts
+++ b/public/components/apm/common/types/prometheus_types.ts
@@ -52,10 +52,12 @@ export interface ExecuteMetricRequestParams {
   startTime: number;
   endTime: number;
   step?: number; // Step interval in seconds; if omitted, OSD core calculates it
+  signal?: AbortSignal; // Optional: cancels the in-flight request when aborted
 }
 
 // ExecuteInstantQuery params - for point-in-time evaluation
 export interface ExecuteInstantQueryParams {
   query: string;
   time: number; // Unix epoch in seconds
+  signal?: AbortSignal; // Optional: cancels the in-flight request when aborted
 }
--- a/public/components/apm/pages/application_map/application_map_page.tsx
+++ b/public/components/apm/pages/application_map/application_map_page.tsx
@@ -229,7 +229,6 @@ export const ApplicationMapPage: React.FC<ApplicationMapPageProps> = ({
     isLoading: mapLoading,
     error: mapError,
     availableGroupByAttributes,
-    refetch: refetchMap,
   } = useServiceMap({
     startTime: parsedTimeRange.startTime,
     endTime: parsedTimeRange.endTime,
@@ -351,12 +350,13 @@ export const ApplicationMapPage: React.FC<ApplicationMapPageProps> = ({
 
   // Handle refresh
   const handleRefresh = useCallback(() => {
+    // Map refetches via the refreshTrigger prop; the metrics hook only exposes
+    // refetch(). Bumping the trigger AND calling refetchMap() double-fetched the map.
     setRefreshTrigger((prev) => prev + 1);
-    refetchMap();
     refetchMetrics();
     // Clear selected edge on refresh
     setSelectedEdge(null);
-  }, [refetchMap, refetchMetrics]);
+  }, [refetchMetrics]);
 
   // Handle filter changes
   const handleFiltersChange = useCallback((newFilters: ApplicationMapFilters) => {
--- a/public/components/apm/pages/service_details/service_dependencies.tsx
+++ b/public/components/apm/pages/service_details/service_dependencies.tsx
@@ -418,27 +418,28 @@ export const ServiceDependencies: React.FC<ServiceDependenciesProps> = ({
     return { min, max: Math.max(max, min + 1) };
   }, [textFilteredDependencies]);
 
-  // Step 3: Reset slider ranges when bounds change
-  // Use functional update to prevent unnecessary re-renders when values haven't changed
+  // Step 3: Track slider ranges to the metric bounds only while the user has not
+  // adjusted them, and only once metrics have settled. Never clears the
+  // user-modified flags, so an active filter survives refreshes and partial loads.
   useEffect(() => {
-    latencyUserModified.current = false;
+    if (metricsLoading || latencyUserModified.current) return;
     setLatencyRange((prev) => {
       if (prev[0] === latencyBounds.min && prev[1] === latencyBounds.max) {
         return prev; // Return same reference to avoid re-render
       }
       return [latencyBounds.min, latencyBounds.max];
     });
-  }, [latencyBounds.min, latencyBounds.max]);
+  }, [latencyBounds.min, latencyBounds.max, metricsLoading]);
 
   useEffect(() => {
-    requestsUserModified.current = false;
+    if (metricsLoading || requestsUserModified.current) return;
     setRequestsRange((prev) => {
       if (prev[0] === requestsBounds.min && prev[1] === requestsBounds.max) {
         return prev; // Return same reference to avoid re-render
       }
       return [requestsBounds.min, requestsBounds.max];
     });
-  }, [requestsBounds.min, requestsBounds.max]);
+  }, [requestsBounds.min, requestsBounds.max, metricsLoading]);
 
   // Cleanup expanded rows when dependencies list changes to remove stale entries
   useEffect(() => {
@@ -477,9 +478,11 @@ export const ServiceDependencies: React.FC<ServiceDependenciesProps> = ({
         }
       }
 
-      // Requests range filter (only if range has been adjusted)
+      // Requests range filter (only if the user actually adjusted the slider;
+      // deriving activity from live bounds misfires during partial metric loads)
       const isRequestsFilterActive =
-        requestsRange[0] > requestsBounds.min || requestsRange[1] < requestsBounds.max;
+        requestsUserModified.current &&
+        (requestsRange[0] > requestsBounds.min || requestsRange[1] < requestsBounds.max);
       if (isRequestsFilterActive) {
         const depRequestCount = dep.requestCount ?? 0;
         if (depRequestCount < requestsRange[0] || depRequestCount > requestsRange[1]) {
--- a/public/components/apm/pages/service_details/service_operations.tsx
+++ b/public/components/apm/pages/service_details/service_operations.tsx
@@ -253,6 +253,10 @@ export const ServiceOperations: React.FC<ServiceOperationsProps> = ({
     }
   }, [timeRange]);
 
+  // The operations list (PPL) and the metrics (PromQL) must query the same
+  // environment, otherwise rows and metrics line up under different env values.
+  const normalizedEnvironment = environment || 'generic:default';
+
   // Fetch operations list from PPL
   const {
     data: operationsData,
@@ -260,7 +264,7 @@ export const ServiceOperations: React.FC<ServiceOperationsProps> = ({
     error: opsError,
   } = useOperations({
     serviceName,
-    environment,
+    environment: normalizedEnvironment,
     startTime: parsedTimeRange.startTime,
     endTime: parsedTimeRange.endTime,
     refreshTrigger,
@@ -270,7 +274,7 @@ export const ServiceOperations: React.FC<ServiceOperationsProps> = ({
   const { metrics: operationMetrics, isLoading: metricsLoading } = useOperationMetrics({
     operations: operationsData || [],
     serviceName,
-    environment: environment || 'generic:default',
+    environment: normalizedEnvironment,
     startTime: parsedTimeRange.startTime,
     endTime: parsedTimeRange.endTime,
     prometheusConnectionId,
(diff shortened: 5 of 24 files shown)


TackAdam:
🔴 **High — Service map metrics always produce invalid PromQL**

`public/components/apm/query_services/query_requests/promql_queries.ts:1012-1045` + `use_service_map_metrics.ts:77`

The three service-map builders still interpolate the filter directly before a comma:

`sum_over_time(request{${serviceFilter},remoteService="",namespace="span_derived"}[${timeRange}])`

`use_service_map_metrics.ts:77` now hardcodes `const serviceFilter = ''`, so every query becomes:

`request{,remoteService="",namespace="span_derived"}`

This produces a leading comma that Prometheus rejects with a parse error. All three `Promise.allSettled` queries reject, so the Topology map renders every node with zero throughput/faults/errors and sets an error.

The parallel services-home builders were fixed by routing through `buildServicesNodeSelector(...).filter(Boolean)` (lines 37-38); these three were missed.

**Fix:** Apply the same selector helper, or use `[serviceFilter, ...].filter(Boolean).join(', ')`, here.

This is a live break, not an edge case — the map's node metrics are always empty now.

---

🟡 **Medium — Sort divergence can fetch sparklines for off-page rows**

`services_home.tsx:774-817`

The `visibleServices` effect re-sorts `displayedServices` with a hand-rolled comparator:

`va < vb ? -1 : va > vb ? 1 : 0`

This is intended to mirror `EuiInMemoryTable`'s page. When metric values tie (common, since many services fall back to `0`), or on string columns, EUI's own `EuiComparators` can order ties/blanks differently.

As a result, the computed slice can diverge from the rendered page: sparklines load for rows that aren't shown, while visible rows stay blank.

**Fix:** Consider deriving the visible slice from the table's actual `onTableChange` state rather than re-sorting independently.

---

🟡 **Low/Medium — Service-map metrics don't retrigger on environment-only change**

`use_service_map_metrics.ts:79-83, 144-149`

`servicesKey` joins only `serviceName`, but `metricsMap` is keyed by `${serviceName}::${environment}` (`nodeId`).

If the service set changes environments while names stay constant, the effect never re-runs and nodes render stale/missing metrics.

Separately, the query is `sum by (service)` with no environment, and `extractServiceData` filters by name only. This means two nodes sharing a service name across environments get identical metrics.

**Fix:** Include `environment` in `servicesKey`.

---

🟡 **Low — Sparkline refetch key is name-only**

`use_services_red_metrics.ts:150-153, 325-423`

`sparklineKey` is built from service names only. This is documented as intentional because the range query returns all environments per name.

However, if paging lands on rows with the same name set but different `(name, env)` keys that aren't cached, both `sparklineKey` and `throughputFailureMap` remain unchanged.

As a result, Effect 3 never runs and those rows show empty sparklines.

This is an edge case, but real when a name spans more environments than fit on one page.

---

🟡 **Low — Instant metrics for undefined-environment services match an arbitrary environment**

`use_services_red_metrics.ts:205-210, 523/542/559`

When a service is listed with `environment === undefined`, `serviceNodeKey` yields `name::`, and `extractServiceData(resp, name, undefined)` skips the environment check.

This returns the first matching series, or blends all environments in the `data_frame` path.

The `services_home` caller always passes `s.environment`, so this is latent today, but the optional-environment API invites incorrect behavior.

**Fix:** Add a guard or document that `environment` is required for correct keying.

joshuali925:
### Code review

Found 5 issues:

1. Invalid PromQL for Application Map node metrics: `use_service_map_metrics.ts` hardcodes `serviceFilter = ''` and passes it to `getQueryServiceMapThroughput/Faults/Errors`, which (unlike the other builders in this PR) were not migrated to `buildServicesNodeSelector` and still interpolate raw. With an empty filter this renders `request{,remoteService="",namespace="span_derived"}` — a leading comma, which is invalid PromQL — so every throughput/faults/errors query on the Application Map fails and `Promise.allSettled` silently blanks the node metrics.

https://github.com/opensearch-project/dashboards-observability/blob/72800d02aabc36e1ab10abe14927f3db218c97d4/public/components/apm/query_services/query_requests/promql_queries.ts#L1011-L1016

https://github.com/opensearch-project/dashboards-observability/blob/72800d02aabc36e1ab10abe14927f3db218c97d4/public/components/apm/shared/hooks/use_service_map_metrics.ts#L76-L112

2. AbortController signal is never threaded into the PPL calls, so in-flight requests are not cancelled. `use_operations.ts` / `use_dependencies.ts` / `use_service_map.ts` create an `AbortController` and guard state on `signal.aborted`, but `pplSearchService.listServiceOperations/listServiceDependencies/getServiceMap` don't accept a `signal` param (only `executeQuery`/`executePPLQuery` do). Rapid param changes keep firing redundant, uncancelled backend queries — inconsistent with the PromQL-backed sibling hooks in this same PR that wire `signal` correctly.

https://github.com/opensearch-project/dashboards-observability/blob/72800d02aabc36e1ab10abe14927f3db218c97d4/public/components/apm/query_services/ppl_search_service.ts#L117-L124

https://github.com/opensearch-project/dashboards-observability/blob/72800d02aabc36e1ab10abe14927f3db218c97d4/public/components/apm/shared/hooks/use_operations.ts#L90-L99

3. Services Home sparkline paging uses an uncontrolled table with no page clamp, so sparklines can silently blank out after filtering. Page/sort is mirrored into local state and used to slice `visibleServices`, but the `EuiInMemoryTable` is left uncontrolled (only `initialPageSize`) with no clamp / `resetPage()` when filters shrink the row set. On page 3+, narrowing filters leaves `tablePageIndex` stale, so `sorted.slice(start, …)` returns empty and no sparklines are fetched for the on-screen rows. The existing `use_controlled_pagination` hook (used by `service_operations.tsx` / `service_dependencies.tsx`) solves exactly this but wasn't adopted here.

https://github.com/opensearch-project/dashboards-observability/blob/72800d02aabc36e1ab10abe14927f3db218c97d4/public/components/apm/pages/services_home/services_home.tsx#L796-L801

https://github.com/opensearch-project/dashboards-observability/blob/72800d02aabc36e1ab10abe14927f3db218c97d4/public/components/apm/pages/services_home/services_home.tsx#L239-L242

4. `| head 1000` cap can silently truncate topology edges with no UI signal. The row cap applies at the edge/row level, but the "too many services" notice only fires on `celestialNodes.length > MAX_RENDERED_NODES` (500) — the already-truncated node count. In a dense mesh (many edges, fewer unique nodes), edges can be cut well before 500 nodes, so the topology renders looking complete while missing dependencies, with no truncation indicator.

https://github.com/opensearch-project/dashboards-observability/blob/72800d02aabc36e1ab10abe14927f3db218c97d4/public/components/apm/query_services/query_requests/ppl_queries.ts#L286-L291

https://github.com/opensearch-project/dashboards-observability/blob/72800d02aabc36e1ab10abe14927f3db218c97d4/public/components/apm/shared/components/service_map/service_map_graph.tsx#L142-L148

5. Metric-range slider filters can be silently reset while metrics stream in. The bounds-sync effect unconditionally calls `setLatencyUserModified(false)` / `setThroughputUserModified(false)` whenever the bound values change. Instant metrics arrive in multiple waves (`use_services_red_metrics.ts` sets the count/failure and latency maps in separate effects), so a bound can go from `0` to a real value in a later wave. If the user narrows a slider during that window, the next wave clears their filter. This is the same concern raised on #2856 (discussion_r3937621666); changing the dependency array fixes the identity-churn variant but not the value-change reset.

https://github.com/opensearch-project/dashboards-observability/blob/72800d02aabc36e1ab10abe14927f3db218c97d4/public/components/apm/pages/services_home/services_home.tsx#L695-L705

ps48:
Thanks @TackAdam and @joshuali925 for the careful reviews — all findings are addressed. Changes are in `bb40bd2` and `97b1063`, and CI is green.

**Invalid PromQL on the Application Map node metrics (leading comma)** — the three `getQueryServiceMap{Throughput,Faults,Errors}` builders still interpolated the filter raw before a comma, so with the now-default empty filter they produced `request{,remoteService="",...}`. They now route through the same `buildServicesNodeSelector` helper the services-home builders use, which drops the filter (and its comma) when empty. Added a builder test asserting no leading comma for empty or populated filters. (`bb40bd2`)

**PPL requests not cancelled** — `listServiceOperations` / `listServiceDependencies` / `getServiceMap` now accept and forward the `AbortSignal` to `executePPLQuery`, and the hooks pass their controller's signal, matching the PromQL-backed hooks. (`97b1063`)

**Services Home pagination — stale page after filtering / sort divergence** — adopted the existing `useControlledPagination` hook (as the service-details tables do). Pagination and sort are now controlled and the page index is clamped to the post-filter row count, so a shrinking result set can't strand the user on an empty page with no sparklines, and the visible-page slice stays in step with the rendered rows. Changing a filter also returns to page 1. (`97b1063`)

**Range-slider filter reset while metrics stream in** — the bounds-sync effect no longer clears the user-modified flags; it only initializes sliders the user hasn't touched (and depends on the primitive bounds, not the object identity), so a bound changing in a later metric wave can't wipe an active filter. (`97b1063`)

**Service-map metrics blended across environments** — the node-metric queries now group by `(environment, service)`, extraction matches on name + environment, and `servicesKey` includes environment, so nodes sharing a name across environments get distinct metrics and refetch on an environment-only change. (`97b1063`)

**Silent topology edge truncation** — `getServiceMap` now flags when the edge query hits its row cap, and the page shows a notice, so a dense mesh (edges cut before the node cap) isn't truncated without any indicator. (`97b1063`)

**Sparkline refetch key / undefined-environment** — the visible-page sparkline key is now `(name, environment)`, so an environment-only page change still triggers the cache-aware fetch (which only queries the not-yet-cached nodes). Documented that `environment` should be provided for correct metric-map keying. (`97b1063`)

Also updated the affected metric-hook unit tests for the new keying and fetch model, and added tests for the query builders and the PromQL/PPL escape helpers.

TackAdam (approved):
One minor follow-up (non-blocking): visibleServices still derives its slice from a hand-rolled comparator rather than EUI's onTableChange order. It's mitigated with a 0-fallback matching the column comparators, so at worst a tie on a string column could put a sparklike on a row that isn't the one rendered, no correctness impact. Fine to track separately.

joshuali925 (approved):

PR #2862, opened 2026-09-03
feat(alerting): copy Explore query into metrics rule flyout + real preview

lezzago:
### Description

Fixes two issues with **Create alert rule** from the Explore **Metrics (PromQL)** page and hardens the surrounding experience.

**A. The Explore query is now copied into the metrics rule flyout.**
Previously the flyout opened with an empty query. It now pre-fills a new, visible/editable **PromQL expression** field from the Explore editor — using the committed query, and falling back to the live `QueryStringManager` value so a not-yet-run query still copies (the metrics multi-query editor isn't wired to the shared `queryInEditor` buffer). The point-and-click builder still assists and writes into the same field.

**B. The preview now shows a real chart instead of hardcoded sample data.**
New server route `POST /api/alerting/prometheus/{dsId}/preview` runs a real range query via the DirectQuery Prometheus backend; `QueryPreviewResults` renders the live series (loading / error / empty / chart). Wired into both the metrics flyout and the Alert Manager "Create metrics rule" flyout, keyed on an explicit **Run preview** token.

**Unified Rules view freshness.** A rule created elsewhere (e.g. from the Metrics page) previously stayed invisible in the Alerts → Rules list until a manual filter toggle, because the list fetched once on mount and Cortex's querier lags the create. The view now refetches when the Rules tab is activated and on window focus, and gains a manual **Refresh** button.

### Hardening (from adversarial review)

- Preview fetch is keyed on the Run-preview token + datasource (snapshots the query) — no per-keystroke range queries.
- Preview uses the non-swallowing `queryRangeMatrix` so real failures surface as an error callout (instead of a misleading "no data"); empty-state copy reworded to not over-claim.
- Preview route bounds inputs: min step, clamp `end` to now, reject `start >= end`, and cap resolution to 1500 points to prevent oversized range queries.
- Shared `stripTrailingComparison` helper (handles `>=`, scientific notation, negatives; preserves compound conditions) used by both the preview route and the rule-detail preview; the route returns the effective plotted query so the caption is accurate.
- Throttled the focus/visibilitychange refetch to dedupe the double-fire.

### Testing

- Jest unit tests: query-seeding, unified-rules refetch on tab activation, `stripTrailingComparison`, updated preview/route-count tests (all green).
- New Cypress API-contract spec (`alerting_test/metrics_rule_preview_api.spec.js`) pinning the preview route against the real Cortex backend, wired into the `cypress-slo-cortex` workflow.
- Manually validated end-to-end against a local Metrics/PromQL datasource: query copies (incl. pre-submit), preview renders real points with the comparison correctly stripped, rule creates, and appears in the unified Rules view after refresh.

### Check List
- [x] All tests pass

_Draft: opened for review; screenshots/GIFs to be attached in a follow-up comment._

---

### Update — immediacy, audit fixes, and a UI test

**Rules-list immediacy.** Beyond refetch-on-tab-activate / focus / manual Refresh, the Rules tab now **silently background-polls every 15s** (no spinner, no error-banner clobber), so a rule created here or from the Metrics page appears on its own within ~15s of Cortex's querier loading it. True zero-latency isn't possible (the unified list reads Cortex's querier, which lags the create by up to the rule's eval interval); this closes the gap without a manual reload.

**Audit fixes (metrics flyout):**
- **WYSIWYG (was a real bug):** the Rule Preview (YAML) and the save payload now share `materializeLabels`/`materializeAnnotations`, so an empty-value label/annotation can no longer be silently written while hidden from the preview.
- Reject an expression that starts with a comparison operator (no series) in the validity check.
- Trim rule name / group name before saving.
- Stop the PromQL builder re-emitting (and normalizing whitespace of) a seeded query on mount — that spuriously marked the form dirty; ownership is still tracked so clearing the metric clears a builder-authored query.

**Deferred (documented, not in this PR):** server-side overwrite-guard for a same-named metric rule created from the Metrics page (pre-existing create-or-replace semantics; needs cross-flow design + tests); group-scoped duplicate-name check; preview honoring the Explore time picker; builder-clobber-of-hand-typed-expression confirm; YAML preview group/interval header; datasource-picker selected-state.

**UI test.** Adds `alerting_test/metrics_create_alert_rule_ui.spec.js` (query-copy + real preview) and a dedicated **workspace-enabled** `cypress-metrics-alerting-ui` Cortex job (the Explore metrics page is workspace-gated, which conflicts with the SLO specs' workspace-off job). The job creates a workspace and best-effort-associates the Prometheus datasource; the spec skips gracefully if the flyout isn't reachable, so the workspace/datasource-association step can be iterated on CI without hard failures. The deterministic preview-route API spec runs in the SLO job regardless.

---

## Before / After

Before = `origin/main`, after = this PR. Changes outlined in red.

### 1. Explore PromQL query is copied into the metrics rule flyout
query copy before vs after

The flyout now has a **Builder / Code toggle** — a copied query lands in **Code** mode as-is, so the point-and-click builder can no longer silently overwrite it (audit finding **#7**). Switching to Builder with a non-representable expression warns first:

code vs builder mode

### 2. Real preview instead of hardcoded sample data
preview before vs after
<sub>The preview also now runs over the Explore time picker's window instead of a fixed last hour (#5) — behavioral, no separate screenshot.</sub>

### 3. Unified Rules — Refresh button + 15s auto-refresh
refresh before vs after

### 4. Rule Preview (YAML) shows namespace, rule group & evaluation interval (#10)
The preview now renders the full rule-group that gets written (namespace → group → rule, including the group-level `interval`), instead of just the rule body.
yaml before vs after

### 5. Datasource picker marks the active selection (#11)
datasource picker before vs after

### Full "Create alert rule from Metrics" workflow

**After (this PR)** — query copied, real preview:

after workflow gif

**Before (`origin/main`)** — empty query, fake sample data:

before workflow gif

<sub>Media hosted on the `pr-2862-media` asset branch of the fork (not part of this PR's diff).</sub>

---

### Corrections to the notes above
- **#2 (overwrite guard) and #3 (group-scoped duplicate check) are now implemented in this PR** — they are no longer deferred. Creating a metric rule whose name already exists in the target group returns a 409 (edit passes `overwrite:true`); the duplicate-name check is keyed on (datasource, group, name).
- The **metrics Create-alert-rule UI Cypress job/spec has been removed from this PR** and deferred to a follow-up (the workspace↔datasource association needs CI iteration). The deterministic `metrics_rule_preview_api.spec.js` remains wired into the SLO Cortex job.

### DIFF
--- a/public/components/alerting/alarms_page.tsx
+++ b/public/components/alerting/alarms_page.tsx
@@ -172,6 +172,12 @@ function buildPrometheusRulePayload(opts: {
   annotations: Record<string, string>;
   enabled: boolean;
   groupName?: string;
+  /**
+   * Allow replacing an existing same-named rule in the group. Edit flows set
+   * this; create flows leave it false so the server rejects a name collision
+   * (409) instead of silently overwriting.
+   */
+  overwrite?: boolean;
 }) {
   return {
     name: opts.name,
@@ -182,6 +188,7 @@ function buildPrometheusRulePayload(opts: {
     annotations: opts.annotations,
     enabled: opts.enabled,
     ...(opts.groupName ? { groupName: opts.groupName } : {}),
+    ...(opts.overwrite ? { overwrite: true } : {}),
   };
 }
 
@@ -407,7 +414,57 @@ export const AlarmsPage: React.FC<AlarmsPageProps> = ({
     setRules,
     setRulesTotal,
     refetch: refetchRules,
+    backgroundRefetch: backgroundRefetchRules,
   } = useRulesData({ selectedDsIds });
+
+  // Keep the rules list fresh without a manual reload. A rule created
+  // elsewhere (e.g. the Metrics page "Create alert rule" flyout) only lands
+  // in the unified list once Cortex's querier has loaded and first-evaluated
+  // it, which can lag the create by up to the rule's evaluation interval — so
+  // the initial mount fetch can miss it and the list would otherwise stay
+  // stale until a filter toggle. Refetch when the user switches INTO the Rules
+  // tab so a freshly-created rule shows up on navigation.
+  const prevTabRef = useRef(activeTab);
+  useEffect(() => {
+    if (activeTab === 'rules' && prevTabRef.current !== 'rules') {
+      refetchRules();
+    }
+    prevTabRef.current = activeTab;
+  }, [activeTab, refetchRules]);
+
+  // Also refetch when the window/tab regains focus, so returning to an
+  // already-open Alerts app re-syncs (e.g. after creating a rule in another
+  // browser tab). Cheap: only re-runs the rules query. `focus` and
+  // `visibilitychange` both fire when refocusing a window, so throttle to
+  // one refetch per second to avoid a redundant double request.
+  const lastFocusRefetchRef = useRef(0);
+  useEffect(() => {
+    const onVisible = () => {
+      if (document.visibilityState !== 'visible') return;
+      const nowMs = Date.now();
+      if (nowMs - lastFocusRefetchRef.current < 1000) return;
+      lastFocusRefetchRef.current = nowMs;
+      refetchRules();
+    };
+    window.addEventListener('focus', onVisible);
+    document.addEventListener('visibilitychange', onVisible);
+    return () => {
+      window.removeEventListener('focus', onVisible);
+      document.removeEventListener('visibilitychange', onVisible);
+    };
+  }, [refetchRules]);
+
+  // While the Rules tab is open and visible, poll silently every 15s so a rule
+  // that was just created — here or from the Metrics page — shows on its own
+  // once Cortex's querier has loaded it, without the user clicking Refresh.
+  // Background refetch = no spinner / no error-banner clobber.
+  useEffect(() => {
+    if (activeTab !== 'rules') return undefined;
+    const intervalId = window.setInterval(() => {
+      if (document.visibilityState === 'visible') backgroundRefetchRules();
+    }, 15000);
+    return () => window.clearInterval(intervalId);
+  }, [activeTab, backgroundRefetchRules]);
 
   const [deletedRuleIds, setDeletedRuleIds] = useState<Set<string>>(new Set());
   const [showCreateMonitor, setShowCreateMonitor] = useState(false);
@@ -425,40 +482,37 @@ export const AlarmsPage: React.FC<AlarmsPageProps> = ({
   // datasource filter has surfaced — saving against an unselected DS could
   // still create a same-name monitor on that DS, but that's a much narrower
   // footgun than the original "two identical names on the same cluster".
-  const rulesByDsName = useMemo(() => {
-    const map = new Map<string, Set<string>>();
-    rules.forEach((r) => {
-      if (deletedRuleIds.has(r.id)) return;
-      const dsKey = r.datasourceId;
-      const names = map.get(dsKey) ?? new Set<string>();
-      names.add(r.name.trim().toLowerCase());
-      map.set(dsKey, names);
-    });
-    return map;
-  }, [rules, deletedRuleIds]);
-
-  const isNameTakenForCreate = useCallback(
-    (name: string, dsId: string) => {
-      const set = rulesByDsName.get(dsId);
-      return !!set?.has(name.trim().toLowerCase());
-    },
-    [rulesByDsName]
-  );
-
-  const buildIsNameTakenForEdit = useCallback(
-    (excludeRuleId: string) => (name: string, dsId: string) => {
+  // Duplicate-rule detection. Prometheus rule identity is (datasource, group,
+  // name), so when a `group` is supplied the collision is scoped to that group
+  // — this lets a user legitimately keep `HighLatency` in two different groups
+  // without a false "already exists". When no group is supplied (OpenSearch/PPL
+  // monitors, which have no group) the check falls back to datasource-wide.
+  const isRuleNameTaken = useCallback(
+    (name: string, dsId: string, group: string | undefined, excludeRuleId?: string) => {
       const trimmed = name.trim().toLowerCase();
-      // Walk the rule list directly so we can exclude the monitor being
-      // edited; the cached map doesn't carry id information.
+      if (!trimmed) return false;
+      const g = group?.trim().toLowerCase();
       return rules.some(
         (r) =>
           r.id !== excludeRuleId &&
           !deletedRuleIds.has(r.id) &&
           r.datasourceId === dsId &&
-          r.name.trim().toLowerCase() === trimmed
+          r.name.trim().toLowerCase() === trimmed &&
+          (g === undefined || g === '' || (r.group ?? '').trim().toLowerCase() === g)
       );
     },
     [rules, deletedRuleIds]
+  );
+
+  const isNameTakenForCreate = useCallback(
+    (name: string, dsId: string, group?: string) => isRuleNameTaken(name, dsId, group),
+    [isRuleNameTaken]
+  );
+
+  const buildIsNameTakenForEdit = useCallback(
+    (excludeRuleId: string) => (name: string, dsId: string, group?: string) =>
+      isRuleNameTaken(name, dsId, group, excludeRuleId),
+    [isRuleNameTaken]
   );
   // The popover's "Logs" entry maps to an OpenSearch monitor and "Metrics"
   // maps to a Prometheus rule. When the user picks Logs the flyout is forced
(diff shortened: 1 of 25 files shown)


TackAdam (approved):
## One minor finding (UX correctness, not a bug)

`query_preview_results.tsx` charts only `series[0]` (`const points = series[0]?.values ?? []`), and the accordion header reads `Results (N)`, where `N` is the sample-point count of that one series, not the number of matching series.

When the expression matches multiple series (e.g. `up` across targets, or a `rate(...)` with several label sets), the preview silently plots one arbitrary series and the count reads like a result-set size.

The PR description acknowledges that "the preview is a single line," so this is intended scope — but the `Results (N)` label is misleading for multi-series expressions. Consider labeling it as points/samples, or noting when `>1` series matched.

## Notes (not blocking)

- Sub-second Explore ranges could floor to `start === end` → `400` (`plugin.tsx` checks `max > min` in ms, then divides by `1000`). Practically impossible for real Explore windows; fine to leave.
- The route comment still says `"Reuses the same promBackend.queryRange path"` but it now uses `queryRangeMatrix` — stale comment.

lezzago:
Thanks for the review! Addressed all three in `2117cc5a`:

- **Multi-series `Results (N)` label (main finding):** renamed the accordion header to **`Samples (N)`** (N was always the sample-point count, never a result-set size), and the preview route now returns `seriesCount` so the UI shows a note when the expression matched more than one series: *"Matched N series — charting the first. The alert evaluates every matching series."* Added a unit test for the note.
- **Sub-second range floor → 400:** `plugin.tsx` now requires a `>= 1s` Explore window before passing a preview range; sub-second windows fall back to the last-hour default so the floor-to-seconds can't collapse to `start === end`.
- **Stale route comment:** updated — the preview uses `queryRangeMatrix` (throws on bad/unreachable queries and caps points per series), no longer `queryRange`.

public/components/alerting/alarms_page.tsx
```diff
@@ -407,8 +414,58 @@ export const AlarmsPage: React.FC<AlarmsPageProps> = ({
...
+  // `visibilitychange` both fire when refocusing a window, so throttle to
+  // one refetch per second to avoid a redundant double request.
+  const lastFocusRefetchRef = useRef(0);
+  useEffect(() => {
+    const onVisible = () => {
+      if (document.visibilityState !== 'visible') return;
+      const nowMs = Date.now();
+      if (nowMs - lastFocusRefetchRef.current < 1000) return;
+      lastFocusRefetchRef.current = nowMs;
+      refetchRules();
```
ps48: ### p1: focus refetch uses the foreground path, clobbering the error banner

This effect's comment says "Cheap: only re-runs the rules query", but `refetchRules()` is the foreground refetch: `fetchRules(background=false)` sets `isLoading(true)` and clears `error` / `warnings` (`use_rules_data.ts:72-76`). So every time the window regains focus (throttled to 1/sec) the table flashes a spinner and a visible fetch-error / warning banner is wiped and re-derived. The 15s poll right below deliberately uses `backgroundRefetchRules()` for exactly this reason ("Background refetch = no spinner / no error-banner clobber", line 460). Could we call `backgroundRefetchRules()` here too so a refocus re-syncs silently and matches the poll's intent?

public/components/alerting/query_preview_results.tsx
```diff
@@ -8,77 +8,256 @@
...
+  const queryRef = useRef(query);
+  queryRef.current = query;
+  // Same treatment for the range: read it at fetch time so a re-render with a
+  // new object identity doesn't re-fire the query (only runToken/ds do).
+  const timeRangeRef = useRef(timeRange);
+  timeRangeRef.current = timeRange;
+
+  useEffect(() => {
+    const trimmed = (queryRef.current || '').trim();
+    if (!datasourceId || !trimmed) {
```
ps48, line 139: ### p2: no-input collapses to the same empty-state as a real empty result

When the query is blank or no datasource is selected this early-returns to `EMPTY_STATE`, which then renders the `resultCount === 0` warning "The query returned no data for this time range, or the data source could not evaluate it." The Run preview button in both parents (`create_metrics_monitor.tsx:450`, `prometheus_form_section.tsx:331`) has no `isDisabled`, so a user can click it with an empty expression / before picking a datasource and be told a valid metric returned nothing or couldn't be evaluated. Could we distinguish "nothing has been run / no input" from "ran and got zero rows" (e.g. a separate not-run state, or gate the Run preview button on a non-empty query + datasource) so the warning only shows for a genuine empty result?

public/components/alerting/create_metrics_monitor.tsx
```diff
@@ -809,6 +1010,8 @@ export const CreateMetricsMonitor: React.FC<CreateMetricsMonitorProps> = ({
   const isValid =
     form.monitorName.trim() !== '' &&
     form.query.trim() !== '' &&
+    // Reject an expression that is only/starts-with a comparison (no series).
+    !startsWithComparison(form.query) &&
```
ps48, line 1018: ### p2: `startsWithComparison` only rejects a leading comparison, not a missing one

`isValid` blocks an expression that *starts* with a comparison (`> 0.5`), which is a good cheap guard, and the comment correctly defers full validation to the server / preview. Worth noting the gap it leaves: a comparison-less expression like `up` or `sum(rate(http_requests[5m]))` is a valid series but not a valid *alert condition* (it evaluates truthy for every returned sample, i.e. an always-firing rule), yet it passes `isValid` and Create is enabled. If the server doesn't enforce a comparison either, users can create always-on rules from a copied Explore query. Could we either require the expression to contain a comparison operator (an actual condition) here, or clarify the field / help text that the copied query still needs a threshold added? Low priority if the server already rejects this.

ps48 (commented):
### copy explore query into metrics rule flyout + real preview

Solid, well-tested PR and the core plumbing checks out. Verified: no new dependencies, no new saved-object types, no stray `console.*`, no vega / elastic-charts (echarts throughout). The one new route (`/api/alerting/prometheus/{dsId}/preview`) sits in the existing alerting `prometheus/*` family rather than being a general search route, is point-budget bounded (`MAX_PREVIEW_POINTS`), forwards the caller's request for credentialed reads, and deliberately uses `queryRangeMatrix` (throws on a bad / unreachable query instead of degrading to `[]`) so the client can tell a real failure from an empty result. `selectedDsIds` is a stable `useState` value so the new 15s poll isn't reset every render, and the Cypress spec hits the real API (no `intercept` / mock).

Credit where due: the latest commit (`2117cc5`, "honest multi-series preview label") already resolves the biggest thing I had — the preview now returns `seriesCount`, relabels the caption "Results (N)" to "Samples (N)", and adds a "Matched N series, charting the first; the alert evaluates every matching series" note. That's a clean fix. The same commit's `plugin.tsx` bounds guard (`max - min >= 1000`) also covers the realistic sub-second / collapsed-range case, so I'm not re-raising the forward-range clamp.

No `p0`. What remains is behavioral / UX-messaging, not crashes:

- `p1` the window-focus refetch uses the foreground path (spinner + clears the error / warning banner), contradicting its own "cheap" comment and the 15s poll's deliberate background refetch.
- `p2` a blank query / no datasource collapses to the same empty-state as a real empty result, so "the query returned no data / the data source could not evaluate it" is shown for missing input.
- `p2` `startsWithComparison` only rejects a leading comparison, so a comparison-less expression (an always-firing condition) still passes client validation.

One thing I looked at and am NOT flagging: the group-scoped duplicate-name check (`alarms_page.tsx:501`) intentionally scopes by `(datasource, group, name)` and create always passes a non-empty group (falling back to the rule name), so it matches Prometheus rule identity rather than being a bypass. Good.

Nice work.

lezzago:
Thanks @ps48 — all three addressed in `7f4661f6`:

- **p1 (focus refetch uses foreground path):** switched the window-focus effect to `backgroundRefetchRules()`, so a refocus re-syncs silently (no spinner, no error/warning-banner clobber) — matching the 15s poll's intent. Updated the misleading "cheap" comment.
- **p2 (no-input collapses to the empty-result warning):** added a distinct not-run/no-input state in `QueryPreviewResults` (neutral "Nothing to preview yet" prompt) so the "returned no data / could not evaluate" warning only shows for a genuine empty result, **and** gated the Run preview button on a non-empty expression + datasource in both flyouts.
- **p2 (`startsWithComparison` allows a comparison-less expression):** I opted for the help-text route rather than requiring a comparison operator — a comparison-less expression is legitimately valid for `absent()`, `vector(1)`, etc., and the point-and-click Builder itself emits `metric{label="value"}` with no comparison, so hard-requiring one would false-reject valid conditions (and break the Builder). The Code-mode help text now says a bare series fires for every returned series and to add a comparison/threshold for a conditional alert.

Added tests: not-run vs. real-empty distinction + the multi-series note (`query_preview_results.test.tsx`), and Run-preview-disabled-until-ready (`create_metrics_monitor.test.tsx`).

PR #2865, opened 2026-09-03
fix(notebooks): bind sample notebook paragraphs to the selected data

### LINKED ISSUE #2454: [BUG] Sample notebooks does not honor the data source selected
**What is the bug?**

`Add sample notebooks` does not honor the data source selected, it is still using the default data source for the data source selector.

https://github.com/user-attachments/assets/465fde85-e83f-434b-8139-6c83d9a781f0

The selector should honor the data source selected

 - Version [e.g. 22] 3.0.0
 - Plugins

Add any other context about the problem.

cnoramut:
### Description

`Add sample notebooks` does not honor the data source selected in its modal. The sample data installs against the chosen data source correctly, but every code block inside the created notebooks points at whichever data source is configured as the default.

**Root cause.** The selected id is dropped at the persistence boundary. `main.tsx` posts only `visIds`, and `addSampleNotes` accepts only `visIds`, so the generated paragraphs are written with no `dataSourceMDSId` and no `dataSourceMDSLabel`:

```ts
body: JSON.stringify({ visIds }),
```

The per-paragraph selector gates its `defaultOption` on `paradataSourceMDSId !== undefined`. With nothing stored that is `false`, so `DataSourceSelector` falls through to `handleDefaultDataSource` and resolves the default data source instead of the one picked. On a real multi-cluster setup the sample data lands on one cluster while the notebook queries another, so the PPL and SQL paragraphs hit a cluster with no sample indices.

**Fix.** Thread the id and label from the modal through the route into `addSampleNotes`, which stamps both onto every paragraph before create. The fields, the writer on paragraph run, and the reader in `default_parser.tsx` all already existed. Only the sample creation path never populated them.

**Notes for reviewers**

1. **Existing sample notebooks are not repaired**, since the missing binding is already persisted. Delete and re-add them, as re-adding creates duplicates rather than replacing.
2. **No unit test, verified manually instead**, with a before and after recording below. A three-case test on `addSampleNotes` was written and validated red then green, can add if necessary.
3. **The empty string fallback is deliberate.** `notebook.tsx` blanks a `QUERY` paragraph's output and raises a danger toast when `dataSourceMDSId` is truthy while multiple data sources are disabled. Local cluster's id is already the empty string, so defaulting to `''` keeps Local cluster and non-MDS installs on their existing path.
4. **Complementary to #2861, not dependent on it.** That fix only changes behavior when the id is falsy while the label is truthy, which is only the Local cluster case. For a real data source both are truthy and the visualization title search string is identical either way. Either can land first.

**Testing.** Verified with MDS enabled inside a workspace holding a wizard-created data source. Before the fix every paragraph across all four sample notebooks had `dataSourceMDSId` and `dataSourceMDSLabel` null. After, every paragraph carries the selected id and label, confirmed by querying the `observability-notebook` saved objects directly. Jest suites for `public/components/notebooks` and `server` pass, lint clean.

**Before**

https://github.com/user-attachments/assets/eb348ca4-1760-43f7-ae28-cf2e6ac370ae

**After**

https://github.com/user-attachments/assets/a1c06392-8724-4ff3-8a80-69e2a60f6370

### Issues Resolved

Fixes #2454

### Check List

### DIFF
--- a/public/components/notebooks/components/main.tsx
+++ b/public/components/notebooks/components/main.tsx
@@ -408,7 +408,7 @@ export class Main extends React.Component<MainProps, MainState> {
         });
       await this.props.http
         .post(`${NOTEBOOKS_API_PREFIX}/note/savedNotebook/addSampleNotebooks`, {
-          body: JSON.stringify({ visIds }),
+          body: JSON.stringify({ visIds, dataSourceMDSId, dataSourceMDSLabel }),
         })
         .then((res) => {
           const newData = res.body.map((notebook: any) => ({
--- a/server/adaptors/notebooks/saved_objects_notebooks_router.tsx
+++ b/server/adaptors/notebooks/saved_objects_notebooks_router.tsx
@@ -68,13 +68,31 @@ export function renameNotebook(noteBookObj: { name: string; noteId: string }) {
 
 export async function addSampleNotes(
   opensearchNotebooksClient: SavedObjectsClientContract,
-  visIds: string[]
+  visIds: string[],
+  dataSourceMDSId?: string,
+  dataSourceMDSLabel?: string
 ) {
   const notebooks = getSampleNotebooks(visIds);
   const sampleNotebooks = [];
   try {
+    // Without a stored binding, each code block's selector falls back to the default data
+    // source rather than the one the sample data was just installed against.
     for (const item of notebooks) {
-      const createdNotebooks = await opensearchNotebooksClient.create(NOTEBOOK_SAVED_OBJECT, item);
+      const notebook = {
+        ...item,
+        savedNotebook: {
+          ...item.savedNotebook,
+          paragraphs: item.savedNotebook.paragraphs.map((paragraph) => ({
+            ...paragraph,
+            dataSourceMDSId: dataSourceMDSId ?? '',
+            dataSourceMDSLabel: dataSourceMDSLabel ?? '',
+          })),
+        },
+      };
+      const createdNotebooks = await opensearchNotebooksClient.create(
+        NOTEBOOK_SAVED_OBJECT,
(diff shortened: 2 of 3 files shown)


TackAdam:
LGTM, just one minor finding worth adding before merge

In `server/routes/notebooks/noteRouter.ts`:

```ts
dataSourceMDSId: schema.maybe(schema.string({ defaultValue: '' })),
dataSourceMDSLabel: schema.maybe(schema.string({ defaultValue: '' })),
```

The `defaultValue: ''` here is dead code. `MaybeType`'s constructor wraps the inner schema with `.default(() => undefined)` (`packages/osd-config-schema/src/types/maybe_type.ts`), so when the client omits either field, the parsed value is `undefined`, never `''`.

The empty-string default is misleading — it reads as if the server sees `''`, but the actual behavior relies on the `?? ''` fallback in `addSampleNotes`.

Please drop the `defaultValue` so the schema reflects the actual behavior:

```ts
dataSourceMDSId: schema.maybe(schema.string()),
dataSourceMDSLabel: schema.maybe(schema.string()),
```

Functionally equivalent since `addSampleNotes` already applies `?? ''`.

cnoramut:
Thank you. Fix with the new push.

Worth flagging that `paraRouter.ts` has the same dead `defaultValue: ''` on these two field names in four places, lines 129, 130, 204, and 205, which is where I copied it from. I can clean that up here or leave it as is?

TackAdam:
Good catch, those four in paraRouter.ts (lines 129, 130, 204, 205) are the same dead default. I'd keep this PR scoped to the sample-notebook fix and clean those up in a separate PR so the back-port stays clean.

TackAdam (approved):

PR #2866, opened 2026-09-04
feat(alerting): optimistic pending-rule cache for the metrics rules list

lezzago:
> ✅ **#2862 merged; this PR was rebased onto `main` (commit d484847c).** The diff is now just the optimistic pending-rule cache — client-side only, no server/route changes.

### Problem
A newly-created Prometheus/metrics alert rule is persisted **synchronously** to Cortex's ruler config store (a 2xx = "persisted"), but the unified rules list reads the **querier** (`/api/v1/rules`), which only surfaces the rule ~60s+ later (ruler poll + one eval interval). Two bugs made it flicker/disappear:

1. The optimistic insert built the row with a throwaway `new-<ts>` id and **no group**, so it never matched the server's canonical id `${dsId}-${group}-${name}`.
2. `useRulesData.fetchRules` **full-replaces** the list on every refetch (15s poll, tab-activate, focus, manual Refresh) — dropping the optimistic row before propagation, and the ad-hoc `setTimeout(refetchRules, 15000)` calls actively triggered the drop.

A poll timeout means **"not propagated yet"**, not "creation failed" — so timeout warns softly, never errors.

### Approach
An optimistic **pending-rule cache** that carries the row for that window and reconciles against each querier refetch — **no new timer** (rides the existing poll + tab-activate + focus refetches).

- **Pure core** `monitors_table/pending_rules.ts`: one match `key` (`dsId|group|name`, trimmed + lowercased, built from the create *payload's* groupName), `isPending` (status `pending` **and** synthetic `new-` id — so a real Cortex-pending rule isn't mistaken for ours), `mergePending`, and `reconcilePending` (evict on confirm / user-delete / TTL 120s / maxAttempts 6; dedupe-before-append so no two-row flash).
- **State** `hooks/use_pending_rules.ts`: in-memory **module singleton** (the `coreRefs` precedent — survives the Explore→Alerts SPA nav; **no** sessionStorage, so a hard reload cleanly falls back to the poll rather than resurrecting a dead row) + a `useSyncExternalStore` hook. Reconcile runs **once per querier refetch** (effect keyed on `rules` alone); the rendered view is a separate pure derivation so an added row appears immediately.
- **Wiring** in `alarms_page.tsx`: layer `mergedRules` into `visibleRules`; route create/clone/metrics/batch through `addPending`; **retire** `refetchTimerRef` + all four `setTimeout(refetchRules, 15000)` (keep the immediate refetch after each create); scope pending to the selection on datasource change; soft **warning** toast (deduped per key) on timeout.
- **UX**: spinner `pendingRuleBadge` in the status cell; Edit/Delete/Clone/Enable disabled in the detail flyout + pending rows excluded from checkbox selection (their synthetic id would 404); Prometheus create toast gains a "may take up to a minute" note.

### Out of scope (future follow-up)
Config-API positive confirmation (`RulerClient.getRuleGroup`) for lag-free confirm + honest write-failed vs querier-lag distinction. v1 is poll-only.

### Tests
- Heavy branch coverage on the pure `reconcilePending` / `key` / `isPending` / `mergePending` (incl. group case/whitespace-insensitivity, confirm-vs-delete/ttl/attempts priority, no-resurrection, ds-scoping).
- Hook tests: immediate display on add, silent evict on confirm, warn-once on timeout, drop-on-datasource-change.
- Thin component tests: pending badge renders, actions disabled, pending row not selectable, Prom-only toast note.

All 465 alerting suites pass; `yarn lint --fix` clean via the pre-commit hook.

### DIFF
--- a/public/components/alerting/__tests__/create_metrics_monitor.test.tsx
+++ b/public/components/alerting/__tests__/create_metrics_monitor.test.tsx
@@ -389,8 +389,13 @@ describe('CreateMetricsMonitor', () => {
     expect(body).not.toHaveProperty('operator');
     expect(body).not.toHaveProperty('threshold');
 
-    // Should call onSave and show success toast
+    // Should call onSave and show success toast, including the querier-lag
+    // note as the toast body (Prometheus creates take ~a minute to appear).
     expect(onSave).toHaveBeenCalled();
-    expect(addToast).toHaveBeenCalledWith(expect.any(String), 'success');
+    expect(addToast).toHaveBeenCalledWith(
+      expect.any(String),
+      'success',
+      expect.stringContaining('take up to a minute')
+    );
   });
 });
--- a/public/components/alerting/__tests__/monitor_detail_flyout.test.tsx
+++ b/public/components/alerting/__tests__/monitor_detail_flyout.test.tsx
@@ -162,6 +162,52 @@ describe('MonitorDetailFlyout', () => {
     labels: { monitor_kind: 'composite', composite_delegates: 'mon-a,mon-b' },
   };
 
+  it('disables Edit / Clone / Delete / Enable for a pending optimistic rule', async () => {
+    const onClone = jest.fn();
+    const onDelete = jest.fn();
+    const onEdit = jest.fn();
+    // ppl/metric OS monitor would normally be editable + toggleable in place;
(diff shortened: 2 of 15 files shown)


TackAdam (approved):
LGTM

PR #2686, opened 2026-05-20
[Alerts] Empty state improvements

TackAdam:
### Description
Improve empty state and make Rules button primary for empty state.
Before:
EmptyState

After:
After1
After2
After3

### Check List
- [ ] New functionality includes testing.
  - [ ] All tests pass, including unit test, integration test and doctest
- [ ] New functionality has been documented.
  - [ ] New functionality has javadoc added
  - [ ] New functionality has user manual doc added
- [ ] Commits are signed per the DCO using --signoff

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #2870, opened 2026-09-04
feat(alerting): point-and-click condition builder for Prometheus metric rules

lezzago:
## Summary

The metrics-rule **Builder** previously emitted only a bare series selector (`metric{label="x"}`), so a *conditional* alert (a comparison/threshold) could only be written by hand in **Code** mode — and Code mode silently allowed an always-firing, condition-less rule. This PR adds a first-class, point-and-click alert-condition experience.

### What changed
- **New pure core** `create_monitor/prom_condition.ts` composes the PromQL `expr` in stacked layers and provides a lossless inverse parser:
  1. **Series** — `metric` + optional `{label OP "value"}`
  2. **Function** — rate / increase / delta and the `_over_time` family (avg/min/max/sum/count/last) over a window
  3. **Aggregate** — sum / avg / min / max / count with optional `by` / `without` grouping
  4. **Condition** — IS ABOVE / BELOW / EQUAL / NOT EQUAL / ABOVE-OR-EQUAL / BELOW-OR-EQUAL / **OUTSIDE RANGE** / **WITHIN RANGE**
  - `buildExpr` ↔ `parseExpr` round-trip so switching Builder↔Code (or re-opening to edit) is lossless for anything the builder could produce; anything more complex (nested aggs, histogram_quantile, arithmetic between series, multi-label matchers) parses to `null`, leaving the builder inert so a hand-written expression is never clobbered. Covers e.g. `sum(rate(metric[4m])) > 6`.
- **`PromQueryBuilder`** gains Function + Aggregate + Condition rows on top of the metric/label picker, composing via the core, with a live composed `Expression` preview.
- **Always-firing guard** — a warning callout in **both** Builder and Code modes when the expression has no comparison. Non-blocking (a condition-less rule is legal PromQL), just surfaced so it isn't created by accident.
- **Preview** — the *Run preview* chart now draws the threshold line(s) and shows a **"Would fire now" / "Would not fire"** badge evaluated against the latest sample.

`parseExpr` supersedes the old `parseBuilderQuery` (now represents comparisons and ranges, not just bare selectors); callers and tests updated.

## Before / after

Same rule (`sum(rate(gen_ai_client_token_usage_total[5m])) > 6`), same viewport, from the unified **Alerts → Rules → Create metrics rule** flyout. Left = `origin/main`, right = this PR; orange box = area before, green box = same area after.

### 1. Point-and-click condition builder (the feature)
On `origin/main` the Builder stops at the metric/label picker and *For duration* — there is **no way to express a condition**, so a threshold alert has to be hand-typed in Code. This PR adds **Function / Aggregate / Condition** rows and a live composed **Expression**.

builder before/after

**Before — full flow (`origin/main`):** pick a metric → the Builder offers no condition → you must switch to Code and hand-write the PromQL.

before flow

**After — full flow (this PR):** pick a metric → set Function (rate) · Aggregate (sum) · Condition (IS ABOVE 6) → the Expression composes itself and Create enables. No raw PromQL required.

after flow

### 2. Rules table — internal scroll (fixes #2686 regression)
Before, the table grew past the viewport → the whole page scrolled and a nested table scrollbar appeared (double-scroll). After, the table region is capped to the viewport and scrolls internally with a sticky header.

table before/after

### 3. No more horizontal "side" scroll
Before, the filters panel and the table each showed a spurious horizontal scrollbar. After, both are gone (a real x-scrollbar appears only when the table genuinely can't fit).

side-scroll before/after

<sub>Captured at matched viewports (builder/flows 1440×900, table 1200×689). Composites/GIFs are hosted on the `evidence/condition-builder` fork branch and are not part of this PR's diff.</sub>

### On reusing the Discover/Explore metrics builder
The Discover/Explore metrics `PromQLBuilder` is **not cleanly reusable** from this plugin: it lives in explore's internal `application/` tree (not exported from any public contract — a deep import would violate cross-plugin import rules), and its `PrometheusClient` is constructed from `ExploreServices` + the query_enhancements-registered `prometheus` resource client. This builder deliberately matches its UX shape and its `onQueryChange(promql)` + seed-from-query contract, so it can be swapped for that component if explore ever exports it.

### Testing
- New `prom_condition.test.ts` — exhaustive build/parse/round-trip + always-firing heuristic.
- Updated `create_metrics_monitor.test.tsx` / `prometheus_form_sync.test.tsx` for the `parseExpr` semantics.
- All alerting suites pass locally; typecheck + lint clean.
- e2e-validated in both surfaces (unified Alerts Rules tab + Discover/Explore metrics "Create alert rule") — correct POST payloads for comparison + range conditions, static + dynamic labels, and templated annotations.

### Checklist
- [x] Unit tests added/updated
- [x] `Signed-off-by` (DCO)
- [x] Manual UI walkthrough / screenshots (before/after above)

### DIFF
--- a/public/components/alerting/__tests__/create_metrics_monitor.test.tsx
+++ b/public/components/alerting/__tests__/create_metrics_monitor.test.tsx
@@ -24,21 +24,22 @@ jest.mock('../promql_editor', () => ({
 jest.mock('../metric_browser', () => ({
   MetricBrowser: () => <div data-test-subj="metricBrowserMock" />,
 }));
-// Stub the shared builder with a button that emits a query, so tests can
-// simulate an explicit builder selection (the form seeds query: '' and the
-// Create button stays disabled until the builder produces one)
-jest.mock('../create_monitor/prom_query_builder', () => ({
-  PromQueryBuilder: ({ onQueryChange }: { onQueryChange: (q: string) => void }) => (
-    <button data-test-subj="mockBuilderSetQuery" onClick={() => onQueryChange('up{job="api"}')} />
-  ),
-  // Lightweight stand-in: builder-representable = a bare metric or metric{...};
-  // anything else (rates, comparisons) returns null. Used by the Query section
-  // to decide the "builder will overwrite your expression" warning.
-  parseBuilderQuery: (q: string) => {
-    const m = (q || '').trim().match(/^([a-zA-Z_:][a-zA-Z0-9_:]*)(\{[^}]*\})?$/);
-    return m ? { metric: m[1] } : null;
-  },
-}));
+// Stub the shared builder *component* with a button that emits a query, but keep
+// the REAL parse/compose core (from prom_condition) so representability decisions
+// in these tests match what the component actually does. Using a hand-rolled
+// stand-in `parseExpr` here previously diverged from the real one (it only
+// recognised bare selectors), which made "non-representable" fixtures pass for
+// the wrong reason. `up{job="api"}` (a bare selector) is builder-representable.
+jest.mock('../create_monitor/prom_query_builder', () => {
+  const realCore = jest.requireActual('../create_monitor/prom_condition');
+  return {
+    PromQueryBuilder: ({ onQueryChange }: { onQueryChange: (q: string) => void }) => (
+      <button data-test-subj="mockBuilderSetQuery" onClick={() => onQueryChange('up{job="api"}')} />
+    ),
+    parseExpr: realCore.parseExpr,
+    buildExpr: realCore.buildExpr,
+  };
+});
 
 global.ResizeObserver = jest.fn().mockImplementation(() => ({
   observe: jest.fn(),
@@ -152,22 +153,22 @@ describe('CreateMetricsMonitor', () => {
   });
 
   it('seeds the query from initialQuery copied off the Explore Metrics page', () => {
+    // A copied expression the builder cannot represent (histogram_quantile) opens
+    // in Code mode, pre-filled, visible/editable — never hidden-yet-submittable.
     render(
       <CreateMetricsMonitor
         onCancel={jest.fn()}
         onSave={jest.fn()}
         datasourceId="prom-1"
-        initialQuery="rate(http_requests_total[5m])"
+        initialQuery="histogram_quantile(0.9, rate(http_requests_total[5m]))"
       />
     );
 
-    // The pre-filled expression is visible/editable (not hidden), so a complex
-    // expression the builder can't represent is never hidden-yet-submittable.
     const expr = document.querySelector(
       '[data-test-subj="metricsMonitorPromQlExpression"]'
     ) as HTMLTextAreaElement;
     expect(expr).not.toBeNull();
-    expect(expr.value).toBe('rate(http_requests_total[5m])');
+    expect(expr.value).toBe('histogram_quantile(0.9, rate(http_requests_total[5m]))');
 
     // With the query seeded, Create enables as soon as a name is entered — no
     // separate builder selection required.
@@ -179,14 +180,15 @@ describe('CreateMetricsMonitor', () => {
     expect(createBtn.disabled).toBe(false);
   });
 
-  it('defaults to Code mode when a query is copied in, Builder mode when empty', () => {
-    // Copied query -> Code mode: raw expression shown, builder hidden.
+  it('defaults to Builder mode, but Code mode for a copied query the builder cannot represent', () => {
+    // Copied query the real parser can't represent (histogram_quantile) -> Code
+    // mode so it shows as-is and is never clobbered.
     const { unmount } = render(
       <CreateMetricsMonitor
         onCancel={jest.fn()}
         onSave={jest.fn()}
         datasourceId="prom-1"
-        initialQuery="rate(http_requests_total[5m]) > 0.5"
+        initialQuery="histogram_quantile(0.9, rate(http_requests_total[5m]))"
       />
     );
     expect(
@@ -195,7 +197,20 @@ describe('CreateMetricsMonitor', () => {
     expect(document.querySelector('[data-test-subj="mockBuilderSetQuery"]')).toBeNull();
     unmount();
 
-    // Empty flyout -> Builder mode: builder shown, raw expression hidden.
+    // A copied query the builder CAN represent (rate + comparison) opens in
+    // Builder mode — the real parseExpr recognises it, so it isn't stuck in Code.
+    const rep = render(
+      <CreateMetricsMonitor
+        onCancel={jest.fn()}
+        onSave={jest.fn()}
+        datasourceId="prom-1"
+        initialQuery="rate(http_requests_total[5m]) > 0.5"
+      />
+    );
+    expect(document.querySelector('[data-test-subj="mockBuilderSetQuery"]')).not.toBeNull();
+    rep.unmount();
+
+    // Empty flyout opens in Builder mode (the default point-and-click experience).
     render(<CreateMetricsMonitor onCancel={jest.fn()} onSave={jest.fn()} datasourceId="prom-1" />);
     expect(document.querySelector('[data-test-subj="mockBuilderSetQuery"]')).not.toBeNull();
     expect(document.querySelector('[data-test-subj="metricsMonitorPromQlExpression"]')).toBeNull();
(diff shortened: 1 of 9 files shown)


lezzago:
## Before / after — Builder UX

Captured in the running dev app (Alert Manager → **Create metrics rule** → Query → Builder), same metric (`gen_ai_client_token_usage_total`) in both.

### Before (`origin/main`)
The Builder emitted only a bare selector — Metric + Label filter, then straight to *For duration*. **No way to express a threshold**, so a conditional alert could only be hand-written in Code mode.

before

### After — Reduce + Condition
Grafana-style **Reduce** (Average over a rolling **Window**) + **Condition** (IS ABOVE) + **Value**, with a live composed expression:
`avg_over_time(gen_ai_client_token_usage_total[5m]) > 5`

after-condition

### After — always-firing guard
Setting Condition back to *No condition* surfaces the warning (both Builder and Code modes):

after-warning

### After — range condition
*IS OUTSIDE RANGE* → From/To fields, composing `(avg_over_time(…[5m]) < 10 or avg_over_time(…[5m]) > 90)`:

after-range

---

**Note on the preview overlay:** the threshold line + "Would fire now" badge could not be shown against live data — this dev instance's Prometheus preview endpoint (`/api/alerting/prometheus/…/preview`) returns **502** (visible as "Could not run preview" in the warning screenshot). That's an environment/infra issue, not a code change; the overlay renders on a successful preview and is covered by the logic changes in `query_preview_results.tsx`.

_Screenshots hosted on the `evidence/grafana-condition-builder` branch of the fork (not part of this PR's diff)._

lezzago:
## Follow-up: Function (rate/increase) + Aggregation layers

Extended the builder so aggregation-over-function expressions like `sum(rate(gen_ai_client_token_usage_total[4m])) > 6` now round-trip — previously they fell through to "can't be represented — switch to Code".

The builder now composes four layers (see `prom_condition.ts`):
1. **Series** — metric + label matcher
2. **Function** — `rate` / `increase` / `delta` + the `_over_time` family, over a window
3. **Aggregate** — `sum` / `avg` / `min` / `max` / `count`, with optional `by (…)` / `without (…)` grouping
4. **Condition** — comparison / range

Verified in the dev app: pasting `sum(rate(gen_ai_client_token_usage_total[4m])) > 6` in Code mode and switching to Builder seeds **Function = Rate**, **Window = 4m**, **Aggregate = Sum**, **Condition = IS ABOVE**, **Value = 6**, and re-emits the identical expression.

after-aggregation

Round-trip + build/parse coverage for rate, aggregation, and grouping added to `prom_condition.test.ts` (all alerting suites green: 37 / 397).

lezzago:
## UX polish + Code-mode default

Ran a UI/UX design-review pass (EUI-only) and made the create-alert-rule experience smoother, especially for users arriving with a query.

**Default to Code mode.** The flyout now opens in Code — it renders any PromQL as-is, so a user who already has a query sees/edits it immediately with no Builder→Code hop. The **Editor** toggle (now labeled) switches to Builder for point-and-click. The "PromQL" tag is a neutral badge (was a BetaBadge, which wrongly implied experimental), and the Code help text is one line (the always-firing callout carries the nuance). The editor is monospaced.

code default

**Builder polish.** Fixed the Window unit control (the compressed `append` select was clipping the unit — now "5 m" reads clearly); added a **Transform (optional)** section so Function/Aggregate/Grouping read as one group; progressive disclosure (empty builder = just the Series row until a metric is picked); **Condition + Value** stay on one line so "IS ABOVE → 6" is one decision; and a labeled live **Expression** line.

builder full

Deferred (lower-value, noted): a full shared column-grid refactor and converting the grouping-labels field to a metadata-seeded combobox.

All alerting suites green (37 / 397); typecheck + lint clean. Commit `ea47ee93`.

lezzago:
## Before / after — highlighted (left = `origin/main`, right = this PR)

Orange box = the area on `origin/main`; green box = the same area after this PR.

### 1. Grafana-style condition builder (the feature)
Before, the metric-rule Builder went straight from the metric/label picker to *For duration* — no way to express a condition. After, it adds **Function** (rate/increase/…), **Aggregate** (sum/avg/… + by/without), and **Condition** (IS ABOVE / ranges / …), with a live composed `Expression`.

builder before/after

Toggle:

builder toggle

### 2. Rules table — internal scroll (fixes #2686 regression)
Before, the table grew past the viewport → the whole page scrolled and a nested table scrollbar appeared (double-scroll). After, the table region is capped to the viewport and scrolls internally with a sticky header.

table before/after

### 3. No more horizontal "side" scroll
Before, the filters panel and the table each showed a spurious horizontal scrollbar (highlighted strip). After, both are gone (a real x-scrollbar now appears only when the table genuinely can't fit, e.g. <1024px).

side-scroll before/after

_Captured at matched viewports (builder/table 1200×689, side-scroll 1440×900). Composites/GIF live on the `evidence/grafana-condition-builder` fork branch (not part of this PR's diff)._

riysaxen-amzn:
Nice design on the pure core — `buildExpr`/`parseExpr` being React-free with an exhaustive round-trip suite is the right call, and I verified the parser fails safe: every mis-capture I could construct (`sum(a) / sum(b) > 1`, `rate(m[5m]) + rate(n[5m])`, `a > 1 or b > 2`) falls through to `parseSelector`, which rejects it → `null` → builder stays inert. The `#2686` double-scroll fix (`height` cap + `min-height: 0` cascade) also reads correctly.

Non-blocking questions/nits:

**1. `prom_condition.ts` — range-threshold ordering:** `parseExpr` accepts `(x < 10 or x > 2)` as `outside` with `thresholdA=10, thresholdB=2` (low > high), and `buildExpr` will happily re-emit it. Is there UI-side validation that A ≤ B for the `outside`/`within` operators? If not, an inverted range silently produces an always-true (`outside`) or never-true (`within`) condition. A validation error (or auto-swap) in the builder row would close this.

**2. `prom_condition.ts` — window regex:** `RANGE_FN_RE` only accepts `\d+[smhdwy]` — compound durations (`1h30m`) and `ms` parse to `null`, dropping the whole expression to Code mode. That's the safe direction, so fine — but worth a one-line comment on the regex so a future "why doesn't my 90-minute window seed the builder" report has an answer. (Prometheus also accepts `90m`, so users have an escape hatch.)

**3. `isAlwaysFiring` heuristic edge:** a comparison with the `bool` modifier (`up > bool 0`) returns a 0/1 sample for every series — it always fires, but the heuristic sees the `>` and stays quiet. Conversely `absent(metric)` has no comparison and gets the warning even though it's a legitimate conditional alert. Both are edge cases and the callout is non-blocking, so no change requested — maybe just mention `bool` in the function's doc comment as a known limitation.

**4. Test comment inconsistency — `create_metrics_monitor.test.tsx`:** one test asserts "Empty flyout opens in Builder mode (the default point-and-click experience)", but the create-flow test around L369 says "The flyout opens in Code mode now, so switch to Builder to reach the mock builder" and clicks the Builder tab. If that second flyout is also empty, one of the two comments is stale — which is the actual default for an empty flyout? (If the second test seeds a query, ignore me — the hunk doesn't show the render props.)

**5. `sum by ()` normalization (observation, no change needed):** `parseExpr` accepts `sum by () (rate(m[5m]))` (grouping=`by`, labels=[]), and `buildExpr` re-emits it as `sum(rate(m[5m]))`. Semantically identical in PromQL, so the rewrite is harmless — just noting it's the one representable input that doesn't round-trip byte-for-byte.

Scope note: I reviewed the `prom_condition` core, both test files, and the SCSS in depth; the `PromQueryBuilder` component and preview changes I validated through the updated tests rather than line-by-line.

lezzago:
Thanks for the thorough read, @riysaxen-amzn — and for independently confirming the parser fails safe. Addressed in `bab332fc`:

**1. Range bounds ordering** — good catch. `buildExpr` now normalizes the bounds to `lo <= hi` for both `outside` and `within`, so an inverted range entered as A > B can no longer silently emit an always-true `outside` (`x < hi or x > lo`) or never-true `within`. Chose normalization over a blocking validation error so the natural "outside the range 2–10" reading just works regardless of which box holds which number; `parseExpr` reads the ordered bounds back. Added a unit test.

**2. Window regex** — added a comment on `RANGE_FN_RE` documenting that only a single-unit integer duration (`\d+[smhdwy]`) seeds the builder; a compound duration (`1h30m`) or `ms` parses to `null` and drops to Code mode by design (Prometheus still accepts `90m`, so there's an escape hatch).

**3. `isAlwaysFiring` edges** — documented the `bool` modifier (`up > bool 0` always fires but the `>` keeps the heuristic quiet) and `absent(metric)` (no comparison → gets the hint though it's a valid conditional) as known, acceptable limitations in the function's doc comment. The callout is a non-blocking hint, so no behavioral change.

**4. Test comment inconsistency** — fixed. The actual default for an empty flyout is **Builder**; the "opens in Code mode now" comment was stale (from when Code was the default). Corrected it — the `Builder` click in that test is now noted as defensive (a no-op when Builder is already selected).

**5. `sum by ()` → `sum()` normalization** — agreed, semantically identical in PromQL and harmless; leaving as-is. It's the one representable input that doesn't round-trip byte-for-byte, as you noted.

riysaxen-amzn (approved):

PR #2871, opened 2026-09-06
fix(alerting): clone + edit monitors in the unified Alerts view

lezzago:
## Summary

Fixes bugs in the unified Alerts view (Alert Manager → **Rules** tab) around **cloning and editing monitors**, hardens the clone flow across every OpenSearch monitor type, and includes small filter-panel/table UI polish. Scoped to the `dashboards-observability` alerting UI + its server rule-detail path.

### Bug fixes
1. **Blank edit page after cloning an OpenSearch monitor.** Cloning a cluster-metrics monitor and clicking **Edit** opened a blank classic-editor page. Root cause: the server rule-detail returned `raw` as the lossy `mapMonitor` projection, which coerces `monitor_type` (→ `query_level_monitor`) and flattens type-specific trigger wrappers; re-POSTing that shape on clone stored an invalid monitor, and the classic editor then crashed reading `inputs[0].search.indices`.
2. **Cloning any per-bucket / per-document monitor failed** with "Incompatible trigger for monitor type …" — same lossy-`raw` root cause (the `bucket_level_trigger` / `document_level_trigger` wrapper + its `parent_bucket_path`/`buckets_path` were dropped).
3. **Duplicate clone names.** Cloning the same monitor twice produced two identically-named copies.

**Fix:** `getOSRuleDetail` now returns the **faithful upstream monitor document** (new `OpenSearchBackend.getMonitorWithSource`, added to the interface + impl), so the clone re-creates the exact `monitor_type` + wrapped triggers for **every** OpenSearch monitor type (query / bucket / doc / ppl / cluster-metrics). The clone strips server-owned / response-derived fields that must not be re-POSTed (`id`, `version`, `owner`, `user`, `data_sources`, `item_type`, `associated_workflows`, `associatedCompositeMonitorCnt`, `last_run_context`, timestamps). `buildUniqueCloneName` yields `(Copy)`, `(Copy 2)`, … (and `-copy`, `-copy-2`, … for Prometheus), de-duped against existing rules **and** names issued in-flight this session (reconcile-based pruning, soft-delete aware), and trims by code points so a length cap can't split a surrogate pair.

### UI polish (unified Alerts view)
- **Forecaster** create/edit flyout showed **"Detector name"** help text/placeholder → now branches on rule type ("Forecaster name").
- Rules table **Datasource** column wrapped a long name mid-word → single-line ellipsis + full-name hover tooltip via the shared `TruncatedLabel`.
- Filter-panel **label-facet counts** read like a superscript (`severity2`) → now `severity (2)` (RTL-safe `margin-inline-start: $euiSizeXS`).
- Long filter **facet keys** (e.g. `deployment_environment`) clipped with no tooltip → now truncate + reveal the full name on hover **and keyboard focus** (WCAG 1.4.13; Esc dismisses; excludes `tabindex=-1` accordion wrappers).

### Under the hood
- Server: `getMonitorWithSource` returns the narrowed projection **and** the faithful upstream doc; `getOSRuleDetail` exposes the latter as `raw` (documented cast — OS `raw` carries wrapped triggers).
- Clone-name reservation set is bounded and reconciled against the fetched rules (mirrors `isRuleNameTaken`'s soft-delete filter).

## Visual evidence

Matched viewport/route/workspace. **BEFORE = `origin/main`, AFTER = this branch.**

### 1. Editing a cloned OpenSearch monitor
_BEFORE: the Edit page is blank (classic editor crashes). AFTER: the full monitor editor renders with the clone's fields._

Before/after — cloned monitor edit renders

| Before flow (origin/main) | After flow (this PR) |
|---|---|
| before flow | after flow |

### 2. Cloning the same monitor twice
_BEFORE: every clone is "… (Copy)" — indistinguishable duplicates. AFTER: unique names — "(Copy)", "(Copy 2)", …_

Before/after — unique clone names

| Before flow (origin/main) | After flow (this PR) |
|---|---|
| before flow | after flow |

### 3. Filter-panel & table polish
_BEFORE: label-facet counts hug the label ("severity2") and long datasource names wrap mid-word. AFTER: spaced "(n)" counts, and single-line datasource names with a hover/keyboard tooltip._

Before/after — filter-panel & table polish

## Testing
- Unit: full alerting + shared-component + server suites green (Jest + react-testing-library). Added coverage for the cloned-monitor edit target, all-OS-type clone (wrapped triggers + strip-list), doc-level clone, clone-name uniqueness + in-flight dedup, `getMonitorWithSource`, and `TruncatedLabel` keyboard reveal (ancestor focus, `tabindex=-1` exclusion, `label.control`, Esc).
- Manual: verified in the running dev app across cluster-metrics / query / bucket / doc / PPL / Prometheus rules — clone, edit-after-clone (editor renders), delete, and the create-from-Explore (Logs + Metrics) flows.

## Checklist
- [x] All commits DCO signed-off
- [x] Unit tests added/updated and passing
- [x] Screenshots / visual evidence attached (above)
- [x] Changes are backwards compatible

### DIFF
--- a/common/types/alerting/opensearch_types.ts
+++ b/common/types/alerting/opensearch_types.ts
@@ -319,6 +319,13 @@ export interface OpenSearchBackend {
   // Monitors — read-only methods.
   getMonitors(client: AlertingOSClient): Promise<OSMonitor[]>;
   getMonitor(client: AlertingOSClient, monitorId: string): Promise<OSMonitor | null>;
+  // Returns the narrowed `monitor` projection AND the faithful upstream `source`
+  // document (real monitor_type + wrapped triggers) — used by the clone flow so
+  // it can re-create a monitor without the lossy `mapMonitor` projection.
+  getMonitorWithSource(
+    client: AlertingOSClient,
+    monitorId: string
+  ): Promise<{ monitor: OSMonitor; source: Record<string, unknown> } | null>;
   runMonitor(client: AlertingOSClient, monitorId: string, dryRun?: boolean): Promise<unknown>;
   searchQuery(
     client: AlertingOSClient,
--- a/public/components/alerting/__tests__/alarms_page.test.tsx
+++ b/public/components/alerting/__tests__/alarms_page.test.tsx
@@ -639,6 +639,168 @@ describe('AlarmsPage', () => {
     expect((inner.actions as Array<Record<string, unknown>>)[0].id).toBeUndefined();
   });
 
+  it('strips server-owned / response-derived fields from the clone payload', async () => {
+    // The faithful upstream doc carries identity/audit/ownership/principal
+    // fields and GET-only enrichments that must NOT be re-POSTed on create.
+    const fakeRaw = {
+      id: 'mon-strip',
+      name: 'Strip Mon',
+      type: 'monitor',
+      monitor_type: 'query_level_monitor',
+      inputs: [{ search: { indices: ['logs-*'], query: {} } }],
+      triggers: [{ query_level_trigger: { id: 't', name: 't', actions: [] } }],
+      // fields that must be dropped:
+      version: 7,
+      last_update_time: 123,
+      enabled_time: 456,
+      schema_version: 5,
+      owner: 'alerting',
+      user: { name: 'creator', backend_roles: ['admin'], roles: ['all_access'] },
+      data_sources: { tenant: 't1' },
+      item_type: 'query_level_monitor',
+      associated_workflows: [{ id: 'wf1' }],
+      associatedCompositeMonitorCnt: 2,
+      last_run_context: { lastFired: 0 },
+    };
+    mockGetRuleDetail.mockResolvedValue({ raw: fakeRaw });
+    mockCreateMonitor.mockResolvedValue({ id: 'new-strip' });
+
+    await act(async () => {
+      render(<AlarmsPage {...defaultProps} />);
+    });
+    fireEvent.click(screen.getByTestId('alertManagerTabs-rules'));
+    const tableProps = mockMonitorsTable.mock.calls[mockMonitorsTable.mock.calls.length - 1][0] as {
+      onClone: (monitor: unknown) => Promise<void>;
+    };
+    await act(async () => {
+      await tableProps.onClone({
+        id: 'mon-strip',
+        name: 'Strip Mon',
+        datasourceId: 'ds-1',
+        definitionType: 'monitor',
+      });
+    });
+
+    const payload = mockCreateMonitor.mock.calls[0][0] as Record<string, unknown>;
+    for (const stripped of [
+      'id',
+      'version',
+      'last_update_time',
+      'enabled_time',
+      'schema_version',
+      'owner',
+      'user',
+      'data_sources',
+      'item_type',
+      'associated_workflows',
+      'associatedCompositeMonitorCnt',
+      'last_run_context',
+    ]) {
+      expect(payload[stripped]).toBeUndefined();
+    }
+    // …while legitimate create fields still round-trip.
+    expect(payload.monitor_type).toBe('query_level_monitor');
+    expect(payload.type).toBe('monitor');
+  });
+
+  it('clones a doc-level monitor preserving the document_level_trigger + doc-only fields', async () => {
+    const fakeRaw = {
+      id: 'mon-doc',
+      name: 'Doc Mon',
+      type: 'monitor',
+      monitor_type: 'doc_level_monitor',
+      // doc-level create-time fields that MUST survive in ...rest:
+      delete_query_index_in_every_run: true,
+      should_create_single_alert_for_findings: false,
+      inputs: [
+        {
+          doc_level_input: {
+            description: '',
+            indices: ['logs-*'],
+            queries: [{ id: 'q1', name: 'q1', query: 'x:1' }],
+          },
+        },
+      ],
+      triggers: [
+        {
+          document_level_trigger: {
+            id: 'trig-doc',
+            name: 'd1',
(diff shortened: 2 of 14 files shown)


lezzago:
Addressed the **medium** security finding on `alert_detail.ts` (RBAC/tenant fields in `raw`) in 822d7ad4.

`getOSRuleDetail` needs the faithful upstream monitor doc as `raw` only for the clone flow (real `monitor_type` + wrapped triggers). The clone never needs the security principal, so `getMonitorWithSource` now strips `user` / `owner` / `data_sources` **server-side** before `source` is returned — those RBAC role assignments and the tenant are no longer transmitted to the browser. The client-side strip before the clone re-POST stays as defense-in-depth. Added a backend test asserting `user`/`owner`/`data_sources` are absent from `source`.

riysaxen-amzn:
Thanks for the thorough PR and the before/after evidence — the root-cause fix is correct. The old code re-POSTed the lossy `mapMonitor` projection (coerced `monitor_type`, flattened trigger wrappers), which explains both the blank edit page and the "Incompatible trigger" failures. Returning the faithful upstream doc via `getMonitorWithSource` and stripping server-owned fields on the clone is the right shape, and the test coverage is solid (all monitor types, the strip-list, the in-flight dedup race, and the a11y regressions).

A few non-blocking questions/nits before I approve:

**1. `opensearch_backend.ts` — `getMonitorWithSource`:** `source` is sanitized (`user`/`owner`/`data_sources` stripped), but the `monitor` projection alongside it is still built from the unstripped body: `this.mapMonitor(resp.body._id, resp.body.monitor)`. Can you confirm `mapMonitor` doesn't pass any of those three fields through into `OSMonitor`? If it's a fixed-field projection this is fine as-is — just want the "server-side removal is the authoritative guard" comment to hold for both returned values. If `mapMonitor` ever spreads unknown keys, passing `safeMonitor` into it would close that path cheaply.

**2. `alarms_page.tsx` — `cloneNameKey`:** the reservation key normalizes with `trim().toLowerCase()`, but it's paired with `isRuleNameTaken(candidate, dsId)`. If `isRuleNameTaken` compares case-sensitively, the two checks have different semantics — worst case a rare duplicate slips through when two names differ only by case. Worth aligning the normalization on both sides (or a one-line comment stating `isRuleNameTaken` is already case-insensitive).

**3. Clone `enabled` state (product-level):** the OpenSearch clone inherits `enabled` from the source (via `...rest`), so cloning an enabled monitor starts a second live monitor firing the same notifications immediately; the Prometheus branch hardcodes `enabled: true`. Was enabled-on-clone deliberate? Starting clones disabled (letting the user review before enabling) is arguably safer — just flagging the double-notification window.

**4. Nit — `alert_detail.ts` raw cast:** the `source as unknown as UnifiedRule['raw']` cast is well-documented, but any *existing* consumer of `detail.raw` for OS rules that assumed the flattened shape would now silently get wrapped triggers. From this diff the only raw readers I can see are the two clone branches (both updated) — if you've grepped for other `detail.raw` readers on the OS path and found none, a one-liner confirming that would close the loop.

**5. Nit — `doc_level_input.queries[].id` not stripped:** the clone strips trigger and action ids but keeps doc-level query ids (your doc-level test keeps `q1`). Query ids are monitor-scoped so re-POSTing them should be fine (and your manual doc-level clone test passing supports that) — just confirming that was intentional rather than an oversight in the id-strip pass.

Everything else — the code-point-safe truncation, the bounded attempt loop with timestamp fallback, the `.altFacetGroupCount` RTL-safe margin scoped to group keys only, and the detector/forecaster copy branch — looks correct. The medium security finding (RBAC/tenant fields in `raw`) is already addressed server-side in 822d7ad with a test asserting absence.

lezzago:
Thanks @riysaxen-amzn — went through all five. Changes in c9a77576.

**1. `getMonitorWithSource` / `mapMonitor`** — confirmed `mapMonitor` is a fixed-field projection (explicit object literal: id/type/monitor_type/name/enabled/schedule/inputs/triggers/last_update_time/schema_version — no `...spread`), so it never passes `user`/`owner`/`data_sources` through today. To make the "server-side removal is authoritative" guarantee hold even if that ever changes, I now build the `monitor` projection from the sanitized body too (`mapMonitor(_id, safeMonitor)`). Belt-and-suspenders as you suggested.

**2. `cloneNameKey` vs `isRuleNameTaken`** — they're aligned: `isRuleNameTaken` compares `r.name.trim().toLowerCase() === name.trim().toLowerCase()` (case-insensitive), same normalization as `cloneNameKey`, so a case-only duplicate can't slip through. Added a one-line comment stating this on `cloneNameKey`.

**3. Clone `enabled` state** — good call, changed. Clones now start **disabled** on both paths (the OpenSearch clone explicitly overrides the source's inherited `enabled`; the Prometheus branch flips `enabled: true` → `false`), so a clone can't fire the same notifications before it's reviewed. Both paths are now consistent, and there are tests asserting `enabled === false` for each.

**4. Other OS-path `detail.raw` readers** — grepped: the only readers are the two clone branches (both updated), `monitor_detail_flyout` (reads `.inputs` / `.monitor_type` only), and `edit_monitor` (reads `raw` **only** in the Prometheus branch as `raw.rules[...]`; OS in-place edit is PPL-only). `grep raw.triggers` on the OS path → zero hits, so nothing consumes the wrapped-vs-flat trigger shape. Loop closed.

**5. `doc_level_input.queries[].id`** — intentional. Query ids are monitor-scoped, so re-POSTing them on create is fine (the `doc_level_monitor` clone test passing confirms it); only trigger/action ids collide on create, which is why just those are stripped.

CI re-running on c9a77576.

riysaxen-amzn (approved):

PR #2873, opened 2026-09-08
[CVE] Bump linkify-it to ^5.0.2 and fast-uri to ^3.1.7 (6 CVEs)

TackAdam:
### Description
Remediates **six** dependency CVEs flagged against this plugin by bumping two transitive packages to their fixed releases via `resolutions`, then regenerating `yarn.lock` through the OSD bootstrap flow (not hand-edited).

| CVE | Package | Was | Now | Severity |
|---|---|---|---|---|
| CVE-2026-48801 | linkify-it | 3.0.3 | 5.0.2 | High |
| CVE-2026-59887 | linkify-it | 3.0.3 | 5.0.2 | High |
| CVE-2026-75931 | fast-uri | 3.1.5 | 3.1.7 | High |
| CVE-2026-75899 | fast-uri | 3.1.5 | 3.1.7 | High |
| CVE-2026-75975 | fast-uri | 3.1.5 | 3.1.7 | High |
| CVE-2026-76172 | fast-uri | 3.1.5 | 3.1.7 | High |

Both packages are transitive; neither is imported directly in `dashboards-observability` source.

### linkify-it (CVE-2026-48801, CVE-2026-59887)
Two high-severity quadratic-complexity / ReDoS advisories:
- **CVE-2026-48801** (GHSA-22p9-wv53-3rq4) — `LinkifyIt#match` scan loop has O(n²) algorithmic complexity on crafted text. Present since the initial 2014 commit through v5.0.0; fixed in **5.0.1**.
- **CVE-2026-59887** (GHSA-v245-v573-v5vm) — quadratic DoS via the `mailto:` validator scan-loop, which re-scans remaining input at every `mailto:` occurrence. Fixed in **5.0.2**.

Pulled in only through:
```
@nteract/outputs → ansi-to-react@^6.0.5 (6.2.6) → linkify-it@^3.0.3
```
The patched `5.0.2` is a major bump that breaches the transitive `^3.0.3` range, so this adds a `resolutions` override (`"linkify-it": "^5.0.2"`) and regenerates the lockfile. The `^5` constraint deliberately stays on the 5.x line and avoids `6.0.0`, which removed the `.pretest()`/`.compile()` APIs.

**Does the major bump break `ansi-to-react`? No — verified against the installed 5.0.2 source:**
- **CJS interop preserved** — `linkify-it@5.0.2` main is `build/index.cjs.js` with `module.exports = LinkifyIt` and the no-`new` factory guard (`if (!(this instanceof LinkifyIt)) return new LinkifyIt(...)`), identical to 3.0.3, so `ansi-to-react`'s `__importDefault(require("linkify-it"))` still resolves.
- **API surface intact** — `ansi-to-react` only calls `linkifyIt({ fuzzyEmail: false }).tlds(["io"], true)` then `.pretest()` / `.match()`, iterating match objects' `.index` / `.lastIndex` / `.url`. All are present in 5.0.2, and the `fuzzyEmail` option is still honored.

### fast-uri (CVE-2026-75931, CVE-2026-75899, CVE-2026-75975, CVE-2026-76172)
Four high-severity host-confusion / SSRF advisories in the URI parser:
- **CVE-2026-75931** (GHSA-5jgf-p345-68v8) — host confusion via skipped IDN canonicalization on scheme-relative references.
- **CVE-2026-75899** (GHSA-fph4-wmhf-6fwf) — SSRF via repeated hostname percent-decoding.
- **CVE-2026-75975** (GHSA-f65p-4m7j-42xc) — SSRF via malformed IPv6 normalization.
- **CVE-2026-76172** (GHSA-jqff-g426-hqxp) — host confusion via percent-encoded scheme normalization.

All four are fixed in the **3.1.6** release of the 3.x line. Pulled in through:
```
ajv@^8.18.0 (8.20.0) → fast-uri@^3.0.1
```
`3.1.7` (latest 3.x, satisfying ajv's `^3.0.1`) is pinned via `resolutions` to guard against lockfile regression. This is a patch-level bump with no API impact on `ajv`.

### Verification
- `yarn osd bootstrap` completes successfully; both packages resolve to a single deduped copy (`linkify-it@5.0.2`, `fast-uri@3.1.7`) with no stale versions in `yarn.lock`.
- Plugin Jest suites covering both real consumers pass — **65 tests, 6 snapshots green**:
  - `public/components/notebooks/.../para_output.test.tsx` (linkify-it path via `@nteract/outputs`/`ansi-to-react`)
  - `server/adaptors/integrations/__test__/validators.test.ts` (fast-uri path via `ajv`)

### Issues Resolved
CVE-2026-48801, CVE-2026-59887, CVE-2026-75931, CVE-2026-75899, CVE-2026-75975, CVE-2026-76172

### Check List
- [x] All tests pass
- [x] New functionality does not require documentation


ps48 (approved):
### bump linkify-it to ^5.0.2 and fast-uri to ^3.1.7 (6 CVEs)

Clean, well-documented CVE remediation. Verified the mechanics against the regenerated lockfile and the dependency tree, and it all holds:

- **`fast-uri` 3.1.5 -> 3.1.7** is a patch-level bump within the 3.x line. Its only consumer here is `ajv@8.20.0` (`fast-uri "^3.0.1"`), which `3.1.7` satisfies, so no API risk. Deduped to a single copy (`fast-uri@^3.0.1, fast-uri@^3.1.7`).
- **`linkify-it` 3.0.3 -> 5.0.2** is the one major bump, and it's the only part with real risk. Confirmed the sole consumer in this plugin's tree is `ansi-to-react@6.2.6` (`linkify-it "^3.0.3"`) via `@nteract/outputs`; there is no `markdown-it` in this tree, so nothing else is forced across the major. `uc.micro` correctly floats to `2.1.0` (linkify-it 5's `^2.0.0` dep) with no orphaned `uc.micro@^1` entry left behind. The `^5` (not `^6`) ceiling to preserve `.pretest()` / `.compile()` is the right call, and the point-in-time verification of `ansi-to-react`'s usage (`.tlds` / `.pretest` / `.match` / `fuzzyEmail`) against 5.0.2 matches what I see in the consumer.
- Both additions land in `resolutions` (not new direct deps), which is correct for forcing a transitive fix, and this plugin carries its own `yarn.lock`, so the override actually takes effect on bootstrap. The `string-width` / `strip-ansi` / `wrap-ansi` `-cjs` alias consolidation in the lockfile is benign regeneration churn, not a version change.

No `p0`, no `p1`. The lockfile is internally consistent (single deduped copy of each bumped package, no stale versions). One optional aside: `package.json` still ends without a trailing newline (pre-existing, and the diff carries the `\ No newline at end of file` marker) — since this PR is already editing the `resolutions` block, adding the newline would retire that perpetual marker, but it's purely cosmetic and not introduced here. LGTM.

PR #2876, opened 2026-09-09
refactor(alerting): dead-code sweep of the unified Alerts view

lezzago:
### Description

Dead-code sweep of the unified Alerts view (Alert Manager). This is **PR 1** of a
sequenced cleanup — pure deletions with **no behavior change**, split so each commit is
independently reviewable. Net **−4,526 lines** across 39 files.

Every removal was grep-verified to have no reachable importer/caller before deletion.

**Removed (by commit):**

1. **`validatePromQL` extraction** — moved the one live export of `promql_editor.tsx` into a
   small `promql_validation.ts` so the editor file could be deleted.
2. **Unreachable monitor-creation UI:**
   - `monitor_template_wizard/` (~1,600 LOC) — gated behind `TEMPLATE_MODE_ENABLED = false`; the
     only `setCreationMode('ai')` call site lived inside that dead block, so the wizard render
     branch and the `onBatchSave`/`handleBatchCreateMonitors` batch-create plumbing were never
     reachable.
   - `promql_monaco_editor.tsx` + the `PromQLEditor` component/autocomplete engine in
     `promql_editor.tsx` — never rendered (Prometheus uses `PromQueryBuilder`, PPL uses
     `PplQueryEditor`); kept alive only by stale `jest.mock` stubs.
   - `metric_browser.tsx` — orphaned when #2870's condition builder replaced the metric-picker UX.
3. **Dead hooks / APIs / scaffolding:** `useRules`, `useAlertDetail`, `useAlertmanagerConfig`
   (no importers); `AlertService.getPaginatedRules`/`getPaginatedAlerts` + `PaginationParams`
   (no callers); `server/services/alerting/constants.ts` (`PLUGIN_ID`/`PLUGIN_NAME`, never
   imported); `UnifiedRule.suppressionRules`/`notificationRouting` + their types (every producer
   wrote `[]`, nothing read them — the Routing tab fetches Alertmanager config directly).
4. **Dead form transforms:** the logs-monitor transform path and `transformMetricsFormToPayload`
   in `form_transforms.ts` (no UI builds a logs/metrics form through them); the live PPL exports
   are untouched. `sortRules` in `filter.ts` (referenced only by its own test).
5. **Tested-only / vestigial:** `validators.formatDuration`; `time_range.parseDateMath` +
   `dateMathToDSLString`; the unselectable `CUSTOM_FEATURE_TYPE` ("Custom expression") option in
   `create_ad_rule_flyout.tsx` — its feature-type `<EuiSelect>` is `disabled` and hard-pinned to
   `SIMPLE_FEATURE_TYPE` with no `onChange`, so the option was never selectable and its value was
   never read.
6. **Unreachable table-column branches:** `visibleColumns` is hard-pinned to `DEFAULT_VISIBLE`
   (name/status/severity/monitorType/healthStatus/datasource) via `useMemo` with no setter, so
   `buildTableColumns`' branches for labels/createdBy/createdAt/lastModified/lastTriggered/
   destinations/query/group/`label:<key>` were never reached. The `createdBy`/`label:` **filter**
   features (`monitors_filters_panel`, `matchesSearch`) are unrelated and untouched.

**Deliberately deferred** (not in this PR): consolidating the divergent `FilterState`/
`matchesSearch`/`matchesFilters` copies (a refactor, consumed via barrel incl. SLO) and the
`buildExpr` re-export (an unused binding, not dead code).

### UI impact / visual evidence

**No visible change.** Both user-facing-looking removals are inert by construction, and the
before/after composites below (BEFORE = `origin/main`, AFTER = this branch, captured on the same
dev server against identical data) confirm the rendered UI is pixel-identical.

**1. Rules table — column set unchanged.** `visibleColumns` is hard-pinned to `DEFAULT_VISIBLE`
(6 columns) with no setter, so the 9 removed `buildTableColumns` branches were never rendered. The
table shows the same Name · Status · Severity · Type · Health · Datasource before and after.

Rules table before/after

**2. AD rule feature-type select — unselectable option removed.** The "Find anomalies based on"
`<EuiSelect>` is `disabled` and pinned to `simple_aggs` ("Field value") with no `onChange`. The
removed "Custom expression" option only ever existed in the DOM `<option>` list (verified:
`options = [Field value, Custom expression]` on `origin/main` → `[Field value]` on this branch),
never visible to users since the disabled select can't be opened.

AD feature-type select before/after

_Flow GIFs are intentionally omitted: with a provably no-visible-change diff, a workflow recording
would show identical frames before and after. The DOM-level proofs above (rendered column set;
`<option>` counts) are the substantive evidence._

### Issues Resolved

N/A — internal cleanup.

### Check List
  - [x] All tests pass (64 alerting jest suites, 899 tests); no type errors introduced.
- [x] New functionality has been documented. (N/A — deletions only.)


TackAdam:
Three low-severity cleanup observations (none blocking):

1. `create_metrics_monitor.tsx:2567` — In Builder mode with a metric selected and no condition, the shared always-firing `EuiCallOut` duplicates `PromQueryBuilder`'s own "No condition set" hint. Two redundant warnings for the same state.  
   **Fix:** Gate the callout to code mode.

2. `query_preview_results.tsx:161` — `conditionMet()` / `thresholdValues()` use `thresholdA ?? 0`, so an op with an undefined threshold evaluates against `0`. Currently safe (only caller is `parseExpr`, which always sets a threshold when `op !== 'none'`), but a future op-only caller would get a misleading "Would fire now" badge.  
   **Fix:** Treat a non-finite threshold as no-evaluation.

3. `truncated_label.tsx:404` — Each `TruncatedLabel` attaches focus/blur listeners to its nearest focusable ancestor; multiple labels under one control would all reveal at once. Fine today (≤1 label per control), but fragile if markup changes.

TackAdam (approved):

lezzago:
Thanks for the review — all three addressed in fc3e2e82.

1. **`create_metrics_monitor.tsx` — duplicate always-firing warning.** Gated the shared `EuiCallOut` to code mode (`queryMode === 'code' && isAlwaysFiring(...)`). In builder mode `PromQueryBuilder` renders its own "No condition set" hint, so the two no longer overlap. Updated the "both modes" comment accordingly.

2. **`query_preview_results.tsx` — `thresholdA ?? 0`.** `conditionMet()` now returns `undefined` ("not evaluable") when the op's threshold(s) aren't finite, instead of comparing against a defaulted `0`. The badge is already hidden when `wouldFire === undefined`, so a future op-only caller gets **no badge** rather than a misleading "would fire now". Range ops (`outside`/`within`) require both `thresholdA` and `thresholdB` finite.

3. **`truncated_label.tsx` — shared focusable ancestor.** Added a guard so at most one `TruncatedLabel` per focusable control wires the focus reveal (first mounted owns it, tracked via a `data-obs-truncated-focus-owner` attribute, cleaned up on unmount). Multiple labels under one control can no longer all reveal on focus; hover reveal stays per-label and unchanged. Defensive today (≤1 label per control), as you noted.

Verified the 3 affected jest suites pass (26 tests). These were pre-existing observations outside the dead-code diff, but small enough to fold in here since they were raised on this PR.

PR #2877, opened 2026-09-09
fix(alerting): stop clone/edit from corrupting Prometheus rule expressions

lezzago:
### Description

The unified Alerts view could silently change or destroy a Prometheus rule's definition while reporting success. Two fixes:

**1. Clone rewrote the PromQL expression.** `handleCloneRule` stripped a trailing comparison off the stored expression and re-appended a separately-parsed operator/threshold. Because the strip matched the *last* comparison while the parse read the *first* (or fell back to `> 0`), any non-trivial expression was corrupted:

| Source expression | Old clone produced |
|---|---|
| `sum(rate(x[5m]))` (no comparison) | `sum(rate(x[5m])) > 0` |
| `(a) > 0.8 and cap > 0` | `(a) > 0.8 and cap > 0.8` |
| `errors > 1e-05` | `errors > 1` |

Clone now sends the stored expression **verbatim**, matching the create/edit paths (the PromQL expression is itself the complete alert condition).

**2. Edit always set `overwrite: true`.** Renaming a rule onto a name already used by a *different* rule in the same group silently overwrote that other rule. Edit now sets `overwrite` only for an **in-place edit** (unchanged group + name — the rule replacing itself). A rename or group move sends no overwrite, so the server's collision guard returns 409 and the conflict surfaces instead of causing silent data loss.

### Before / after evidence

Reproduced on a live Prometheus datasource by cloning `genairule` (`sum(rate(gen_ai_client_token_usage_total[4m]))`) and reading the clone's stored expression back from the rule detail (and the ruler API).

**1. Cloning a Prometheus rule — the stored PromQL expression.** BEFORE (origin/main) the clone gains a spurious `> 0`; AFTER (this PR) it is preserved verbatim.

clone before/after

**Flow — BEFORE (origin/main):** clone appends `> 0`.

before flow

**Flow — AFTER (this PR):** clone preserves the expression.

after flow

> The edit `overwrite` fix has no clean single-screen visual (it prevents a cross-datasource collision from silently clobbering a different rule). It is verified at the API layer — an `overwrite: true` re-POST replaces a same-named rule (`up == 1` → `up == 2`) while a non-overwrite collision correctly returns **409** — and by the regression tests below.

### Testing

`prometheus_rule_lifecycle.test.tsx` — added regression tests:
- Clone preserves an expression with no comparison (no spurious `> 0`), an interior-comparison expression (trailing threshold unchanged), and a scientific-notation threshold.
- Edit overwrites for an in-place edit; a rename does **not** overwrite and removes the old copy.

All alerting suites pass; typecheck + lint clean.

### Follow-ups

- #2878 — Prometheus create/edit/clone only round-trip a fixed field set (unmodeled ruler fields like `keep_firing_for` are dropped). Out of scope here; the expression corruption is fixed in this PR.

### Check List
- [x] All tests pass
- [x] Screenshots/GIFs of before & after included

### DIFF
--- a/public/components/alerting/__tests__/prometheus_rule_lifecycle.test.tsx
+++ b/public/components/alerting/__tests__/prometheus_rule_lifecycle.test.tsx
@@ -403,6 +403,17 @@ describe('Prometheus rule edit — overwrite only for an in-place edit', () => {
     // Rename removes the old copy after the new one is created.
     expect(mockDeletePrometheusRule).toHaveBeenCalledWith('ds-1', 'OldName', 'OldName');
   });
+
+  it('does NOT overwrite (and does not delete) when the edited rule is not in the loaded list', async () => {
+    // Edge (stale list / background-refetch race): if `rules.find(id)` misses,
+    // we must default to the SAFE path — no overwrite (avoid clobbering a rule
+    // that occupies the target name) and no delete (a guessed old-name delete
+    // could remove the rule we just created).
+    seedRule({ id: 'someOtherRule', name: 'Unrelated', group: 'Unrelated', datasourceId: 'ds-1' });
+    const payload = await editAndGetPayload(promForm('GhostRule'), 'missing-id');
+    expect(payload.overwrite).toBeUndefined();
+    expect(mockDeletePrometheusRule).not.toHaveBeenCalled();
+  });
 });
 
 describe('Prometheus rule delete', () => {
--- a/public/components/alerting/__tests__/prometheus_rule_lifecycle.test.tsx
+++ b/public/components/alerting/__tests__/prometheus_rule_lifecycle.test.tsx
@@ -72,7 +72,13 @@ jest.mock('../notification_routing_panel', () => ({
   NotificationRoutingPanel: () => <div data-test-subj="routingPanel" />,
 }));
 jest.mock('../create_monitor', () => ({ CreateMonitor: () => null }));
-jest.mock('../create_monitor/edit_monitor', () => ({ EditMonitor: () => null }));
+const mockEditMonitor = jest.fn();
+jest.mock('../create_monitor/edit_monitor', () => ({
+  EditMonitor: (props: unknown) => {
+    mockEditMonitor(props);
+    return <div data-test-subj="editMonitor" />;
+  },
+}));
 jest.mock('../alert_detail_flyout', () => ({ AlertDetailFlyout: () => null }));
 
 const mockGetRuleDetail = jest.fn();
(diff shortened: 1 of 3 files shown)


riysaxen-amzn:
This is a clean, surgical fix — I verified the removed strip-regex (`/\s*(>|>=|<|<=|==|!=)\s*[\d.]+(?:[eE][+-]?\d+)?\s*$/`) against the three corruption examples in the description and they all reproduce, and the new `isInPlaceEdit` gate composes correctly with the existing create-first/delete-second rename flow (a 409 on rename now leaves the original rule untouched — fail-safe). Approving with four non-blocking questions/nits:

**1. Clone payload no longer carries `operator`/`threshold` — server contract check:** the tests assert both keys are absent from the clone POST. Can you confirm the server's create route treats them as purely optional (i.e., it never composed `query + operator + threshold` into the stored expr)? Your live before/after evidence strongly suggests yes — a one-liner confirming the server ignores absent operator/threshold would close the loop for future readers.

**2. `originalRule` not-found fallback defaults to in-place:** if `rules.find((r) => r.id === ruleId)` ever misses (stale list, background refetch race), the fallbacks make `originalName = promForm.name`, so a rename in that edge is classified as in-place → `overwrite: true` → the exact silent-clobber this PR fixes. The comment says the rule "is always present in the loaded `rules`", which is probably true today — but since the safe default costs nothing, consider falling back to `overwrite: false` (worst case: an in-place edit surfaces a spurious 409 the user can retry) rather than `overwrite: true` (worst case: silent data loss).

**3. 409 surfacing UX:** with overwrite off for renames, the collision now reaches the user — does the edit error path map the 409 to something readable ("a rule named X already exists in this group") or does it show a generic "Failed to update rule" toast? If generic, a follow-up to special-case 409 would make the new guard actually actionable.

**4. Observation (no change requested):** the verbatim principle now covers the expression, but the clone still re-derives `forDuration`/`evaluationInterval` with `'5m'`/`'1m'` fallbacks and rebuilds labels/annotations from `raw`. If `detail`/`raw` ever lack those fields, a clone could still differ from its source in evaluation cadence. Fine for this PR's scope — just noting the remaining non-verbatim surface in case a future "clone changed my rule" report lands there.

riysaxen-amzn (approved):

lezzago:
Thanks for the careful review! Addressed in `fc9aa493`:

**1. Server contract — confirmed.** `buildRuleGroup` composes `query + operator + threshold` only when *both* are defined, otherwise it uses `payload.query` verbatim (`server/routes/alerting/mutations/prometheus_handlers.ts:60-63`), and both are `schema.maybe` in the route body (`prometheus_routes.ts`). So the clone POST omitting them stores `expr = query` unchanged — matching the live before/after.

**2. `originalRule` not-found fallback — fixed (good catch).** You're right that the old fallback classified that edge as in-place → `overwrite: true`. Reworked to a three-state gate: overwrite **only** when the original rule is found *and* its `(group, name)` is unchanged; a rename/move **or** a not-found lookup now sends no overwrite. I also gated the old-copy delete on `originalRule` being known — otherwise defaulting overwrite off would have made the delete run against the guessed `(newName, newName)` and remove the just-created rule. Added a regression test for the not-in-loaded-list edge (`overwrite` undefined, no delete).

**3. 409 surfacing — improved.** The edit catch now special-cases a collision (HTTP 409 / "already exists") into an actionable title — *`A rule named "X" already exists in this group. Rename it or pick a different group.`* — while keeping the raw server message as the toast detail.

**4. Non-verbatim cadence/labels surface — filed as follow-up #2878.** Agreed it's out of scope here. As you noted the `for`/`evaluationInterval`/labels/annotations already come from the source (the `5m`/`1m` fallbacks are defensive and the server applies the same defaults). The broader gap — create/edit/clone only round-trip a fixed field set, so unmodeled ruler fields like `keep_firing_for` are dropped on any edit — is tracked in #2878.

PR #2879, opened 2026-09-09
[Alerting] Consolidate unified-alerting a11y + theme + i18n (supersedes #2827, #2825, #2839)

lezzago:
## Summary

Consolidates the three overlapping "unified Alerting a11y + theme + i18n" drafts — **#2827** (Alerts list), **#2825** (Alert flyout + SLO pivot), **#2839** (Rules-table + facet labels) — into a single branch rebased cleanly onto current `main`.

Those three were all cut from a shared base and **never rebased**, so they conflict with each other (and with `main`) on the shared primitives `alert_colors.ts` / `enum_labels.ts` / `time_format.ts`. Those primitives have since landed on `main` in a **superset** form (identical exports + extra `getMonitorStateLabel`), so this PR keeps `main`'s versions and re-applies only the **render-wiring** on top — no duplicate primitives.

## What this does
- **Alerts list** (`alerts_dashboard.tsx`): labeled severity badge (fixes the color-only WCAG 1.4.1 / 4.1.2 gap), Title-cased state, theme-aware colors via `getSeverityHex`/`getStateHex`, responsive columns; honors a `slo_id:<id>` label deep-link.
- **Alert flyout** (`alert_detail_flyout.tsx`): human-readable enums, timezone-aware timestamps, named alert source, linkified runbook annotations; adds `linkify_annotation.tsx`.
- **Rules table** (`monitors_table_columns.tsx`, `monitors_filters_panel.tsx`): Title-cases the Status/Severity/Health **cells and facet filters** to match the Alerts tab. Preserves `main`'s optimistic "pending" row badge.

## Visual evidence
BEFORE = `origin/main` (merge base), AFTER = this branch. Identical viewport (1440×900), route, data, and datasource selection in each pair.

### 1. Alerts list — severity is a labeled badge, not a color-only dot
Alerts list before/after

### 2. Alert detail flyout — human-readable enums + timezone-aware timestamps
Alert flyout before/after

### 3. Rules table & facets — Title-cased Status / Severity / Health
Rules table before/after

### Flow — walking Alerts list → alert detail → Rules table
| BEFORE (origin/main) | AFTER (this PR) |
|---|---|
| before flow | after flow |

## Supersedes (do NOT merge those)
Supersedes **#2827**, **#2825**, **#2839** — they will be **closed after this PR merges** (kept open until then so nothing is lost). This PR is the rebased, conflict-free union of all three.

## Testing
- Plugin Jest suite (`jest --config ./test/jest.config.js` over the alerting + apm/slos trees): **108 suites / 1173 tests pass**; full CI (build-linux/windows/macos, Lint, DCO, Cypress, FTR) green.
- `node scripts/eslint` clean on the hand-merged files.
- Two test helpers updated to match `main`'s pending-row status renderer (row `item` arg; badge-scoped "Pending" assertion).

## Notes for reviewers
- Opened as **draft**. Commit authorship from the original three PRs is preserved.
- Evidence media is hosted on a fork asset branch (`evidence/2879-alerts-a11y`), off the reviewable diff.

### DIFF
--- a/common/services/alerting/__tests__/filter.test.ts
+++ b/common/services/alerting/__tests__/filter.test.ts
@@ -168,8 +168,18 @@ describe('filter', () => {
     it('matches a single labelKey:value term against the label, not as a substring', () => {
       expect(alertMatchesSearch(alert, 'slo_id:abc-123')).toBe(true);
       expect(alertMatchesSearch(alert, 'slo_id:abc')).toBe(true); // includes()
-      expect(alertMatchesSearch(alert, 'slo_id:zzz')).toBe(false);
-      expect(alertMatchesSearch(alert, 'other:abc-123')).toBe(false); // wrong key
+      expect(alertMatchesSearch(alert, 'slo_id:zzz')).toBe(false); // real key, no value match
+    });
+
+    it('falls back to substring match when the colon-term key is not a label', () => {
+      // `other` is not a label on the alert, so the term is treated as free text.
+      const withColonInMessage = {
+        name: 'HighCpu',
+        message: 'saw error:timeout in logs',
+        labels: { slo_id: 'abc-123' },
+      };
+      expect(alertMatchesSearch(withColonInMessage, 'error:timeout')).toBe(true); // substring hit
+      expect(alertMatchesSearch(alert, 'other:abc-123')).toBe(false); // no label, no substring
     });
 
     it('keeps whole-string substring behavior for free-text (with a space)', () => {
--- a/common/services/alerting/__tests__/filter.test.ts
+++ b/common/services/alerting/__tests__/filter.test.ts
@@ -3,7 +3,14 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import { emptyFilters, filterAlerts, matchesFilters, matchesSearch } from '../filter';
+import {
+  alertMatchesSearch,
+  emptyFilters,
+  filterAlerts,
+  matchesFilters,
+  matchesSearch,
+  sortRules,
+} from '../filter';
 
 describe('filter', () => {
   describe('emptyFilters', () => {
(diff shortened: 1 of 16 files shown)


riysaxen-amzn:
Reviewed the consolidated branch. The consolidation itself is the right call — #2827/#2825/#2839 were cut from a shared base and conflicted on the shared primitives, and keeping `main`'s superset primitives while re-applying only the render wiring avoids the duplicate-primitives trap. Evidence is strong (before/after at identical viewports, 108 suites / 1173 tests, CI green, original authorship preserved). Approving with items below — only the first needs a change before merge:

1. **`sourceDisplayName` precedence for SLO alerts.** The fallback chain prefers `monitor_name` unconditionally, so an SLO alert carrying both `slo_id` and a copied `monitor_name` renders the monitor's name in a row titled "Source SLO" — inconsistent with the "Open SLO" action. Suggest: when `sloId` is present, resolve `slo_name ?? sloId` first, and only fall back to `monitor_name` for non-SLO alerts.

2. **`labelKey:value` search heuristic can swallow free-text queries.** Any whitespace-free token containing a colon is now treated as a label search, so pasting a URL fragment or message text like `error:timeout` into the search box silently matches nothing instead of substring-matching. Consider falling back to substring match when the parsed key doesn't exist in any alert's labels, or documenting the syntax in the search placeholder.

3. **Nit:** `linkify_annotation` renders `trimmed` in the link path but raw `value` in the plain-text fallback — inconsistent when the annotation is whitespace-padded. Rendering `trimmed` in both branches also keeps test text-matching stable.

4. **Nit:** the synthetic `hashchange` dispatched immediately after `navigateToApp` fires while the source app's router is still mounted and the target's may not be. Chaining it off the promise `navigateToApp` returns (`.then(() => window.dispatchEvent(...))`) would remove the race.

Appreciate the "supersedes, close after merge" hygiene on the three source PRs.

riysaxen-amzn (approved):

PR #2880, opened 2026-09-09
[APM Services] Improve filters' user experience

TackAdam:
## Description

Brings the APM Services Environment filter up to parity with the Attributes filter, and adds truncation with hover tooltips so long environment/attribute names no longer break the sidebar layout.

Previously the Environment filter was a bare checkbox list: no search, no result cap, and long values (e.g. verbose k8s namespaces or cluster identifiers) overflowed the narrow filter panel. Attribute values could be searched and capped at 5, but expanding a long list stretched the panel instead of scrolling, and long labels/titles had no truncation.

https://github.com/user-attachments/assets/f23bc5b6-1008-4219-acd3-18e4d681abda

### Environment filter — now matches the Attributes filter

- Per-filter search box (`EuiFieldSearch`) that filters the environment list client-side.
- Select all / Clear all links (reuse the existing `selectAll` / `clearAll` i18n strings).
- Shows the first 5 values by default with a `+N more` / `Show less` toggle (reuses `ATTRIBUTE_VALUES_INITIAL_LIMIT`).
- Empty states: `noEnvironments` when there are none at all; `noMatchingValues` when a search filters everything out.

### Scroll-when-expanded (both filters)

- When a value list is expanded past the default 5, it now scrolls within a fixed max height instead of growing the sidebar.
- New constant `FILTER_VALUES_EXPANDED_MAX_HEIGHT = 240` (px); the wrapper only applies `max-height` + `overflow-y: auto` in the expanded state.

### Long-name truncation + hover tooltip (both filters)

- Environment labels, attribute value labels, and the attribute-group accordion title now render through the shared `TruncatedLabel` component (single-line ellipsis + instant hover/focus tooltip that appears only when the text is actually clipped — the same component used in Alert Manager and the SLO/facet filters).
- Added the `min-width: 0` CSS cascade required for `text-overflow: ellipsis` to engage inside checkbox rows and the accordion header.

## Implementation notes

- CSS selectors cover both eui* and oui* class names. Components are imported under their Eui* names, but the class prefix in the rendered DOM depends on which EUI build the runtime serves, so the truncation/min-width: 0 rules list both prefixes to stay robust. Documented inline in apm_common.scss.

## Files changed

- `public/components/apm/pages/services_home/services_home.tsx` — Environment filter rewrite; `TruncatedLabel` + scroll wrappers on both filters.
- `public/components/apm/common/constants.ts` — `FILTER_VALUES_EXPANDED_MAX_HEIGHT`.
- `public/components/apm/shared/styles/apm_common.scss` — truncation/min-width cascade for checkbox groups and the attribute accordion.

## Test subjects (for functional tests)

`environmentSearch`, `environmentSelectAll`, `environmentClearAll`, `environmentShowMore`, `environmentCheckboxGroup` (Environment); existing `attribute-${attrPath}-*` hooks unchanged.
### Check List


public/components/apm/pages/services_home/services_home.tsx
```diff
@@ -1241,15 +1256,107 @@ export const ServicesHome: React.FC<ServicesHomeProps> = ({
...
+                                              APM_CONSTANTS.ATTRIBUTE_VALUES_INITIAL_LIMIT
+                                            )
+                                        ).map((env) => ({
+                                          id: env,
+                                          label: <TruncatedLabel text={env} />,
+                                        }))}
+                                        idToSelectedMap={selectedEnvironments}
+                                        onChange={onEnvironmentChange}
+                                        compressed
+                                        data-test-subj="environmentCheckboxGroup"
```
ps48, line 1331: ### p1: the new environment filter has no unit coverage

This adds a fair amount of new interactive logic (client-side `filteredEnvironments`, select-all / clear-all, the `+N more` / Show less slice toggle, and the two distinct empty states) but no unit tests, and the PR checklist leaves "New functionality includes testing" unchecked. `apm/__tests__/services.test.tsx` only touches environment via a URL-encoding assertion, so none of this is exercised. Per our bar of full unit coverage for new component logic, could we add tests that drive the `data-test-subj` hooks you already wired up: search narrows `environmentCheckboxGroup` (case-insensitive) and shows `noMatchingValues` when nothing matches vs `noEnvironments` when the list is empty; `environmentSelectAll` selects the currently-filtered subset; `environmentClearAll` clears; and `environmentShowMore` flips between the 5-item slice and the full scrollable list?

public/components/apm/pages/services_home/services_home.tsx
```diff
@@ -485,6 +486,20 @@ export const ServicesHome: React.FC<ServicesHomeProps> = ({
     }));
   }, []);
 
+  const handleSelectAllEnvironments = useCallback(() => {
```
ps48, line 489: ### p2: select-all diverges from the attributes filter under an active search

`handleSelectAllEnvironments` merges the filtered subset into the previous selection (`{ ...prev }` then set each filtered env true), so environments selected under a different search term stay selected. The attributes equivalent, `handleSelectAllForAttribute`, instead replaces the whole map (`[attrPath]: newSelections`), which drops any previously-selected value that the current search filters out. Since this PR is specifically about bringing the two filters to parity, the two Select-all semantics now differ. The environment (merge) behavior is the better one, so the cleaner fix is arguably to align the attributes handler to merge too, but at minimum worth a deliberate call on which semantics both should use. Note the attributes side is pre-existing.

ps48 (commented):
### [apm services] improve filters

Nice parity work, and it's well-aligned with the APM filter-sidebar paradigm the checklist cares about, since the whole point is bringing the Environment filter to match the Attributes filter (search, select / clear all, `+N more`, empty states) plus truncation and scroll-when-expanded on both. Verified the mechanics: `filteredEnvironments` filters client-side (case-insensitive), the list is capped at `ATTRIBUTE_VALUES_INITIAL_LIMIT` with a `+N more` / Show less toggle, and expanded lists scroll within `FILTER_VALUES_EXPANDED_MAX_HEIGHT` rather than stretching the panel. The new constant lives in `apm/common/constants.ts`, `TruncatedLabel` is the shared component already used elsewhere, OUI throughout (`EuiFieldSearch` / `EuiCheckboxGroup` / `EuiLink`), and no new dependencies, routes, saved objects, or stray `console.*`. The `eui*` / `oui*` dual-prefix `min-width: 0` cascade is a real gotcha and it's documented inline. The environment list is bounded by the service set and display-capped, so no unbounded concern.

No `p0`. One `p1`: the new Environment-filter logic has no unit coverage. One `p2`: the two filters' Select-all now behave differently under an active search. Both inline.

public/components/apm/pages/services_home/services_home.tsx
TackAdam: ### Added `services_home.test.tsx`

Added 5 tests that drive the `data-test-subj` hooks directly:

- **Search filtering**
  - Narrows `environmentCheckboxGroup` case-insensitively (`EC` → `ec2` / `ecs`).
  - Shows `noMatchingValues` when nothing matches.
  - Shows `noEnvironments` when services exist but none carry an environment.

- **Select All**
  - `environmentSelectAll` selects the currently filtered subset.
  - Preserves selections made under a prior search (see the P2 fix below).

- **Clear All**
  - `environmentClearAll` clears the map.

- **Show More**
  - `environmentShowMore` flips between the 5-item slice and the full list.

public/components/apm/pages/services_home/services_home.tsx
TackAdam: ### Aligned Select-All Merge Semantics

Good catch. Aligned the two on the preferred merge semantics.

`handleSelectAllForAttribute` now merges the filtered values into the existing per-attribute selection instead of replacing the map, matching `handleSelectAllEnvironments`. This ensures values selected under a previous search term are preserved.

The parity test in `services_home.test.tsx` asserts this merge behavior for the environment filter.

ps48 (approved):

PR #2881, opened 2026-09-09
docs: add AGENTS.md and CLAUDE.md agent guidance

ps48:
## Description

Adds `AGENTS.md` (and a thin `CLAUDE.md` that points to it) so AI coding and review agents working in this repo follow the plugin's conventions automatically, not just when a particular tooling skill happens to be loaded.

`AGENTS.md` covers:
- **Repo shape + build/test/lint** — this is an OSD plugin (scripts shell out to the OSD root); node version from OSD `.node-version`, `yarn osd bootstrap`, `yarn test` (jest), `yarn lint`, `yarn cypress:run`.
- **Conventions for dev agents** — feature-flag airtightness, `common/` placement for shared constants/types, no new deps/backend query routes/saved-object types, datasources from saved objects + datasets, logging hygiene, error surfacing, bounded/paginated data paths, OUI + echarts (no vega), APM filter-sidebar parity, and tests-ship-with-code (real-backend Cypress).
- **Conventions for review agents** — the hard rule to scope only to `gh pr diff --name-only` (never `git diff main`, which is stale on stacked branches and yields false findings), verify-before-asserting, the p0/p1/p2 + lowercase-h3 comment style, and the summary-review-plus-individual-inline-comments posting model anchored to the head SHA.

`CLAUDE.md` just defers to `AGENTS.md` to avoid drift.

Docs-only; no code or dependency changes.

### Check List


CLAUDE.md
```diff
@@ -0,0 +1,7 @@
+# CLAUDE.md
+
+This repo's agent guidance lives in [`AGENTS.md`](./AGENTS.md). Read it first and follow it — it
+covers build/test/lint, the code conventions for dev agents, and the review process and posting
+rules for review agents.
+
+Nothing Claude-specific overrides `AGENTS.md`; if the two ever disagree, `AGENTS.md` wins.
```
joshuali925, line 7: just use a symlink? so it doesn't cause claude code model to actively read agents.md every time

PR #2882, opened 2026-09-10
fix(alerting): edit metrics flyout — Builder/Code toggle + overwrite guard (no expr data-loss)

lezzago:
## Summary

Fixes a **silent data-loss footgun** when editing a Prometheus rule whose expression the point-and-click builder can't represent.

The edit flyout (`PrometheusFormSection`) rendered the builder **only**. Opening a rule with any non-trivial PromQL — `or`/`and`, arithmetic, multi-comparison — showed a **blank builder** (the parser can't seed it), and the raw expression was visible only in the read-only YAML preview. Selecting any metric then **overwrote the stored expression with no warning**. e.g. editing `OtelCollectorExportFailures` and picking a metric rewrote its 3-clause `or` expression to `expr: up`.

This ports the create flyout's Builder ⇄ Code toggle + overwrite guard (`create_metrics_monitor.tsx`) into the edit flyout:
- **Defaults to Code** when the seeded expression is builder-unrepresentable (`parseExpr(query) === null`), so it opens **shown as-is** instead of blank.
- **Code mode**: a raw PromQL textarea bound to the query (plus the always-firing hint for a condition-less expression).
- **Builder mode**: warns — with a switch-to-Code affordance — before a metric pick would replace an unrepresentable expression.

The expression is now only rewritten when the user explicitly edits it.

## Visual evidence
BEFORE = `origin/main` (merge base), AFTER = this branch. Same rule (`OtelCollectorExportFailures`), viewport (1440×900), and scroll.

### Edit flyout — blank builder (no Code editor) → Code mode with the full expression
Edit flyout before/after

### Flow — BEFORE: picking a metric silently wipes the expression
before flow

### Flow — AFTER: opens in Code with the expression; Builder is guarded
after flow

## Relationship to #2879
Independent and **conflict-free** — this touches only `create_monitor/prometheus_form_section.tsx` (+ its test), which #2879 does not modify. Either can merge first.

## Testing
- Plugin Jest (`jest --config ./test/jest.config.js public/components/alerting`): **50 suites / 546 tests pass**, including new `prometheus_form_sync` cases: unrepresentable rule opens in Code showing the raw expr; representable rule opens in Builder; Code edits bypass the builder; switching to Builder surfaces the overwrite guard.
- `node scripts/eslint` clean.

## Notes
- Opened as **draft** pending full CI.

### DIFF
--- a/public/components/alerting/__tests__/prometheus_form_sync.test.tsx
+++ b/public/components/alerting/__tests__/prometheus_form_sync.test.tsx
@@ -21,6 +21,14 @@ import { render, screen, fireEvent } from '@testing-library/react';
 import { PrometheusFormSection } from '../create_monitor/prometheus_form_section';
 import { parseExpr } from '../create_monitor/prom_query_builder';
 import type { PrometheusFormState } from '../create_monitor/create_monitor_types';
+import type { LabelEntry } from '../monitor_form_components';
+
+/** The subset of LabelEditor props the sync tests capture and assert against. */
+interface CapturedLabelEditorProps {
+  labels: LabelEntry[];
+  onChange: (labels: LabelEntry[]) => void;
+  context?: { service?: string; team?: string };
+}
 
 // Mock dependencies that PrometheusFormSection uses
 jest.mock('../monitor_form_components', () => ({
@@ -291,10 +299,10 @@ describe('PrometheusFormSection — rule group', () => {
 
   it('hides _ruleGroup from the label editor and preserves it through label edits', () => {
     const onUpdate = jest.fn();
-    const labelEditorProps: any[] = [];
+    const labelEditorProps: CapturedLabelEditorProps[] = [];
     // Capture what LabelEditor receives via the module mock
     const { LabelEditor } = jest.requireMock('../monitor_form_components');
-    LabelEditor.mockImplementation((props: any) => {
+    LabelEditor.mockImplementation((props: CapturedLabelEditorProps) => {
       labelEditorProps.push(props);
       return <div data-test-subj="label-editor" />;
     });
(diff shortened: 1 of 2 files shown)


riysaxen-amzn:
Reviewed the toggle + overwrite guard end to end — this closes the exact data-loss class from the audit (a rule the builder can't represent used to open in a blank builder and get silently clobbered on save). The shape of the fix is right: mode seeded from `parseExpr(query) === null`, raw expression shown as-is in Code, warning callout guarding the Builder side, and `data-test-subj` hooks wired for functional tests. Approving with a few non-blocking items:

1. **Preservation guarantee depends on `PromQueryBuilder` mount behavior.** The comment says the raw expression is "preserved until the user actually picks a metric" — that only holds if `PromQueryBuilder` never emits `onQueryChange` on mount when handed an unparseable query. Can you confirm that, or add a test asserting `onUpdate` is *not* called when Builder mounts with `complexExpr`? That would pin the guarantee the warning callout promises.

2. **Null-safety inconsistency.** The mode initializer uses `(form.query ?? '').trim()` but `builderWouldOverwrite` calls `form.query.trim()` directly — this throws if `query` is ever undefined. Suggest `(form.query ?? '').trim() !== ''` for consistency.

3. **CI Lint warnings.** The Lint job is flagging two `Unexpected any` in `prometheus_form_sync.test.tsx` (`labelEditorProps: any[]` and the `props: any` mock param) — worth cleaning up before merge since they're new in this PR.

4. **Observation, no action needed:** the `isAlwaysFiring` callout in Code mode inherits the bool-modifier edge case already noted on #2870 — not new here.

Nice work wiring the test hooks — this scenario is a strong candidate for the regression catalog.

lezzago:
Thanks for the thorough review @riysaxen-amzn! Addressed 1–3 in 97b13b50:

**1. Preservation guarantee — added the test you asked for.** Confirmed the guarantee holds by design: on mount `PromQueryBuilder` returns early via `didMountRef` without emitting, and because a builder-unrepresentable query gives `parseExpr → null`, `builderOwnsQuery` stays `false` so a later clear can't wipe it either. The existing `does not clobber a non-representable seeded query on mount` test only exercised the Code-mode default (the builder isn't even mounted there), so I added `does not emit a query update when the Builder mounts with an unrepresentable expression` — it toggles into Builder with a complex expr and asserts `onUpdate` is never called with `'query'`, pinning exactly the guarantee the callout promises.

**2. Null-safety — fixed.** `builderWouldOverwrite` now uses `(form.query ?? '').trim()` / `parseExpr(form.query ?? '')`, matching the mode initializer.

**3. `any` in the sync test — cleaned up.** Replaced `any[]`/`any` in the `LabelEditor` mock with a typed `CapturedLabelEditorProps` interface. Minor note: these two `any`s actually predate this PR (introduced in #2796, per `git blame`) and the Lint check was already green, so this is opportunistic cleanup rather than a new-in-PR fix — but agree they're worth removing.

**4.** Noted, no action — `isAlwaysFiring` bool-modifier edge case tracked on #2870.

Verified locally: 30/30 tests in `prometheus_form_sync.test.tsx` pass, ESLint clean on both files.

riysaxen-amzn (approved):

PR #2885, opened 2026-09-10
fix(alerting): left-align facet filter titles and keep group count inline

lezzago:
## Description

The facet-header toggle button in the shared `FacetFilterGroup` rendered its title **centered** and wrapped the group `(N)` count onto the line **below** the title. Two coupled causes:

1. `EuiButtonEmpty` defaults to `text-align: center`, and the button's text span was `display: block` — so the span stretched full-width and centered the inline title inside it. (The parent `contentProps` justify only positions the span, not the text within it.)
2. `TruncatedLabel`'s inner span is `display: block; width: 100%`, so inside the block text span it consumed the whole line and pushed the `(N)` count underneath.

### Fix
Make the button's text span a **left-aligned flex row** (`display: flex`, `textAlign: 'start'`) with the title `<strong>` as a `flex: 1` child and the count as a fixed-width (`flex-shrink: 0`) sibling. `textAlign: 'start'` is required *in addition to* the flex row because `EuiButtonEmpty`'s center alignment is inherited into `TruncatedLabel`'s full-width inner span, so short titles (e.g. `Type`, `monitor_name`) would otherwise still render centered. `start` (not `left`) keeps it RTL-safe.

Affects the **Alerts**, **Monitors/Rules**, and **SLO** filter panels that share `FacetFilterGroup`.

## Visual Evidence

BEFORE = merge base (`origin/main`); AFTER = this PR. Same dev server, route, data, viewport, and theme — only the component code differs.

### 1. Facet titles: centered → flush-left

The facet titles (`Type`, `Severity`, `State`) were centered under the toggle button; they now pin flush-left, aligned with the `Filters` heading.

Facet title alignment before/after

### Flow — BEFORE (origin/main): titles centered

Opening the Alerts filter panel and expanding each facet — titles stay centered throughout.

Before flow

### Flow — AFTER (this PR): titles flush-left

Same interaction on this PR — titles are flush-left at every step.

After flow

> Note: the `(N)` group-count facets (label-key facets that set `showOptionCount`) only render when there are alerts carrying labels; this dev environment has no reachable alerting backend, so those facets aren't shown above. The inline-count behavior is covered by the unit tests below.

## Issues Resolved
Facet filter titles were centered and the group count wrapped to a second line.

## Testing
Added regression tests to `facet_filter_panel.test.tsx` (jsdom preserves inline styles):
- header text span is a left-aligned flex row (`display: flex`, `textAlign: start`, not `block`)
- title `<strong>` flex-grows and can truncate
- group `(N)` count renders inline as a **sibling** of the title (not nested/wrapped below)
- count is omitted when `showOptionCount` is false (default)

All 19 tests in the suite pass; ESLint clean.

## Check List
- [x] All tests pass
- [x] New functionality has been documented (n/a — UI-only style fix)
- [x] Commit changes are listed out in CHANGELOG.md file (n/a — UI-only style fix)
- [x] Commit changes are signed off (`Signed-off-by`)
- [x] Screenshots / visual evidence attached (before/after + flow GIFs above)

### DIFF
--- a/public/components/alerting/__tests__/facet_filter_panel.test.tsx
+++ b/public/components/alerting/__tests__/facet_filter_panel.test.tsx
@@ -217,6 +217,65 @@ describe('useFacetCollapse', () => {
   });
 });
 
+describe('FacetFilterGroup — header title alignment and inline group count', () => {
+  beforeEach(() => jest.clearAllMocks());
+
+  // Regression: the facet header title used to render centered (EuiButtonEmpty
+  // defaults to text-align:center and its text span was display:block), and the
+  // group `(N)` count wrapped onto the line below the title. The fix switches
+  // the button's text span to a left-aligned flex row so the title pins
+  // flush-left and the count sits inline. jsdom does not apply stylesheets, but
+  // it preserves inline styles — which is exactly where these load-bearing
+  // props live — so we assert them directly on the DOM.
+  const getTextSpan = (toggle: HTMLElement) =>
+    toggle.querySelector<HTMLElement>('.euiButtonEmpty__text');
+
+  it('left-aligns the header title via a flex row (not a centered block span)', () => {
+    const { getByTestId } = render(<FacetFilterGroup {...defaultProps} />);
+    const textSpan = getTextSpan(getByTestId('facetGroup-status-toggle'));
+    expect(textSpan).not.toBeNull();
+    // Load-bearing for left alignment: a flex row + start alignment, NOT a
+    // display:block span that would let EuiButtonEmpty's text-align:center
+    // center the inline title.
+    expect(textSpan!.style.display).toBe('flex');
+    expect(textSpan!.style.textAlign).toBe('start');
+    expect(textSpan!.style.display).not.toBe('block');
+  });
+
+  it('lets the title <strong> flex-grow so it pins flush-left and truncates', () => {
+    const { getByTestId } = render(<FacetFilterGroup {...defaultProps} />);
+    const textSpan = getTextSpan(getByTestId('facetGroup-status-toggle'));
+    const strong = textSpan!.querySelector<HTMLElement>('strong');
+    expect(strong).not.toBeNull();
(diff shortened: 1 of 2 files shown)


riysaxen-amzn:
1). textOverflow: 'ellipsis' and whiteSpace: 'nowrap' are retained on the span that is now a flex container, where text-overflow has no effect (truncation actually happens inside TruncatedLabel and the <strong>). Harmless leftovers, could be dropped for clarity.
2). The tests are intentionally implementation-coupled (asserting inline style values). Fine given the jsdom constraint the test comment already explains, but they will need updating if this ever migrates to a CSS class. Also the count's flex-shrink: 0 lives in .altFacetGroupCount CSS, which jsdom can't verify, so that half of the inline guarantee rests on the visual evidence.

lezzago:
Thanks @riysaxen-amzn!

**1) `textOverflow: 'ellipsis'` / `whiteSpace: 'nowrap'` leftovers** — good catch, and they're already gone in this PR. Removing exactly those two (plus `overflow`) from the now-flex `textProps` span was the second half of the change (the truncation genuinely happens inside `TruncatedLabel` / the `<strong>`, as you noted). The current head only keeps `overflow: 'hidden'` on the `<strong>` wrapper, which is load-bearing for clipping the truncating child. Nothing left to drop.

**2) Implementation-coupled tests + the `flex-shrink: 0` gap** — agreed the inline-style assertions are coupled by the jsdom constraint (called out in the test comment); they'd need revisiting only if this ever moves to a CSS class. On the `flex-shrink: 0` that lives in `.altFacetGroupCount`: jsdom can't compute it, but it *can* confirm the class is applied, so I added an assertion that the count element carries `altFacetGroupCount` (same pattern the suite already uses for `.altFacetTarget`). That closes the "rests only on visual evidence" gap — the class binding is now unit-tested, and the visual evidence backs the rendered result. Pushed in the amended commit.

riysaxen-amzn (approved):

PR #2887, opened 2026-09-10
[APM Topology Map] Improve environment filter UX

TackAdam:
## Description

Brings the **Topology Map** Environment filter up to parity with the APM Services filters, applying the same UX shipped in #2880 to the topology sidebar.

Previously the Topology Map's Environment filter was a bare checkbox list: no search, no result cap, and long environment values (e.g. verbose k8s namespaces or cluster identifiers) overflowed the narrow filter panel and stretched the sidebar. It now behaves exactly like the Services page's Environment filter.

### Environment filter — now matches the Services filter

- **Per-filter search box** (`EuiFieldSearch`) that filters the environment list client-side, matching against the display name (prefix before `:`), case-insensitively.
- **Select all / Clear all** links (reuse `selectAll` / `clearAll` i18n strings). Select all **merges** the currently-filtered subset into the existing selection so values selected under a previous search term survive; Clear all empties the selection.
- Shows the first 5 values by default with a `+N more` / `Show less` toggle (reuses `APM_CONSTANTS.ATTRIBUTE_VALUES_INITIAL_LIMIT`).
- **Empty states**: `noEnvironments` when there are none at all; `noMatchingValues` when a search filters everything out.

### Scroll-when-expanded

- When the value list is expanded past the default 5, it scrolls within `FILTER_VALUES_EXPANDED_MAX_HEIGHT` (240px) instead of growing the sidebar.

### Long-name truncation + hover tooltip

- Environment labels render through the shared `TruncatedLabel` component (single-line ellipsis + instant hover/focus tooltip that appears only when the text is actually clipped) — the same component used in the Services filters, Alert Manager, and the SLO/facet filters.
- Reuses the existing `apmFilterCheckboxGroup` `min-width: 0` CSS cascade from `apm_common.scss` (added in #2880) required for `text-overflow: ellipsis` to engage inside checkbox rows.

## Implementation notes

- The Topology Map sidebar stores selected environments as a `string[]` (`filters.environments`), unlike the Services page's `Record<string, boolean>` map, so select-all / clear-all are implemented via set-union / empty-array over that array.
- Group By / Fault Rate / Error Rate filters are untouched — #2880 only reworked value-list (checkbox) filters, and those controls aren't of that shape.

## Files changed

- `public/components/apm/shared/components/service_map/service_map_sidebar.tsx` — Environment filter rewrite (search, select/clear all, show more/less, scroll wrapper, `TruncatedLabel`).
- `public/components/apm/pages/application_map/application_map_i18n.ts` — new filter strings: `selectAll`, `clearAll`, `showLess`, `showMore`, `noMatchingValues`, `noEnvironments`.
- `public/components/apm/shared/components/service_map/__tests__/service_map_sidebar.test.tsx` — new test suite.

## Testing

New unit tests (6, all passing) cover: cap-at-5 + `+N more`/`Show less` toggle, case-insensitive search, `noMatchingValues`, `noEnvironments`, Select all merge semantics, and Clear all.

Test subjects (for functional tests): `environmentSearch`, `environmentSelectAll`, `environmentClearAll`, `environmentShowMore`, `environmentCheckboxGroup`.

### Check List


public/components/apm/shared/components/service_map/service_map_sidebar.tsx
```diff
@@ -237,14 +260,110 @@ export const ServiceMapSidebar: React.FC<ServiceMapSidebarProps> = ({
...
-        <EuiCheckboxGroup
-          options={environmentCheckboxes}
-          idToSelectedMap={environmentSelectionMap}
-          onChange={handleEnvironmentChange}
-          compressed
-          disabled={isLoading}
-          data-test-subj="environmentCheckboxGroup"
-        />
+
+        {availableEnvironments.length > 0 ? (
```
lezzago: **Misleading empty state during load.** `availableEnvironments` derives from `nodes` in the parent (`application_map_page.tsx`), and `isLoading = mapLoading || metricsLoading`. On initial load `nodes` is empty, so `availableEnvironments` is `[]` and this `length > 0 ? … : <noEnvironments>` check falls through to **"No environments available"** while the map is still loading. There's no loading branch, so users see that misleading text on every cold load. Consider a `isLoading && availableEnvironments.length === 0` branch that renders a loading affordance instead.

public/components/apm/shared/components/service_map/service_map_sidebar.tsx
```diff
@@ -237,14 +260,110 @@ export const ServiceMapSidebar: React.FC<ServiceMapSidebarProps> = ({
...
+            />
+
+            <EuiSpacer size="s" />
+
+            {/* Select all / Clear all links */}
+            {filteredEnvironments.length > 0 && (
+              <>
+                <EuiFlexGroup gutterSize="s" justifyContent="spaceBetween">
+                  <EuiFlexItem grow={false}>
+                    <EuiLink
```
lezzago, line 296: _Minor:_ the **Select all / Clear all** links (and **Show more** at ~340) have no `disabled` guard, while the search box and `EuiCheckboxGroup` both get `disabled={isLoading}`. During a refetch (time-range change/refresh) where previous `nodes` are still retained, these links stay clickable and can mutate `filters.environments` even though the rest of the panel is locked. Gate them on `isLoading` for consistency.

public/components/apm/shared/components/service_map/service_map_sidebar.tsx
```diff
@@ -237,14 +260,110 @@ export const ServiceMapSidebar: React.FC<ServiceMapSidebarProps> = ({
...
+                  style={
+                    environmentExpanded
+                      ? {
+                          maxBlockSize: APM_CONSTANTS.FILTER_VALUES_EXPANDED_MAX_HEIGHT,
+                          overflowY: 'auto',
+                        }
+                      : undefined
+                  }
+                >
+                  <EuiCheckboxGroup
```
lezzago, line 331: _Cleanup (non-blocking):_ this search + select/clear-all + show-more + checkbox-list block largely duplicates the widget in `filter_sidebar.tsx` (`FilterSidebar`), and `services_home.tsx` hand-rolls its own copy too. Worth extracting a shared filter-value-list component at some point to avoid maintaining three+ near-identical variants.

public/components/apm/shared/components/service_map/service_map_sidebar.tsx
```diff
@@ -62,13 +68,14 @@ export const ServiceMapSidebar: React.FC<ServiceMapSidebarProps> = ({
...
 
-  // Environment checkbox options - dynamically built from available environments
-  const environmentCheckboxes = useMemo(() => {
-    return availableEnvironments.map((env) => ({
-      id: env,
-      label: getEnvironmentDisplayName(env),
-    }));
-  }, [availableEnvironments]);
+  // Environments narrowed by the per-filter search box (matches the display name).
+  const filteredEnvironments = useMemo(() => {
```
lezzago, line 73: _Edge case (low impact):_ `availableEnvironments` is deduped on the **raw** env string in the parent, but the list/search key on `getEnvironmentDisplayName(env)` (prefix before `:`). Two distinct raw envs sharing a prefix — e.g. `eks:cluster1` and `eks:cluster2` — would render as two identical `eks` checkboxes. The Services filter avoids this by deduping on the parsed platform. Not blocking, but worth a note if such env strings are possible.

lezzago (commented):
A few UX/consistency notes from a follow-up review of the environment filter, ranked. #1 is the only one with real user-visible impact.

public/components/apm/shared/components/service_map/service_map_sidebar.tsx
TackAdam: Fixed in 9ff3bd3b. Added an \`isLoading && availableEnvironments.length === 0\` branch that renders a spinner + "Loading environments…" (new \`loadingEnvironments\` i18n string) instead of the "No environments available" text. During a refetch where previous nodes are retained, \`availableEnvironments\` is non-empty so the checkbox list stays rendered rather than flashing the loader. Covered by two new tests.

public/components/apm/shared/components/service_map/service_map_sidebar.tsx
TackAdam: Fixed in 9ff3bd3b. Added \`disabled={isLoading}\` to the Select all, Clear all, and Show more/less links so they lock in step with the search box and checkbox group during a refetch.

public/components/apm/shared/components/service_map/service_map_sidebar.tsx
TackAdam: Fixed in 9ff3bd3b. The parent memo now dedupes on \`getEnvironmentDisplayName(env)\` (keeping one representative raw env per display name) instead of the raw string, so \`eks:cluster1\`/\`eks:cluster2\` collapse into a single "eks" checkbox — matching how service_map_graph already filters by display name.

public/components/apm/shared/components/service_map/service_map_sidebar.tsx
TackAdam: Agreed this is worth doing. Leaving it out of this PR since it is non-blocking and spans three files (service_map_sidebar, filter_sidebar, services_home) — will follow up with a dedicated PR to extract a shared filter-value-list component.

public/components/apm/shared/components/service_map/service_map_sidebar.tsx
```diff
@@ -237,14 +261,124 @@ export const ServiceMapSidebar: React.FC<ServiceMapSidebarProps> = ({
...
+          <>
+            {/* Search box */}
+            <EuiFieldSearch
+              placeholder=""
+              value={environmentSearchQuery}
+              onChange={(e) => setEnvironmentSearchQuery(e.target.value)}
+              isClearable
+              fullWidth
+              compressed
+              disabled={isLoading}
```
ps48: ### p2: the whole filter is disabled during any refetch, not just the first load

`disabled={isLoading}` is applied to the search box here and to Select all / Clear all, the checkbox group, and Show more below. The empty-first-load case already has its own spinner branch, so this `disabled` only bites during a background refetch when a previous list is retained (the "keeps showing the retained list" path your test covers) - there, a user mid-search or mid-toggle has the controls go inert until the refetch settles. Since the list is intentionally retained to stay interactive, do we need to disable the controls at all in that state? Dropping `disabled` on the search / links / group (keeping the retained list live) would avoid the interaction hiccup. Minor.

ps48 (commented):
### [apm topology map] improve environment filter ux

Nice, and it closes the loop on the services-home filter review: full unit coverage for the new behavior (cap / `+N more`, case-insensitive search, `noMatchingValues` vs `noEnvironments`, the loading affordance, Select-all-merges, Clear-all) and the Select-all now merges the filtered subset into the existing selection instead of dropping off-search selections. Verified the display-name dedupe is correct end to end: `availableEnvironments` collapses raw envs by display name, and the graph filter (`service_map_graph.tsx:255,320`) maps both the selected envs and each node's env through `getEnvironmentDisplayName` before comparing, so selecting the deduped "eks" matches every raw `eks:*` node. OUI throughout, `TruncatedLabel` reused, `FILTER_VALUES_EXPANDED_MAX_HEIGHT` scroll wrapper, no new deps / routes / saved objects / `console.*`. The column-width rebalance is documented.

No `p0`, no `p1`. One optional `p2` inline.

public/components/apm/shared/components/service_map/service_map_sidebar.tsx
TackAdam: Good call — done in 516d618f. You are right that the cold-load case renders the spinner branch, so \`disabled={isLoading}\` only ever bit during a background refetch with a retained list, which is exactly the state we keep interactive on purpose. Dropped it from the search box, Select all / Clear all / Show more links, and the checkbox group so a user mid-search or mid-toggle no longer freezes until the refetch settles. Updated the retained-list test to assert the search box and checkboxes stay enabled during \`isLoading\`.

ps48 (approved):

PR #2891, opened 2026-09-11
fix(apm): correlated logs/spans filters search all results, not just visible page

TackAdam:
## Summary

Fixes a bug on the APM **Services** page correlated-data flyout (`ServiceCorrelationsFlyout`) where the **Correlated logs** and **Correlated spans** filters only searched the currently-fetched page of rows instead of the full result set.

Both tabs fetched a capped page from the backend (`head 50` for spans, `head 10` per dataset for logs) and then applied the dropdown filter **client-side** over that page. The filter state was also deliberately excluded from the fetch `useEffect` dependency arrays, so changing a filter never re-queried. Net effect: applying a filter (e.g. log level = `error`, or span status = `error`) only searched the ~10/50 rows already on screen and silently missed matching rows beyond the cap — inaccurate results.

## Changes

**Spans — status filter now searches all spans**
- `error` / `ok` / `unset` are pushed into the PPL `WHERE` clause on the canonical `` `status.code` `` field, so the backend returns matching rows across all spans (not just the visible page).
- The derived status clause + fetch limit are added to the fetch `useEffect` deps so changing the dropdown re-queries.
- HTTP-status buckets (`2xx`–`5xx`) **cannot** be pushed into PPL — the HTTP status field name varies per dataset (`attributes.http.status_code`, `http.status_code`, `span.attributes.http.status_code`, `attributes['http.response.status_code']`) and referencing a missing field errors in PPL. For those, the fetch widens (`SPANS_FILTERED = 1000`) and the existing client-side filter narrows it. All HTTP buckets produce the same query, so switching among them does not trigger a redundant refetch.

**Logs — level filter now searches all logs**
- Log severity fields are not schema-mapped and vary across datasets (`severityText` / `severityNumber` / `severity` / `level`) — the same reliability problem as HTTP status — so a `WHERE` clause on level would error on datasets using a different field.
- Instead, when a level filter is active the per-dataset fetch widens (`LOGS_PER_DATASET_FILTERED = 1000`) and the client-side filter narrows it; the fetch limit is added to the logs `useEffect` deps.

**Supporting changes**
- `constants.ts`: added `SPANS_FILTERED` and `LOGS_PER_DATASET_FILTERED` under `APM_CONSTANTS.QUERY_LIMITS`; the previously hard-coded `50` / `10` now reference these constants.
- i18n: the spans/logs descriptions no longer promise "up to 50 / up to 10" and now state that filtering searches across recent spans/logs, not just the visible page.
- Converted an ineffective `useMemo` around `tabs` (its JSX-fragment deps changed every render, so it never memoized and tripped `react-hooks/exhaustive-deps`) into a plain inline IIFE. Behavior-identical; unblocks lint.

## Behavior note
Because the spans fetch also derives the traceIds / time window used for log correlation, changing the **span status filter** in the operations view (where logs correlate by traceId) will now also refresh the correlated logs to match the filtered spans. This is reasonable, but it is a new cross-tab coupling worth flagging for review.

## Testing
- `service_correlations_flyout.test.tsx` and `service_correlations_flyout_datasource.test.tsx` pass (14/14).
- Lint clean (no new errors).

### Manual test
1. Open the APM Services page, open a service's correlated-data flyout.
2. On **Correlated logs**, apply a level filter (e.g. `error`) → results now reflect all matching recent logs, not just the visible page.
3. On **Correlated spans**, apply a status filter (`error`/`ok`/`unset`) → backend re-queries and returns matching spans across the full set; HTTP-code buckets filter over a widened fetch.


public/components/apm/shared/components/service_correlations_flyout.tsx
```diff
@@ -265,6 +265,27 @@ export const ServiceCorrelationsFlyout: React.FC<ServiceCorrelationsFlyoutProps>
...
+    if (statusFilter === 'error') return ' | where `status.code` = 2';
+    if (statusFilter === 'ok') return ' | where `status.code` = 0';
+    if (statusFilter === 'unset') return ' | where `status.code` != 0 AND `status.code` != 2';
+    return '';
+  }, [statusFilter]);
+
+  const spanFetchLimit = useMemo(
+    () =>
+      statusFilter.startsWith('http-')
+        ? APM_CONSTANTS.QUERY_LIMITS.SPANS_FILTERED
```
ps48: ### p1: filter the http buckets in ppl instead of widening + client-filtering

The reason given for widening to `SPANS_FILTERED` is that the HTTP field name varies per dataset. But client-side filtering only ever narrows the *recent* window (here the 1000 most-recent spans), and during an RCA the span you are hunting is usually rare and not recent - so it falls outside the window and the filter shows nothing exactly when it matters. Pushing the bucket into the `WHERE` searches the whole selected time range instead.

The robust way to handle the varying field name is schema-resolved-then-coalesced: use the dataset's mapped fields (the same schemaMappings mechanism this flyout already uses for `serviceName` / `timestamp` / `traceId`) to know which HTTP-status fields exist, then `coalesce` across that set so a fully-unmapped field is never referenced (no PPL resolution error) while services that use different field names in the same dataset both match:
```
| where cast(coalesce(`attributes.http.response.status_code`, `attributes.http.status_code`) as int) >= 200
    AND cast(coalesce(`attributes.http.response.status_code`, `attributes.http.status_code`) as int) < 300
```
(`http.response.status_code` is the current stable OTel attr, `http.status_code` the deprecated one - coalescing both is future-proof; `cast(... as int)` covers keyword-typed indices.) That lets spans keep `head 50`, drops `SPANS_FILTERED`, and removes the unbounded render below.

public/components/apm/shared/components/service_correlations_flyout.tsx
```diff
@@ -437,7 +480,7 @@ export const ServiceCorrelationsFlyout: React.FC<ServiceCorrelationsFlyoutProps>
             pplQuery += ` | where (\`${traceIdFieldValue}\` IN (${traceIdList}) OR \`${traceIdFieldValue}\` = '' OR isnull(\`${traceIdFieldValue}\`))`;
           }
 
-          pplQuery += ` | sort - \`${timestampField}\` | head 10`;
+          pplQuery += ` | sort - \`${timestampField}\` | head ${logFetchLimit}`;
```
ps48: ### p1: log level can be pushed into ppl too - client-side is the wrong tool for an rca

Same RCA problem as the spans, and worse here because it is per-dataset: `head ${logFetchLimit}` pulls up to `LOGS_PER_DATASET_FILTERED = 1000` per dataset into the browser and the level filter runs in JS over only that recent slice, so an older matching log (the one an RCA is looking for) is invisible. The widened rows also render into an unpaginated `EuiBasicTable` (`:923`), so a broad match mounts up to `1000 x datasets` expandable rows.

It is pushable with the same schema-gated coalesce, the only extra wrinkle being the number-vs-text split that `normalizeLogLevel` bridges (it prefers `severityNumber`, else substring-matches `severityText`). Mirror that with a gated OR, referencing only fields the schema says are mapped:
```
| where (severityNumber >= 17 AND severityNumber <= 20)
     OR (isnull(severityNumber) AND LOWER(coalesce(`severityText`, `level`)) LIKE '%error%')
```
The `coalesce(severityText, level)` is what handles service A emitting `severityText` while service B emits `level` in the *same* dataset - resolve one field and you would silently drop the other service's rows. Generating the predicate from `normalizeLogLevel`'s own thresholds/keywords keeps one source of truth. That drops `LOGS_PER_DATASET_FILTERED`, keeps `head 50` of matching rows, and removes the unpaginated-render concern. One caveat to validate on the target OpenSearch version: that a mapped-but-referenced-through-coalesce field set doesn't error when a candidate is entirely absent.

public/components/apm/common/constants.ts
```diff
@@ -150,6 +150,12 @@ export const APM_CONSTANTS = {
   QUERY_LIMITS: {
     SPANS: 50,
     LOGS_PER_DATASET: 10,
+    // When a filter can only be applied client-side (HTTP-status buckets on spans, or
+    // any log-level filter — those fields' names vary per dataset and can't be pushed
+    // into a PPL WHERE clause safely), the base page is fetched wider so the filter
+    // searches across many recent rows instead of only the default page.
+    SPANS_FILTERED: 1000,
+    LOGS_PER_DATASET_FILTERED: 1000,
```
ps48: ### p2: these `_FILTERED` widen limits go away if the filters are pushed server-side

If the HTTP and level filters move into the PPL `WHERE` (see the flyout comments), both `SPANS_FILTERED` and `LOGS_PER_DATASET_FILTERED` become unnecessary - the query returns the most-recent *matching* rows across the full time range at the normal `head 50` / `head 10`. If for some reason a client-side path is kept, note that `1000` is a recent-window cap, not "all": a match older than the 1000th row is still invisible, so the empty state should say only the most recent N were scanned rather than implying none exist.

ps48 (commented):
### fix(apm): correlated logs/spans filters search across recent results

Good, correctly-reasoned fix, and it's a real improvement over filtering only the visible page. Verified the mechanics: pushable filters go server-side (`spanStatusWhere` on the canonical `status.code`), the ones whose field names vary per dataset are widened + narrowed client-side (`filteredSpans` line 238, `filteredLogs` 905), constants live in `common/`, effect deps are updated, and the `spanStatusWhere` literals are injection-safe. The i18n was made honest ("recent … not just the visible page"). No new deps / routes / saved objects / `console.*`.

No `p0`. One `p1` on the core approach for RCA, and one `p2`. Some unrelated reformatting rides along (the `createToggleRowHandler` and toast-text reflows) - worth dropping to keep the bugfix diff focused.

public/components/apm/shared/components/service_correlations_flyout.tsx
ps48: Version note: `coalesce` needs OpenSearch 3.1+. On older domains use `ifnull` instead - the PPL condition docs call this out explicitly ("For OpenSearch versions prior to 3.1, coalesce-like functionality can be achieved using nested ifnull statements", https://github.com/opensearch-project/sql/blob/main/docs/user/ppl/functions/condition.md). Since there are only two HTTP candidates a single `ifnull` is the drop-in:
```
| where cast(ifnull(`attributes.http.response.status_code`, `attributes.http.status_code`) as int) >= 200
    AND cast(ifnull(`attributes.http.response.status_code`, `attributes.http.status_code`) as int) < 300
```
(nest as `ifnull(f1, ifnull(f2, f3))` if more candidate fields are ever added). Still schema-gate the field set so a fully-unmapped candidate isn't referenced.

public/components/apm/shared/components/service_correlations_flyout.tsx
ps48: Same version note as the spans thread: `coalesce` is 3.1+, so on older domains use `ifnull` per the PPL condition docs (https://github.com/opensearch-project/sql/blob/main/docs/user/ppl/functions/condition.md, "prior to 3.1 ... nested ifnull"):
```
| where (severityNumber >= 17 AND severityNumber <= 20)
     OR (isnull(severityNumber) AND LOWER(ifnull(`severityText`, `level`)) LIKE '%error%')
```
Nest (`ifnull(severityText, ifnull(level, ...))`) if more text-field spellings need covering, and keep the whole predicate generated from `normalizeLogLevel` so the thresholds/keywords stay single-source.

public/components/apm/shared/components/service_correlations_flyout.tsx
TackAdam: Done in 58fe41a3 — pushed the HTTP buckets into the PPL `WHERE` (`buildHttpStatusPplWhere`), dropped `SPANS_FILTERED` back to `head 50`, and removed the client-side span filtering (now server-side):

```
| where cast(coalesce(`attributes.http.response.status_code`, `attributes.http.status_code`) as int) >= 400
    AND cast(coalesce(`attributes.http.response.status_code`, `attributes.http.status_code`) as int) < 500
```

On schema-resolution vs coalesce: validated against a live 3.8 cluster (`otel-v1-apm-span*`). Confirmed the key property — a **bare** reference to a fully-absent field throws `FIELD_NOT_FOUND`, but `coalesce(absent, present)` does **not** error. So `coalesce` *is* the schema-gate here: we never need the field mapped, and services using either OTel name in the same index both match. `cast(... as int)` is required because the field is keyword-typed on some indices and long on others.

Re the version note (`coalesce` is 3.1+): I tested `ifnull` as the pre-3.1 fallback and it is **not** an equivalent drop-in — `ifnull(absent, present)` also throws `FIELD_NOT_FOUND` (only `coalesce` tolerates a fully-absent candidate). A bare `ifnull` swap would reintroduce the absent-field error on clusters missing one candidate; making it safe needs the schema-gating we're deliberately avoiding here. So keeping `coalesce` and targeting OpenSearch 3.1+; if <3.1 must be supported that's a schema-gated follow-up, not a bare ifnull.

public/components/apm/shared/components/service_correlations_flyout.tsx
TackAdam: Done in 58fe41a3 — pushed the level filter into PPL via `buildLogLevelPplWhere`, generated from `normalizeLogLevel`'s own thresholds/keywords (single source of truth), dropped `LOGS_PER_DATASET_FILTERED` back to `head 10`, and removed the client-side level filter plus the unpaginated widened render. Predicate mirrors `normalizeLogLevel`:

```
| where (cast(coalesce(`severityNumber`, `severity.number`) as int) >= 17
         AND cast(coalesce(`severityNumber`, `severity.number`) as int) <= 20)
     OR ((isnull(cast(coalesce(`severityNumber`, `severity.number`) as int))
          OR cast(coalesce(`severityNumber`, `severity.number`) as int) <= 0)
         AND LOWER(coalesce(`severityText`, `severity.text`, `level`)) LIKE '%error%')
```

Validated on a live 3.8 cluster (`logs-otel-v1*`): error=367, info=349448, fatal=0. Two gotchas hit and handled: (1) the bare `severity` field is a **struct** — `LOWER(severity)` throws `got STRUCT`, so I coalesce the nested `severity.text`/`severity.number`, never bare `severity`; (2) coalesce the number fields too, since some datasets nest it.

Same version note as the spans thread: `ifnull` is **not** absent-safe here — `ifnull(absent, present)` throws `FIELD_NOT_FOUND` (verified against the fully-absent `level` field), only `coalesce` tolerates it. So keeping `coalesce`/3.1+ rather than a bare `ifnull` swap.

public/components/apm/common/constants.ts
TackAdam: Done in 58fe41a3 — both `SPANS_FILTERED` and `LOGS_PER_DATASET_FILTERED` are removed now that the HTTP and level filters are pushed into the PPL `WHERE`. Queries return the most-recent matching rows across the full time range at `head 50` / `head 10`, so the widen caps and the "only the most recent N scanned" empty-state caveat no longer apply.

TackAdam:
## Update: server-side filter push + graceful pre-3.1 fallback

Addressed the two p1s and the p2 across two commits, and validated the PPL against a live OpenSearch 3.8 cluster (`otel-v1-apm-span*` / `logs-otel-v1*`).

### `58fe41a3` — push both filters into PPL WHERE
- **Spans HTTP buckets** → `cast(coalesce(\`attributes.http.response.status_code\`, \`attributes.http.status_code\`) as int)` per bucket. `cast` is required — the field is `keyword` on some indices, `long` on others.
- **Log level** → generated from `normalizeLogLevel`'s own thresholds/keywords (single source of truth), coalescing `severityNumber`/`severity.number` and `severityText`/`severity.text`/`level`. Note: bare `severity` is a **struct** (`LOWER(severity)` throws `got STRUCT`), so only the nested `severity.text` is referenced.
- Removed the `_FILTERED` widen limits, the client-side filtering, and the unbounded render.
- Verified: a bare reference to an absent field throws `FIELD_NOT_FOUND`, but `coalesce(absent, present)` does not — so `coalesce` itself is the schema-gate; no field needs to be mapped.

### `b6ea3d7f` — graceful fallback for OpenSearch < 3.1
`coalesce` was added in 3.1. Rather than swap to `ifnull` (verified **not** absent-field-safe: `ifnull(absent, present)` also throws `FIELD_NOT_FOUND` — only `coalesce` tolerates it), we detect the coalesce-unsupported error and fall back to the previous widen-fetch + client-side filter, per span-filter and per log-dataset. An `EuiCallOut` banner then tells the user only the most recent rows were scanned and points them to Explore / an upgrade. On 3.1+ nothing changes — the server-side push is used directly.

### Notes
- The pre-3.1 error string couldn't be validated (test cluster is 3.8+); detection matches `coalesce` in the error message/body, which is reliable since the filter queries only ever reference that function.
- Tests added for the new PPL builders and the error detector; lint/prettier clean.

TackAdam:
banner2

ps48 (approved):
### re-review: response commit pushes the filters into ppl — verified

Strong response commit, and I verified the core mechanics end to end. HTTP buckets and log level are now pushed into the PPL `WHERE` (`buildHttpStatusPplWhere` / `buildLogLevelPplWhere`), so both filter across the full time range instead of only the fetched page, which is what an RCA needs. The log predicate is generated from a shared `LOG_LEVEL_DEFS` — the single source of truth for both `normalizeLogLevel` and the builder, so client classification and server filter can't drift — and it uses nested `severity.text` / `severity.number` (never the bare struct) plus a `level` fallback, with the `isnull(num) OR num <= 0` gate mirroring the client's number-priority.

I stood up OpenSearch 3.8 and confirmed the load-bearing assumption behind the whole approach: `coalesce` over a field that is entirely absent from the index mapping returns null rather than erroring (a bare reference to the same field throws `Field not found`), so `coalesce(response.status_code, status_code)` works on a modern-only index, and the log-level predicate matches on both a `severityText`-only index and a `severityNumber`-only index. That also means two services with different field conventions in the same dataset both match. Pre-3.1 domains (no `coalesce`) are handled by the `isCoalesceUnsupportedError` fallback to the widened client-side filter, so `SPANS_FILTERED` / `LOGS_PER_DATASET_FILTERED` are now genuinely fallback-only. Both builders have unit tests including the field coalescing and `normalizeLogLevel` parity.

No `p0`, no `p1` — this resolves my earlier review. Nice work.

PR #2892, opened 2026-09-14
[APM] Correlated dashboards (experimental)

ps48:
## Summary

Adds an optional, experimental way to open the user's own OpenSearch dashboards from an APM service (and from a topology map node), scoped to the current APM time range. This is an interim step toward a native Infrastructure tab: users get one-click access to any dashboard they've already authored (infrastructure or business) without waiting for a per-service correlation model.

The feature is inert unless a user configures it — no dashboards ⇒ no UI change beyond the entry point.

## Data model

Extends the existing APM config saved object (the shared `correlations` SO, registered by the core `data` plugin) with an optional list of dashboard ids stored under `attributes.entities.correlatedDashboards`, plus new `references[]` entries of `type: 'dashboard'`. Because `entities` is mapped `object / enabled: false`, this is a **non-breaking additive change** — no migration, no mapping change. Old configs (no dashboards) read back with the field `undefined`, so users who don't opt in see APM byte-for-byte the same as today.

Titles are resolved from the dashboard saved objects at load, not stored.

## UI

- **Settings modal** — new `EuiComboBox` (async, multi-select) picker under Window duration. Marked "experimental".
- **Service overview** — the existing "Correlated data" card gets a 4th row: **"View correlated dashboards (N)"**. Opens the shared correlations flyout on a new **Dashboards** tab. The card stays a fixed 4-row list so the sibling "top services by fault rate" panel never stretches with the number of dashboards.
- **Topology map** — the node panel (`ServiceDetailsPanel`) gains an icon action next to Spans/Logs (uses the `dashboardApp` icon), which opens the same flyout tab. Node-only, not edges.
- **Empty state** — if no dashboards are configured, the flyout Dashboards tab renders an `EuiEmptyPrompt` with a primary "Add correlated dashboards" CTA that opens APM settings.
- **Open behavior** — a link opens `/app/dashboards#/view/<id>` in a **new browser tab**, carrying the current APM time range as `_g` (no per-service filter in v0). Uses `basePath.prepend`, so workspace context is preserved.
Screenshot 2026-09-14 at 11 28 59 AM
Screenshot 2026-09-14 at 11 29 22 AM

## Tests

- SO client: create with dashboards, create without (feature inert), update preserves existing ids when omitted, update replaces with a new list, update clears with an empty list.
- Navigation helpers: deep-link shape, workspace `basePath.prepend`, special-char id encoding, no-time-range case, and the APM settings marker path.

No new integration tests, per the experimental scope.


ps48:
Thanks for the analysis — addressed in dfd4adc:

**Fixed**

- **Ensure hashchange fires for settings marker** — `openApmSettings` now dispatches a `HashChangeEvent` (matching the existing `navigateToSloSuggest` in the same file) so the marker hook reacts even when the hash is assigned the same value, and defaults the path to `/` so an empty hash can't produce `#?...`. I preserved the existing query params verbatim rather than rebuilding them with `URLSearchParams.toString()`, since that would percent-encode any rison `_g`/`_a` values — the same care `useOpenOnUrlMarker` takes when it strips the marker. Added unit tests for the dispatch, the empty-path guard, rison preservation, and marker de-duplication.
- **Tighten placeholder regex with anchors** — anchored the reference-placeholder regex to `/^references\[(\d+)\]\.id$/` in both parse sites.

**Not changed (with rationale)**

- **Encode time range values in rison** — kept as-is. This matches the existing trace/log deep links in this same file, which interpolate the time range into `time:(from:…,to:…)` directly. Wrapping the values in `encodeURIComponent(...)` inside rison single-quotes would produce non-standard rison (as the suggestion itself notes), and the values come from internal time-picker state, not untrusted input. The destination is also always same-origin via `basePath.prepend`.
- **Remove eslint-disable comments** — kept. Removing them reintroduces 3 `react-hooks/exhaustive-deps` errors (verified locally): the `spans`/`logs`/`attributes` tab-content consts are plain JSX used in the tabs `useMemo` dependency array. They predate this PR and only surface here because this PR touches the file, so the disables are load-bearing for the lint step rather than no-ops.

joshuali925:
### Code review

Found 4 issues:

1. The new dashboard-search `loadDashboards` is wired to `onSearchChange` (fires per keystroke) but has no cancellation/ordering guard, so a slower earlier request can resolve last and overwrite `dashboardOptions` with stale results. Every sibling async fetch in this APM module (`useDatasets`, `usePrometheusDataSources`, `useCorrelatedLogs`) uses an `AbortController` guard for exactly this reason. The same concern was raised on #2557 (dismissed there only because that call fired on mount/refresh, not per keystroke).

https://github.com/opensearch-project/dashboards-observability/blob/c870641756e4fd08ca6ddb97e752dc60638aa91c/public/components/apm/config/apm_settings_modal.tsx#L108-L129

2. The Dashboards tab destructures only `config` from `useApmConfig()` and ignores the `loading` flag. On mount `config` is `null` until the fetch resolves, so `dashboards.length === 0` is transiently true and the "No correlated dashboards" empty state flashes even for services that have dashboards configured. #2887 fixed this exact class of bug in `service_map_sidebar.tsx` by adding an `isLoading && list.length === 0` loading branch.

https://github.com/opensearch-project/dashboards-observability/blob/c870641756e4fd08ca6ddb97e752dc60638aa91c/public/components/apm/shared/components/service_correlations_flyout.tsx#L974-L977

3. The new `correlatedDashboards` entity stores an array of placeholders, diverging from the documented single-object `{ id: 'references[N].id' }` entity contract that `parseEntityReference` handles. As a result it's silently skipped by the shared resolution path (`entityValue.id` is undefined on an array) and re-parsed ad hoc with a duplicated regex in `getBulkWithResolvedReferences`. Consider extending the documented resolution path instead.

https://github.com/opensearch-project/dashboards-observability/blob/c870641756e4fd08ca6ddb97e752dc60638aa91c/public/services/saved_objects/saved_object_client/osd_saved_objects/apm_config.ts#L91-L97

4. Three `eslint-disable-next-line react-hooks/exhaustive-deps` comments were added above plain JSX const assignments (`spansTabContent`, `logsTabContent`, `attributesTabContent`), which are not hook calls — the rule doesn't apply to them, so the directives are inert and misleading (likely meant for the `tabs` `useMemo` below). Same at L763 and L912.

https://github.com/opensearch-project/dashboards-observability/blob/c870641756e4fd08ca6ddb97e752dc60638aa91c/public/components/apm/shared/components/service_correlations_flyout.tsx#L669-L672

TackAdam:
## Additional Findings

### N1 — Low — Same missing-loading-flag root cause 

`service_overview.tsx:77`

The overview card's count pill derives:

`correlatedDashboardCount = config?.correlatedDashboards?.length ?? 0`

On mount, `config` is `null`, so the count is `0` and the `(N)` pill is briefly absent. It then pops in once the fetch resolves.

This is a cosmetic flicker. loading gate in the context/consumer should cover this case as well.

### N2 — Low — Exported marker constant not used by its reader

`apm_settings_modal.tsx:144,157`

`APM_SETTINGS_FOCUS_MARKER = '_apmSettingsFocus'` is defined in `navigation_utils.ts:410` and used by the writer (`openApmSettings`), but the reader (the settings modal) hardcodes the literal `'_apmSettingsFocus'` in two places.

This allows the writer and reader to drift if the string ever changes.

**Fix:** Import and use `APM_SETTINGS_FOCUS_MARKER` in the modal instead of hardcoding the literal.

ps48:
Thanks for the review — both focus areas addressed in 4f5c966:

1. **Reference-index coupling** — `createEntities` no longer hard-codes the dashboard start slot. It now derives it from the actual references array that `createReferences` produced (`references.length - correlatedDashboardIds.length`, since dashboards are appended after the fixed dataset/datasource refs), so the entity placeholders stay in lock-step with the reference positions and won't drift if a fixed ref is ever added ahead of them. The hard-coded `DASHBOARD_REF_START` constant is gone.

2. **Dead `tabs` memoization + no-op eslint-disable** — you're right that the `tabs` `useMemo` never hit its cache (its deps include the non-memoized `spans`/`logs`/`attributes` tab-content consts, which are new each render) and that the `eslint-disable react-hooks/exhaustive-deps` comments on those plain consts had no effect. Converted `tabs` to a plain IIFE (behavior-identical, since it already rebuilt every render) and removed the three disable comments.

No behavior change; the saved-object client unit tests remain green.

ps48:
Thanks @TackAdam!

**N2 — fixed** in d5395dc: the settings modal now imports and uses `APM_SETTINGS_FOCUS_MARKER` from `navigation_utils` instead of hardcoding `'_apmSettingsFocus'` in the two reader sites, so writer and reader can't drift.

**N1 — I'd like to leave the count as-is.** Deriving `correlatedDashboardCount` from `config` (rather than gating on `loading`) is deliberate: on a manual refresh the context keeps `config` at its last value while `loading` flips true, so the count stays stable and doesn't blink. Gating the pill on `loading` would make it disappear on every refetch. The only artifact is the one-time pop-in on first mount, which is inherent to the async load and would need a skeleton to hide. Happy to add a skeleton/placeholder if you'd prefer that over the pop-in.

ps48:
Thanks @joshuali925 — all 4 addressed:

1. **Search race** — `loadDashboards` now uses a monotonic sequence token; a response is applied (and clears the loading flag) only if it's still the latest search, so an earlier slow request can't overwrite newer results. (fb94c0c)
2. **Empty-state flash** — the Dashboards tab now reads the config `loading` flag and shows a spinner while `config` is still `null`, instead of transiently flashing the "No correlated dashboards" empty state. (fb94c0c)
3. **Duplicated placeholder regex / entity contract** — extracted a single `parseReferenceIndex('references[N].id') → N` helper, now used by both the 1:1 `parseEntityReference` path and the `correlatedDashboards` resolution (no more duplicated regex). Documented that `correlatedDashboards` is intentionally an *array* of placeholders — a service can link many dashboards — which is why it's resolved as a list rather than through the 1:1 single-value path. (fb94c0c)
4. **Inert eslint-disable comments** — already removed in 4f5c966; the defeated `tabs` `useMemo` (its deps included the non-memoized tab-content consts) was converted to a plain IIFE, so there's no misleading directive left.

Separately, per product feedback I also dropped the "(N)" count from the overview "View correlated dashboards" row to match the count-less spans/logs rows.

TackAdam (approved):

joshuali925 (approved):

PR #2899, opened 2026-09-17
PPL look-back window + notification throttling for logs rules

lezzago:
### Description

Adds two capabilities to the unified-alerts **logs (PPL) rule** flow, plus graceful handling of large queries and the UX/a11y polish surfaced while building them.

**Look-back window** (mirrors alerting-dashboards [#1516](https://github.com/opensearch-project/alerting-dashboards-plugin/pull/1516))
- Toggle + amount + unit (min/hr/day) + timestamp-field picker in the Schedule panel; 1-minute–7-day bound; default-on at 1 hour, auto-anchored to a detected date field (falls back to the toolbar time field); disabled with a visible reason when no date field exists.
- Realized by injecting a sliding `| where <ts> > DATE_SUB(NOW(), INTERVAL n u)` filter into the query on save and stripping/re-injecting idempotently. `ppl_monitor` persists no `ui_metadata`, so the edit round-trip reconstructs the window by parsing the filter back out of the query and shows the clean query.

**Per-action notification throttling**
- Checkbox + "only send every N minutes" per action → `throttle_enabled` + `throttle:{value,unit:MINUTES}`; off by default; round-tripped on edit.

**Large-query handling (backend-authoritative)**
- No client-side length cap: the alerting backend enforces `ppl_monitor_max_query_length` (a cluster may raise it) and we can't read cluster settings, so blocking on a hard-coded default would lock out those users. The OSD proxy body-schema bound is a generous abuse guard only; the backend's limit-naming error is surfaced inline; a non-blocking char-count advisory replaces any block. The query editor auto-expands to fit large queries, then scrolls.

**UX / a11y** — number fields are the sole `EuiFormRow` child so EUI wires the error into `aria-describedby` (WCAG 3.3.1); disabled-reason rendered as visible helpText (4.1.2); tidy throttle layout; removed a leaked cluster-setting name from help text.

### Screenshots

Before = `origin/main` (merge base), after = this PR. Captured at the same viewport/route/data.

**1. Look-back window in the logs-rule Schedule panel** — before: Schedule has only Run every / Unit; after: a look-back window (default 1h) is added and auto-anchored to a detected date field.

Look-back window before/after

**2. Per-action notification throttling** — before: an action notifies on every match; after: a per-action "only send every N minutes" throttle.

Throttle before/after

**Flow — create a logs rule (this PR):** open → pick index (look-back auto-activates) → add action + enable throttle.

After flow

**Flow — same steps on `origin/main`:** no look-back in Schedule, no throttle on the action.

Before flow

Responsive: both new controls reflow cleanly down to ~560px (amount+unit stay paired; throttle field stays compact); verified via layout probe + screenshots.

### Issues Resolved
N/A — feature work.

### Check List


lezzago:
Thanks for the automated review — each finding was verified against the code (and, where relevant, a live cluster):

**Code-Diff-Analyzer**

1. **`body_schema.ts` — `PPL_QUERY_MAX` 2000 → 100000 (medium): not an issue.** "No other size guard exists" doesn't hold: OSD core caps every request body at `server.maxPayload` (1 MiB default, `src/core/server/http/http_config.ts`), and these routes set no override. The schema already accepts far more than 100k chars via `triggers` (50 × `unknowns: 'allow'`). The real limit is the alerting backend's `plugins.alerting.ppl_monitor_max_query_length`, which is cluster-configurable with no upper bound (verified: it accepts 2147483647) — a hard 2000 at the proxy would block clusters that raised it, and OSD can't read the setting without extra permissions. The backend's rejection ("PPL Query length must be at most N but was M") is surfaced inline.
2. **`ppl_lookback.ts` — timestamp field interpolated into PPL (low): not a security issue.** The same user authors the entire PPL query, which runs under their own permissions, so a crafted field name can't do anything they couldn't type directly. Non-identifier field names are now backtick-quoted (`formatPplField`) so they produce valid PPL.

**PR Reviewer Guide**

3. **Parser misses plural units / narrower field class:** the plural half is incorrect — the regex already accepts `(MINUTE|HOUR|DAY)S?` (tested). The field-class half was valid for backticked names and is fixed: parse now accepts backtick-quoted fields, matching what the injector writes.
4. **Re-save reorders the clause relative to user filters:** the clause the plugin writes round-trips exactly. The real issue behind this was hand-written `DATE_SUB` clauses elsewhere in the query being adopted and moved — fixed: the plugin now owns only the whole first pipe segment in the exact shape it writes, save never removes or rewrites user text, and user time filters stay verbatim (12-case test matrix on save and on edit).
5. **Silent 10-minute throttle default:** unreachable — both callers of `transformPplFormToPayload` go through `validatePplForm`, which rejects an invalid throttle before save (and the field shows the error). It mirrors the existing defensive defaults in the same file.

Changed-code coverage is now 100% of lines and branches (183/183, 200/200).

lezzago:
Follow-up on the findings raised against `f722b7d` that weren't covered above (each verified):

**PR Reviewer Guide**

1. **Backtick escaping in `formatPplField` is lossy:** not worth changing. This only applies to a time field whose name contains a literal backtick. In that unlikely case it fails loudly: the clause names a field that doesn't exist, and the backend's PPL validation rejects it inline. The edit parser also won't match it, so the clause stays visible in the query rather than being silently adopted.
2. **Potential ReDoS on `INJECTED_SEGMENT_RE`:** not an issue. The pattern is anchored, and its whitespace quantifiers are separated by literal tokens with no nested quantifiers. Benchmarked on 100k-char pathological segments (whitespace runs, unterminated backticks, long tails): ≤1.1 ms each.

**PR Code Suggestions**

3. **Validate the throttle unit before adopting the value:** unreachable. The alerting backend rejects any non-`MINUTES` unit. Verified on a live cluster: `{"value":2,"unit":"HOURS"}` returns `400 "Only support MINUTES throttle unit currently"`, so no stored action can carry another unit.
4. **Clamp negative look-back amounts:** not needed. The input has `min={1}`, and `validatePplForm` flags anything below 1 inline. Clamping to 0 would show the same error.
5. **Guard fractional throttle values:** no change needed. As the suggestion's own rationale notes, the value is already floored before the `>= 1` check.

riysaxen-amzn:
* formatPplField backtick-stripping is lossy (a`b → `ab`). The author's rationale (fails loud at the backend, unlikely field name) is reasonable, but silently mapping two field names to one is the kind of thing that bites later. Fine to leave, worth a one-line acknowledgement in code.
* The throttle default of 10 is a magic number in buildPplActionPayload while DEFAULT_THROTTLE_VALUE = 10 exists in create_monitor_types.ts — but common/ can't import from public/, so it's duplicated. A named const in form_transforms.ts would read better than the bare 10.

riysaxen-amzn (approved):

PR #2900, opened 2026-09-18
[APM] Chart interactivity: synced crosshair, time brush, legend isolate, x-axis fix, open-in-metrics

ps48:
### Description

Improves interactivity of the shared PromQL line chart used across the APM
service-details pages (Overview, Operations, Dependencies) and the topology
service-insights flyout. These pages render many Prometheus charts, and today
they are static.

**Changes**
- **Time brush** — drag-select on a chart's x-axis to zoom. On the
  service-details tabs this updates the whole page time range (every chart
  re-queries and the URL syncs), wired through a new optional
  `onTimeRangeChange` prop.
- **Synced crosshair** — hovering one chart draws a synced vertical guide and
  per-series dots on the sibling charts on the same page/flyout, via a small
  pub/sub cursor bus (`ApmCursorContext`) keyed on the hovered timestamp. The
  hovered chart keeps the native tooltip; remote charts show only the guide.
  Adapted from the Explore "metrics" flavor implementation.
- **Legend isolate** — clicking a series in a multi-series chart hides the
  others; re-clicking the isolated series restores all.
- **X-axis label overlap** — sets `axisLabel.hideOverlap` so long time ranges
  drop colliding tick labels instead of overprinting them.
- **Open in metrics** — a top-right button on each chart deep-links to the
  Explore metrics query view with the chart's PromQL, Prometheus data source,
  and current time range pre-loaded.

**Notes**
- No new dependencies (`echarts` is already used by the chart).
- No new server routes, saved-object mappings, or migrations.

### Testing

- `yarn lint` clean; the plugin bundle compiles.
- Jest: cursor bus (subscribe/publish/unsubscribe/isolation), the brush→ISO
  callback, legend isolation dispatch, the `hideOverlap` option, and the
  deep-link button. All affected suites pass.
- UI screenshots to follow.

### Check List
- [x] All tests pass
- [x] New functionality includes unit tests
- [ ] New functionality has been documented


TackAdam:
## Code review — chart interactivity

Reviewed the full diff at high effort. The rison/URL escaping for the query and time fields is solid and well-tested, and the synced-crosshair effect lifecycle (setOption-before-interaction ordering, `isDisposed()` cleanup guards, shape teardown/re-add) is sound. Below are the residual findings, ranked.

### Action items (correctness)

**1. `connectionId` is not rison-escaped in the deep-link dataset — `navigation_utils.ts:82`**
The query and time values are carefully rison-escaped, but `connectionId` is interpolated raw into the single-quoted rison strings:
```ts
const dataset = `dataset:(id:'${connectionId}',title:'${connectionId}',...)`;
```
A Prometheus connection whose id contains a rison-special char (`'` or `!`) produces malformed rison. Explore's `redux_persistence` then discards the corrupted dataset, so "Open in metrics" opens with no data source — silently defeating the robustness added for the query/time fields. Apply the same `!`→`!!` then `'`→`!'` escaping to `connectionId`.

**2. Right-to-left brush produces an inverted time range — `promql_line_chart.tsx:368`**
`onBrushEnd` uses `range[0]`/`range[1]` directly. If ECharts reports `coordRange` in drag order, a right-to-left drag yields `from > to`; every chart then re-queries an inverted range and returns empty/erroring results, and the inverted range is written to the URL. Sort the two endpoints before converting to ISO.

### Lower priority (edge-case / UX / convention)

**3. Crosshair not cleared when leaving the plot rect but staying on canvas — `promql_line_chart.tsx:512`**
`onZrMouseMove` returns early when the pointer is over the canvas but outside the grid (legend/axis area) without publishing `null` or clearing `isLocalHoverRef`. Sibling charts keep a stale crosshair and this chart keeps suppressing its own overlay until the pointer fully leaves the canvas. Publish `null` / reset the local-hover flag on the out-of-grid path.

**4. Legend-isolate `series-${i}` fallback won't match ECharts' auto-names — `promql_line_chart.tsx:388`**
For unnamed series ECharts generates names like `series0`, but the handler dispatches for `series-${i}` (`series-0`), which matches nothing — clicking those legend entries is a silent no-op. Either name every series explicitly on the `setOption` side or mirror ECharts' actual naming.

**5. "Open in metrics" button overlays error/empty states — `promql_line_chart.tsx:596`**
`canOpenInMetrics` only checks `promqlQuery`/`connectionId`, not chart state, so the absolutely-positioned button still renders (z-index 1, top-right) over the error/empty placeholder and can intercept clicks in that corner. Gate it on the loaded-with-data state too.

**6. Pervasive `any` in the new interaction code — `promql_line_chart.tsx:368`+**
`(params: any)`, `(e: any)`, `(inst.getModel() as any)`, `as number[] | null` casts throughout. Repo `CLAUDE.md` says "Avoid `any` (use `unknown` or generics)." Typing the ECharts event payloads and `convert*` return shapes would restore type-checking on the data this feature relies on.

---
*Items 1 and 2 are the ones worth blocking on; the rest are polish. Automated review — please sanity-check before acting.*

PR #2902, opened 2026-09-18
fix(trace-analytics): defer initial fetch until MDS data source resolves

### LINKED ISSUE #2901: [BUG] Trace Analytics Services/Traces error with "No Living connections" on first load when MDS is enabled without a local cluster
**Describe the bug**
When multi-data-source (MDS) support is enabled and the deployment has no local cluster (local cluster hidden / an MDS-only setup), the Trace Analytics Services and Traces pages fire their initial requests on first load before a data source has been resolved. Those requests are sent without a `dataSourceMDSId` query parameter, so the server routes them to the local cluster, which does not exist, and they fail.

Failing requests observed on first load:
- `POST /api/observability/trace_analytics/query` returns 400 `{"error":"Bad Request","message":"No Living connections"}` (services aggregation, service-map aggregation, and traces list).
- `POST /api/observability/trace_analytics/jaeger_indices` returns 400 with the same message.
- `GET /api/dsl/fieldCaps` returns 503 `{"error":"Service Unavailable","message":"No Living connections"}`.

The errors happen only on the first load of these pages. Once a data source is selected in the data source picker, the requests are re-issued with a valid `dataSourceMDSId` and succeed.

**To Reproduce**
1. Enable the data source (MDS) plugin and configure the deployment without a local cluster (local cluster hidden).
2. Register at least one remote data source.
3. Open Trace Analytics and land on the Services or Traces page as the first load.
4. Observe the failed network requests (400/503 with "No Living connections") in the browser dev tools.

**Expected behavior**
On first load with MDS enabled and no local cluster, Trace Analytics should not query the nonexistent local cluster. It should wait until the data source picker resolves a data source, then issue the requests against that data source. No "No Living connections" errors should appear.

**Root cause**
The Trace Analytics data source state starts with an undefined id before the data source picker asynchronously resolves and selects a default. The Services, Traces, and Dashboard content components fire their initial fetch effects on mount while the id is still undefined. Because the core HTTP client strips undefined query values, the `dataSourceMDSId` parameter is omitted and the server falls back to the local cluster client. There is no guard that defers the initial fetch until a data source id has been resolved. The Metrics page already uses this "defer until a data source is selected" pattern.

Relevant source:
- `public/components/trace_analytics/home.tsx` (jaeger_indices and fieldCaps calls on mount)
- `public/components/trace_analytics/components/services/services_content.tsx`
- `public/components/trace_analytics/components/traces/traces_content.tsx`
- `public/components/trace_analytics/components/dashboard/dashboard_content.tsx`
- `public/components/metrics/index.tsx` (existing guard pattern to mirror)

**Suggested fix**
Gate the initial fetch effects in the Trace Analytics content components so that, when MDS is enabled, requests are deferred until a data source id has been resolved. When MDS is disabled, behavior is unchanged and requests fire immediately against the local cluster.

**OpenSearch Dashboards version**
main (and applicable maintained branches).

**Plugins**
dashboards-observability (Trace Analytics).

**Host/Environment**
Any deployment with the data source (MDS) plugin enabled and the local cluster hidden.

joshuali925-osdbot:
### Description

When multi-data-source (MDS) is enabled and the deployment has **no local cluster** (local cluster hidden), the Trace Analytics **Services**, **Traces** and **Dashboard** content components fire their initial fetch effects on mount *before* the data source picker has asynchronously resolved and selected a default data source.

Because the data source id is still `undefined` at that point, the core HTTP client strips the `dataSourceMDSId` query param, so the server falls back to the (nonexistent) local cluster and the requests fail:

- `POST /api/observability/trace_analytics/query` → `400 {"message":"No Living connections"}` (services aggregation, service-map aggregation, traces list)
- `POST /api/observability/trace_analytics/jaeger_indices` → `400` same message
- `GET /api/dsl/fieldCaps` → `503 {"message":"No Living connections"}`

Once a data source is selected the requests are re-issued with a valid `dataSourceMDSId` and succeed, so the errors only appear on first load.

#### Fix

Add a small shared guard `shouldFetchTraceData(dataSourceEnabled, dataSourceMDSId)` (mirroring the existing "defer until a data source is selected" pattern used by the Metrics page) and use it to gate the initial fetch effects in:

- `home.tsx` (`jaeger_indices` + `fieldCaps` on mount)
- `services_content.tsx`
- `traces_content.tsx`
- `dashboard_content.tsx`

When MDS is enabled, requests are deferred until a data source id has been resolved. When MDS is disabled, behavior is unchanged and requests fire immediately against the local cluster.

### Verification

Reproduced locally with MDS enabled + `data_source.hideLocalCluster: true` and one registered remote data source, landing on **Services** as the first load.

**Before (upstream):** initial requests fire on first load with **no** `dataSourceMDSId` (these route to the local cluster and return `No Living connections` when it is absent):
```
POST /api/observability/trace_analytics/jaeger_indices          <-- no dataSourceMDSId
GET  /api/dsl/fieldCaps?index=otel-v1-apm-span-*...             <-- no dataSourceMDSId
POST /api/observability/trace_analytics/query                   <-- no dataSourceMDSId
POST /api/observability/trace_analytics/query                   <-- no dataSourceMDSId
```

**After (this PR):** every trace analytics request on first load carries the resolved `dataSourceMDSId`; no param-less request is made:
```
POST /api/observability/trace_analytics/query?dataSourceMDSId=<id>
POST /api/observability/trace_analytics/jaeger_indices?dataSourceMDSId=<id>
GET  /api/dsl/fieldCaps?index=otel-v1-apm-span-*...&dataSourceMDSId=<id>
```

Services page loading correctly against the selected MDS data source (no `No Living connections` errors):

services-after

### Issues Resolved

Fixes #2901

### Check List


public/components/trace_analytics/components/common/helper_functions.tsx
```diff
@@ -121,6 +121,19 @@ export function processTimeStamp(time: string, mode: TraceAnalyticsMode, isEndTi
...
+// asynchronously. Until it does, the data source id is undefined and the core HTTP client
+// strips the undefined `dataSourceMDSId` query param, causing the request to be routed to a
+// (possibly nonexistent) local cluster and fail with "No Living connections". This guard
+// defers the initial fetch until a data source id has been resolved. When MDS is disabled,
+// requests fire immediately against the local cluster (unchanged behavior).
+export function shouldFetchTraceData(
+  dataSourceEnabled: boolean,
+  dataSourceMDSId: string | undefined
+): boolean {
+  return dataSourceEnabled ? dataSourceMDSId !== undefined : true;
```
ps48: **[1] Silent permanent empty state when the data source id never resolves.** When MDS is enabled and the data source picker never resolves an id (no default data source, hidden local cluster, or a picker load error), `shouldFetchTraceData` stays `false` forever, so no request ever fires and no error is shown.

_Failure scenario:_ if `dataSourceEnabled=true` and `DataSourceSelectable` never calls `onSelectedDataSources`, `dataSourceMDSId[0].id` remains `undefined`, so the traces/services/dashboard pages sit permanently blank with no fetch and no error. Previously a request fired and surfaced "No Living connections", giving the user feedback; now the failure is silent.

public/components/trace_analytics/components/traces/traces_content.tsx
```diff
@@ -185,7 +190,7 @@ export function TracesContent(props: TracesProps) {
     setFilteredService(newFilteredService);
     if (!redirect && (mode === 'data_prepper' || (mode === 'jaeger' && jaegerIndicesExist)))
       props.setDataSourceMenuSelectable?.(true);
-    refresh();
+    if (shouldFetchTraceData(props.dataSourceEnabled, props.dataSourceMDSId[0].id)) refresh();
```
ps48: **[2] `refresh()` here is gated only by `shouldFetchTraceData`.** Unlike `dashboard_content`/`services_content` (which gate on `!redirect && (mode==='data_prepper' || (mode==='jaeger' && jaegerIndicesExist))`), this call fires during the redirect phase and in jaeger mode even when `jaegerIndicesExist` is `false`.

_Failure scenario:_ in jaeger mode with no jaeger indices, or while `redirect` is still true on first render, once the id resolves `refresh()` runs `handleTracesRequest` against nonexistent jaeger indices / during redirect, producing an error or a wasted request. Pre-existing (refresh was previously unconditional) but re-exposed by this touched guard line.

public/components/trace_analytics/home.tsx
```diff
@@ -2,6 +2,7 @@
  * Copyright OpenSearch Contributors
  * SPDX-License-Identifier: Apache-2.0
  */
+/* eslint-disable react-hooks/exhaustive-deps */
```
ps48: **[3] File-wide `eslint-disable react-hooks/exhaustive-deps`.** This disables missing-dependency linting for every effect in `home.tsx`, rather than narrowly fixing the deps the new guard touched.

_Failure scenario:_ future edits to any of `home.tsx`'s effects can silently omit dependencies (e.g., forget to add a new prop the effect reads) with no lint warning, letting stale-closure / missing-refetch bugs ship unnoticed. Prefer per-effect disables or adding the actual deps.

public/components/trace_analytics/components/dashboard/dashboard_content.tsx
```diff
@@ -88,9 +89,22 @@ export function DashboardContent(props: DashboardProps) {
...
-  }, [filters, startTime, endTime, appConfigs, redirect, mode, jaegerIndicesExist]);
+  }, [
+    filters,
+    startTime,
+    endTime,
+    appConfigs,
+    redirect,
+    mode,
+    jaegerIndicesExist,
+    dataSourceMDSId,
```
ps48: **[4] Adding `dataSourceMDSId` to these deps introduces a redundant duplicate fetch.** This dep was absent before the PR; adding it makes the dashboard re-run `refresh()` whenever the shared id+label array object changes — including when only the label resolves via home's `getDatasourceAttributes`.

_Failure scenario:_ loading a dashboard from a URL with `datasourceId` set — id is present so `refresh()` fires once; then `getDatasourceAttributes` updates the label (same id) and replaces the `dataSourceMDSId` array, changing its identity → the effect re-runs → a second, identical set of dashboard/throughput/error-rate requests fires.

public/components/trace_analytics/components/dashboard/dashboard_content.tsx
```diff
@@ -88,9 +89,22 @@ export function DashboardContent(props: DashboardProps) {
   }, []);
 
   useEffect(() => {
-    if (!redirect && (mode === 'data_prepper' || (mode === 'jaeger' && jaegerIndicesExist)))
+    if (
+      !redirect &&
+      (mode === 'data_prepper' || (mode === 'jaeger' && jaegerIndicesExist)) &&
+      shouldFetchTraceData(props.dataSourceEnabled, dataSourceMDSId[0].id)
```
ps48: **[5] Guard reads `props.dataSourceEnabled` but the deps array omits it** (same pattern in `services_content.tsx` and `traces_content.tsx`). It relies on `exhaustive-deps` being disabled and on `dataSourceEnabled` being constant.

_Failure scenario:_ if `dataSourceEnabled` ever changed at runtime from true→false while id is still undefined, the deferred fetch would not re-trigger because the effect doesn't observe that prop. Low risk today since the flag is effectively fixed per session.

public/components/trace_analytics/components/traces/traces_content.tsx
```diff
@@ -202,6 +207,7 @@ export function TracesContent(props: TracesProps) {
 
   useEffect(() => {
     if (tracesTableMode !== 'traces') return;
+    if (!shouldFetchTraceData(props.dataSourceEnabled, props.dataSourceMDSId[0].id)) return;
```
ps48: **[6] This guard is effectively dead for the deferral case.** The maxTraces effect now reads `shouldFetchTraceData(props.dataSourceEnabled, props.dataSourceMDSId[0].id)`, but its dependency array is still `[maxTraces]`, so it cannot re-run when the id resolves; initial 'traces' table loading depends entirely on the primary effect.

_Failure scenario:_ on mount with MDS enabled and id undefined this effect returns early (correct); when the id later resolves it does not re-run (id not a dep). It works only because the primary effect reloads the traces table — the guard here just adds an inert branch and a latent fragility if the primary effect's coverage ever narrows.

ps48 (commented):
Automated recall-focused correctness review of the MDS first-load guard. 6 findings below, ranked most-severe first. Posted as plain comments (no approve / request-changes).

public/components/trace_analytics/components/traces/traces_content.tsx
ps48: ### p1: the traces-table effect defers but never re-fires when the data source resolves

The guard is correct, but unlike the other effects this PR gates, this one's dependency array is still just `[maxTraces]` (`:219`) - it does not include `props.dataSourceMDSId`. So on a first MDS load with an unresolved id, `shouldFetchTraceData` returns false and this returns early (good), but when the id then resolves, the effect does not re-run (the id isn't a dep), so `refreshTracesTableData` is never called and the traces table stays empty until something else changes `maxTraces`. That reintroduces a blank-table version of the same first-load problem this PR fixes. Could we add `props.dataSourceMDSId` (and `props.dataSourceEnabled`) to this effect's deps, matching what you did for the sibling effect just above and in `services_content` / `dashboard_content`?

public/components/trace_analytics/home.tsx
ps48: ### p2: prefer a scoped disable over file-wide `exhaustive-deps`

This adds `/* eslint-disable react-hooks/exhaustive-deps */` for the whole file. Since this very fix is about an effect that needs a dependency (`dataSourceMDSId`) to re-fire, disabling that lint across the entire file removes the guard that would catch the next missing-dep of exactly this kind (it's the same class as the `traces_content` gap noted above). Could we scope it to the specific effect(s) that need it (a per-line `// eslint-disable-next-line`) so the linter keeps checking the rest of the file?

ps48 (commented):
### fix(trace-analytics): defer initial fetch until MDS data source resolves

Good, well-targeted fix for the exact "No Living connections" first-load class, and it's the right shape: a shared `shouldFetchTraceData(dataSourceEnabled, dataSourceMDSId)` guard in `common/helper_functions`, MDS-only (unchanged behavior when disabled), and it correctly treats the local cluster's empty-string id as "resolved" (only `undefined` defers). Unit tests cover the matrix. I verified the re-fire contract on most of the gated effects: `services_content`, `traces_content`'s first effect, `dashboard_content`, and both `home.tsx` effects all add `dataSourceMDSId` (and `dataSourceEnabled`) to their dependency arrays, so the deferred fetch correctly re-runs once the id resolves.

No `p0`. One `p1`: one gated effect is missing that dependency, so its deferred fetch never re-fires. One `p2` on a file-wide lint suppression.

TackAdam:
### Additional review (net-new items not already raised)

I reviewed on top of the existing feedback and did **not** re-raise anything already covered — @ps48's 6 recall-review findings + the `p1` (traces-table effect deps still `[maxTraces]`) and `p2` (file-wide `exhaustive-deps` disable), plus the PR-Agent bot notes (missing `props.dataSourceEnabled` dep, optional chaining on `dataSourceMDSId[0].id`) all stand and should be addressed. Below are only issues I did not see mentioned anywhere in the thread.

**1. (highest value) The same first-load bug is unfixed on the detail-page deep-link path — `service_view.tsx` and `trace_view.tsx`.**
The fix gates the three landing content pages + `home.tsx`, but the detail views have the identical ungated pattern:

- `service_view.tsx:257`
  ```ts
  useEffect(() => {
    if (!redirect) refresh();               // refresh() -> checkValidServiceName / handleServiceViewRequest with props.dataSourceMDSId[0].id
  }, [props.startTime, props.endTime, props.serviceName, props.mode]);
  ```
- `trace_view.tsx:371`
  ```ts
  useEffect(() => {
    ...
    refresh();                              // refresh() -> handlePayloadRequest / handleServiceMapRequest with props.dataSourceMDSId[0].id
  }, [props.traceId, props.mode, props.setDataSourceMenuSelectable]);
  ```

Neither call is guarded by `shouldFetchTraceData`, and neither effect lists `dataSourceMDSId` in its deps. So refreshing/bookmarking directly onto a service (`#/services/<name>`) or trace (`#/traces/<id>`) detail URL with MDS enabled + `hideLocalCluster: true` will fire before the id resolves and hit the exact `No Living connections` failure this PR targets — and because `dataSourceMDSId` isn't a dep, it won't recover when the id resolves either. Recommend extending the same guard (and dep) here so the fix is complete.

**2. Test coverage gap for the actual behavior being fixed.**
The new unit tests only exercise the pure `shouldFetchTraceData` matrix. None of the component effects' defer/re-fire behavior is tested. Notably the single Codecov-uncovered line is the `traces_content` guard branch (`traces_content.tsx:193`) — which is exactly where the `p1` "never re-fires" bug lives. A component-level test asserting "no request fires while id is `undefined`, request fires once id resolves" would both close the coverage gap and regression-guard `p1`.

**3. (minor) Unrelated reformatting bundled into `home.tsx`.**
The `getDataSourceMenu<...>()` generic reformats and the `(x as unknown as TraceAnalyticsMode)` cast rewrites are Prettier churn unrelated to the fix. They enlarge the diff and can cause conflicts; consider dropping them or splitting into a separate cleanup.

---

@joshuali925-osdbot could you please address all of the above along with the outstanding review comments (the `p1`/`p2` and the six inline findings) and push an update? Happy to re-review once the changes are in.

